PR #27, opened 2020-04-18
Add lastFailureReason and Details to ActivityTaskStartedEvent

samarabbas:


event/message.proto
```diff
@@ -155,6 +155,8 @@ message ActivityTaskStartedEventAttributes {
     string identity = 2;
     string requestId = 3;
     int32 attempt = 4;
+    string lastFailureReason = 5;
+    bytes lastFailureDetails = 6;
```
alexshtin, line 159: One more `bytes details`.

PR #30, opened 2020-04-20
Refactored RetryPolicy

mfateev:
Proto changes for https://github.com/temporalio/temporal/issues/290

### DIFF
--- a/common/message.proto
+++ b/common/message.proto
@@ -53,26 +53,19 @@ message Header {
     map<string, bytes> fields = 1;
 }
 
-message WorkflowRetryPolicy {
-    // Interval of the first retry. If coefficient is 1.0 then it is used for all retries.
+message RetryPolicy {
+    // Interval of the first retry. If retryBackoffCoefficient is 1.0 then it is used for all retries.
     int32 initialIntervalInSeconds = 1;
-
     // Coefficient used to calculate the next retry interval.
     // The next retry interval is previous interval multiplied by the coefficient.
     // Must be 1 or larger.
     double backoffCoefficient = 2;
-
     // Maximum interval between retries. Exponential backoff leads to interval increase.
-    // This value is the cap of the increase. Default is 100x of initial interval.
+    // This value is the cap of the increase. Default is 100x of the initial interval.
     int32 maximumIntervalInSeconds = 3;
-
     // Maximum number of attempts. When exceeded the retries stop even if not expired yet.
-    // Must be 1 or bigger. Default is unlimited.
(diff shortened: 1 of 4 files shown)


decision/message.proto
```diff
@@ -37,13 +37,35 @@ message ScheduleActivityTaskDecisionAttributes {
...
     string namespace = 3;
     tasklist.TaskList taskList = 4;
-    bytes input = 5;
-    int32 scheduleToCloseTimeoutSeconds = 6;
-    int32 scheduleToStartTimeoutSeconds = 7;
-    int32 startToCloseTimeoutSeconds = 8;
-    int32 heartbeatTimeoutSeconds = 9;
-    common.RetryPolicy retryPolicy = 10;
-    common.Header header = 11;
+    common.Header header = 5;
```
samarabbas, line 40: Are you doing this only for activity in this change?  There are other payloads like Heartbeat, Signal, etc where we should add them too.

samarabbas: Ah.  Just realized you moved it around.  You can ignore this comment.

decision/message.proto
```diff
@@ -37,13 +37,35 @@ message ScheduleActivityTaskDecisionAttributes {
...
+    // This timeout is always non retriable as all a retry would achieve is to put it back into the same queue.
+    // Defaults to scheduleToCloseTimeoutSeconds or workflow execution timeout if not specified.
+    int32 scheduleToStartTimeoutSeconds = 8;
+    // Maximum time an activity is allowed to execute after a pick up by a worker.
+    // This timeout is always retriable. Either this or scheduleToCloseTimeoutSeconds is required.
+    int32 startToCloseTimeoutSeconds = 9;
+    // Maximum time between successful worker heartbeats.
+    int32 heartbeatTimeoutSeconds = 10;
+    // Interval of the first retry. If retryBackoffCoefficient is 1.0 then it is used for all retries.
+    int32 initialRetryIntervalInSeconds = 11;
```
samarabbas, line 56: I don't like that we are flattening the structure.  The only difference is now we don't have expiration interval.  Let's have a quick discussion on the implications on user experience.

common/message.proto
```diff
@@ -53,7 +53,7 @@ message Header {
     map<string, bytes> fields = 1;
 }
 
-message RetryPolicy {
+message WorkflowRetryPolicy {
```
samarabbas, line 56: RetryPolicy is should be same for all kind of retries done by server.  There is nothing specific here about Workflow Vs Activity.

PR #35, opened 2020-04-29
Removed expiration from RetryPolicy and added one more workflow timeout

mfateev:
`RetryPolicy.expirationIntervalInSeconds` specified for how long the retry policy should keep retrying. The problem is that activity already has `ScheduleToClose` timeout that specifies for how long a workflow is willing to wait for an activity completion. This required extending the `ScheduleToClose` up to `RetryPolicy.expirationInterval` by the server which was extremely confusing for users. 
For workflows `RetryPolicy.expirationIntervalInSeconds` limited the duration of retries. At the same time there was no way to limit workflow duration os a cron workflow or a workflow that keeps calling _continue as new_.

This change:

1. Removes `expirationIntervalInSeconds` from `RetryPolicy`
2. Renames workflow `executionStartToCloseTimeoutSeconds` to `workflowRunTimeoutSeconds`
3. Adds to workflow `workflowExecutionTimeoutSeconds`

From now on the server is expected to retry activities up to `ScheduleToClose` timeout.
The workflow duration is limited to `workflowExecutionTimeoutSeconds` across all the retries and calls to _continue as new_. The duration of a single run is limited up to `workflowRunTimeoutSeconds`.

The other expected server change is that an activity should be given a default retry policy and retried by default unless the `retryPolicy. maximumAttempts` is not set to 1.

### DIFF
--- a/decision/message.proto
+++ b/decision/message.proto
@@ -37,13 +37,25 @@ message ScheduleActivityTaskDecisionAttributes {
     common.ActivityType activityType = 2;
     string namespace = 3;
     tasklist.TaskList taskList = 4;
-    bytes input = 5;
-    int32 scheduleToCloseTimeoutSeconds = 6;
-    int32 scheduleToStartTimeoutSeconds = 7;
-    int32 startToCloseTimeoutSeconds = 8;
-    int32 heartbeatTimeoutSeconds = 9;
-    common.RetryPolicy retryPolicy = 10;
-    common.Header header = 11;
+    common.Header header = 5;
+    bytes input = 6;
+    // Indicates how long the caller is willing to wait for an activity completion.
+    // Limits for how long retries are happening. Either this or startToCloseTimeoutSeconds is required.
+    // When not specified defaults to the workflow execution timeout.
+    int32 scheduleToCloseTimeoutSeconds = 7;
+    // Limits time an activity task can stay in a task queue before a worker picks it up.
+    // This timeout is always non retriable as all a retry would achieve is to put it back into the same queue.
+    // Defaults to scheduleToCloseTimeoutSeconds or workflow execution timeout if not specified.
+    int32 scheduleToStartTimeoutSeconds = 8;
+    // Maximum time an activity is allowed to execute after a pick up by a worker.
+    // This timeout is always retriable. Either this or scheduleToCloseTimeoutSeconds is required.
+    int32 startToCloseTimeoutSeconds = 9;
+    // Maximum time between successful worker heartbeats.
+    int32 heartbeatTimeoutSeconds = 10;
+    // Retry parameters. Note that activity is retried by default according to a default retry policy.
+    // To disable retries provide a retry policy with maximumAttempts equals to 1.
+    // The retries happen up to scheduleToCloseTimeout.
+    common.RetryPolicy retryPolicy = 11;
 }
 
 message RequestCancelActivityTaskDecisionAttributes {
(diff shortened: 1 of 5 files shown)


decision/message.proto
```diff
@@ -37,13 +37,25 @@ message ScheduleActivityTaskDecisionAttributes {
...
+    // Limits time an activity task can stay in a task queue before a worker picks it up.
+    // This timeout is always non retriable as all a retry would achieve is to put it back into the same queue.
+    // Defaults to scheduleToCloseTimeoutSeconds or workflow execution timeout if not specified.
+    int32 scheduleToStartTimeoutSeconds = 8;
+    // Maximum time an activity is allowed to execute after a pick up by a worker.
+    // This timeout is always retriable. Either this or scheduleToCloseTimeoutSeconds is required.
+    int32 startToCloseTimeoutSeconds = 9;
+    // Maximum time between successful worker heartbeats.
+    int32 heartbeatTimeoutSeconds = 10;
+    // Retry parameters. Note that activity is retried by default according to a default retry policy.
```
alexshtin, line 55:
```suggestion
    // Activity retry policy. Note that activity is retried by default according to a default retry policy.
```

decision/message.proto
```diff
@@ -37,13 +37,25 @@ message ScheduleActivityTaskDecisionAttributes {
...
+    // When not specified defaults to the workflow execution timeout.
+    int32 scheduleToCloseTimeoutSeconds = 7;
+    // Limits time an activity task can stay in a task queue before a worker picks it up.
+    // This timeout is always non retriable as all a retry would achieve is to put it back into the same queue.
+    // Defaults to scheduleToCloseTimeoutSeconds or workflow execution timeout if not specified.
+    int32 scheduleToStartTimeoutSeconds = 8;
+    // Maximum time an activity is allowed to execute after a pick up by a worker.
+    // This timeout is always retriable. Either this or scheduleToCloseTimeoutSeconds is required.
+    int32 startToCloseTimeoutSeconds = 9;
+    // Maximum time between successful worker heartbeats.
```
alexshtin, line 53: I would add required/optional and if optional what is default? The same way as you did for all 3 above.

event/message.proto
```diff
@@ -39,27 +39,31 @@ message WorkflowExecutionStartedEventAttributes {
...
-    string cronSchedule = 20;
-    int32 firstDecisionTaskBackoffSeconds = 21;
-    common.Memo memo = 22;
-    common.SearchAttributes searchAttributes = 23;
-    execution.ResetPoints prevAutoResetPoints = 24;
-    common.Header header = 25;
+    string firstExecutionRunId = 17;
+    common.RetryPolicy retryPolicy = 18;
+    int32 attempt = 19;
+    int64 workflowExecutionTimeoutTimestamp = 20;
```
alexshtin, line 60: Timeout is an interval. `TimeoutTimestamp` looks weird to me. I believe "expiration" is a better word here.

workflowservice/request_response.proto
```diff
@@ -118,16 +118,21 @@ message StartWorkflowExecutionRequest {
...
+    // Total workflow execution timeout including retries and continue as new
+    int32 workflowExecutionTimeoutSeconds = 6;
+    // Timeout of a single workflow run
+    int32 workflowRunTimeoutSeconds = 7;
+    // Timeout of a single workflow task
+    int32 workflowTaskTimeoutSeconds = 8;
+    string identity = 9;
+    string requestId = 10;
+    common.WorkflowIdReusePolicy workflowIdReusePolicy = 11;
+    // Retries up to workflowExecutionTimeout
```
alexshtin, line 130:
```suggestion
    // Retries up to workflowExecutionTimeoutSeconds
```

workflowservice/request_response.proto
```diff
@@ -210,21 +215,25 @@ message PollForActivityTaskRequest {
...
+    int64 scheduledTimestamp = 10;
+    int64 scheduledTimestampOfThisAttempt = 11;
+    int64 startedTimestamp = 12;
+    int32 attempt = 13;
+    int32 scheduleToCloseTimeoutSeconds = 14;
+    int32 startToCloseTimeoutSeconds = 15;
+    int32 heartbeatTimeoutSeconds = 16;
+    // This is an actual retry policy the service uses.
+    // It can be different from the one provided (or not) during activity scheduling
+    // as the service can override the provided one with default values.
```
alexshtin, line 235: Usually it is vice versa: "provided values override default values". I think better explanation why and when service ignores provided values.

PR #36, opened 2020-04-29
Rename Payload to Payloads

alexshtin:
and `PayloadItem` to `Payload`.

### DIFF
--- a/common/message.proto
+++ b/common/message.proto
@@ -33,11 +33,11 @@ message DataBlob {
     bytes data = 2;
 }
 
-message Payload {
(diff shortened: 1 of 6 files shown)


common/message.proto
```diff
@@ -51,15 +51,15 @@ message ActivityType {
 }
 
 message SearchAttributes {
-    map<string, Payload> indexedFields = 1;
+    map<string, Payloads> indexedFields = 1;
```
mfateev, line 54: I think this one should be a single Payload

alexshtin: This is next step.

PR #40, opened 2020-05-06
Add Failure and FailureInfo

alexshtin:


common/enum.proto
```diff
@@ -61,3 +61,10 @@ enum ContinueAsNewInitiator {
    Retry = 1;
    CronSchedule = 2;
 }
+
+enum TimeoutType {
```
alexshtin, line 65: Had to move it here to avoid package cycle.

alexshtin: Will extract this move to separate PR.

failure/message.proto
```diff
@@ -92,3 +98,63 @@ message QueryFailed {
...
+
+message ActivityTaskFailureInfo {
+    int64 scheduledEventId = 1;
+    int64 startedEventId = 2;
+    string identity = 3;
+}
+
+message ChildWorkflowExecutionFailureInfo {
+    string namespace = 1;
+    // execution.WorkflowExecution workflowExecution = 2;
```
alexshtin, line 139: I will uncomment this later. Need to fix another package cycle for it.

PR #42, opened 2020-05-14
Move types between packages

alexshtin:
- `WorkflowExecution` from `execution` to `common`.
 - `TimeoutType` from `event` to `common`.
 - `failure` package to `errordetails`.
- "retriable" to "retryable".


common/enum.proto
```diff
@@ -61,3 +61,10 @@ enum ContinueAsNewInitiator {
    Retry = 1;
    CronSchedule = 2;
 }
+
+enum TimeoutType {
```
mfateev, line 65: Let's break activity timeouts from workflow timeouts right away. It will be hard in the future

common/message.proto
```diff
@@ -28,6 +28,7 @@ option java_multiple_files = true;
 
 import "common/enum.proto";
 
+// Storage messages
```
mfateev, line 31: What is storage message?

common/enum.proto
alexshtin: I will do it in separate PR.

PR #43, opened 2020-05-19
Use Failure and FailureInfo for errors

alexshtin:
`string reason`/`common.Payloads details` pairs were replaced with one `failure.Failure failure` field which holds all information about errors on SDK side.


common/enum.proto
```diff
@@ -57,9 +57,9 @@ enum IndexedValueType {
 }
 
 enum ContinueAsNewInitiator {
-   Decider = 0;
-   Retry = 1;
-   CronSchedule = 2;
+    Decider = 0;
```
samarabbas, line 60: I thought we are not using 0 as real value?

alexshtin: Yes, but this is another PR (I just fixed padding here).

PR #46, opened 2020-05-30
Add RetryStatus enum and use it

alexshtin:
Adding `RetryStatus` enum and field of this type to `Failure`. Generally, it indicates the reason why server stopped retrying (can be useful for debugging). There is also some extra logic on server based on this field.

### DIFF
--- a/common/enum.proto
+++ b/common/enum.proto
@@ -74,7 +74,6 @@ enum TimeoutType {
 enum NoRetryReason{
     NotSet = 0;
     ErrorIsNonRetryable = 1;
-    RetryIsDisabled = 2;
-    Timeout = 3;
-    MaximumAttemptsExceeded = 4;
+    Timeout = 2;
+    MaximumAttemptsExceeded = 3;
 }
--- a/common/enum.proto
+++ b/common/enum.proto
@@ -70,3 +70,11 @@ enum TimeoutType {
     ScheduleToClose = 2;
     Heartbeat = 3;
 }
+
+enum NoRetryReason{
+    NotSet = 0;
+    ErrorIsNonRetryable = 1;
+    RetryIsDisabled = 2;
+    Timeout = 3;
+    MaxAttemptsExceeded = 4;
+}
--- a/common/enum.proto
+++ b/common/enum.proto
@@ -71,9 +71,12 @@ enum TimeoutType {
     Heartbeat = 3;
 }
 
-enum NoRetryReason{
-    NotSet = 0;
+enum RetryStatus{
+    Retrying = 0;
     ErrorIsNonRetryable = 1;
     Timeout = 2;
     MaximumAttemptsExceeded = 3;
+    RetryDisabled = 4;
+    InternalError = 5;
+    CancelRequested = 6;
 }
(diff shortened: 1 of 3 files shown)


event/message.proto
```diff
@@ -191,13 +191,10 @@ message ActivityTaskFailedEventAttributes {
...
 message ActivityTaskTimedOutEventAttributes {
-    common.Payloads lastHeartbeatDetails = 1;
     int64 scheduledEventId = 2;
     int64 startedEventId = 3;
-    common.TimeoutType timeoutType = 4;
-    // For retry activity, it may have a failure before timeout. It's important to keep those information for debug.
-    // Client can also provide the info for making next decision
-    failure.Failure lastFailure = 5;
+    // For retry activity, it may have a failure before timeout. It is stored as cause in timeoutFailure.
+    failure.Failure timeoutFailure = 5;
```
alexshtin, line 197: This always will be failure with `TimeoutFailureInfo`. It has `lastHeartbeatDetails` and `timeoutType`. `lastFailure` will go to `cause`.

mfateev: It is not clear. What is top level failure returned?

alexshtin: Top level is always `TimeoutFailureInfo` (I guess, with `ScheduleToClose` type always). `cause` might be another `TimeoutFailureInfo` with `StartToClose` or something else from `mutableState`.

common/enum.proto
```diff
@@ -70,3 +70,13 @@ enum TimeoutType {
...
 }
+
+enum RetryStatus {
+    Retrying = 0; // or InProgress, or Active
+    NonRetryableError = 1;
+    Timeout = 2;
+    MaximumAttemptsExceeded = 3; // or MaximumAttemptsExhausted
+    Disabled = 4; // or RetryDisabled, or NotConfigured
+    InternalServerError = 5; // InternalError, ServiceError
+    CancelRequested = 6;
```
alexshtin, line 81: Need some input on these names.

failure/message.proto
```diff
@@ -88,4 +87,5 @@ message Failure {
         ActivityTaskFailureInfo activityTaskFailureInfo = 11;
         ChildWorkflowExecutionFailureInfo childWorkflowExecutionFailureInfo = 12;
     }
+    common.RetryStatus retryStatus = 13;
```
mfateev, line 90: This doesn't belong to every error. It should be part of ActivityTaskFailureInfo and  ChildWorkflowExecutionFailureInfo.

event/message.proto
```diff
@@ -381,6 +382,7 @@ message ChildWorkflowExecutionFailedEventAttributes {
     common.WorkflowType workflowType = 4;
     int64 initiatedEventId = 5;
     int64 startedEventId = 6;
+    common.RetryStatus retryStatus = 7;
```
mfateev, line 385: We never retry cancelled activities. So I dont think it should be here.

event/message.proto
```diff
@@ -399,6 +401,7 @@ message ChildWorkflowExecutionTimedOutEventAttributes {
...
```
mfateev, line 404: Cancelled children are never retried.

PR #48, opened 2020-06-01
Added fields to ActivityTaskFailureInfo and failure to the marker

mfateev:
FailureInfo has to contain enough information to recreate the error without looking into the history.

Added failure to the marker decision and the corresponding event. It simplifies storing result of a local activity.


mfateev:
Fixes https://github.com/temporalio/temporal-proto/issues/47

PR #61, opened 2020-06-20
Add ruby_package definitions

antstorm:
Hey 👋 

I'm working on a Ruby SDK for Temporal (https://github.com/coinbase/temporal-ruby) and need `ruby_package` definitions in the proto files in order to use this repo as a submodule. Hope that works.

### DIFF
--- a/temporal/common/v1/message.proto
+++ b/temporal/common/v1/message.proto
@@ -28,6 +28,7 @@ option go_package = "go.temporal.io/temporal-proto/common/v1;common";
 option java_package = "io.temporal.common.v1";
 option java_multiple_files = true;
 option java_outer_classname = "MessageProto";
+option ruby_package = "Temporal::Api::Common::V1";
 
 import "temporal/enums/v1/common.proto";
 
--- a/temporal/decision/v1/message.proto
+++ b/temporal/decision/v1/message.proto
@@ -28,6 +28,7 @@ option go_package = "go.temporal.io/temporal-proto/decision/v1;decision";
 option java_package = "io.temporal.decision.v1";
 option java_multiple_files = true;
 option java_outer_classname = "MessageProto";
+option ruby_package = "Temporal::Api::Decision::V1";
 
 import "temporal/enums/v1/workflow.proto";
 import "temporal/enums/v1/decision_type.proto";
--- a/temporal/enums/v1/common.proto
+++ b/temporal/enums/v1/common.proto
@@ -28,6 +28,7 @@ option go_package = "go.temporal.io/temporal-proto/enums/v1;enums";
 option java_package = "io.temporal.enums.v1";
 option java_multiple_files = true;
 option java_outer_classname = "CommonProto";
+option ruby_package = "Temporal::Api::Enums::V1";
 
 enum EncodingType {
     ENCODING_TYPE_UNSPECIFIED = 0;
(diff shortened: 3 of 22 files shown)


temporal/common/v1/message.proto
```diff
@@ -28,6 +28,7 @@ option go_package = "go.temporal.io/temporal-proto/common/v1;common";
 option java_package = "io.temporal.common.v1";
 option java_multiple_files = true;
 option java_outer_classname = "MessageProto";
+option ruby_package = "Temporal::Proto::Common::V1";
```
alexshtin, line 31: This looks good to me. Run our linters for this change and they are happy. Can you remove `Proto` from path to match it with actual proto package path? I.e. make it `Temporal::Common::V1`?

antstorm: The reason I'm steering away from this is that I'm using `Temporal` as the top-level namespace for the SDK code and don't want any potential overlap with the generated code. I can change the top-level namespace to `TemporalProto` here if that makes any difference

alexshtin: This makes sense. Java has kinda the same problem. And we had to remove `proto` from `java_package` name because linter requires match with proto package name (`temporal.common.v1` in this case). For Ruby it doesn't have this requirement but I would follow the same rule here.
Also we try to follow googleapis [repo](https://github.com/googleapis/googleapis/search?q=ruby_package&unscoped_q=ruby_package) practices.
I guess `V1` suffix will serve as a good distinguisher between proto generated and SDK code.
And I am planning to remove `proto` word from `go_package` name also. The best candidate for now is `go.temporal.io/api/common/v1`.

I understand your point and I hope these are good arguments to convince you.

antstorm: @alexshtin this makes sense, but there are some practical inconveniences with this approach — since `V1` suffix comes after a 2nd level namespace (`Temporal::Workflow::V1` instead of `Temporal::V1::Workflow`) I can not use `Temporal::Workflow` class now (which is being used for workflow definitions), because Protobuf generated code expects these to be modules, instead of classes (both can be used for namespacing in Ruby, but have different properties). This is what generated code looks like:

```ruby
module Temporal
  module Workflow
    module V1
      ...
    end
  end
end
```

This is just one of the examples why SDK code should really be isolated from the Protobuf generated code.

Any ideas if there's any other feasible approach? I can of course do some renaming of the generated code in my repo, but that I think won't be a good solution.

alexshtin: What's about `api` (or `Api`) instead of `Proto`? I might rename `temporal` dir to `api` and fix Java package name accordingly.

antstorm: @alexshtin that would work perfectly 👍

alexshtin: Will do it this week and let you know.

antstorm: @alexshtin thanks! I've updated this PR accordingly

alexshtin:
Also please run `make` to make sure all linters are happy.

alexshtin:
I enabled automatic linting and building on server. You need to recreate this PR. Anyway, I think it would be easier for you to create new one rather than rebasing this.

antstorm:
@alexshtin yep, thank you for making the changes. Here's a new PR — #71

PR #66, opened 2020-07-08
Adopt Google api-linter

alexshtin:
Added config for `api-linter` and added both `api-linter` and `buf` to `Makefile`.

Run `make` once to install all linters and plugins and compile proto files.

Run `make proto` to compile proto files.


api-linter.yaml
```diff
@@ -0,0 +1,37 @@
...
+  disabled_rules:
+    - 'core::0192::has-comments'
+
+- included_paths:
+    - '**/message.proto'
+  disabled_rules:
+    - 'core::0122::name-suffix'
+    - 'core::0123::resource-annotation'
+    - 'core::0140::prepositions'    # should be gone after time fields conversion
+    - 'core::0142::time-field-type' # should be gone after time fields conversion
```
alexshtin, line 12: @shawnhathaway, these are for you to resolve.

temporal/api/enums/v1/namespace.proto
```diff
@@ -29,13 +29,17 @@ option java_package = "io.temporal.api.enums.v1";
 option java_multiple_files = true;
 option java_outer_classname = "NamespaceProto";
 
+// (-- api-linter: core::0216::synonyms=disabled
+//     aip.dev/not-precedent: TODO: Rename to NamespaceState. --)
```
alexshtin, line 33: Will resolve all TODOs in another PR.

PR #68, opened 2020-07-10
Onboard Buildkite

alexshtin:
Buildkite will run linters and build proto files to make sure they are buildable and validated.

### DIFF
--- a/docker/buildkite/Dockerfile
+++ b/docker/buildkite/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.14
+
+RUN apt update && apt install -y --no-install-recommends \
+      protobuf-compiler \
+      && rm -rf /var/lib/apt/lists/* \
+# https://github.com/docker-library/golang/blob/c1baf037d71331eb0b8d4c70cff4c29cf124c5e0/1.4/Dockerfile
+      && mkdir -p /temporal
+
+WORKDIR /temporal
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,10 @@
+steps:
+  - label: ":proto: build"
+    agents:
+      queue: "default"
+      docker: "*"
+    command: "make"
+    plugins:
+      - docker-compose#v3.1.0:
+          run: build
+          config: docker/buildkite/docker-compose.yml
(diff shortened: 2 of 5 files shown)


docker/buildkite/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.14
+
+RUN apt update && apt install -y --no-install-recommends \
+      protobuf-compiler \
```
markmark206, line 4: I would vote for moving this `apt-get install` behind a `make` target, and invoking that target in `command:` in pipeline.yml (e. g.  `- command: "make install-protobuff-compiler && make install"`)

I think this would help us maintain consistency in the version of protobuff compiler across all the contexts in which we might use it – local builds, buildkite, concourse, ...)

alexshtin: This is very good point. My only concern is Mac and possible Windows. Essentially we will neep platform depended installation from `Makefile`. Is it good practice?

alexshtin: https://stackoverflow.com/questions/714100/os-detecting-makefile

markmark206: I am not sure about it being a universal practice, but it's been working out nicely for us in installing kubernetes-related tools on dev machines (macs, anyway;) and in our test  pipelines (which run things in linux-based containers). 

Here is an example of a `Makefile` target that installs a specific version of kubectl (in this case, we are doing a direct `wget` download, rather than an apt-get command, so the details are a little different, but the idea is the same): 

https://github.com/temporalio/helm-chart-experiments/blob/a171d4597b0459a1ad0a62d0e98c56cc982d71bf/Makefile#L206

and here is the script that it uses to perform the actual download: https://github.com/temporalio/helm-chart-experiments/blob/master/tools/download_kubectl.sh
 
In this setup, one can run the same`make install-tools` command on a mac and on linux, and get the same version of things.

markmark206: that stack overflow thread is a good discussion. oh, and i put that logic in a separate script for simplicity (since running shell commands in a makefile has its own set of considerations).

alexshtin: I will try it in separate PR.

PR #89, opened 2020-09-22
Switch braking changes detector to PACKAGE level

alexshtin:


.gitignore
```diff
@@ -1,3 +1,2 @@
 /.idea
-/.gen
-/image.bin
```
samarabbas, line 3: Why did you remove this from gitignore?

samarabbas (approved):

.gitignore
alexshtin: not `.gen`. Just `image.bin`. I change the way validator works and it will not be created any more.

PR #98, opened 2020-11-03
Set SHELL explicitly for PATH to work on Mac - 2

alexshtin:


Makefile
```diff
@@ -13,8 +13,7 @@ GOPATH := $(shell go env GOPATH)
 endif
 
 GOBIN := $(if $(shell go env GOBIN),$(shell go env GOBIN),$(GOPATH)/bin)
-SHELL := /bin/bash
-export PATH := $(GOBIN):$(PATH)
```
sergeybykov, line 17: Breaks on MacOS without this `export` line.

sergeybykov: Never mind. It works for me.

sergeybykov (approved):

PR #100, opened 2020-11-04
Switch bash back to sh

alexshtin:


wxing1292 (approved):
why moving from bash to sh?

alexshtin:
It was `sh` originally. I switched it to `bash` today and now realized that we don't need it.

PR #101, opened 2020-12-04
Add namespace to RespondActivityTask* requests

sergeybykov:
This is how client will pass the namespace it is configured for as part of those requests. Will help us detect mismatches between that namespace and one in the token.


wxing1292:
also include `RecordActivityTaskHeartbeatRequest`

wxing1292 (approved):

PR #103, opened 2021-01-05
Re-order import so @grpc/proto-loader works

sunshineo:
In [this post](https://community.temporal.io/t/how-to-do-operations-tctl-can-do-from-any-programming-language/1256/4), @mfateev told me to use this project with Node grpc to call the service.
I followed [the tutorial](https://www.grpc.io/docs/languages/node/basics/) but found this strange problem.
I added this project using git submodule to a folder called `api` and used the following code to load package definition. It fails out of the box but works with the changes in the pull request, which did nothing but re-order the import

```javascript
const grpc = require('@grpc/grpc-js')
const protoLoader = require('@grpc/proto-loader');

const PROTO_PATH = './api/temporal/api/workflowservice/v1/service.proto'

const packageDefinition = protoLoader.loadSync(
    PROTO_PATH,
    {
        keepCase: true,
        longs: String,
        enums: String,
        defaults: true,
        oneofs: true,
        includeDirs: ['./api']
    }
)
const protoDescriptor = grpc.loadPackageDefinition(packageDefinition).temporal.api.workflowservice.v1
```

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -30,11 +30,6 @@ option java_multiple_files = true;
 option java_outer_classname = "RequestResponseProto";
 option ruby_package = "Temporal::Api::WorkflowService::V1";
 
-import "google/protobuf/duration.proto";
-import "google/protobuf/timestamp.proto";
-
-import "dependencies/gogoproto/gogo.proto";
-
 import "temporal/api/enums/v1/workflow.proto";
 import "temporal/api/enums/v1/namespace.proto";
 import "temporal/api/enums/v1/failed_cause.proto";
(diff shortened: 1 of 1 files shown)


feedmeapples:
@sunshineo could you provide more context, how did you come up with the re-ordering and fixing the issue this way?

In Temporal Web we take a similar approach, except it uses an older `grpc` node module. I wonder why the newer `@grpc/grpc-js` only works after re-ordering

alexshtin (approved):
This is weird but it shouldn't affect any other language.

feedmeapples (approved):

sunshineo:
@feedmeapples It was error out saying cannot import, so I start to copy paste everything into one file piece by piece. And through pure luck I noticed the order matters. I agree it must be the new grpc library doing something weird. I'm still stuck after 2 hard days. Please help https://community.temporal.io/t/how-to-do-operations-tctl-can-do-from-any-programming-language/1256/4

PR #112, opened 2021-04-22
[DO NOT MERGE] Add scheduled status to WorkflowExecutionStatus

Ardagan:
**What changed?**
Added WORKFLOW_EXECUTION_STATUS_SCHEDULED to WorkflowExecutionStatus

**Why?**
We can have a case when cron scheduled workflow is not started yet.

**How did you test it?**
This adds a new field to protobuf that is a backwards compatible change.

**Potential risks**
N/A


temporal/api/enums/v1/workflow.proto
```diff
@@ -71,6 +71,7 @@ enum WorkflowExecutionStatus {
     WORKFLOW_EXECUTION_STATUS_TERMINATED = 5;
     WORKFLOW_EXECUTION_STATUS_CONTINUED_AS_NEW = 6;
     WORKFLOW_EXECUTION_STATUS_TIMED_OUT = 7;
+    WORKFLOW_EXECUTION_STATUS_SCHEDULED = 8;
```
wxing1292, line 74: What about created instead of scheduled?

PR #117, opened 2021-06-17
Add metadata_only to workflow listing queries

dnr:
**What changed?**

Adding a `metadata_only` field to list workflows without fetching the `memo` fields, which can be very large.

**Why?**

To speed up workflow queries.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -440,6 +440,7 @@ message ListOpenWorkflowExecutionsRequest {
         temporal.api.filter.v1.WorkflowExecutionFilter execution_filter = 5;
         temporal.api.filter.v1.WorkflowTypeFilter type_filter = 6;
     }
+    bool omit_memo = 7;
 }
 
 message ListOpenWorkflowExecutionsResponse {
@@ -457,6 +458,7 @@ message ListClosedWorkflowExecutionsRequest {
         temporal.api.filter.v1.WorkflowTypeFilter type_filter = 6;
         temporal.api.filter.v1.StatusFilter status_filter = 7;
     }
+    bool omit_memo = 8;
 }
 
 message ListClosedWorkflowExecutionsResponse {
@@ -469,6 +471,7 @@ message ListWorkflowExecutionsRequest {
     int32 page_size = 2;
     bytes next_page_token = 3;
     string query = 4;
+    bool omit_memo = 5;
 }
 
 message ListWorkflowExecutionsResponse {
@@ -481,6 +484,7 @@ message ListArchivedWorkflowExecutionsRequest {
     int32 page_size = 2;
     bytes next_page_token = 3;
     string query = 4;
+    bool omit_memo = 5;
 }
 
 message ListArchivedWorkflowExecutionsResponse {
@@ -493,6 +497,7 @@ message ScanWorkflowExecutionsRequest {
     int32 page_size = 2;
     bytes next_page_token = 3;
     string query = 4;
+    bool omit_memo = 5;
 }
 
 message ScanWorkflowExecutionsResponse {
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -57,6 +57,8 @@ message SearchAttributes {
 
 message Memo {
     map<string, Payload> fields = 1;
+    // Memo is present in visibility database, but was not returned in this response.
+    bool is_omitted = 2;
 }
 
 message Header {


temporal/api/workflow/v1/message.proto
```diff
@@ -52,6 +52,7 @@ message WorkflowExecutionInfo {
     temporal.api.common.v1.WorkflowExecution parent_execution = 8;
     google.protobuf.Timestamp execution_time = 9 [(gogoproto.stdtime) = true];
     temporal.api.common.v1.Memo memo = 10;
+    bool has_memo = 15;
```
wxing1292, line 55: the return value should not contain `has_memo`
if the request is asking for all data, then return all data
if the request is asking for only metadata, then only return metadata

wxing1292 (changes requested):
see my above comments

temporal/api/workflow/v1/message.proto
dnr: @feedmeapples specifically requested an indication that the memo field is present even when it's not being returned. Is that still needed?

wxing1292: > indication that the memo field is present even when it's not being returned
what is the use case?

feedmeapples: use case is to show users placeholder boxes for payloads/memos etc, fields that can be large in size (mbs). Then we can fetch them if user clicks on the placeholder box.
This is similar how github shows diff for big files: 
image
 
 
As for the API design: i think we should do this carefully and make sure it covers Web UI needs (able to show placeholders for inputs, results, payloads, memos).
let's sync on this and keep the PR in draft until we have a solid agreement on this API

cc @stevekinney @sw-yx ^

feedmeapples (changes requested):

alexshtin (changes requested):
I would suggest the following:
1. Rename `metadata_only` to `omit_memo` because it is not clear what is "metadata" in this case. Are `search_attributes` metadata? Or `auto_reset_points`? We will need to define "metadata".
2. For response field, I have two options:
  - Add `is_omitted` field to `Memo` struct. The cons is that we use same struct to persist memo and this field will go to db (minimum overhead, but still).
  - Clear `data` field in every field payload and set encoding to some specific value such as `omitted`.

alexshtin:
@mfateev what do you think about this?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -440,6 +440,7 @@ message ListOpenWorkflowExecutionsRequest {
         temporal.api.filter.v1.WorkflowExecutionFilter execution_filter = 5;
         temporal.api.filter.v1.WorkflowTypeFilter type_filter = 6;
     }
+    bool metadata_only = 7;
```
mfateev, line 443: Why are we calling search attributes metadata?
Or we don't return the search attribute values?

temporal/api/workflow/v1/message.proto
mfateev: If we do this then this field should be part of the Memo structure itself in my opinion. 
Another option is to return memo_size instead of has_memo.

temporal/api/workflowservice/v1/request_response.proto
alexshtin: It is for `Memo` not for `SearchAttributes` and this is exactly the problem I pointed out in my [comment](https://github.com/temporalio/api/pull/117#pullrequestreview-688900085). It is not clear what is "metadata" here.

dnr:
> I would suggest the following:
> 
> 1. Rename `metadata_only` to `omit_memo` because it is not clear what is "metadata" in this case. Are `search_attributes` metadata? Or `auto_reset_points`? We will need to define "metadata" (which is already overloaded term).

Agreed, I like `omit_memo` better.

> 2. For response field, I have two options:
> 
> * Add `is_omitted` field to `Memo` struct. The cons is that we use same struct to persist memo and this field will go to db (minimum overhead, but still).

Sure, this works. I think having a field in Memo that only makes sense in certain contexts isn't a big deal. No extra data is persisted if it's not set.

> * Clear `data` and `metadata` fields in every memo field payload struct and add special key to `Payload.metadata` such as `omitted`. I would vote for this.

The main goal (as I understand it) is to avoid returning large unnecessary data to clients, but if we keep only a single bool present/not-present, I think we can avoid fetching it from the database also. Getting the field names would mean we have to fetch it and then discard it, right? Special key names also feels more clunky than a dedicated field in the proto.

Anyway, updated with new names/fields now.

mfateev:
I propose to limit the size of the metadata field to less than 1k to completely eliminate this problem. ElasticSearch or any other indexing engines are not designed as blob stores. A user can always store a link to a blob in a real blob store like S3 in the metadata field.

temporal/api/workflow/v1/message.proto
feedmeapples: How about the following design:

Requests:
1. Add a field to requests "max_payload" so that we can use it as a filter to avoid too large payloads. 
**Reasonings**: 
 - allows to still show `Payload`s that are smaller in size on UI without requiring users to do additional clicking (only actually big payloads will be hidden behind click placeholder) 
 - Instead of depending on server defaults, Allows clients to control how big payloads can be depending on the case
 - Gives flexibility to change the allowed payload size over future

Responses:
1. Always return `Payload` objects
2. Add a field `Size` to `Payload.Metadata equal to `size(Payload.Data`
2. Set `Payload.Data` to `nil` if `size(Payload.Data) > max_payload `

dnr:
(decided not to go in this direction)

PR #118, opened 2021-07-29
Add non deterministic as failed cause for workflow task

yiminc:
**What changed?**
Add new enum value to WorkflowTaskFailedCause for non deterministic error.

**Why?**
SDK could report non deterministic error properly, and server side could emit metrics accordingly. 

**How did you test it?**
eye_balls

**Potential risks**
No


temporal/api/enums/v1/failed_cause.proto
```diff
@@ -56,6 +56,7 @@ enum WorkflowTaskFailedCause {
     WORKFLOW_TASK_FAILED_CAUSE_BAD_BINARY = 21;
     WORKFLOW_TASK_FAILED_CAUSE_SCHEDULE_ACTIVITY_DUPLICATE_ID = 22;
     WORKFLOW_TASK_FAILED_CAUSE_BAD_SEARCH_ATTRIBUTES = 23;
+    WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR = 24;
```
wxing1292, line 59: WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC?

wxing1292 (approved):

Sushisource:
@yiminc I'm trying to find the issue/pr where we were talking about this but I couldn't - but we should really consider separating the causes that SDKs can produce from the causes server can produce. As it stands it's very confusing what can come from where.

So, now we have a value for this which is better than nothing, but we should try to address that problem as well long run, which might be hard to do in a backwards compat way.

Sushisource:
Ah, here: https://github.com/temporalio/temporal/pull/1726

PR #119, opened 2021-08-27
Add new_execution_run_id to some workflow completion events

dnr:
**What changed?**

Adding a field for completed/failed/timed-out workflow completion events to link to a next execution, in the case of retry or cron. (Currently canceled or terminated workflows can't have a next execution in that way, but if that changes we can add it there also.)

**Why?**

1. We want to be able to link to a potential next execution of retried or cron workflows once they're no longer using the continue-as-new event.
2. We need to know whether there is a next execution to tell whether to notify a parent workflow.

**How did you test it?**

**Potential risks**

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -57,6 +57,7 @@ message WorkflowExecutionStartedEventAttributes {
     google.protobuf.Duration workflow_run_timeout = 8 [(gogoproto.stdduration) = true];
     // Timeout of a single workflow task.
     google.protobuf.Duration workflow_task_timeout = 9 [(gogoproto.stdduration) = true];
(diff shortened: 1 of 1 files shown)


wxing1292 (approved):

yiminc (approved):

temporal/api/history/v1/message.proto
```diff
@@ -82,16 +83,22 @@ message WorkflowExecutionStartedEventAttributes {
 message WorkflowExecutionCompletedEventAttributes {
     temporal.api.common.v1.Payloads result = 1;
     int64 workflow_task_completed_event_id = 2;
+    // If another run is started by cron, this contains the new run id.
+    string new_execution_run_id = 3;
```
mfateev, line 87: What about next_run_id?

PR #127, opened 2021-11-08
Additional header support for signals and queries

cretz:
**What changed?**

One needed field for headers was added and one deprecated/removed.

**Why?**

#124 added headers for signals and queries, yet missed one and added one too many.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -549,9 +549,9 @@ message QueryWorkflowRequest {
     // QueryRejectCondition can used to reject the query if workflow state does not satisfy condition.
     // Default: QUERY_REJECT_CONDITION_NONE.
     temporal.api.enums.v1.QueryRejectCondition query_reject_condition = 4;
-    // Headers that are passed with the query to the processing workflow.
-    // These can include things like auth or tracing tokens.
-    temporal.api.common.v1.Header header = 5;
+
+    // Removal of unused, deprecated field
+    reserved 5;
 }
 
 message QueryWorkflowResponse {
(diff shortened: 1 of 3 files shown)


cretz:
The build breaks because I removed a field and marked its index as reserved. This field was unused. I will let reviewers determine if that's acceptable.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -549,9 +549,9 @@ message QueryWorkflowRequest {
     // QueryRejectCondition can used to reject the query if workflow state does not satisfy condition.
     // Default: QUERY_REJECT_CONDITION_NONE.
     temporal.api.enums.v1.QueryRejectCondition query_reject_condition = 4;
-    // Headers that are passed with the query to the processing workflow.
-    // These can include things like auth or tracing tokens.
-    temporal.api.common.v1.Header header = 5;
+
+    // Removal of unused, deprecated field
+    reserved 5;
```
wxing1292, line 554: why this header is removed?

cretz: Because `header` is already inside the `query` message passed. Sorry I did not make that more clear here in PR.

alexshtin: It was added recently, right? And there is no code which actually use it, right? Then I wouldn't use `reserved` and just remove it.

cretz: Works for me. Will do shortly.

alexshtin (approved):

temporal/api/workflowservice/v1/request_response.proto
cretz: Done

PR #134, opened 2021-12-13
New GetSystemInfo RPC

cretz:
**What changed?**

Added new `WorkflowService.GetSystemInfo` RPC call.

**Why?**

Needed for server to support namespace-based system info.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -739,6 +739,34 @@ message GetClusterInfoResponse {
...
+message GetSystemInfoRequest {
+    string namespace = 1;
+}
+
+message GetSystemInfoResponse {
+    // Version of the server.
+    string server_version = 1;
+
+    // Version info for the server.
+    temporal.api.version.v1.VersionInfo version_info = 2;
```
cretz, line 751: Per @sergeybykov - Removing this for now

Sushisource (approved):
Makes sense to me. Someone from server should approve too.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -739,6 +739,31 @@ message GetClusterInfoResponse {
     string visibility_store = 8;
 }
 
+message GetSystemInfoRequest {
+    string namespace = 1;
```
alexshtin, line 743: I don't like field `namespace` here. It looks weird to me. Or API should be called something like `GetNamespaceCapabilities`. I would change it either of these:
1. Add `Capabilities capabilities = 2;` to the `GetClusterInfo` and make it namespace independent, or
2. Add `Capabilities capabilities = 2;` to the `DescribeNamespace` and return const capabilities for now, which might be different for different namespaces in future.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -739,6 +739,31 @@ message GetClusterInfoResponse {
...
+
+message GetSystemInfoResponse {
+    // Version of the server.
+    string server_version = 1;
+
+    // All capabilities the system supports.
+    Capabilities capabilities = 2;
+
+    // System capability details.
+    message Capabilities {
```
alexshtin, line 754: We have `supported-features` header which is set of features that SDK supports. Isn't it an inconsistently to have "Server capabilities" and "SDK features"? I like word "capabilities" better. It seems like `supported-features` header is not used by SDKs and can be renamed to `supported-capabilities`.

From other side, there should be some distinguish between server and SDK capabilities because they are completely different sets. As far as this `message Capabilities` is inside `GetSystemInfoRespose` it is good enough, I guess.

temporal/api/workflowservice/v1/request_response.proto
alexshtin: For `server_version` SDK should be able to call `GetClusterInfo`.

cretz: > I don't like field `namespace` here. It looks weird to me.

`namespace` is present on many other calls like `ListOpenWorkflowExecutions`. This is consistent with those.

> Add `Capabilities capabilities = 2;` to the `GetClusterInfo` and make it namespace independent

I was told Temporal cloud SDK users cannot call this API.

> Add `Capabilities capabilities = 2;` to the `DescribeNamespace` and return const capabilities for now, which might be different for different namespaces in future

This is fine if others want this as well (it deviates from the internal proposal)

> For `server_version` SDK should be able to call `GetClusterInfo`.

I was told Temporal cloud SDK users cannot call this API, but arguably they shouldn't care about version.

temporal/api/workflowservice/v1/request_response.proto
cretz: > We have `supported-features` header which is set of features that SDK supports. Isn't it an inconsistently to have "Server capabilities" and "SDK features"? I like word "capabilities" better. It seems like `supported-features` header is not used by SDKs and can be renamed to `supported-capabilities`.

If consensus is reached here, I can change that, yes.

bergundy: We all agreed that the name capabilities is better.
I'd rename the `supported-features` header to `supported-capabilities` but since it's still unused it's not high priority.
Let's open an issue to track it.

temporal/api/workflowservice/v1/request_response.proto
bergundy: > Add Capabilities capabilities = 2; to the GetClusterInfo and make it namespace independent

As per internal discussion this is not a valid option for Cloud.

> Add Capabilities capabilities = 2; to the DescribeNamespace and return const capabilities for now, which might be different for different namespaces in future.

DescribeNamespace will eventually be deprecated and will be moved out of `WorkflowService`, so that's not a valid option.

bergundy: @alexshtin I feel like the time to make such general design comments was in the original proposal that was submitted over 3 weeks ago.

bergundy (approved):
Let's get this merged, I think we've iterated on this long enough.

yiminc (approved):

PR #135, opened 2021-12-17
Add DeleteWorkflowExecution API to operator service

alexshtin:
**What changed?**
Add `DeleteWorkflowExecution` API to `operatorservice`.

**Why?**
To add ability to delete workflow execution from the persistence.

**How did you test it?**
N/A

**Potential risks**
No risks.

### DIFF
--- a/api-linter.yaml
+++ b/api-linter.yaml
@@ -20,6 +20,8 @@
     - 'core::0132::request-unknown-fields'
     - 'core::0132::response-unknown-fields'
     - 'core::0134::request-unknown-fields'
+    - 'core::0135::request-name-required'
+    - 'core::0135::request-unknown-fields'
     - 'core::0158::request-page-size-field'
     - 'core::0158::request-page-token-field'
     - 'core::0158::response-next-page-token-field'
@@ -33,6 +35,8 @@
     - 'core::0127::http-annotation'
     - 'core::0131::method-signature'
     - 'core::0131::response-message-name'
+    - 'core::0135::method-signature'
+    - 'core::0135::response-message-name'
 
 - included_paths:
     - 'dependencies/gogoproto/gogo.proto'
--- a/api-linter.yaml
+++ b/api-linter.yaml
@@ -20,8 +20,6 @@
     - 'core::0132::request-unknown-fields'
     - 'core::0132::response-unknown-fields'
     - 'core::0134::request-unknown-fields'
-    - 'core::0135::request-name-required'
-    - 'core::0135::request-unknown-fields'
     - 'core::0158::request-page-size-field'
     - 'core::0158::request-page-token-field'
     - 'core::0158::response-next-page-token-field'
@@ -35,8 +33,6 @@
     - 'core::0127::http-annotation'
     - 'core::0131::method-signature'
     - 'core::0131::response-message-name'
-    - 'core::0135::method-signature'
-    - 'core::0135::response-message-name'
 
 - included_paths:
     - 'dependencies/gogoproto/gogo.proto'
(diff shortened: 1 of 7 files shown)


api-linter.yaml
```diff
@@ -19,6 +19,8 @@
     - 'core::0132::request-unknown-fields'
     - 'core::0132::response-unknown-fields'
     - 'core::0134::request-unknown-fields'
+    - 'core::0135::request-name-required'
```
alexshtin, line 22: These exceptions because google linter has strict rules for `Delete` methods like https://linter.aip.dev/135/response-message-name or https://linter.aip.dev/135/method-signature which we can't/don't want to follow.

wxing1292 (approved):

temporal/api/workflowservice/v1/service.proto
```diff
@@ -250,6 +250,12 @@ service WorkflowService {
     rpc TerminateWorkflowExecution (TerminateWorkflowExecutionRequest) returns (TerminateWorkflowExecutionResponse) {
     }
 
+    // DeleteWorkflowExecution delete closed workflow execution (workflow must be completed or terminated before) and its history synchronously.
+    // Visibility record is deleted asynchronously.
+    // If call to DeleteWorkflowExecution fails, it must be retried to prevent orphan DB records.
```
yiminc, line 255: I don't like this.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -250,6 +250,12 @@ service WorkflowService {
     rpc TerminateWorkflowExecution (TerminateWorkflowExecutionRequest) returns (TerminateWorkflowExecutionResponse) {
     }
 
+    // DeleteWorkflowExecution delete closed workflow execution (workflow must be completed or terminated before) and its history synchronously.
```
yiminc, line 253: Why cannot delete open workflow? Can we terminate it internally if it is still open?

alexshtin: I plan to implement this logic in `tctl` command. It will query workflow first and if it is still running ask if user wants to terminate it first. This interactive experience is not possible with single API.

temporal/api/workflowservice/v1/service.proto
alexshtin: We can't avoid this. Delete itself is multi stage operation. If first step went through and second failed, what can we do? Retry. And there will be retries for transient errors but if it is not transient (db completely down) I can't do anything. The other possible option would be to implement entire delete as async which I don't want to do.

temporal/api/workflowservice/v1/service.proto
mfateev: Should we add `terminateIfOpen` flag to the DeleteWorkflowExecutionRequest then?

Or we should rename the call to `DeleteClosedWorkflowExecution`

temporal/api/workflowservice/v1/service.proto
mfateev: I think we must leave DB in consistent state always. We cannot rely on an external retry.

temporal/api/workflowservice/v1/service.proto
alexshtin: I like the idea of renaming it to `DeleteClosedWorkflowExecution`.

temporal/api/workflowservice/v1/service.proto
yiminc: My point is I don't like the comment that it require callers to retry on errors. We cannot rely on that. We need some other means to make sure DB will eventually be consistent.

temporal/api/workflowservice/v1/service.proto
yiminc: I would rather this logic handled by the API internally. The end result is the same, but user experience is very different. It would be a much nicer to be able to delete without having to query its status and terminate it if it is still open.

yiminc (approved):

PR #139, opened 2022-01-25
Add PendingWorkflowTaskInfo in DescribeWorkflowExecutionResponse

yycptt:
**What changed?**
- Add PendingWorkflowTaskInfo in DescribeWorkflowExecutionResponse

**Why?**
- Increase visibility into workflow task retrying. Current when workflow task is keep retrying, there's no progress in workflow history. There's also no info about it when describing the workflow. It's hard for user to know if workflow is making progress or not (workflow task is retrying or not)

**How did you test it?**

**Potential risks**


temporal/api/workflow/v1/message.proto
```diff
@@ -89,6 +89,14 @@ message PendingChildExecutionInfo {
     temporal.api.enums.v1.ParentClosePolicy parent_close_policy = 5;
 }
 
+message PendingWorkflowTaskInfo {
+    temporal.api.enums.v1.PendingWorkflowTaskState state = 1;
+    google.protobuf.Timestamp scheduled_time = 2 [(gogoproto.stdtime) = true];
+    google.protobuf.Timestamp original_scheduled_time = 3 [(gogoproto.stdtime) = true];
```
Sushisource, line 95: Since it may not immediately be obvious what the distinction is, I would love it if we added some brief comments to these.

I went through all these protos and added comments to as many not-immediately-clear things as I could a while back and it'd be awesome to keep that up. Thanks!

wxing1292 (approved):

temporal/api/workflow/v1/message.proto
yycptt: Sure. Just added the comment for that field.

PR #147, opened 2022-01-27
Add GetWorkflowHistoryReverse API

Ardagan:
**What changed?**
Add fields to support retrieving workflow history in reverse order.

**Why?**
More efficient way to query large histories.

**How did you test it?**
Manual, Automated tests

**Potential risks**
N/A

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -202,7 +202,6 @@ message GetWorkflowExecutionHistoryReverseResponse {
     temporal.api.history.v1.History history = 1;
     // Will be set if there are more history events than were included in this response
     bytes next_page_token = 3;
-    bool archived = 4;
 }
 
 message PollWorkflowTaskQueueRequest {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -195,10 +195,7 @@ message GetWorkflowExecutionHistoryReverseRequest {
     string namespace = 1;
     temporal.api.common.v1.WorkflowExecution execution = 2;
     int32 maximum_page_size = 3;
-    // If a `GetWorkflowExecutionHistoryResponse` or a `PollWorkflowTaskQueueResponse` had one of
-    // these, it should be passed here to fetch the next page.
     bytes next_page_token = 4;
-    bool skip_archival = 7;
 }
 
 message GetWorkflowExecutionHistoryReverseResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -191,6 +191,23 @@ message GetWorkflowExecutionHistoryResponse {
     bool archived = 4;
 }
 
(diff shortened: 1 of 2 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -642,4 +642,5 @@ message HistoryEvent {
 
 message History {
     repeated HistoryEvent events = 1;
+    int64 transaction_id = 2;
```
yiminc, line 645: I don't think we need this to be exposed in public api.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -191,6 +191,23 @@ message GetWorkflowExecutionHistoryResponse {
...
 }
 
+message GetWorkflowExecutionHistoryReverseRequest {
+    string namespace = 1;
+    temporal.api.common.v1.WorkflowExecution execution = 2;
+    int32 maximum_page_size = 3;
+    // If a `GetWorkflowExecutionHistoryResponse` or a `PollWorkflowTaskQueueResponse` had one of
+    // these, it should be passed here to fetch the next page.
+    bytes next_page_token = 4;
+    bool skip_archival = 7;
```
cretz, line 201: Intentionally keeping same index for `skip_archival` as `GetWorkflowExecutionHistoryRequest`?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -94,6 +94,12 @@ service WorkflowService {
     // `NotFound` if the specified workflow execution is unknown to the service.
     rpc GetWorkflowExecutionHistory (GetWorkflowExecutionHistoryRequest) returns (GetWorkflowExecutionHistoryResponse) {
     }
+    
+    // GetWorkflowExecutionHistoryReverse returns the history of specified workflow execution in reverse 
+    // order (starting from last event). Fails with`NotFound` if the specified workflow execution is 
+    // unknown to the service.
+    rpc GetWorkflowExecutionHistoryReverse (GetWorkflowExecutionHistoryReverseRequest) returns (GetWorkflowExecutionHistoryReverseResponse) {
```
cretz, line 101: Is this not clearer as just a boolean on `GetWorkflowExecutionHistoryRequest` (and each successive page request after)? Could even embed the direction in the token (even if just for verifying the bool is the same/expected on each page call, not necessarily for affecting the direction).

temporal/api/workflowservice/v1/request_response.proto
Ardagan: Yes. Keeping IDs compatible will not make it worse.

temporal/api/workflowservice/v1/service.proto
Ardagan: Getting history in reverse has less functionality, performance and is effectively a separate code path at this point. It is cleaner to have a separate api.

yiminc: cc @stevekinney

temporal/api/workflowservice/v1/request_response.proto
Ardagan: you are correct.

yycptt (approved):

PR #150, opened 2022-02-08
Add details to TerminatedFailureInfo

lorensr:
**Why?**

It's present in `WorkflowExecutionTerminatedEventAttributes`:

https://github.com/temporalio/api/blob/ed7d7427692cada52fc5915925ca2c7cde26d649/temporal/api/history/v1/message.proto#L388

**How did you test it?**

Ran `make`

**Potential risks**

Adding an optional field is backcompat

### DIFF
--- a/temporal/api/failure/v1/message.proto
+++ b/temporal/api/failure/v1/message.proto
@@ -50,6 +50,7 @@ message CanceledFailureInfo {
 }
 
 message TerminatedFailureInfo {
+    temporal.api.common.v1.Payloads details = 1;
 }
 
 message ServerFailureInfo {


cretz:
I think there may need to be an accompanying server issue to actually populate this field. We may learn that it was absent by intention.

yiminc (approved):

alexshtin:
I think it wasn't added because it wasn't needed. Why do you need it? What's about `reson` field? I would like to see other PRs (server and SDKs) linked to this one, before you merge this.

lorensr:
I don't have a need for it. It would be useful if a string wasn't enough for a user—if they wanted to attach non-string data to a termination, so that a client that gets a `WorkflowFailedError` with `TerminatedFailure` could see the data via `error.cause.details`. We can leave open or close until a user has such a request

yiminc:
close for now, will revisit when there is real use case.

PR #151, opened 2022-02-11
Add first_execution_run_id to missing requests and responses

bergundy:
As discussed on slack, we need `first_execution_run_id` in all of these requests to make them safer.
Also added this to `SignalWithStartWorkflowExecutionResponse` to make using stubs / handles created this way safer.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -519,6 +519,10 @@ message SignalWorkflowExecutionRequest {
     // Headers that are passed with the signal to the processing workflow.
     // These can include things like auth or tracing tokens.
     temporal.api.common.v1.Header header = 8;
+    // If set, this call will error if the most recent (if no run id is set on
+    // `workflow_execution`), or specified (if it is) workflow execution is not part of the same
+    // execution chain as this id.
+    string first_execution_run_id = 9;
 }
 
 message SignalWorkflowExecutionResponse {
@@ -560,6 +564,10 @@ message SignalWithStartWorkflowExecutionRequest {
 
 message SignalWithStartWorkflowExecutionResponse {
     string run_id = 1;
+    // The first run_id of the current execution chain.
+    // This can be used in calls to signal, query, terminate, and cancel this Workflow to ensure
+    // that those requests only affect this execution chain.
+    string first_execution_run_id = 2;
 }
 
 message ResetWorkflowExecutionRequest {
@@ -708,6 +716,10 @@ message QueryWorkflowRequest {
     // QueryRejectCondition can used to reject the query if workflow state does not satisfy condition.
     // Default: QUERY_REJECT_CONDITION_NONE.
     temporal.api.enums.v1.QueryRejectCondition query_reject_condition = 4;
+    // If set, this call will error if the most recent (if no run id is set on
+    // `workflow_execution`), or specified (if it is) workflow execution is not part of the same
+    // execution chain as this id.
+    string first_execution_run_id = 5;
 }
 
 message QueryWorkflowResponse {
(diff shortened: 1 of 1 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -519,6 +519,10 @@ message SignalWorkflowExecutionRequest {
     // Headers that are passed with the signal to the processing workflow.
     // These can include things like auth or tracing tokens.
     temporal.api.common.v1.Header header = 8;
+    // If set, this call will error if the most recent (if no run id is set on
+    // `workflow_execution`), or specified (if it is) workflow execution is not part of the same
+    // execution chain as this id.
```
cretz, line 524: Is not part of the same execution chain, or explicitly if this was not the first in the chain? Can you clarify the doc a bit? Alternative description/approach:

> If set, this will limit the workflow signaled to be in a chain of workflows started with this run ID. This check occurs regardless of `workflow_execution.run_id`, therefore, an error will occur if this is present and `workflow_execution.run_id` is not in that chain.

I presume that's how the `cancel` and `terminate` work too?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -519,6 +519,10 @@ message SignalWorkflowExecutionRequest {
...
+    string first_execution_run_id = 9;
```
cretz, line 525: Hrmm, I would have figured this could be used to get the latest for `workflow_execution.workflow_id` instead of just validate.

Is this just for validation? Does that mean there is no reasonable way to say "signal the latest workflow in this chain"? I think that makes our SDK a bit rough. Basically you'll tell people "If you get this error, it means you started a newer workflow of the same workflow ID. Therefore you _somehow_ have to find the run ID for this specific one and go recreate a handle."

cretz (commented):
Would rather not merge this until we get the server impl at least POC'd to confirm no more changes/quirks are needed.

temporal/api/workflowservice/v1/request_response.proto
bergundy: > Does that mean there is no reasonable way to say "signal the latest workflow in this chain"?

To signal the latest, pass in `workflow_id` and `first_execution_run_id`.

Creating a new handle won't change the fact that the current execution chain has already completed.
You can't signal a non-running workflow.

> Therefore you somehow have to find the run ID for this specific one and go recreate a handle.

That's what we want, we're adding this for safety.

cretz: > To signal the latest, pass in workflow_id and first_execution_run_id.

So it's not for validation only, it's also used to root the chain? If so, that's perfect. I was just concerned this only validated that the lasted for the workflow ID started with this run, but if it actually uses this run as a root for the chain, perfect.

bergundy: yes, IIUC

temporal/api/workflowservice/v1/request_response.proto
yiminc, line 524: I found this comment hard to read. Better to break it into 2 sentences. For example:
If set, this call will error if the target workflow is not part of the same execution chain as this id. Target workflow is specified by `workflow_execution` if run_id is set there, or the most recent run if run_id is not set on `workflow_execution`.

yiminc (approved):

macrogreg (changes requested):
The PR only addresses a part of the "workflow chain overflow" scenario.
I think we need to hold off on merging it and instead address it holistically with the part of returning the "workflow chain IDs" back to the user.

See here for details: https://github.com/temporalio/temporal/issues/2691

bergundy:
Yes, I agree @macrogreg.
I didn't want to merge this before verifying the server can implement these API changes anyways.
Let's see if we can get this prioritized.

bergundy:
We're not doing this anytime soon.

PR #152, opened 2022-02-22
Introducing generic SystemWorkflowFailure

Spikhalskiy:
Rename AddSearchAttributesFailure into SystemWorkflowFailure, so this error can be reused for other endpoints that use underlying workflows

### DIFF
--- a/temporal/api/errordetails/v1/message.proto
+++ b/temporal/api/errordetails/v1/message.proto
@@ -81,11 +81,10 @@ message ResourceExhaustedFailure {
     temporal.api.enums.v1.ResourceExhaustedCause cause = 1;
 }
 
-message AddSearchAttributesFailure {
-    // WorkflowId and RunId of the Temporal system workflow performing registration of the search attributes.
+message SystemWorkflowFailure {
+    // WorkflowId and RunId of the Temporal system workflow performing the underlying operation.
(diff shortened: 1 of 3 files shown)


temporal/api/errordetails/v1/message.proto
```diff
@@ -81,11 +81,10 @@ message ResourceExhaustedFailure {
...
+message SystemWorkflowFailure {
+    // WorkflowId and RunId of the Temporal system workflow performing the underlying operation.
     // Looking up the info of the system workflow run may help identify the issue causing the failure.
-    // WorkflowId is always "temporal-sys-add-search-attributes-workflow".
-    temporal.api.common.v1.WorkflowExecution system_workflow_execution = 1;
-    // Serialized error returned by the system workflow performing registration
-    string system_workflow_error = 2;
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 1;
+    // Serialized error returned by the system workflow performing the underlying operation.
+    string workflow_error = 2;
```
Sushisource, line 89: It might make sense for this to be a `oneof` with specific types, so that specific errors can include well-typed details that can be matched against if the need ever arises.

alexshtin: I think it is too much for now/

alexshtin (approved):

Issue #154, opened 2022-03-03
Document SearchAttributes parsing


lorensr:
### Is your feature request related to a problem? Please describe.

`SearchAttributes`:

https://github.com/temporalio/api/blob/148f6186cf063ddac5600aa53549ec8c0afe5eaa/temporal/api/common/v1/message.proto#L58-L62

I don't know from reading this proto how to specify search attribute types in `StartWorkflowExecution`. Possible types are:

https://github.com/temporalio/api/blob/df58bcd5eb85434a8dc5561d55e597d0ad031fc5/temporal/api/enums/v1/common.proto#L40-L48

![image](https://user-images.githubusercontent.com/251288/156648148-d3771749-e72f-41a0-8d6d-13b81994cf45.png)

https://docs.temporal.io/docs/concepts/what-is-a-search-attribute/#custom-search-attributes

### Describe the solution you'd like

A comment explaining how the Payloads are parsed, for example:

// Each Payload has a `metadata.type: IndexedValueType`
// If `payload.metata.type` is `INDEXED_VALUE_TYPE_DATETIME`, `payload.data` should be an integer with milliseconds since epoch.

PR #155, opened 2022-03-03
Specify that Datetime values should be ISO strings

lorensr:
Fixes https://github.com/temporalio/api/issues/154

(I think Datetime representation is the only ambiguity—the rest isn't needed, since precise field type info is only needed in `AddSearchAttributesRequest`)

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -56,6 +56,7 @@ message Payload {
 }
 
 // A user-defined set of *indexed* fields that are used/exposed when listing/searching workflows.
+// Datetime fields are represented as ISO strings.
 // The payload is not serialized in a user-defined way.
 message SearchAttributes {
     map<string, Payload> indexed_fields = 1;
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -56,7 +56,7 @@ message Payload {
 }
 
 // A user-defined set of *indexed* fields that are used/exposed when listing/searching workflows.
-// Datetime fields are represented as ISO strings.
+// Datetime fields are represented as RFC 3339 strings.
 // The payload is not serialized in a user-defined way.
 message SearchAttributes {
     map<string, Payload> indexed_fields = 1;


temporal/api/common/v1/message.proto
```diff
@@ -56,6 +56,7 @@ message Payload {
 }
 
 // A user-defined set of *indexed* fields that are used/exposed when listing/searching workflows.
+// Datetime fields are represented as ISO strings.
```
cretz, line 59: I think there is even more clarity than this needed. ISO8601 has multiple string formats (not to be confused w/ RFC3389 and other additions many things make to those). Can you specify exact format here? If unsure, we can start with where they are converted to date (Go? ES?) and then get the exact format from there.

lorensr: I think Go does

```go
var val time.Time
json.Unmarshal(payload.GetData(), &val)
```

image

and `json.Unmarshal` calls this? 

https://pkg.go.dev/time#Time.UnmarshalJSON

which uses an RFC 3339 string. I'll edit to say that.

Other research, for posterity:


image

On ES side, Server has a v7 client and v6 client. v7 is [date_nanos](https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html), and v6 is [date](https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html):
image

v6:
```go
		case enumspb.INDEXED_VALUE_TYPE_DATETIME:
			typeMap = map[string]interface{}{"type": "date"}
		}
```

Neither appears to customize format, so 

- default `date_nanos` format: `"strict_date_optional_time_nanos||epoch_millis"`
image
image

- default `date` format:   `"strict_date_optional_time||epoch_millis"`

image

What's working for me in `tctl` and web UI:

```
tctl workflow list --query 'StartTime < "2022-03-04T16:11:33.388Z"'
tctl workflow list --query 'StartTime < "2022-03-04T16:11:33.388000Z"'
tctl workflow list --query 'StartTime < "1646410293388"'
```

Doesn't work:

```
tctl workflow list --query 'StartTime < 1646410293388'
```

Works for me in TS SDK:

```
"2022-03-04T16:11:33.388Z"
"2022-03-04T16:11:33.388000Z"
```

Doesn't work:

```
1646410293388 // gRPC error '1.646410293388e+12 is not a valid value for search attribute CustomDatetimeField of type Datetime',
"1646410293388" // gRPC Error '1646410293388 is not a valid value for search attribute CustomDatetimeField of type Datetime',
```

cretz: Does `2001-02-03T04:05:06+0000` which is a perfectly valid ISO-8601 string work?

It does not in Go using `RFC3339` format. If this does not work, we need to document that this is not exactly ISO-8601. Notice https://docs.temporal.io/visibility states:

> a string in [RFC3339Nano format](https://pkg.go.dev/time#pkg-constants) (such as "2006-01-02T15:04:05.999999999Z07:00").

Need to clarify the same in docs here if it applies.

PR #156, opened 2022-03-05
Adding failures to RespondActivityTaskFailureResponse

jbreiding:
**What changed?**
Addition of failures to response object

**Why?**
Indicate level of success achieved by request, include any failures sent upstream back downstream.

**How did you test it?**
Will be tested in temporal repo

**Potential risks**
None

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -438,6 +438,9 @@ message RespondActivityTaskFailedRequest {
 }
 
 message RespondActivityTaskFailedResponse {
+    // Server validation failures could include
+    // last_heartbeat_details payload is too large, request failure is too large
+    repeated temporal.api.failure.v1.Failure failures = 1;
 }
 
 message RespondActivityTaskFailedByIdRequest {
@@ -458,6 +461,9 @@ message RespondActivityTaskFailedByIdRequest {
 }
 
 message RespondActivityTaskFailedByIdResponse {
+    // Server validation failures could include
+    // last_heartbeat_details payload is too large, request failure is too large
+    repeated temporal.api.failure.v1.Failure failures = 1;
 }
 
 message RespondActivityTaskCanceledRequest {


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -438,6 +438,10 @@ message RespondActivityTaskFailedRequest {
 }
 
 message RespondActivityTaskFailedResponse {
+    // Will be set to true if the activity last heartbeat details was updated.
+    bool last_heartbeat_details_updated = 1;
+    // Will be set to true if the failure was truncated due to size constraints
+    bool failure_truncated = 2;
```
Sushisource, line 444: Might make sense to break these out into an `ActivityTaskFailureDetails` message or something so they can be de-duped.

On the comments - it would be good to list why the last heartbeat details might not be updated - them not being sent in the request is clear, and I assume them being too large is the other case?

And is `failure_truncated` talking about the `failure` itself from the request?

Lastly - how do I know the heartbeat details weren't updated _because_ it was too large? Might make sense to do a `last_heartbeat_details_not_updated_reason` field instead, which could be typed as an enum.

jbreiding: Good points. Ack on struct over separate booleans.

The failure is the failure itself from the request being truncated.

The only reason the heartbeat details wouldn't be updated is the payload being too large, no other reason.

I suppose then `last_heartbeat_details_updated` should become `last_heartbeat_details_too_large`?

jbreiding: Or maybe instead of booleans there should be failure objects, since the api handler constructs a failure object to send upstream.

That same failure object can be returned to downstream?

jbreiding: Updated to have the failure object in the response object.

temporal/api/workflow/v1/message.proto
```diff
@@ -117,3 +117,8 @@ message ResetPointInfo {
     // false if the reset point has pending childWFs/reqCancels/signalExternals.
     bool resettable = 6;
 }
+
+message ActivityTaskFailureDetails {
```
cretz, line 121: I wonder if the workflow package is the best for this. Seems all other `RespondActivityX` stuff is inline in the request/response message. Not sure there is value in breaking this off, and even if so, not sure this is the best package.

jbreiding: This seemed to follow the pattern in use. 
request/response proto had only top level request & response objects.

Other messages not directly tied to a service were included in a file like this one.

jbreiding: Maybe I misunderstood the de-dupe here?

https://github.com/temporalio/api/pull/156#discussion_r820001960

Sushisource: I think this follows the pattern and that's fine. The pattern itself doesn't exactly make a ton of sense to me, but that's a different story.

The value in making this a separate message is that if there are more details that need to be added later to activity failure related things we ensure they get added here rather than manually keeping the different task failure responses in sync.

temporal/api/workflow/v1/message.proto
```diff
@@ -117,3 +117,8 @@ message ResetPointInfo {
...
+    temporal.api.failure.v1.Failure request_failure_details = 1;
+    temporal.api.failure.v1.Failure request_heartbeat_details = 2;
```
Sushisource, line 123: `details` postfix here doesn't tell me much.

I almost want `request_failure_failure` and `request_heartbeat_failure` but the first reads funny obviously. Maybe `request_details_failure` for the first.

Sushisource (approved):
This makes sense to me. Naming details not a huge concern.

Probably worth adding per-field comments to `ActivityTaskFailureDetails` too, explaining what populates them.

Thanks!

temporal/api/workflow/v1/message.proto
cretz: Agreed about the de-dupe separate message thing, I was just concerned about it being in the "workflow" package when it's completely unrelated to workflows.

jbreiding: Okay, I see your point and better understand the usage concern.
I can extract this message into the workflow service request/response.

On the surface it seemed similar enough to
https://github.com/temporalio/api/blob/60c7fe0fb3f45c94985803ed610d18d708b303ba/temporal/api/workflow/v1/message.proto#L68

But its really only related to the request/response for workflowservice.

jbreiding: @cretz does this fit better?
Updated.

temporal/api/workflow/v1/message.proto
jbreiding: how about just
`request_failure`
`request_last_heartbeat_details`

This tracks back to the request field name?

jbreiding: so usage at call site might look like
```
resp.GetDetails().GetRequestFailure() != nil
resp.GetDetails().GetLastHeartbeatDetails() != nil

PR #157, opened 2022-03-09
Add Schedule messages

dnr:
**What changed?**
Add messages for Schedule and related requests. GRPC interfaces will come in the next PR.

**Why?**
These are how the SDKs and frontends will communicate with the new schedule service.


temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    // If set, trigger one action immediately, overriding the schedule description.
+    // Overrides paused.
+    // Automatically unset after being used once.
+    bool trigger_immediately = x;
+
+    // If set, runs though the specified time period(s) and takes actions as if that time
+    // passed by right now, all at once.
+    // Overrides paused.
+    // Automatically unset after being used once.
+    repeated BackfillRequest backfill_request = x;
```
dnr, line 181: One idea here would be to make trigger immediately and backfill into separate grpc apis, instead of having them here and using the Update api. The benefit of having them here is is that you can create a schedule, have it backfill some runs, and/or run immediately, in a single Create call. Having separate apis would be more explicit, though.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    ScheduleDescription description = x;
+    ScheduleAction action = x;
+    SchedulePolicies policies = x;
+    ScheduleState state = x;
+}
+
+// ScheduleListEntry is returned by ListSchedules.
+message ScheduleListEntry {
+    string schedule_id = x;
+    // TODO: how much can we get from visibility?
```
dnr, line 205: This part needs more work. What values do people want to see here?

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,112 @@
...
+    // The namespace the schedule should be created in.
+    string namespace = x;
+    // The id of the new schedule.
+    string schedule_id = x;
+    temporal.api.schedule.v1.Schedule schedule = x;
+    // The identity of the client who initiated this request.
+    string identity = x;
+    // A unique identifier for this create request. Typically UUIDv4.
+    string request_id = x;
+}
```
dnr, line 46: Do we want headers here too? We could propagate them to the created workflows, but I'm not sure if that's desired.

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,112 @@
...
+    // The namespace of the schedule to update.
+    string namespace = x;
+    // The id of the schedule to update.
+    string schedule_id = x;
+    // The new schedule. A Schedule has four fields (description, action, policies,
+    // state), each of which can be present or absent from this message. If a field is
+    // absent here, it is not modified. If it's present, it's replaced entirely.
+    // This makes it easier to change state fields like "paused" or "trigger_immediately"
+    // without modifying the rest of the schedule.
+    temporal.api.schedule.v1.Schedule schedule = x;
```
dnr, line 76: This is a sort of hybrid semantics in between "replace all" and "patch fields that are present only". Is it better to just go with full replace all? Or try to figure out how to represent a patch (including un-setting fields)?

temporal/api/enums/v1/schedule.proto
```diff
@@ -0,0 +1,62 @@
...
+    // completes, the next scheduled event after that time will be considered.
+    OVERLAP_SKIP = 1;
+    // OVERLAP_BUFFER_ONE means start the workflow again soon as the current one
+    // completes, but only buffer one start in this way. If another start is
+    // supposed to happen when the workflow is running, and one is already
+    // buffered, then only one will be started after this one finishes.
+    OVERLAP_BUFFER_ONE = 2;
+    // OVERLAP_BUFFER_ALL means buffer up any number of starts to all happen
+    // immediately after the running workflow completes.
+    OVERLAP_BUFFER_ALL = 3;
```
yiminc, line 49: Are they run in parallel or sequential? Should we give user choice?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+// similar to a traditional cron specification. Each field can be one of:
+//   *: matches always
+//   x: matches when the field equals x
+//   x/y : matches when the field equals x+n*y where n is an integer
+//   x-z: matches when the field is between x and z inclusive
+//   w,x,y,...: matches when the field is one of the listed values
+// Each x, y, z, ... is either a decimal integer, or a month or day of week name
+// or abbreviation (in the appropriate fields).
+// A second in time matches if all fields match.
+// day_of_month uses 0 or 7 as Sunday
```
yiminc, line 45: do you mean day_of_week

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    string minute = x;           // default: 0
+    string hour = x;             // default: 0
+    string day_of_month = x;     // default: *
+    string month = x;            // default: *
+    string year = x;             // default: *
+    string day_of_week = x;      // default: *
+}
+
+// IntervalSpec matches times that can be expressed as:
+// epoch + n * interval + phase
```
yiminc, line 59: A concrete example would help

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    google.protobuf.Duration phase = x [(gogoproto.stdduration) = true];
+}
+
+// ScheduleDescription is a complete description of a set of absolute timestamps
+// (possibly infinite) that an action should occur at. The meaning of a
+// ScheduleDescription depends only on its contents and never changes, except
+// that the definition of a time zone can change over time (most commonly, when
+// daylight saving time policy changes for an area). To create a totally
+// self-contained ScheduleDescription, use UTC or include timezone_data.
+message ScheduleDescription {
```
yiminc, line 75: name suggestion: ScheduleSpec.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    //
+    // If your application requires more control over the time zone definition
+    // used, it may pass in a complete definition in the form of a TZif file
+    // from the time zone database. If present, this will be used instead of
+    // loading anything from the environment. You are then responsible for
+    // updating timezome_data when the definition changes.
+    //
+    // There is _no special handling_ of DST: if you write a calendar spec that
+    // fires at 2:30am and specify a time zone that follows DST, that action may
+    // be triggered zero, one, or two times in a day.
```
yiminc, line 108: nit: maybe say it explicitly that is on the days when time adjust for DST.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    string timezone_name = x;
+    bytes timezone_data = x;
+}
+
+message SchedulePolicies {
+    // Policy for overlaps:
+    temporal.api.enums.v1.ScheduleOverlapPolicy overlap_policy = x;
+
+    // Policy for catchups:
+    // If the system misses an action due to being down, and comes back up,
```
yiminc, line 118: the system is 'temporal server' or 'user worker'? From user's POV, what is the difference?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    // SignalWorkflowExecutionRequest apply here. The exceptions are:
+    // - namespace must be omitted or match the namespace of the schedule.
+    // - identity and request_id must be omitted (they will be set by the schedule).
+    // - control must be omitted.
+    temporal.api.workflowservice.v1.SignalWorkflowExecutionRequest signal_workflow = x;
+
+    // TODO: should we add signal with start also?
+
+    // If append_time_argument is true, append the nominal trigger time of each
+    // schedule as an extra input argument to start_workflow and signal_workflow.
```
yiminc, line 150: I don't understand. How do you pass extra input argument to start/signal workflow?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    //   WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE.
+    // - cron_schedule must be omitted.
+    temporal.api.workflowservice.v1.StartWorkflowExecutionRequest start_workflow = x;
+
+    // If present, signal a workflow. Most fields of
+    // SignalWorkflowExecutionRequest apply here. The exceptions are:
+    // - namespace must be omitted or match the namespace of the schedule.
+    // - identity and request_id must be omitted (they will be set by the schedule).
+    // - control must be omitted.
+    temporal.api.workflowservice.v1.SignalWorkflowExecutionRequest signal_workflow = x;
```
yiminc, line 145: Is there a requirement for this? I thought we are only going to support StartWorkflow as action for now. Later we might add StartActivity as another action. But I did not know we plan to have signal as action.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    // Automatically unset after being used once.
+    bool trigger_immediately = x;
+
+    // If set, runs though the specified time period(s) and takes actions as if that time
+    // passed by right now, all at once.
+    // Overrides paused.
+    // Automatically unset after being used once.
+    repeated BackfillRequest backfill_request = x;
+
+    // Information-only fields (read-only, ignored on create/update):
```
yiminc, line 183: I think we should break ScheduleState into 2 message types, one for displaying info, another for updating.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+message Schedule {
+    ScheduleDescription description = x;
+    ScheduleAction action = x;
+    SchedulePolicies policies = x;
+    ScheduleState state = x;
+}
+
+// ScheduleListEntry is returned by ListSchedules.
+message ScheduleListEntry {
+    string schedule_id = x;
```
yiminc, line 204: what is a schedule_id? do user specify schedule_id when they create a schedule, or do system return one on create?

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,112 @@
...
+option ruby_package = "Temporal::Api::ScheduleService::V1";
+option csharp_namespace = "Temporal.Api.ScheduleService.V1";
+
+import "temporal/api/schedule/v1/message.proto";
+
+message CreateScheduleRequest {
+    // The namespace the schedule should be created in.
+    string namespace = x;
+    // The id of the new schedule.
+    string schedule_id = x;
```
yiminc, line 40: is this akin to WorkflowID? What happen if id collide?

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,112 @@
...
+    temporal.api.schedule.v1.Schedule schedule = x;
+    // The identity of the client who initiated this request.
+    string identity = x;
+    // A unique identifier for this create request. Typically UUIDv4.
+    string request_id = x;
```
yiminc, line 45: Can this be non UUID? Any length limit?

temporal/api/scheduleservice/v1/request_response.proto
yiminc: The update namespace tried to do the "patch if present" way, and it is not looking good. 
I would prefer a simple replace all approach. User can get the current value and update fields as needed if they want to only patch for some fields. But we need some kind of version number to make sure update only overwrite the right one.

temporal/api/schedule/v1/message.proto
cretz, line 75: Might need to make clear that only positive times are allowed

temporal/api/schedule/v1/message.proto
cretz: Yeah, there are several complications here w/ error handling (i.e. absence of workflow), whether start and signal in the same action constitute schedule-with-start or are separate actions, how workflow ID is interpreted, what overlap policy means for signal-only actions, etc.

temporal/api/schedule/v1/message.proto
cretz: Sounds like a JSON-string appended payload to input of either action. The problem here is that payloads are often encoded in various opaque ways client-side and therefore server-side appending is not feasible.

We offer the user the event time of `WorkflowTaskStarted` which is probably good enough here. If not, we need to provide another way to get this metadata to them. I don't think we can mess with payloads.

temporal/api/scheduleservice/v1/request_response.proto
cretz: Same.

Patch is incredibly confusing and difficult to get right, especially on scalars/repeated with how protobuf handles defaults. You essentially would mimic protobuf's proto merging feature which is very limited with regards to collection item deletion and other things. It's also difficult to understand whether the absence of a value represents removal or ignore.

Always go with replacement as much as possible (for many reasons) and ask that users are smart enough to choose what to change. If there are concerns about atomicity of replacement and there are good use cases for updating individual values without fear of overwriting something unintentional with full-replacement, I would strongly suggest completely separate proto messages defining exact items to update explicitly.

cretz, line 76: This should make clear what values are mutable and what values are system-controlled and therefore ignored. One might not expect to be able to mutate `remaining_actions`. As much as possible there should be no fields that are both user-mutable and system-mutable to avoid surprising update conflicts.

Even better if you can take the k8s-style approach and separate a user-mutable "spec" from its system-mutable "status"/"state" and not try to combine them into one proto message.

cretz (commented):
Nobody really asked for my review, but I gave it anyways :-)

temporal/api/scheduleservice/v1/request_response.proto
yiminc: The challenge is how to expose it in user friendly way through tctl.

temporal/api/schedule/v1/message.proto
dnr: No requirement, but it's pretty trivial to add and implement, so I thought we might as well. Mostly as an illustration of how to add additional actions.

If it doesn't make much sense, I can remove it.

dnr: I don't think it's that complicated:
Errors: there may be errors, like anywhere else. The errors should show up in server logs, and we'll also have metrics. Do we need a way to get errors through grpc also?
Signal-with-start: no, if we want that we can add a new action for it.
Workflow ID: how else could it be interpreted?
Overlap: signals are instantaneous so no overlap applies. Note that the comments on overlap policy specifically refer to starting workflows, not to actions in general.

temporal/api/schedule/v1/message.proto
dnr: Payloads have metadata attached to describe how they are encoded/transformed, right? So why can't we add one that specifies it's encoded as json, with no other transformations. User dataconverters should pass through payloads if they don't have metadata indicating that they were converted with the corresponding dataconverter. Sure, we can't guaranteee they're written that way. But in that case the user will get an error and can fix their dataconverter, right?

As I understand it, the default decoders for all the sdks should be able to understand a json string.

dnr: I'd like to provide the nominal schedule time (before jitter, system delays, retries, etc.) to the workflow. The event time of WorkflowTaskStarted would change with all those, I think. The input seems like the most logical place to put it.

temporal/api/schedule/v1/message.proto
dnr: Sure, the information-only stuff can be split out. The rest is all updateable though.

temporal/api/schedule/v1/message.proto
dnr: The user specifies it, like a workflow id

temporal/api/scheduleservice/v1/request_response.proto
dnr: Yes, it's like a workflow id. That would return an error from CreateSchedule.

temporal/api/scheduleservice/v1/request_response.proto
dnr: Yes, this is just for de-duping grpc-level create requests, same as request_id in other rpcs. I didn't dig into how exactly these are implemented/used yet, but this is an sdk-internal thing.

temporal/api/scheduleservice/v1/request_response.proto
dnr: Well, it does eventually start a workflow, so that does reach the user in some sense. But yeah, the start workflow action can have its own headers, so there's already a way to get headers in the workflow. I'm leaning toward leaving it at that.

For metadata/details, probably a Memo field would make sense.

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,112 @@
...
+    // - some fields in the state are modified automatically
+    // - the schedule may have been modified by UpdateSchedule
+    temporal.api.schedule.v1.Schedule schedule = x;
+}
+
+message UpdateScheduleRequest {
+    // The namespace of the schedule to update.
+    string namespace = x;
+    // The id of the schedule to update.
+    string schedule_id = x;
```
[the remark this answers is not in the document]
dnr: Hmm. Our existing apis seem to keep the identifiers out of the thing they point to. I could see this one both ways though.

temporal/api/scheduleservice/v1/request_response.proto
dnr: Replace-all is definitely easier. I'll do that, with a serial number to avoid conflicts.

However, I think things like "pause", "unpause", "trigger immediately" are common operations that could benefit from a simple way to express at the rpc level (without a loop around describe/modifiy/update). Do we think it's worth adding specific rpcs for those?

dnr: Everything that isn't explicitly commented as read-only is modifiable. `remaining_actions` seems very much like a field I would want to modify. It is true that it can change without an explicit user update, but so can `paused` (with `pause_on_failure`). If we add a serial number to avoid conflicts, it doesn't seem terribly confusing.

I don't immediately see how to implement pause-on-failure with completely separate user-mutable and system-mutable messages, unless we make the pause/unpause/paused-due-to-failure into serial numbers or similar. I guess there are some advantages to doing it that way, but it feels like overkill for this case.

temporal/api/schedule/v1/message.proto
cretz: > Payloads have metadata attached to describe how they are encoded/transformed, right? So why can't we add one that specifies it's encoded as json, with no other transformations.

Because all (non-search-attributes) payloads are completely under control of the user. They choose the data converter, and many companies have a data converter that will _only_ accept a payload with their expected metadata (e.g. for encryption).

> User dataconverters should pass through payloads if they don't have metadata indicating that they were converted with the corresponding dataconverter

We cannot require this. Many do not want to pass through to avoid leaking unencrypted payloads unintentionally. They have strict converters.

> As I understand it, the default decoders for all the sdks should be able to understand a json string.

They are, but those are the defaults, _not_ something the server can expect will be in use.

> The input seems like the most logical place to put it.

We need to completely leave input alone on the server. The user can encode it in anyway they want and we have never required that they encode it a certain way (except for search attribute payloads).

temporal/api/schedule/v1/message.proto
cretz: > Errors: there may be errors, like anywhere else. The errors should show up in server logs, and we'll also have metrics. Do we need a way to get errors through grpc also?

I do not know necessarily what we want here. It should just be known that this is the first time we've allowed signals to be scheduled actions. In the past people always explicitly signalled and got errors when, say, the workflow didn't exist. Just has to be considered...not a deal breaker. One wonders if the MVP needs this though.

> Signal-with-start: no, if we want that we can add a new action for it.

If we're ok w/ leaving that off, I wonder if it's easier to also leave signals off at first.

> Workflow ID: how else could it be interpreted?

If we aren't treating signal + start workflow as signal-with-start, it doesn't apply.

> Overlap: signals are instantaneous so no overlap applies. Note that the comments on overlap policy specifically refer to starting workflows, not to actions in general.

:+1:

If signals is not hard and a needed feature, no prob keeping.

temporal/api/schedule/v1/message.proto
cretz: Yeah, sorry, just that the intervals are non-negative. But that is obvious, nevermind. No need to add docs saying that.

temporal/api/enums/v1/schedule.proto
yiminc: Agree, please update the comment to make it clear.

temporal/api/schedule/v1/message.proto
yiminc: Why this cannot be handled the same way using OVERLAP_POLICY? Say, if user's worker is down, even if the scheduler is able to schedule a run, there won't be any worker running to make progress and the workflow will keep in running state until worker come back online. From user's POV, the result is the same either temporal server was not able to schedule a new run or user worker was not able to make progress on that scheduled run.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,212 @@
...
+
+message ScheduleAction {
+    // If present, start a workflow. Most fields of
+    // StartWorkflowExecutionRequest apply here. The exceptions are:
+    // - namespace must be omitted or match the namespace of the schedule.
+    // - identity and request_id must be omitted (they will be set by the schedule).
+    // - workflow_id_reuse_policy must be omitted or set to
+    //   WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE.
+    // - cron_schedule must be omitted.
+    temporal.api.workflowservice.v1.StartWorkflowExecutionRequest start_workflow = x;
```
yycptt, line 144: Can start and signal be specified at the same time or only one should be specified?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,212 @@
...
+}
+
+message ScheduleState {
+    // If present and true, do not take any actions based on the schedule description.
+    // Actions may be taken based on explicit requests from the fields below.
+    bool paused = x;
+
+    // If present and equal to zero, don't schedule again.
+    // Automatically decrements on each action.
+    int64 remaining_actions = x;
```
yycptt, line 176: So caller must specify a number when creating a schedule? I mean this field will always present (if ScheduleState is present) and if not specified, it will be 0 and don't schedule again.

temporal/api/schedule/v1/message.proto
dnr: The intention was that generally only one would be specified, but if both were specified for some reason, then both actions would be taken. It's easier (at the code level) to just have a bunch of optional things than use `oneof`, as long as they don't conflict.

But the signal thing is apparently confusing and not needed, so I'll take it out for now.

temporal/api/schedule/v1/message.proto
dnr: Let's leave it off for now.

temporal/api/schedule/v1/message.proto
dnr: I think we'll have to add a new attribute to all the start workflow requests/events to pass in the time. I'll do it in another PR.

temporal/api/schedule/v1/message.proto
dnr: This is one of the things I think was a big step backwards in proto2/3 compared to proto1, presence info for all fields, and customizable defaults :frowning_face:  Either feature would make this easy to fix.

I guess the options are:
1. add a bool field whether to update/respect the action count
2. let -1 mean no limit and ask everyone to set it for normal unlimited schedules
3. let 0 mean no limit, 1 mean "no actions left", 2 mean "1 action left", etc.

Preferences? I feel like (1) is cleanest.

temporal/api/scheduleservice/v1/request_response.proto
dnr: Changed things around here

temporal/api/scheduleservice/v1/request_response.proto
dnr: added memo and searchattributes for now

temporal/api/schedule/v1/message.proto
dnr: split this into three

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,114 @@
...
+
+message DeleteScheduleRequest {
+    // The namespace of the schedule to delete.
+    string namespace = x;
+    // The id of the schedule to delete.
+    string schedule_id = x;
+    // The identity of the client who initiated this request.
+    string identity = x;
+    // A unique identifier for this update request. Typically UUIDv4.
+    string request_id = x;
```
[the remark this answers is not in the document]
dnr: I guess that's true. There's still an ABA problem in theory if a delete by id gets delayed and retried, while another delete goes through and then the id is reused. But pretty unlikely to cause problems in practice. We'd have to give schedules uuids to avoid that.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+//   *: matches always
+//   x: matches when the field equals x
+//   x/y : matches when the field equals x+n*y where n is an integer
+//   x-z: matches when the field is between x and z inclusive
+//   w,x,y,...: matches when the field is one of the listed values
+// Each x, y, z, ... is either a decimal integer, or a month or day of week name
+// or abbreviation (in the appropriate fields).
+// A second in time matches if all fields match.
+// day_of_week uses 0 or 7 as Sunday
+// TODO: add relative-to-end-of-month
```
sw-yx, line 46: standard for this this would be the Quartz `L` character, but if you wanna do something fancier than that i havent come across it

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    google.protobuf.Duration jitter = x [(gogoproto.stdduration) = true];
+
+    // Time zone to interpret all CalendarSpecs in.
+    //
+    // If unset, defaults to UTC. We recommend using UTC for your application if
+    // at all possible, to avoid various unexpected properties of time zones.
+    //
+    // Time zones may be provided by name, corresponding to names in the IANA
+    // time zone database. The definition will be loaded by the Temporal server
+    // from the environment it runs in.
```
sw-yx, line 104: is there a standard link we can leave in the comments for people to look up?

temporal/api/schedule/v1/message.proto
sw-yx: do we have to decide now? i can't even think of anything. maybe by default we expose # of attempts made?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+
+message ScheduleState {
+    // If true, do not take any actions based on the schedule description.
+    bool paused = x;
+
+    // If true, decrement remaining_actions after each action, and do not take
+    // any more scheduled actions if remaining_actions is zero. (Actions may
+    // still be taken by explicit request.)
+    bool limited_actions = x;
+    int64 remaining_actions = x;
```
sw-yx, line 155: general comment about calling them "actions" - this will likely conflict with our pricing model, which as far as i know is also based on "actions" -> which in that context means # of activities executed + # of signals etc

sw-yx:
so just thinking through one basic task, and one edge case here to do with updating schedules:

- basic: if i just want to pause a schedule, will it be very complicated? i think we want to expose some nice high level api in the server. i'm 99% sure this is the case, but just wanna spell it out.
- edge case: lets say i have 2 scheduled workflows running under `OVERLAP_ALLOW_ALL`, and i update my policy to `OVERLAP_CANCEL_OTHER`. does it immediately cancel one of those workflows? i'll guess no, but is that something we wanna communicate or hash out any further?

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,121 @@
...
+    // The id of the schedule to describe.
+    string schedule_id = x;
+}
+
+message DescribeScheduleResponse {
+    // The complete current schedule details. This may not match the schedule as
+    // created because:
+    // - some fields in the state are modified automatically
+    // - the schedule may have been modified by UpdateSchedule
+    temporal.api.schedule.v1.Schedule schedule = x;
```
sw-yx, line 66: this wont show up in the api but just flagging that we'll likely want a "predictschedule" as well - given the current time and current schedule, show me the next N times that the schedule is set to trigger

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    // skipped.
+    repeated CalendarSpec exclude_calendar = x;
+    // Any timestamps before not_before will be skipped. Together, not_before
+    // and not_after make an inclusive interval.
+    google.protobuf.Timestamp not_before = x [(gogoproto.stdtime) = true];
+    // Any timestamps after not_after will be skipped.
+    google.protobuf.Timestamp not_after = y [(gogoproto.stdtime) = true];
+    // All timestamps will be incremented by a random value from 0 to this
+    // amount of jitter. Default: 60 seconds
+    google.protobuf.Duration jitter = x [(gogoproto.stdduration) = true];
```
sw-yx, line 95: this default seems high. very beneficial for us for high load usecases, but will look buggy for people who are testing and unaware that we apply default 1 minute jitter. is it feasible to bring this default down to 1? just asking, i dont have strong opinion here just thinking it thru

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+// saving time policy changes for an area). To create a totally self-contained
+// ScheduleDescription, use UTC or include timezone_data.
+message ScheduleSpec {
+    // Calendar-based specifications of times.
+    repeated CalendarSpec calendar = x;
+    // Interval-based specifications of times.
+    repeated IntervalSpec interval = x;
+    // Any timestamps matching any of the exclude_calendar specs will be
+    // skipped.
+    repeated CalendarSpec exclude_calendar = x;
```
sw-yx, line 87: not likely to be useful in this format if we want people to use a list of holiday dates or something. what's a popular source of holiday data, and can we reshape this field to make that easy to take in?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,206 @@
...
+    // If set, trigger one action immediately.
+    bool trigger_immediately = x;
+
+    // If set, runs though the specified time period(s) and takes actions as if that time
+    // passed by right now, all at once. The overlap policy can be overridden for the
+    // scope of the backfill.
+    repeated BackfillRequest backfill_request = x;
+}
+
+message ScheduleInfo {
```
sw-yx, line 174: worth adding created_at, updated_at, deleted_at?

dnr: Sure, those seem generally useful and low-cost.

I'm not sure about deleted, since the idea was that Describe would return a not found error for a deleted schedule. In fact though, we will (probably) preserve the data until the retention interval and could expose it somehow.

temporal/api/schedule/v1/message.proto
dnr: It's basically a UI question: when a user lists schedules and we present them (e.g. in a table), what do they want to see? I'll figure out how to plumb it through visibility

temporal/api/scheduleservice/v1/request_response.proto
dnr: It's in the ScheduleInfo inside Schedule. Maybe we should move it out here? From a proto message design pov, I kinda like it inside Schedule, since it makes something like `repeated Schedule` more sensible (instead of e.g. parallel arrays or a second message).

dnr:
> * basic: if i just want to pause a schedule, will it be very complicated? i think we want to expose some nice high level api in the server. i'm 99% sure this is the case, but just wanna spell it out.

Yeah, that was in this comment thread: https://github.com/temporalio/api/pull/157#discussion_r822959538  I agree we probably need that.

A couple ideas:

1. Add a `pause` and `resume` to `ScheduleRequest`, and then add a new grpc api called `SignalSchedule` or similar that's just `UpdateSchedule` with only a `ScheduleRequest` field. So you don't have to worry about conflicts or doing a Describe first (though we can support the conflict token there too).

2. Just add explicit apis for PauseSchedule, ResumeSchedule, TriggerImmediately, etc. And maybe get rid of `ScheduleRequest` then? Though I kinda like it for the reasons mentioned here https://github.com/temporalio/api/pull/157#discussion_r822206520

temporal/api/schedule/v1/message.proto
sw-yx: we dont have that much info to share tbh, unless we can easily report aggregate numbers across all instantiations. this is all we're showing rn

image

temporal/api/scheduleservice/v1/request_response.proto
```diff
@@ -0,0 +1,137 @@
...
+    // The id of the new schedule.
+    string schedule_id = 2;
+    temporal.api.schedule.v1.Schedule schedule = 3;
+    // The identity of the client who initiated this request.
+    string identity = 4;
+    // A unique identifier for this create request. Typically UUIDv4.
+    string request_id = 5;
+
+    temporal.api.common.v1.Memo memo = 6;
+    temporal.api.common.v1.SearchAttributes search_attributes = 7;
```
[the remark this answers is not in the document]
dnr: They're for the schedule itself, not the workflows created by the schedule. We're going to use visibility to keep track of these, so we might as well expose some useful features. They're returned in ListSchedules, and I'll add them to DescribeSchedule too. There's no QuerySchedules yet but it would be possible.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,264 @@
...
+
+message ScheduleAction {
+    // If present, start a workflow. Most fields of
+    // StartWorkflowExecutionRequest apply here. The exceptions are:
+    // - namespace must be omitted or match the namespace of the schedule.
+    // - identity and request_id must be omitted (they will be set by the schedule).
+    // - workflow_id_reuse_policy must be omitted or set to
+    //   WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE.
+    // - cron_schedule must be omitted.
+    temporal.api.workflowservice.v1.StartWorkflowExecutionRequest start_workflow = 1;
```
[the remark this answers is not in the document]
dnr: What's the concern? That we might end up with a dependency loop? We can just repeat the contents here, but (this is a controversial opinion) there's already _way_ too much repetition in our api and we should reuse messages much more than we do now.

What I would have done if I could change things arbitrarily is: extract a lot of the workflow execution metadata into a common message and then use that in various places instead of repeating the same 15-odd fields in six different proto messages. Right now rpc-level stuff is mixed with workflow-level stuff and (imo) it's a bit of a mess.

Semantically, this particular message makes a lot of sense here: this is a request to start a workflow, it's just delayed.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,264 @@
...
+    string invalid_schedule_error = 8;
+}
+
+message Schedule {
+    // These fields are used by CreateSchedule and can be modified by UpdateSchedule.
+    // They are returned by DescribeSchedule.
+    ScheduleSpec spec = 1;
+    ScheduleAction action = 2;
+    SchedulePolicies policies = 3;
+    ScheduleState state = 4;
```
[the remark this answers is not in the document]
dnr: Yes. As you said, trying to do a general "patch" is implausible. So I'm imagining two different ways to update a schedule: full replacement (including state.paused), and specific patch requests that we imagine are useful (in request). If you're doing a full replacement, you take responsibility for changing the notes field along with paused, or not.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,264 @@
...
+
+    // Write-only, only used by CreateSchedule/UpdateSchedule, not returned by DescribeSchedule:
+    ScheduleRequest request = 5;
```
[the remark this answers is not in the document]
dnr: That's fair, I can move it into Create/Update. I think internally we'll have to create structs or messages to wrap a pair of Schedule and ScheduleRequest, but I suppose it's cleaner at this level.

dnr:
This is in a much better state so I'll move it to ready now

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,258 @@
...
+
+// CalendarSpec describes an event specification relative to the calendar,
+// similar to a traditional cron specification. Each field can be one of:
+//   *: matches always
+//   x: matches when the field equals x
+//   x/y : matches when the field equals x+n*y where n is an integer
+//   x-z: matches when the field is between x and z inclusive
+//   w,x,y,...: matches when the field is one of the listed values
+// Each x, y, z, ... is either a decimal integer, or a month or day of week name
+// or abbreviation (in the appropriate fields).
```
robzienert, line 51: Any thoughts on supporting a fuzzing character (`H` in Jenkins)? This would allow schedules to not all trigger at 0 seconds (for example) and cause a spike, but rather distribute across 0-59. Reasonable for users to implement this logic themselves, but could be handy to have out-of-the-box.

robzienert: Ah, I see the `ScheduleSpec` already has a `jitter` field.

temporal/api/scheduleservice/v1/request_response.proto
cretz: Marking resolved

temporal/api/enums/v1/schedule.proto
```diff
@@ -0,0 +1,62 @@
...
+    // OVERLAP_CANCEL_OTHER means that if there is another workflow running,
+    // cancel it, and start the new one after the old one completes
+    // cancellation.
```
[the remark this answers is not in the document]
cretz: Marking as resolved assuming that we will clarify this.

temporal/api/schedule/v1/message.proto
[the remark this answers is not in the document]
cretz: The concern is mixing general purpose models vs API contract models. But it's not a big concern.

temporal/api/enums/v1/schedule.proto
```diff
@@ -0,0 +1,61 @@
...
+    // running, cancel it, and start the new one after the old one completes cancellation.
+    SCHEDULE_OVERLAP_POLICY_CANCEL_OTHER = 4;
+    // SCHEDULE_OVERLAP_POLICY_TERMINATE_OTHER means that if there is another workflow
+    // running, terminate it and start the new one immediately.
+    SCHEDULE_OVERLAP_POLICY_TERMINATE_OTHER = 5;
+    // SCHEDULE_OVERLAP_POLICY_ALLOW_ALL means start any number of concurrent workflows.
+    // This means that all runs must have unique workflow ids. To allow this, the workflow
+    // id in the start request has a start time appended to it (always in this mode,
+    // whether overlaps happen in reality or not).
+    SCHEDULE_OVERLAP_POLICY_ALLOW_ALL = 6;
```
mfateev, line 60: I think how IDs are assigned is an implementation detail and is not part of this API. We can enforce any of these policies even if all runs always have different ids. And I propose always giving them different ids.

temporal/api/schedule/v1/message.proto
mfateev: I think it is weird to use * and other symbols when modeling Caalendar as protos. Can we model it in a strongly typed manner? For example, change all these values to integer values and use oneof to specify if value is optional.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,258 @@
...
+// (possibly infinite) that an action should occur at. The meaning of a
+// ScheduleSpec depends only on its contents and never changes, except that the
+// definition of a time zone can change over time (most commonly, when daylight
+// saving time policy changes for an area). To create a totally self-contained
+// ScheduleSpec, use UTC or include timezone_data.
+message ScheduleSpec {
+    // Calendar-based specifications of times.
+    repeated CalendarSpec calendar = 1;
+    // Interval-based specifications of times.
+    repeated IntervalSpec interval = 2;
```
mfateev, line 92: Should we allow oneof of calendar and interval?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,258 @@
...
+    // Any timestamps matching any of the exclude_calendar specs will be
+    // skipped.
+    repeated CalendarSpec exclude_calendar = 3;
+    // Any timestamps before not_before will be skipped. Together, not_before
+    // and not_after make an inclusive interval.
+    google.protobuf.Timestamp not_before = 4 [(gogoproto.stdtime) = true];
+    // Any timestamps after not_after will be skipped.
+    google.protobuf.Timestamp not_after = 5 [(gogoproto.stdtime) = true];
+    // All timestamps will be incremented by a random value from 0 to this
+    // amount of jitter. Default: 1 second
```
mfateev, line 102: 1 second jitter is too low to be useful. I would set the default jitter to 0.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,258 @@
...
+    // Policy for overlaps.
+    // Note that this can be changed after a schedule has taken some actions, and we can't
+    // provide 100% sensible semantics for all changes. The most confusing case would be
+    // changes to/from ALLOW_ALL: when using that policy, workflow ids have a timestamp
+    // appended to provide uniqueness. For all other policies, workflows started by a
+    // schedule all use the same id. Changing between these two types of policies will
+    // leave all workflows with the previous id style alone. E.g., if changing from
+    // ALLOW_ALL to CANCEL_OTHER, and there are workflows running, those workflows will
+    // not be cancelled. If changing from ALLOW_ALL to SKIP with workflows running, the
+    // running workflows will not cause the next action to be skipped.
```
mfateev, line 139: I think we should use different IDs for each run and rely on the controller workflow to enforce all the policies.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,258 @@
...
+    // This value defaults to 60 seconds, and can't be less than 10 seconds.
+    google.protobuf.Duration catchup_window = 2 [(gogoproto.stdduration) = true];
+
+    // If true, and a workflow run fails or times out, turn on "paused".
+    // This applies after retry policies: the full chain of retries must fail to
+    // trigger a pause here.
+    bool pause_after_failure = 3;
+}
+
+message ScheduleAction {
```
mfateev, line 155: I would make this oneof to support adding new actions in a backward compatible manner in the future.

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,258 @@
...
+    string notes = 1;
+
+    // If true, do not take any actions based on the schedule spec.
+    bool paused = 2;
+
+    // If true, decrement remaining_actions after each action, and do not take
+    // any more scheduled actions if remaining_actions is zero. (Actions may
+    // still be taken by explicit request.)
+    bool limited_actions = 3;
+    int64 remaining_actions = 4;
```
mfateev, line 190: should we make the remaining_actions optional instead?

temporal/api/schedule/v1/message.proto
```diff
@@ -0,0 +1,258 @@
...
+    // Override overlap policy for this one request.
+    temporal.api.enums.v1.ScheduleOverlapPolicy overlap_policy = 1;
+}
+
+message BackfillRequest {
+    // Time range to evaluate schedule at.
+    google.protobuf.Timestamp from_time = 1 [(gogoproto.stdtime) = true];
+    google.protobuf.Timestamp to_time = 2 [(gogoproto.stdtime) = true];
+    // Override overlap policy for this request.
+    temporal.api.enums.v1.ScheduleOverlapPolicy overlap_policy = 3;
```
mfateev, line 203: I think we should add "dry run" option to see what is going to be scheduled. Or have a separate operation to query this information given a BackfillRequest.

temporal/api/schedule/v1/message.proto
dnr: The scheduling resolution is 1 second, and we don't promise to start on second boundaries anyway, so 1s isn't even really observable. It's not a huge thing but 1s seems better than 0.

The go sdk comments say that the resolution of timers is 1s and something does `Ceil(d.Seconds())` to get the timer expiry. But I can't find anywhere that Ceil is happening, and I suspect the comment is out of date. If the resolution really is 1s, then 1s jitter is useless since we're going to implement it with timers. But if not, it's slightly useful.

temporal/api/schedule/v1/message.proto
dnr: I commented above, I'm okay with it (especially if it simplifies the code, which I'll try out)

temporal/api/schedule/v1/message.proto
dnr: Sure

temporal/api/schedule/v1/message.proto
dnr: I had it that way at first, but as far as I know, there's no way to do that in proto3: everything is optional and the default value is always zero, so there's no way to have an int64 field and be able to distinguish zero from missing. (IMO this is a major regression from proto1 and proto2)

temporal/api/schedule/v1/message.proto
dnr: I like the idea. Doing it with a separate grpc method seems better since it's going to have to get translated to a different thing internally (query vs signal to actually do backfill).

temporal/api/enums/v1/schedule.proto
yiminc: For many cases that does not need any overlapping, having only one workflowID is a very nice thing.

temporal/api/schedule/v1/message.proto
yiminc: I kind agree with David here. It would become annoy to have to type such verbose spec if we go with the example with Range above. The proposal in this PR seems a good balance.

temporal/api/schedule/v1/message.proto
yiminc: With the planed multi cursor change, i think this does not matter as much. I would go with 0 as default.

temporal/api/schedule/v1/message.proto
yiminc: It would be nice to have a fixed workflowID for non overlapping case. But I don't feel strong about it.

yiminc (approved):
Looks good to me.

temporal/api/schedule/v1/message.proto
mfateev: The scheduler workflow can calculate the aggregates and update the search index with them.

PR #159, opened 2022-03-18
Activity local dispatch

yycptt:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -79,6 +79,8 @@ message ScheduleActivityTaskCommandAttributes {
     // dynamic configuration. Retries will be attempted until `schedule_to_close_timeout` has
     // elapsed. To disable retries set retry_policy.maximum_attempts to 1.
     temporal.api.common.v1.RetryPolicy retry_policy = 11;
+    // Request to start the activity directly bypassing matching service and worker polling
+    bool request_start = 12;
 }
 
 message RequestCancelActivityTaskCommandAttributes {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -281,6 +281,8 @@ message RespondWorkflowTaskCompletedRequest {
 message RespondWorkflowTaskCompletedResponse {
     // See `RespondWorkflowTaskCompletedResponse::return_new_workflow_task`
     PollWorkflowTaskQueueResponse workflow_task = 1;
+    // See `ScheduleActivityTaskCommandAttributes::request_start`
+    repeated PollActivityTaskQueueResponse activity_tasks = 2;
 }
 
 message RespondWorkflowTaskFailedRequest {


yiminc (approved):

Sushisource:
This is neat, but requires some nontrivial SDK changes - was there any proposal around this? I think it's a good idea but we'll have to fit in the SDK work somewhere

temporal/api/command/v1/message.proto
```diff
@@ -79,6 +79,8 @@ message ScheduleActivityTaskCommandAttributes {
     // dynamic configuration. Retries will be attempted until `schedule_to_close_timeout` has
     // elapsed. To disable retries set retry_policy.maximum_attempts to 1.
     temporal.api.common.v1.RetryPolicy retry_policy = 11;
+    // Request to start the activity directly bypassing matching service and worker polling
+    bool request_start = 12;
```
cretz, line 83: Can you help me understand the use case? Is coming back on an activity poller instead of workflow poller that more expensive?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -281,6 +281,8 @@ message RespondWorkflowTaskCompletedRequest {
 message RespondWorkflowTaskCompletedResponse {
     // See `RespondWorkflowTaskCompletedResponse::return_new_workflow_task`
     PollWorkflowTaskQueueResponse workflow_task = 1;
+    // See `ScheduleActivityTaskCommandAttributes::request_start`
+    repeated PollActivityTaskQueueResponse activity_tasks = 2;
```
cretz, line 285: What if I have activities disabled on my worker? Is this expected to still be processed?

temporal/api/command/v1/message.proto
cretz, line 83: What if this is scheduled on a task queue that has no workflow poller?

temporal/api/workflowservice/v1/request_response.proto
cretz, line 285: Is this expected to be subjected to max concurrent activities? What if I receive this and I'm already at max concurrent activities?

temporal/api/command/v1/message.proto
yycptt: Since the flag is set by SDK, sever will expected that if it's set, sdk has the capability for running this activity. 

Server can check if the activity is cross namespace or on a different task queue and reject the request, but I think SDK should guarantee that if it request to start the activity, it should also be able to process it.

temporal/api/workflowservice/v1/request_response.proto
yycptt: Those activity tasks will only contain those requested to start by SDK. So if activity worker is disabled, then none of the `request_start` should be set and this field will always be empty.

yycptt: I think the idea is that SDK only request to start an activity when it has the capacity. 

If by the time this response is received, max concurrent activities is hit, then those activities should be blocked I think. We can have more discussion about this behavior. Another question we can discuss is whether those activities count towards the task queue rps limit.
cc @yiminc

temporal/api/command/v1/message.proto
cretz: It is unfortunate that we have to expose such an optimization to a user instead of optimizing internally.

temporal/api/workflowservice/v1/request_response.proto
cretz: I am confused. Say I have worker A that has a workflow and activity worker and sends `request_start = true` for an activity. Then worker B only has a workflow worker. It can receive a value here for `activity_tasks`. What is workflow-worker-only worker B supposed to do?

temporal/api/command/v1/message.proto
cretz: > Server can check if the activity is cross namespace or on a different task queue and reject the request, but I think SDK should guarantee that if it request to start the activity, it should also be able to process it.

This does not take into account that workers come and go with different settings. It sounds like at the least this feature cannot be enabled for cross-task-queue activities (we can't do some kind of global check that there is a worker running for another task queue).

yycptt: User won't have direct access to this field. SDK need to set the field when it's able to and has the capacity (memory, cpu, etc.) to execute the activity. Of course there can be a flag in worker option to enable/disable this optimization.

yycptt: From server point of view, it doesn't know or need to know what a worker can/cannot do. Today a worker is only responsible for one task queue, in the future it may support multiple task queues, or even cross namespace. But the point is sever doesn't need to know that, it can trust SDK sending the right value for the `request_start` field. i.e. as long as the value is true, the caller (worker) that's making the `RespondWorkflowTaskComplete` RPC call is able to start/run the activity in some way. And some activities may be returned in the response of the RPC call, to start immediately.

Back to the question, true that this feature can't be enabled for cross-task-queue activities, and value for `request_start` should always be false. This needs to be guarantee by SDK implementation, and not relying on server to reject such requests.

temporal/api/workflowservice/v1/request_response.proto
yycptt: The case you described won't happen and it's guaranteed by server. The `request_start` is in the request of `RespondWorkflowTaskComplete` and the `activity_tasks` are in the response. It's the same RPC call, and only the worker request to start activities will receive them.

yycptt: - Like I described above, server guarantees it's the same worker.
- There's task queue level RPS limit: https://github.com/temporalio/sdk-go/blob/master/internal/worker.go#L73

temporal/api/command/v1/message.proto
cretz: This makes sense now that I understand it's same-worker response.

cretz: :+1: Do you expect we would always enable by default?

temporal/api/workflowservice/v1/request_response.proto
cretz: Makes sense, thanks!

cretz (approved):

temporal/api/command/v1/message.proto
yycptt: Yes I think it can be enabled by default. Server side also has per namespace feature flag for this.

yycptt:
> This is neat, but requires some nontrivial SDK changes - was there any proposal around this? I think it's a good idea but we'll have to fit in the SDK work somewhere

Sorry I don't have a written proposal for this. The server side change is quite simple and a detailed proposal seems unnecessary. I am not sure about the plan on SDK side though, and a proposal may be necessary for SDK changes. 

I think the discussion in this PR already covers the contract for the two new added fields. But if things are still unclear, I am happy to fill in the contract part in SDK change proposal.

PR #168, opened 2022-03-25
Add reason field for workflow cancellation

yycptt:
**What changed?**
- Add reason field for request cancel workflow API
- Add reason field for request cancel workflow command
- Add reason field to `RequestCancelExternalWorkflowExecutionInitiatedEventAttributes`
- `WorkflowExecutionCancelRequestedEventAttributes` already contains a `cause` field

- NOTE: 
- SDK need to populate the "reason" field in request cancel workflow command.
- [When seeing a `WorkflowExecutionCancelRequestedEvent`](https://github.com/temporalio/sdk-go/blob/master/internal/internal_event_handlers.go#L856), SDK should create a [CancelledError](https://github.com/temporalio/sdk-go/blob/master/internal/error.go#L143) using the reason in the history event. Currently I think it's using [`ErrCanceled`](https://github.com/temporalio/sdk-go/blob/master/internal/context.go#L196), which is created [without any details](https://github.com/temporalio/sdk-go/blob/master/internal/context.go#L176), and the [details field in workflow execution cancelled event](https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_handlers.go#L1463) is empty.

**Why?**
- https://github.com/temporalio/temporal/issues/2426

**How did you test it?**

**Potential risks**

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -344,7 +344,8 @@ message TimerCanceledEventAttributes {
 
 message WorkflowExecutionCancelRequestedEventAttributes {
     // User provided reason for requesting cancellation
-    string cause = 1;
+    // TODO: shall we create a new field with name "reason" and deprecate this one? 
+    string cause = 1; 
     // TODO: Is this the ID of the event in the workflow which initiated this cancel, if there was one?
     int64 external_initiated_event_id = 2;
     temporal.api.common.v1.WorkflowExecution external_workflow_execution = 3;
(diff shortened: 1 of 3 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -344,7 +344,8 @@ message TimerCanceledEventAttributes {
 
 message WorkflowExecutionCancelRequestedEventAttributes {
     // User provided reason for requesting cancellation
-    string cause = 1;
+    // TODO: shall we create a new field with name "reason" and deprecate this one?
```
yycptt, line 347: Shall we create a new field with name "reason" and deprecate this one so that the names are in sync?

yiminc (approved):

temporal/api/history/v1/message.proto
yycptt: I understand changing field is wire backward compatible. But I am kinda unsure if any code in SDK or TCTL or UI is depending on the generated code like the json tag. 

Had a sync with @yiminc yesterday on this and we decided to keep the existing name for now.

PR #174, opened 2022-04-20
Remove namespace field from ScheduleActivityTaskCommandAttributes message

alexshtin:
**What changed?**
Remove `namespace` field from `ScheduleActivityTaskCommandAttributes` message.

**Why?**
To remove ability of cross namespace activities calls. Initial motivation behind cross namespace activities was to allow different teams to own their own namespaces with activities and other teams to build workflows that use these activities in form of "library". Unfortunately, we found out that "cross namespace activity" is a wrong abstraction to achieve this goal. It adds complexity which is not well covered with test and not actually used (all official SDKs doesn't even expose it). For example, who should control activity heartbeat timeout?

And by the way, we are working on designing new feature/abstraction which will enable this functionality.

**How did you test it?**
N/A

**Potential risks**
Low risk: `namespace` field is not exposed for the most SDKs but there might be some unofficial SDK which will lose ability to start activity in another namespace.

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -10,5 +10,6 @@ lint:
 breaking:
   ignore:
     - temporal/api/command/v1
+    - temporal/api/history/v1
   use:
     - PACKAGE
--- a/buf.yaml
+++ b/buf.yaml
@@ -9,5 +9,6 @@ lint:
     - DEFAULT
 breaking:
   ignore:
+    - temporal/api/command/v1
   use:
     - PACKAGE
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -44,6 +44,7 @@ import "temporal/api/taskqueue/v1/message.proto";
 message ScheduleActivityTaskCommandAttributes {
     string activity_id = 1;
     temporal.api.common.v1.ActivityType activity_type = 2;
+    // This used to be a `namespace` field which allowed to schedule activity in another namespace.
     reserved 3;
     temporal.api.taskqueue.v1.TaskQueue task_queue = 4;
     temporal.api.common.v1.Header header = 5;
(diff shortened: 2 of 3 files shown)


bergundy:
FTR, we removed `namespace` from TS `ActivityOptions` recently so AFAIK all SDKs no longer expose this

temporal/api/command/v1/message.proto
```diff
@@ -44,7 +44,7 @@ import "temporal/api/taskqueue/v1/message.proto";
 message ScheduleActivityTaskCommandAttributes {
     string activity_id = 1;
     temporal.api.common.v1.ActivityType activity_type = 2;
-    string namespace = 3;
+    reserved 3;
```
bergundy, line 47: Let's add a comment explaining what was here before.

bergundy (approved):
Approved with a comment

antstorm:
@alexshtin can you please expand on the "why?" of this change a bit more?

buf.yaml
```diff
@@ -9,5 +9,7 @@ lint:
     - DEFAULT
 breaking:
   ignore:
+    - temporal/api/command/v1
+    - temporal/api/history/v1
```
cretz, line 13: I assume you'll come back in a separate PR and remove these two lines

cretz (approved):

buf.yaml
alexshtin: yes!

bergundy:
> @alexshtin can you please expand on the "why?" of this change a bit more?

We want to support cross namespace calls via another mechanism, one that doesn't require the caller to set options which are implementation details of the activity (e.g. timeouts and retry policy).

alexshtin:
> @alexshtin can you please expand on the "why?" of this change a bit more?

Updated. Does PHP SDK expose cross namespace activities?

PR #176, opened 2022-04-22
Add schedule api to WorkflowService

dnr:
**What changed?**
Adding new methods for schedule api to WorkflowService

**Why?**
This is the main point of entry for sdks and ui to manipulate schedules.


dnr:
(See #157 for more details + discussion)

buf.yaml
```diff
@@ -9,5 +9,6 @@ lint:
     - DEFAULT
 breaking:
   ignore:
+    - temporal/api/schedule/v1
```
dnr, line 12: need to do this temporarily to change the start workflow action. none of this is used yet so there are no compatibility concerns

temporal/api/schedule/v1/message.proto
```diff
@@ -166,18 +166,31 @@ message SchedulePolicies {
...
+    google.protobuf.Duration workflow_execution_timeout = 5 [(gogoproto.stdduration) = true];
+    // Timeout of a single workflow run.
+    google.protobuf.Duration workflow_run_timeout = 6 [(gogoproto.stdduration) = true];
+    // Timeout of a single workflow task.
+    google.protobuf.Duration workflow_task_timeout = 7 [(gogoproto.stdduration) = true];
+    // The retry policy for the workflow. Will never exceed `workflow_execution_timeout`.
+    temporal.api.common.v1.RetryPolicy retry_policy = 8;
+    temporal.api.common.v1.Memo memo = 9;
+    temporal.api.common.v1.SearchAttributes search_attributes = 10;
+    temporal.api.common.v1.Header header = 11;
```
dnr, line 191: question: does header make sense here?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -324,4 +324,32 @@ service WorkflowService {
...
+    // Returns the schedule description and current state of an existing schedule.
+    rpc DescribeSchedule (DescribeScheduleRequest) returns (DescribeScheduleResponse) {
+    }
+
+    // Changes the configuration or state of an existing schedule.
+    rpc UpdateSchedule (UpdateScheduleRequest) returns (UpdateScheduleResponse) {
+    }
+
+    // Makes a request to a schedule (without updating it).
+    rpc MakeScheduleRequest (MakeScheduleRequestRequest) returns (MakeScheduleRequestResponse) {
```
dnr, line 341: this name is not great. any other ideas? PokeSchedule?

dnr (commented):
moved into WorkflowService as discussed

temporal/api/workflowservice/v1/service.proto
cretz: Can you help describe the purpose of this? I was under the impression that schedules were automatically scheduled, is this for manually invoking a scheduled action? Why wouldn't I use another API to make a call instead of triggering the schedule? This is basically turning scheduling into templating?

If this call does what I think it does, `TriggerSchedule` or `ManuallyTriggerSchedule` or `InvokeSchedule` or `ApplyScheduleRequest` or something similar.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -808,3 +809,126 @@ message ListTaskQueuePartitionsResponse {
...
+    string schedule_id = 2;
+    // The schedule spec, policies, action, and initial state.
+    temporal.api.schedule.v1.Schedule schedule = 3;
+    // Optional initial request (e.g. to run the action once immediately).
+    temporal.api.schedule.v1.ScheduleRequest initial_request = 4;
+    // The identity of the client who initiated this request.
+    string identity = 5;
+    // A unique identifier for this create request for idempotence. Typically UUIDv4.
+    string request_id = 6;
+    // Memo and search attributes to attach to the schedule itself.
```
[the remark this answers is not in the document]
dnr: What do you mean by "the workflow" here? The schedule is implemented by a (possibly-hidden) workflow, but at the level of this api we don't expose that directly. We just say the schedule itself has a memo and search attributes. You can get them back from Describe and also List.

At the moment in this PR there isn't a way to search for schedules by search attribute. I think we eventually want to add a query field to ListSchedules but I was putting it off until the more basic stuff is done. We've also discussed simply not hiding the scheduler workflows and leaving them visible to users, in which case users could the regular visibility api to search. There's a meeting tomorrow to talk about this actually.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -808,3 +809,126 @@ message ListTaskQueuePartitionsResponse {
...
+    temporal.api.common.v1.Memo memo = 7;
+    temporal.api.common.v1.SearchAttributes search_attributes = 8;
+}
+
+message CreateScheduleResponse {
```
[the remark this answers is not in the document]
dnr: agreed, it should

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -808,3 +809,126 @@ message ListTaskQueuePartitionsResponse {
...
+    // The identity of the client who initiated this request.
+    string identity = 5;
+    // A unique identifier for this update request for idempotence. Typically UUIDv4.
+    string request_id = 6;
+}
+
+message UpdateScheduleResponse {
+}
+
+message MakeScheduleRequestRequest {
```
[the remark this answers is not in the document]
dnr: This one I'm less sure about. There's two things you can do with this:

1. pause/unpause: The idea is to make a shortcut for read/modify/write where only the paused field is changed, since it's a common operation and we want to expose it to the web UI, etc. If you're doing a read/modify/write over rpc, you need a conflict_token to make it atomic. But here we're pushing the read/modify/write all the way down, so in theory we don't need it.

2. trigger immediate/backfill: This isn't modifying the state at all, so there shouldn't be a need for atomicity. But I could imagine a situation where you do an Update and then a MakeRequest to do a backfill, and you want to be sure that the backfill operates on exactly the result of the Update. But that suggests instead maybe we should add an optional ScheduleRequest to the Update, so you can do all that in one call.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -808,3 +809,126 @@ message ListTaskQueuePartitionsResponse {
...
+    string identity = 4;
+    // A unique identifier for this update request for idempotence. Typically UUIDv4.
+    string request_id = 5;
+}
+
+message MakeScheduleRequestResponse {
+}
+
+message ListMatchingTimesRequest {
+    // The namespace of the schedule to update.
```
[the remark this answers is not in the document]
dnr: oops

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -808,3 +809,126 @@ message ListTaskQueuePartitionsResponse {
...
+    // Time range to query.
+    google.protobuf.Timestamp start_time = 3 [(gogoproto.stdtime) = true];
+    google.protobuf.Timestamp end_time = 4 [(gogoproto.stdtime) = true];
+}
+
+message ListMatchingTimesResponse {
+    repeated google.protobuf.Timestamp start_time = 1 [(gogoproto.stdtime) = true];
+}
+
+message DeleteScheduleRequest {
```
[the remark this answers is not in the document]
dnr: I dunno here either. I'm thinking if I want to delete a schedule, I don't really care if it was updated recently. Is that really going to change my mind?

Is the idea like you're writing a batch deleter where you scan all schedules, evaluate some predicate, and delete ones that match, and you want to detect intervening updates that ? I could see that. The problem though is there's no good way to implement it: "delete schedule" translates to "cancel/terminate scheduler workflow", and that doesn't go through the workflow logic itself, so it can't atomically check the conflict_token.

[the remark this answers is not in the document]
dnr: I think I had this here and took it out after review comments on the last PR. Delete is already idempotent, why do we need an idempotency token?

temporal/api/workflowservice/v1/service.proto
dnr: As I wrote in the other comment, it does two things: manually triggering (once or backfill), and single-rpc pause/unpause (without going through read/modify/write). We can add other "patch-style" updates here too.

If you say those two concepts should be split into separate rpcs, I might be sympathetic to that.

Why use this instead of just calling the target workflow directly: because this way you're guaranteed that it'll run with the same inputs, same timeouts, etc. as what the scheduler will do. You can do it in one rpc call that only mentions the schedule id, you don't need to coordinate to also have the same inputs and everything else available to that code.

Yes, it's templating. Scheduling necessarily needs to include templating (it's just running the same thing over and over with a small tweak), and we might as well expose it directly. It's been requested by users whenever this comes up.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -324,4 +324,32 @@ service WorkflowService {
...
+    }
+
+    // Lists matching times within a range.
+    rpc ListMatchingTimes (ListMatchingTimesRequest) returns (ListMatchingTimesResponse) {
```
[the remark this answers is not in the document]
dnr: yeah, I noticed that too when I was going over it. will do

temporal/api/workflowservice/v1/service.proto
alexshtin: So it is `PatchSchedule`. `...RequestRequest` is definitely not great.

buf.yaml
alexshtin: We start to act more aggressively :-)

temporal/api/schedule/v1/message.proto
```diff
@@ -166,18 +166,31 @@ message SchedulePolicies {
...
-        // - workflow_id_reuse_policy must be omitted.
-        // - cron_schedule must be omitted.
-        // The workflow id of the started workflow may not match the id in start_workflow
-        // exactly, it may have a timestamp appended for uniqueness.
-        temporal.api.workflowservice.v1.StartWorkflowExecutionRequest start_workflow = 1;
+        ScheduleActionStartWorkflow start_workflow = 1;
     }
 }
 
+message ScheduleActionStartWorkflow {
```
alexshtin, line 173: What if we extract all fields required to start new workflow execution to some common message like `NewWorkflowExecutionInfo` and use it here? 
Later we can start using this message with other start executions APIs like `StartWorkflowExecution` and `SignalWithStartWorkflowExecution`. Of course, old fields need to be supported for a long time even as deprecated (probably years) but we can fairly quickly cleanup internal APIs (history service) and make documentation clear.

temporal/api/workflowservice/v1/service.proto
dnr: I'm good with Patch, I'll change it

temporal/api/schedule/v1/message.proto
dnr: Interesting idea :laughing:   Alright, let's try it that way and see what people think.

First question: should parent close policy be in the new message? It's only used when dealing with child workflows, of course, so we could just leave it in those messages (StartChildWorkflowExecutionInitiatedEventAttributes and StartChildWorkflowExecutionCommandAttributes), or we could put it here to be shared.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -808,3 +809,141 @@ message ListTaskQueuePartitionsResponse {
...
+    // Extra schedule state info.
+    temporal.api.schedule.v1.ScheduleInfo info = 2;
+    // The memo and search attributes that the schedule was created with.
+    temporal.api.common.v1.Memo memo = 3;
+    temporal.api.common.v1.SearchAttributes search_attributes = 4;
+
+    // This value can be passed back to UpdateSchedule to ensure that the
+    // schedule was not modified between a Describe and an Update, which could
+    // lead to lost updates and other confusion.
+    bytes conflict_token = 5;
```
yiminc, line 865: so this is used for version check. maybe name it as version_token?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -808,3 +809,141 @@ message ListTaskQueuePartitionsResponse {
...
+    // The identity of the client who initiated this request.
+    string identity = 5;
+    // A unique identifier for this update request for idempotence. Typically UUIDv4.
+    string request_id = 6;
+}
+
+message UpdateScheduleResponse {
+}
+
+message PatchScheduleRequest {
```
yiminc, line 892: It took me a moment to realize patch is incremental update, and update is replace as a whole.

temporal/api/schedule/v1/message.proto
yiminc: I'm not sure it is worth the effect. Practically, the existing APIs are going to stay as is, there is not enough gain for the pain to try to deprecate and migrate to this new struct.

yiminc (approved):

temporal/api/workflowservice/v1/request_response.proto
dnr: Yeah, basically. I think "version" would imply that it's always linear. And in fact the current implementation is just linear, it's an incrementing integer. But I figured the api should not imply that. I'm fine with either name though.

temporal/api/workflowservice/v1/request_response.proto
dnr: I could copy the comment from service.proto here?

temporal/api/schedule/v1/message.proto
dnr: I'm not volunteering to change any existing apis. I agree it's probably more trouble than it's worth. Do you think even putting this non-schedule-specific message in the workflow package is too much? Or can we leave it there and use it for new apis?

temporal/api/schedule/v1/message.proto
```diff
@@ -244,6 +241,10 @@ message ScheduleInfo {
     // Number of skipped actions due to overlap.
     int64 overlap_skipped = 3;
 
+    // IDs of currently-running workflows started by this schedule. (There might
+    // be more than one if the overlap policy allows overlaps.)
+    repeated string running_workflow_ids = 9;
```
cretz, line 246: I wonder if this would better as `repeated temporal.api.common.v1.WorkflowExecution running_workflows` so that you have the run ID and it shares the type we often use for relaying executions back to the user.

dnr: I've been thinking about this as I'm implementing it. The problem is that it doesn't necessarily know the runid. It knows the original runid, but the started workflow might retry, or continue-as-new, or even get reset, so the runid might be wrong. Is it still useful then?

cretz: Definitely not useful if wrong. The question becomes is it worth obtaining correctly from the server side. If it is not worth it, makes sense to leave this as is. The problem is if you or someone else decides later it _is_ important, this collection of strings is no good. You could use `WorkflowExecution` and just document that run ID is unset currently, but I can see how that's annoying too.

Your call; tradeoffs all around.

dnr: If it does change later, deprecating this field and adding a new one isn't a big deal.

But overall, it does seem possible that the original run id could be useful in some edge cases, so for completeness and consistency I'll include it.

cretz (approved):

PR #179, opened 2022-04-29
Add NamespaceInvalidStateFailure and NamespaceNotFoundFailure

alexshtin:
**What changed?**
Add `NamespaceInvalidStateFailure` and `NamespaceNotFoundFailure`.

**Why?**
If namespace is in deleted state, namespace validator on server returns `InvalidArgument` error for any request to this namespace and if namespace is not found it returns `NotFound` error. Go SDK reacts on these errors differently. If poller gets `InvalidArgument` it exits the process [immediately](https://github.com/temporalio/sdk-go/blob/cb9c7de44f8774ab6681f4469ecbed8243d082dd/internal/internal_worker_base.go#L320) (not only worker) but `NotFound` error are considered to be [retryable](https://github.com/temporalio/sdk-go/blob/cb9c7de44f8774ab6681f4469ecbed8243d082dd/internal/internal_worker_base.go#L367). Therefore, worker behaviour in case of deleted namespace is unpredictable and depends on namespace registry refresh and SDK poll interval: it will either kill the process or keeps retrying forever.

Adding 2 new failures will allow to differentiate namespace related errors and handle them accordingly.

**How did you test it?**
N/A

**Potential risks**
No risks.

### DIFF
--- a/temporal/api/errordetails/v1/message.proto
+++ b/temporal/api/errordetails/v1/message.proto
@@ -63,6 +63,9 @@ message NamespaceNotFoundFailure {
     string namespace = 1;
 }
 
+message NamespaceAlreadyExistsFailure {
+}
+
 message ClientVersionNotSupportedFailure {
     string client_version = 1;
     string client_name = 2;
@@ -72,9 +75,6 @@ message ClientVersionNotSupportedFailure {
 message ServerVersionNotSupportedFailure {
     string server_version = 1;
     string client_supported_server_versions = 2;
-}
-
-message NamespaceAlreadyExistsFailure {
 }
 
 message CancellationAlreadyRequestedFailure {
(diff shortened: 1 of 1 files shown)


temporal/api/errordetails/v1/message.proto
```diff
@@ -53,6 +53,19 @@ message NamespaceNotActiveFailure {
     string active_cluster = 3;
 }
 
+message NamespaceInvalidStateFailure {
+    string namespace = 1;
+    string state = 2;
+    repeated string allowed_states = 3;
```
cretz, line 59: Can you document these two fields?

alexshtin: Added. Also decided to convert these to enums.

cretz (approved):

PR #182, opened 2022-05-18
Improve WorkflowExecutionStartEventAttributes documentation

Spikhalskiy:
**What changed?**
Clarified and added some comments to `WorkflowExecutionStartEventAttributes`.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -70,7 +70,7 @@ message WorkflowExecutionStartedEventAttributes {
     temporal.api.failure.v1.Failure continued_failure = 12;
     temporal.api.common.v1.Payloads last_completion_result = 13;
     // This is the run id when the WorkflowExecutionStarted event was written.
-    // Reset changes the execution runId, but preserves this field.
+    // A workflow reset changes the execution run_id, but preserves this field.
     string original_execution_run_id = 14;
     // Identity of the client who requested this execution
     string identity = 15;
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -49,6 +49,8 @@ message WorkflowExecutionStartedEventAttributes {
     temporal.api.common.v1.WorkflowType workflow_type = 1;
     // If this workflow is a child, the namespace our parent lives in
     string parent_workflow_namespace = 2;
+    // Contains information about parent workflow execution that initiated the child workflow these attributes belong to.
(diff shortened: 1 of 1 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -67,11 +69,13 @@ message WorkflowExecutionStartedEventAttributes {
...
     temporal.api.failure.v1.Failure continued_failure = 12;
     temporal.api.common.v1.Payloads last_completion_result = 13;
-    // This is the run id when the WorkflowExecutionStarted event was written
+    // This is the run id when the WorkflowExecutionStarted event was written.
+    // Reset changes the execution runId, but preserves this field.
     string original_execution_run_id = 14;
     // Identity of the client who requested this execution
     string identity = 15;
     // This is the very first runId along the chain of ContinueAsNew and Reset.
+    // Workflow Retry creates a new chain.
```
dnr, line 78: Retry (and cron) preserve this and don't create a new chain

Spikhalskiy: Thanks! I thought workflow retries create new chain. Is there anything you are aware of that breaks the chain now then?

temporal/api/history/v1/message.proto
```diff
@@ -67,11 +69,13 @@ message WorkflowExecutionStartedEventAttributes {
     temporal.api.enums.v1.ContinueAsNewInitiator initiator = 11;
     temporal.api.failure.v1.Failure continued_failure = 12;
     temporal.api.common.v1.Payloads last_completion_result = 13;
-    // This is the run id when the WorkflowExecutionStarted event was written
+    // This is the run id when the WorkflowExecutionStarted event was written.
+    // Reset changes the execution runId, but preserves this field.
```
Sushisource, line 73:
```suggestion
    // A workflow reset changes the execution run_id, but preserves this field.
```
I figure snake case looks a little more consistent in proto land

Sushisource (approved):

temporal/api/history/v1/message.proto
dnr: In my mental model, continuing a chain is the "exceptional case", limited to retry, [old] cron, continue-as-new, and reset, and "everything else" creates a new chain. But I guess there isn't that much else: start workflow (duplicate or not), signal-with-start, start child workflow. I can't think of any others.

Spikhalskiy: Thanks for confirming. 
So basically only one of the explicit start commands (all versions of it) can lead to the creation of the new chain. Other operations always continue the chain.

PR #183, opened 2022-05-21
Fix doc for RequestCancelWorkflowExecution

yiminc:
**What changed?**
Update doc for RequestCancelWorkflowExecution to reflect actual behavior. The behavior is expected.

**Why?**
RequestCancelWorkflow should not return error when target workflow already closed.

**How did you test it?**

**Potential risks**

### DIFF
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -215,9 +215,9 @@ service WorkflowService {
     // RequestCancelWorkflowExecution is called by workers when they want to request cancellation of
     // a workflow execution.
     //
-    // This result in a new `WORKFLOW_EXECUTION_CANCEL_REQUESTED` event being written to the
-    // workflow history and a new workflow task created for the workflow. It returns success if requested
-    // workflow already closed. It fails with 'NotFound' if the requested workflow doesn't exist.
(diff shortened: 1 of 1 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -216,8 +216,8 @@ service WorkflowService {
     // a workflow execution.
     //
     // This result in a new `WORKFLOW_EXECUTION_CANCEL_REQUESTED` event being written to the
```
Sushisource, line 218:
```suggestion
    // This results in a new `WORKFLOW_EXECUTION_CANCEL_REQUESTED` event being written to the
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -216,8 +216,8 @@ service WorkflowService {
...
-    // workflow history and a new workflow task created for the workflow. Fails with `NotFound` if
-    // the workflow is already completed or doesn't exist.
+    // workflow history and a new workflow task created for the workflow. It returns success if requested
+    // workflow already closed. It fails with 'NotFound' if the requested workflow doesn't exist.
```
Sushisource, line 220:
```suggestion
    // workflow history and a new workflow task created for the workflow. It returns success if the requested
    // workflow is already closed. It fails with 'NotFound' if the requested workflow doesn't exist.
```

PR #184, opened 2022-05-25
Fix ResetReapplyType comments

elb3k:
**What changed?** 

Some of the comments related to `ResetReapplyType`.
Fixed comments on `ResetReapplyType`.
1. Wrote most sensible description I could come up to `enum ResetReapplyType`
1. `// Should be removed. Appears unused.` was also misleading. Because it is clearly being used by [temporal](https://github.com/temporalio/temporal) and [tctl](https://github.com/temporalio/tctl)

**Why?** It was little misleading

**How did you test it?** It is just a comment fix

**Potential risks** None


alexshtin (approved):
Thanks for fixing this.

PR #186, opened 2022-05-26
Add supports_schedules flag

dnr:
**What changed?**
Adds a bool field in NamespaceInfo for whether the new scheduled workflow feature is supported on the namespace.

**Why?**
This is intended for the web UI to use to control whether to display schedule-related elements.

### DIFF
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -47,6 +47,10 @@ message NamespaceInfo {
     // A key-value map for any customized purpose.
     map<string, string> data = 5;
     string id = 6;
+
+    // Whether scheduled workflows are supported on this namespace. This is only needed
+    // temporarily while the feature is experimental, so we can give it a high tag.
+    bool supports_schedules = 100;
 }
 
 message NamespaceConfig {


cretz:
While doing this, you can also add a `scheduling` flag to `temporal.api.workflowservice.v1.GetSystemInfoResponse.Capabilities` that you always set to `true`. This can allow the SDK to know ahead of time whether the version of server they have supports scheduling.

Granted the SDK can also just react to `NotFound` when attempting to invoke the calls on older servers. Or maybe we're not wanting to have any high-level client wrappers for these scheduling calls. Still, having the boolean in the capabilities helps for all features only newer server versions support.

dnr:
> While doing this, you can also add a `scheduling` flag to `temporal.api.workflowservice.v1.GetSystemInfoResponse.Capabilities` that you always set to `true` (feel free to come up with a better name). This can allow the SDK to know ahead of time whether the version of server they have supports scheduling.
> 
> Granted the SDK can also just react to `NotFound` when attempting to invoke the calls on older servers. Or maybe we're not wanting to have any high-level client wrappers for these scheduling calls. Still, having the boolean in the capabilities helps for all features only newer server versions support.

Sure, makes sense. We haven't really talked about higher level wrappers yet but I have some ideas.

cretz (approved):

yiminc (approved):

PR #190, opened 2022-06-09
Add worker-build-id-based versioning api

Sushisource:
Add the API as described here https://github.com/temporalio/proposals/pull/54 with a few minor changes needed while working on the initial implementation.

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -117,3 +117,28 @@ message ResetPointInfo {
     // false if the reset point has pending childWFs/reqCancels/signalExternals.
     bool resettable = 6;
 }
+
+// NewWorkflowExecutionInfo is a shared message that encapsulates all the
+// required arguments to starting a workflow in different contexts.
+message NewWorkflowExecutionInfo {
+    string workflow_id = 1;
+    temporal.api.common.v1.WorkflowType workflow_type = 2;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 3;
+    // Serialized arguments to the workflow.
+    temporal.api.common.v1.Payloads input = 4;
+    // Total workflow execution timeout including retries and continue as new.
+    google.protobuf.Duration workflow_execution_timeout = 5 [(gogoproto.stdduration) = true];
+    // Timeout of a single workflow run.
+    google.protobuf.Duration workflow_run_timeout = 6 [(gogoproto.stdduration) = true];
+    // Timeout of a single workflow task.
+    google.protobuf.Duration workflow_task_timeout = 7 [(gogoproto.stdduration) = true];
+    // Default: WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE.
+    temporal.api.enums.v1.WorkflowIdReusePolicy workflow_id_reuse_policy = 8;
+    // The retry policy for the workflow. Will never exceed `workflow_execution_timeout`.
+    temporal.api.common.v1.RetryPolicy retry_policy = 9;
+    // See https://docs.temporal.io/docs/content/what-is-a-temporal-cron-job/
+    string cron_schedule = 10;
+    temporal.api.common.v1.Memo memo = 11;
+    temporal.api.common.v1.SearchAttributes search_attributes = 12;
+    temporal.api.common.v1.Header header = 13;
+}
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -142,3 +142,22 @@ message NewWorkflowExecutionInfo {
     temporal.api.common.v1.SearchAttributes search_attributes = 12;
     temporal.api.common.v1.Header header = 13;
 }
+
+// Used by the worker versioning APIs, represents a node in the version graph for a particular
+// task queue
+message VersionIdNode {
+    VersionId version = 1;
+    // A pointer to the previous version this version is considered to be compatible with
+    VersionIdNode previous_compatible = 2;
+    // A pointer to the last incompatible version (previous major version)
+    VersionIdNode previous_incompatible = 3;
+}
+
+// Used by the worker versioning APIs, represents a specific version of something
+message VersionId {
+    oneof version {
+        // An opaque whole-worker identifier
+        string worker_build_id = 1;
+    }
+}
+
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -143,21 +143,3 @@ message NewWorkflowExecutionInfo {
     temporal.api.common.v1.Header header = 13;
 }
 
-// Used by the worker versioning APIs, represents a node in the version graph for a particular
-// task queue
-message VersionIdNode {
-    VersionId version = 1;
-    // A pointer to the previous version this version is considered to be compatible with
-    VersionIdNode previous_compatible = 2;
-    // A pointer to the last incompatible version (previous major version)
-    VersionIdNode previous_incompatible = 3;
-}
-
-// Used by the worker versioning APIs, represents a specific version of something
-message VersionId {
-    oneof version {
-        // An opaque whole-worker identifier
-        string worker_build_id = 1;
-    }
-}
-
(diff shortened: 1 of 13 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -142,3 +142,22 @@ message NewWorkflowExecutionInfo {
...
+    VersionId version = 1;
+    // A pointer to the previous version this version is considered to be compatible with
+    VersionIdNode previous_compatible = 2;
+    // A pointer to the last incompatible version (previous major version)
+    VersionIdNode previous_incompatible = 3;
+}
+
+// Used by the worker versioning APIs, represents a specific version of something
+message VersionId {
+    oneof version {
```
cretz, line 158: Should remove the `oneof`. Nnot only do we avoid it in our APIs for whatever reason (which I'd likely disagree with), but it is a backwards/forwards compatible change to introduce it in the future when you add your next field in here. See https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility_issues.

temporal/api/workflow/v1/message.proto
```diff
@@ -142,3 +142,22 @@ message NewWorkflowExecutionInfo {
     temporal.api.common.v1.SearchAttributes search_attributes = 12;
     temporal.api.common.v1.Header header = 13;
 }
+
+// Used by the worker versioning APIs, represents a node in the version graph for a particular
+// task queue
+message VersionIdNode {
```
cretz, line 148: I don't think this should be in `temporal.api.workflow.v1` because it is not about workflows necessarily (though kind is in a roundabout way, from the API POV this is about workers). I'd be hesitant to put it in `temporal.api.version.v1` because that is more about server versioning, but I wouldn't be against it. Otherwise, either `temporal.api.namespace.v1` or `temporal.api.taskqueue.v1` since those are the levels it applies at or `temporal.api.common.v1` if nowhere else works.

temporal/api/workflow/v1/message.proto
Sushisource: I know that's true in the proto sense, but it's not true in the codegen sense. That said it may very well be a long long time before this gets used for anything else, so I'd be OK taking it out.

If anyone else has an opinion chime in

mmcshane: I had a similar understanding that we avoided `oneof` until yesterday when it was recommended to me by someone on the OSS team. Maybe @yiminc could clarify the acceptibility of this feature?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -950,3 +950,47 @@ message ListSchedulesResponse {
...
+message UpdateWorkerBuildIdOrderingResponse {}
+
+// (-- api-linter: core::0134::request-resource-required=disabled
+//     aip.dev/not-precedent: GetWorkerBuildIdOrderingRequest RPC doesn't follow Google API format. --)
+message GetWorkerBuildIdOrderingRequest {
+    string namespace = 1;
+    // Must be set, the task queue to interrogate about worker id ordering
+    string task_queue = 2;
+    // Limits how deep the returned DAG will go. 1 will return only the
+    // default build id.
```
mmcshane, line 984: Maybe specify the behavior for 0 here in this comment

mmcshane (approved):

temporal/api/workflow/v1/message.proto
yiminc: We didn't use oneof because of historical reason (thrift). But I'm not aware of practical reason not to use now.

temporal/api/workflow/v1/message.proto
yiminc: is this VersionId only apply to task queue? if so, it probably should go to temporal.api.taskqueue.v1

temporal/api/workflowservice/v1/request_response.proto
yiminc: +1 for 0 value behavior

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -950,3 +950,47 @@ message ListSchedulesResponse {
...
+    bool become_default = 5;
+}
+message UpdateWorkerBuildIdOrderingResponse {}
+
+// (-- api-linter: core::0134::request-resource-required=disabled
+//     aip.dev/not-precedent: GetWorkerBuildIdOrderingRequest RPC doesn't follow Google API format. --)
+message GetWorkerBuildIdOrderingRequest {
+    string namespace = 1;
+    // Must be set, the task queue to interrogate about worker id ordering
+    string task_queue = 2;
```
yiminc, line 982: do you need task queue type here as well?

temporal/api/workflow/v1/message.proto
yiminc: if we only have one field for foreseeable future, and adding oneof in future is backward compatible change, then I would prefer to not have it now.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -950,3 +950,47 @@ message ListSchedulesResponse {
...
+    // Limits how deep the returned DAG will go. 1 will return only the
+    // default build id.
+    int32 max_depth = 3;
+}
+message GetWorkerBuildIdOrderingResponse {
+    // The currently established default ids. There will be one per version type as defined in
+    // the `VersionId` message.
+    repeated temporal.api.workflow.v1.VersionIdNode current_defaults = 1;
```
yiminc, line 990: can you have multiple defaults?

temporal/api/workflow/v1/message.proto
Sushisource: In theory it could end up applying in other places if we support workflow bundling later on. But, even then, there's at least still a reasonable connection to task queues so I'll move it there.

temporal/api/workflow/v1/message.proto
Sushisource: Sure, I'll get rid of it for now.

Adding the oneof later _will_ be a compile-breaking change like I mentioned (even if it is wire compatible), but, I expect few users will be interacting with the message directly so perhaps that's a non-issue.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -950,3 +950,47 @@ message ListSchedulesResponse {
...
 }
+
+// (-- api-linter: core::0134::request-mask-required=disabled
+//     aip.dev/not-precedent: UpdateWorkerBuildIdOrderingRequest doesn't follow Google API format --)
+// (-- api-linter: core::0134::request-resource-required=disabled
+//     aip.dev/not-precedent: UpdateWorkerBuildIdOrderingRequest RPC doesn't follow Google API format. --)
+message UpdateWorkerBuildIdOrderingRequest {
+    string namespace = 1;
+    // Must be set, the task queue to apply changes to.
+    string task_queue = 2;
```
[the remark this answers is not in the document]
Sushisource: I don't think it matters. Everything needs to apply to both types of queues.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Nope, same reason.

temporal/api/workflow/v1/message.proto
dnr: I don't have a strong opinion but `taskqueue` seems good to me too

temporal/api/workflowservice/v1/request_response.proto
dnr: Talked in person, but for reference:

Even if we add another version type, it seems to me like the current default should be one thing, of one type or another. If we keep track of a default per-type then it's unclear what a new workflow will actually run with.

Maybe prevent mixing types on a single tq at all?

dnr (approved):

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
yiminc: I think the major use case is for workflow task queue because the replay of the existing history. Activity does not have such problem. Granted, there are use case for activity to be versioned, but it is a much less of a problem compare to workflow.

PR #194, opened 2022-06-15
WorkflowUpdate and attendant types

mmcshane:
**What changed?**
Adds the WorkflowService.UpdateWorkflow RPC

**Why?**
New feature

**How did you test it?**
`make proto`

**Potential risks**


temporal/api/history/v1/message.proto
```diff
@@ -687,6 +709,9 @@ message HistoryEvent {
         SignalExternalWorkflowExecutionFailedEventAttributes signal_external_workflow_execution_failed_event_attributes = 43;
         ExternalWorkflowExecutionSignaledEventAttributes external_workflow_execution_signaled_event_attributes = 44;
         UpsertWorkflowSearchAttributesEventAttributes upsert_workflow_search_attributes_event_attributes = 45;
+        UpdateWorkflowReceivedEventAttributes update_workflow_received_event_attributes = 46;
```
cretz, line 712: Before the worker receives this event, is it visible during normal client get history requests? And is this event removed from the history if the worker does _not_ accept it?

If the answers are "yes" and "yes", is this the first non-reset time we are rewriting past history?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -950,3 +953,43 @@ message ListSchedulesResponse {
...
+    // Headers that are returned from the update to the processing workflow
+    temporal.api.common.v1.Header header = 1;
+    // An opaque token that can be used to retrieve the update result via
+    // polling if it is not returned as part of the gRPC response
+    string update_token = 2;
+    // The success or failure status of the update
+    oneof result {
+        temporal.api.success.v1.Success success = 3;
+        temporal.api.failure.v1.Failure failure = 4;
+    }
```
cretz, line 994: On query failure we let that be a gRPC failure (unless it's a query rejection). Do we not want that to be the case here too?

temporal/api/success/v1/message.proto
```diff
@@ -0,0 +1,38 @@
...
+option go_package = "go.temporal.io/api/success/v1;success";
+option java_package = "io.temporal.api.success.v1";
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporal::Api::Success::V1";
+option csharp_namespace = "Temporal.Api.Success.V1";
+
+import "temporal/api/common/v1/message.proto";
+
+message Success {
```
cretz, line 36: I assume this is a whole separate object because you think there may be more than just payloads in the future? Instead of a whole new generic "success" package and object that is only used in one place, maybe a `temporal.api.update.v1.WorkflowUpdateResponse` would be more tailored to the future use of this. Otherwise, if this is gonna remain generic, there probably won't be any extra fields ever added for "all successes" that won't be in the actual API response.

temporal/api/update/v1/message.proto
```diff
@@ -0,0 +1,39 @@
...
+option go_package = "go.temporal.io/api/update/v1;update";
+option java_package = "io.temporal.api.update.v1";
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporal::Api::Update::V1";
+option csharp_namespace = "Temporal.Api.Update.V1";
+
+import "temporal/api/common/v1/message.proto";
+
+message WorkflowUpdate {
```
cretz, line 36: So I noticed signal name and input we just put in the gRPC request, but query name and input _and_ headers we put in a separate object in a separate package. Here we are putting name and input in a separate object but headers in the gRPC request. Any reason why not headers here instead?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -950,3 +953,43 @@ message ListSchedulesResponse {
...
+message UpdateWorkflowRequest {
+    // Headers that are passed with the update to the processing workflow.
+    // These can include things like auth or tracing tokens.
+    temporal.api.common.v1.Header header = 1;
+
+    // A unique token for this logical request
+    string request_id = 2;
+
+    // The manner in which the update result will be accessed
+    temporal.api.enums.v1.UpdateWorkflowResultAccessStyle result_access_style = 3;
```
cretz, line 968: Can you clarify what happens if left to default/unset?

temporal/api/enums/v1/update.proto
```diff
@@ -0,0 +1,40 @@
...
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "UpdateProto";
+option ruby_package = "Temporal::Api::Enums::V1";
+option csharp_namespace = "Temporal.Api.Enums.V1";
+
+enum UpdateWorkflowResultAccessStyle {
+    UPDATE_WORKFLOW_RESULT_ACCESS_STYLE_UNSPECIFIED = 0;
+    UPDATE_WORKFLOW_RESULT_ACCESS_STYLE_PREFER_INLINE = 1;
```
cretz, line 36: If it's a simple description, just to help me out, can you clarify when "prefer inline" turns to "poll" and what the response looks like there? I see no timeout on the update workflow to affect this.

Or, is the absence of a separate call for "poll workflow update" mean that this doesn't have value yet?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -950,3 +953,43 @@ message ListSchedulesResponse {
...
+    // arguments to pass to said function.
+    temporal.api.update.v1.WorkflowUpdate update = 7;
+}
+
+message UpdateWorkflowResponse {
+    // Headers that are returned from the update to the processing workflow
+    temporal.api.common.v1.Header header = 1;
+    // An opaque token that can be used to retrieve the update result via
+    // polling if it is not returned as part of the gRPC response
+    string update_token = 2;
```
cretz, line 989: I assume "poll" is a not-yet-present API call coming in the future?

cretz (commented):
Sorry if I wasn't supposed to review yet! (I know my name wasn't added, but figured I'd give feedback anyways).

Also, where are the "commands"? Are they coming later?

temporal/api/enums/v1/update.proto
Sushisource: +1 to some docstrings on these bad boys

temporal/api/success/v1/message.proto
mmcshane: Just need a sanely-named type to serve as the other side of a Failure object in an `anyof`. For me, the Success and Failure types would ideally have been in a `result` package (or similar) but we decided to follow the established pattern. Going forward, `anyof { Success | Failure }` will be the expected style for tunneled things that can fail.

temporal/api/enums/v1/update.proto
mmcshane: Correct that only REQUIRE_INLINE will be supported in the first drop.

I'm just going to remove the other two values for the time being to decrease noise.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: This is correct.

mmcshane:
> [...] where are the "commands"? Are they coming later?

Yes, first drop will just be external fast-path invocations (i.e. gRPC clients, not workflow-to-workflow) so commands are not needed yet.

temporal/api/history/v1/message.proto
cretz: No prob, if I as a user can't get it from history until it's legitimately accepted then my concerns are alleviated.

temporal/api/workflowservice/v1/request_response.proto
cretz: Makes perfect sense to me to have worker error be represented as a failure here. It might be worth documenting when you get a gRPC failure vs when you get a gRPC success + failure in response.

temporal/api/success/v1/message.proto
cretz: > Just need a sanely-named type to serve as the other side of a Failure object in an anyof.

I think this sane name can be the field name, it doesn't have to be the message name. Not convinced we need this single message wrapper here.

cretz:
> Yes, first drop will just be external fast-path invocations (i.e. gRPC clients, not workflow-to-workflow) so commands are not needed yet.

I would think it'd need commands no matter what. How does a worker provide a response for an update back to the server?

temporal/api/workflowservice/v1/request_response.proto
alexshtin: Another open question here is if we need new failure info for this or not. Update handler code need to have an ability to indicate that server shouldn't persist the update reject. It means that server needs to be aware of it similar to `NonRetryable` field in `ApplicationFailureInfo`. I see two options:
1. add something like `UpdateWorkflowFailureInfo` (with corresponding error/exception in all SDKs), or
2. add `SkipPersistence` (or better name) field to `AppicationFailureInfo`.

I like well defined separate failure info (1) more, but it might be more confusing for users (update handler authors) that some specific error must be returned to indicate `SkipPersistence` behaviour. With option (2) we can just add one more `ApplicationError` constructor and leave it `false` by default.

cretz: > So ideally, serviceerror.QueryFailed shouldn't exist but QueryWorkflowResponse should have failure field on it.

If query failed _only_ represents the actual error from query handler the worker I agree. Otherwise, all other failures should remain gRPC failures. I think everyone's on the same page here. I think we just need to document gRPC failure vs gRPC success + response failure.

> Another open question here is if we need new failure info for this or not. [...]

I was under the impression from the SDK side that the side-effect-free "update validator" that is _optional_ and runs _prior_ to the actual update would tell whether the worker rejects the update pre-persistence. I don't think, at least from the SDK POV, that we should combine the pre-process-update/validate code and the process-update code for multiple reasons:

1. The pre-process code must be side effect free and it can be hard for users to properly do that if you're asking them "just make sure you don't do side effects until _after_ you've returned this special kind of error"
2. Many users, myself included, will probably never have pre-process code. We don't for signals and it works just fine for almost everyone.
3. Having an update handler have essentially three return choices of no-persist error, persist error, and persist success is confusing.

temporal/api/success/v1/message.proto
mmcshane: I think the "symmetry" was in the ability to add fields to the value being used. We can add to Failure but not really to Payloads. It still feels strange to me to have a a typed and future-extensible value on the failure side and effectively "whatever" on the success side (fully admitting that `success.v1.Success` is also weird but at least consistent with how Failure is defined). But I can take a hint - I'll remove the `success.v1.Success` and drop this field to `Payloads`

temporal/api/workflowservice/v1/request_response.proto
mmcshane: > I was under the impression from the SDK side that the side-effect-free "update validator" that is optional and runs prior to the actual update would tell whether the worker rejects the update pre-persistence.

The requirement for SDKs is that it be _possible_ to indicate a side-effect-free rejection and request that the update (and rejection) not be persisted. What you describe is one way of achieving that but it's not the only way. There are SDK owners who dislike the two-function approach.

When update rejection is transmitted back to the server, we definitely need a skip/dont-skip flag for persistence in that communication.

cretz: True, but the comment I was responding to was talking about handler authors and error constructors and such.

From the pure API POV, I can reserve comment until the command-side appears (I assume that's a separate PR?).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -995,3 +997,38 @@ message GetWorkerBuildIdOrderingResponse {
...
+}
+
+message UpdateWorkflowResponse {
+    // An opaque token that can be used to retrieve the update result via
+    // polling if it is not returned as part of the gRPC response
+    string update_token = 1;
+    // The success or failure status of the update
+    oneof result {
+        temporal.api.common.v1.Payloads success = 2;
+        temporal.api.failure.v1.Failure failure = 3;
```
cretz, line 1032: Similar to how it should be in command response, I think it's good to separate non-persisted rejection from persisted failure, even if we don't care in the SDK. Can change this and the SDK to have a third result field of `temporal.api.failure.v1.Failure rejection` or can use that enum.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -995,3 +997,38 @@ message GetWorkerBuildIdOrderingResponse {
...
+
+    // The name under which the workflow update function is registered and the
+    // arguments to pass to said function.
+    temporal.api.update.v1.WorkflowUpdate update = 6;
+}
+
+message UpdateWorkflowResponse {
+    // An opaque token that can be used to retrieve the update result via
+    // polling if it is not returned as part of the gRPC response
+    string update_token = 1;
```
cretz, line 1028: Question - is `UpdateWorkflowRequest.request_id` the same as history `UpdateWorkflowXAttributes.update_id` and are either related to this `update_token` at all? Seems the user specified identifier could be used for all 3 of these and could be inside the `WorkflowUpdate` message.

temporal/api/workflowservice/v1/request_response.proto
cretz: Marking resolved under the impression that SDK returned failures are the failures here and that actual server failures are normal gRPC failures.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -995,3 +997,38 @@ message GetWorkerBuildIdOrderingResponse {
...
+
+    // The manner in which the update result will be accessed.
+    // This field requires a non-default value; the default value of the enum
+    // will result in an error.
+    temporal.api.enums.v1.UpdateWorkflowResultAccessStyle result_access_style = 2;
+
+    // The namespace name of the target workflow
+    string namespace = 3;
+    // The target workflow id and a specific run thereof
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 4;
```
Sushisource, line 1014: This semi-implies you must specify a run id. Will it work the way most stuff does where it'll select most recent if run is left off? Might be worth changing the comment a bit if so.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Agree, that info could be useful in some situations to a user.

Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
mmcshane: The update_token includes the request_id but also the namespace+workflow+run_id info so that polling with just the token can be routed to the right history shard. 

I think request_id could be omitted from the event attributes - I had done that when I thought we were doing real deduplication.

alexshtin: I would do the opposite and use `request_id` in both places. And yes, I think it can be useful for update handler authors.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -995,3 +997,38 @@ message GetWorkerBuildIdOrderingResponse {
     repeated temporal.api.taskqueue.v1.VersionIdNode compatible_leaves = 2;
 }
 
+// (-- api-linter: core::0134=disabled
+//     aip.dev/not-precedent: Updae RPCs don't follow Google API format. --)
+message UpdateWorkflowRequest {
+    // A unique token for this logical request
```
alexshtin, line 1003:
```suggestion
    // A unique id for this logical request that is passed to the worker.
```

cretz: I think it should be a unique ID for this "update" instead of for this "request" since it is for the lifetime of the update and travels with the update. Similar to `activity_id` instead of `schedule_id`.

mmcshane: It was a typo to say "token" in the comment here - ID is correct. But this ID is not passed to the worker.

Since the original RFC and even prior, the update ID has been server-owned and generated so that we can guarantee uniqueness, and that if the client has received an update ID (or token, from the client's perspective they're the same thing) then the update request either has already been delivered or is guaranteed to be delivered to the workflow code.

cretz: I wonder, when we design the poll RPC call, if this decision will be revisited. It can be reasonable to ask "Why can't I poll for an update response using info I gave on update request?"

temporal/api/workflowservice/v1/request_response.proto
mmcshane: Clarified that run_id is optional.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: If that comes up then we can enhance the Poll call. But my preference is to start with the smaller thing first.

cretz: And just to confirm, we are ok with users not being able to tie an update in history to its request (without using custom arguments and checking the input manually)?

temporal/api/workflowservice/v1/request_response.proto
cretz: Just a user-defined update ID everywhere is fine IMO. Doesn't need to have anything to do with a request. (though we may just be talking about naming)

PR #197, opened 2022-07-07
Add batch operation APIs to operator

yux0:
**What changed?**
Add batch operation APIs to operator

**Why?**

**How did you test it?**

**Potential risks**


temporal/api/batch/v1/message.proto
```diff
@@ -0,0 +1,41 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
+
+package temporal.api.batch.v1;
```
cretz, line 25: Do these tiny messages deserve their own package? They seem like they are just messages for starting batch and have no other use. Is there another expected use? Can they not live with the start request? Is the only reason they are in a separate package because we, for whatever reason, are unwilling to use nested messages and, for whatever reason, are unwilling to have additional needed messages that are non-request/response in the request/response proto?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,48 @@ message ListClusterMembersResponse {
...
     bytes next_page_token = 2;
 }
+
+message StartBatchOperationRequest {
+    string namespace = 1;
+    temporal.api.enums.v1.BatchOperationType operation_type = 2;
+    string workflow_query = 3;
+    int32 rps = 4;
+    int32 concurrency = 5;
+    oneof parameter {
```
cretz, line 171: To me, "kind", "type", "operation", etc make more sense than "parameter". But obviously that's subjective and doesn't matter much.

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,48 @@ message ListClusterMembersResponse {
     repeated temporal.api.cluster.v1.ClusterMember active_members = 1;
     bytes next_page_token = 2;
 }
+
+message StartBatchOperationRequest {
+    string namespace = 1;
+    temporal.api.enums.v1.BatchOperationType operation_type = 2;
+    string workflow_query = 3;
```
cretz, line 168: Can you add some documentation on what these fields mean and can accept?

temporal/api/operatorservice/v1/service.proto
```diff
@@ -96,4 +96,20 @@ service OperatorService {
     // ListClusterMembers returns information about Temporal cluster members.
     rpc ListClusterMembers(ListClusterMembersRequest) returns (ListClusterMembersResponse) {
     }
+
+    // StartBatchOperation starts a new batch operation
+    rpc StartBatchOperation(StartBatchOperationRequest) returns (StartBatchOperationResponse){
```
cretz, line 101: Suggest space between final paren and open brace

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,48 @@ message ListClusterMembersResponse {
...
+message DescribeBatchOperationResponse {
+    string operation_type = 1;
+    string total_workflow_count = 2;
+    string processed_workflow_count = 3;
+}
+
+message ListBatchOperationsRequest {
+    string namespace = 1;
+    int32 page_size = 2;
+    bytes next_page_token = 3;
```
cretz, line 203: What can this value be if not returned in `ListBatchOperationsResponse`/

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,48 @@ message ListClusterMembersResponse {
...
+message StopBatchOperationResponse {
+}
+
+message DescribeBatchOperationRequest {
+    string namespace = 1;
+    string job_id = 2;
+}
+
+message DescribeBatchOperationResponse {
+    string operation_type = 1;
```
cretz, line 195: Why string and not the enum used in start?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,48 @@ message ListClusterMembersResponse {
...
+    string total_workflow_count = 2;
+    string processed_workflow_count = 3;
```
cretz, line 197: Will operations always be workflows? Maybe `total_operation_count` and `complete_operation_count` instead?

temporal/api/batch/v1/message.proto
```diff
@@ -0,0 +1,41 @@
...
+option ruby_package = "Temporal::Api::Batch::V1";
+option csharp_namespace = "Temporal.Api.Batch.V1";
+
+message BatchOperationTermination {
+  string reason = 1;
+}
+
+message BatchOperationSignal {
+  string name = 1;
+  string input = 2;
```
cretz, line 40: This should probably be payloads. Should we just make sure this accepts everything `temporal.api.workflowservice.v1.SignalWorkflowExecutionRequest` does except the specific workflow execution? (that also justifies moving this to that package, see later comment)

temporal/api/operatorservice/v1/service.proto
cretz, line 101: Why not make these part of the workflow service? Not only could you reuse some of the operation request/response values there, it just makes sense to have the RPC calls for, say, signalling one or signalling many workflows together.

temporal/api/operatorservice/v1/request_response.proto
yux0: Updated to enum

temporal/api/operatorservice/v1/service.proto
yux0: Cleaned up

cretz: Strange to me that I can do the non-batch thing from one service and have to go to the other to do the batch thing. Signalling workflows should be in the workflow service, whether it's one or many workflows IMO.

dnr: I think this should be in workflowservice. It's for regular users, not cluster administrators. Basically, if it's something that a user could do themselves (perhaps badly) using the other apis in workflowservice, it should be in workflowservice.

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,49 @@ message ListClusterMembersResponse {
...
+}
+
+message ListBatchOperationsRequest {
+    string namespace = 1;
+    int32 page_size = 2;
+    bytes next_page_token = 3;
+}
+
+message ListBatchOperationsResponse {
+    repeated string job_id = 1;
```
dnr, line 207: Might we ever want to return extra information here?

temporal/api/operatorservice/v1/request_response.proto
yiminc: I would change it to visibility_query, or list_workflow_query, to avoid confusing with workflow query.

temporal/api/batch/v1/message.proto
```diff
@@ -0,0 +1,41 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporal::Api::Batch::V1";
+option csharp_namespace = "Temporal.Api.Batch.V1";
+
+message BatchOperationTermination {
+  string reason = 1;
+}
+
+message BatchOperationSignal {
```
yiminc, line 38: do you need one for cancel?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,48 @@ message ListClusterMembersResponse {
     repeated temporal.api.cluster.v1.ClusterMember active_members = 1;
     bytes next_page_token = 2;
 }
+
+message StartBatchOperationRequest {
+    string namespace = 1;
+    temporal.api.enums.v1.BatchOperationType operation_type = 2;
+    string workflow_query = 3;
+    int32 rps = 4;
+    int32 concurrency = 5;
```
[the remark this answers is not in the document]
yiminc: Maybe some reasonable default value and server side dynamic config is better?

temporal/api/operatorservice/v1/service.proto
yiminc: We discussed this with multiple parties (prod / cloud), and we think operator is a better place. Because we don't expect this to be a common operation that workflow needs. And the scope of this operation could be big so we don't want this to be exposed to everyone. The operator does not need to be cluster admin, it can be namespace admin.

cretz: Are y'all gonna make sure this has parity with all the options from workflow service? For example, headers.

I missed the discussion. Personally, I don't think services should be separated by "who might execute them" but "what the service does". If decided already, ok. But I would guess, if we give these a high-level interface in the SDK, they will sit right next to the RPC calls that are in a completely different service in the protos.

temporal/api/operatorservice/v1/request_response.proto
yux0: Good point. I will return a struct and flexible to add extra info.

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +161,65 @@ message ListClusterMembersResponse {
...
+}
+
+message ListBatchOperationsRequest {
+    string namespace = 1;
+    int32 page_size = 2;
+    bytes next_page_token = 3;
+}
+
+message ListBatchOperationsResponse {
+    repeated temporal.api.workflow.v1.WorkflowExecutionInfo operations = 1;
```
dnr, line 206: I meant info specific to the batch operation, like maybe an approximate progress indicator, any errors encountered, etc. Stuff that could be useful when listing these in the UI. But maybe we don't need that.

temporal/api/batch/v1/message.proto
yux0: I added a placeholder for cancel. Currently, it doesn't require extra input.

temporal/api/operatorservice/v1/request_response.proto
yux0: It will be in the describe batch operation and return per batch operation job. For list, I updated to list with info: id, state, start time and close time.

yiminc (approved):

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +160,97 @@ message ListClusterMembersResponse {
...
+
+message BatchOperationSignal {
+    // The workflow author-defined name of the signal to send to the workflow
+    string signal_name = 1;
+    // Serialized value(s) to provide with the signal
+    temporal.api.common.v1.Payloads input = 2;
+    // Headers that are passed with the signal to the processing workflow.
+    // These can include things like auth or tracing tokens.
+    temporal.api.common.v1.Header header = 3;
+}
```
cretz, line 251: This is missing the `identity` value (though maybe you want it at the batch level, unsure). Also, can we add a comment here and in the API to keep this and `temporal.api.workflowservice.v1.SignalWorkflowExecutionRequest` in sync with each other?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +160,97 @@ message ListClusterMembersResponse {
...
+    // Batch operation start time
+    google.protobuf.Timestamp start_time = 3 [(gogoproto.stdtime) = true];
+    // Batch operation close time
+    google.protobuf.Timestamp close_time = 4 [(gogoproto.stdtime) = true];
+}
+
+message BatchOperationTermination {
+    // Reason of terminate workflows
+    string reason = 1;
+}
```
cretz, line 241: This is missing the ability to provide `details`, `identity`, and `first_execution_run_id` (or we can explicitly comment we don't want that here since that's single-termination specific and not here). Also, can we add a comment here and in the API to keep this and `temporal.api.workflowservice.v1.TerminateWorkflowExecutionRequest` in sync with each other?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +160,97 @@ message ListClusterMembersResponse {
...
+    temporal.api.common.v1.Payloads input = 2;
+    // Headers that are passed with the signal to the processing workflow.
+    // These can include things like auth or tracing tokens.
+    temporal.api.common.v1.Header header = 3;
+}
+
+message BatchOperationCancellation {
+    // Reason of cancel workflows
+    string reason = 1;
+}
```
cretz, line 256: This is missing the `identity` value (though maybe you want it at the batch level, unsure). Also, can we add a comment here and in the API to keep this and `temporal.api.workflowservice.v1.RequestCancelWorkflowExecutionRequest` in sync with each other?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +160,97 @@ message ListClusterMembersResponse {
     repeated temporal.api.cluster.v1.ClusterMember active_members = 1;
     bytes next_page_token = 2;
 }
+
+message StartBatchOperationRequest {
+    // Namespace contains the batch operation
+    string namespace = 1;
+    // Batch operation type
+    temporal.api.enums.v1.BatchOperationType operation_type = 2;
```
cretz, line 168: This enum appears redundant. You already have the operation, why do you need a type too?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +160,97 @@ message ListClusterMembersResponse {
...
+message BatchOperationInfo {
+    // Batch job ID
+    string job_id = 1;
+    // Batch operation type
+    temporal.api.enums.v1.BatchOperationState state = 2;
+    // Batch operation start time
+    google.protobuf.Timestamp start_time = 3 [(gogoproto.stdtime) = true];
+    // Batch operation close time
+    google.protobuf.Timestamp close_time = 4 [(gogoproto.stdtime) = true];
+}
```
cretz, line 236: I think you should provide the failure here, and in the future when we offer something like batch update/query, you might want to offer success values too.

temporal/api/operatorservice/v1/request_response.proto
alexshtin: At least in go, it is inconvenient to get operation type using `oneof` field. This is why we use similar approach for events and commands. In this case, it is probably not that bad.

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -159,3 +160,97 @@ message ListClusterMembersResponse {
...
+    bytes next_page_token = 3;
+}
+
+message ListBatchOperationsResponse {
+    // BatchOperationInfo contains the basic info about batch operation
+    repeated BatchOperationInfo operation_info = 1;
+    bytes next_page_token = 2;
+}
+
+message BatchOperationInfo {
```
alexshtin, line 227: This file called `request_response` because it contains only `Request` and `Response` objects. Can you move all the rest out? I am ok with `batch_operation.proto` in the same package, but general approach is to move them to special package.

alexshtin, line 227: Why this is different from `DescribeBatchOperationResponse`?

alexshtin (changes requested):

temporal/api/operatorservice/v1/request_response.proto
cretz: We shouldn't alter our protos because people don't want to do a switch. Generated protos in most languages have easy ways to determine this. Having said that, I know we have this redundancy elsewhere, so I can understand an argument for consistency. It's still unnecessary though.

temporal/api/operatorservice/v1/request_response.proto
cretz: I figured it was called "request_response" because it contains request and response objects, and extra needed messages. Why move things out that are never used anywhere else? I don't understand some of these rules. Can the developer make this a child message of the only place it's used instead of unnecessarily moving this once-used item to another package?

yux0: I think it is clean to have in a different proto. Updated.

yux0: This uses in list api and return basic info about the batch operation. The describe batch operation returns the details.

temporal/api/operatorservice/v1/request_response.proto
yux0: Do you mean the success/ failure of each operation? They will be provided in the describe response.

temporal/api/operatorservice/v1/request_response.proto
yux0: Updated

temporal/api/operatorservice/v1/request_response.proto
yux0: Updated

temporal/api/operatorservice/v1/request_response.proto
yux0: Updated

temporal/api/operatorservice/v1/request_response.proto
cretz: My mistake here, I thought this represented a single piece of the batch. If we stick with termination, cancellation, and signals then we probably don't need to record batch errors separately. However if we do synchronous updates which have unpersisted errors sometimes.

temporal/api/operatorservice/v1/request_response.proto
Sushisource: I know this is "resolved" but I'd like to throw my 2c out there that I think this is a _really_ bad idea. It allows you to construct invalid proto messages where type says one thing and the operation says another.

I have a hard time interpreting that any other way than as poor design.

Consistency is, I think, a very poor reason to accept bad design. Backwards compatibility, sure, that's a good reason, but we shouldn't repeat the same bad patterns now just because we used them in the past.

cretz: > poor design

Yup. Unfortunately we have constantly accepted poor design in favor of consistency (which I disagree with, because I almost always disagree with poor design). This also occurs for messages that are only needed to properly structure request/response types, but sadly we disallow them to live next to their parent request/response type. We also disallow nested messages for no good reason.

temporal/api/operatorservice/v1/request_response.proto
cretz: Disagree personally, but not a big deal I guess.

temporal/api/operatorservice/v1/request_response.proto
yux0: I will remove this to only keep the operation.

temporal/api/enums/v1/batch_operation.proto
```diff
@@ -0,0 +1,46 @@
...
+package temporal.api.enums.v1;
+
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "BatchOperationProto";
+option ruby_package = "Temporal::Api::Enums::V1";
+option csharp_namespace = "Temporal.Api.Enums.V1";
+
+enum BatchOperationType {
```
dnr, line 34: can this be removed now that we're just using the `oneof`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1037,3 +1039,72 @@ message UpdateWorkflowResponse {
...
+    // Total operation count
+    int64 total_operation_count = 6;
+    // Complete operation count
+    int64 complete_operation_count = 7;
+    // Failure operation count
+    int64 failure_operation_count = 8;
+}
+
+message ListBatchOperationsRequest {
+    // Namespace contains the batch operation
```
dnr, line 1098:
```suggestion
    // Namespace that contains the batch operation
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1037,3 +1039,72 @@ message UpdateWorkflowResponse {
...
+    // Batch job id
+    string job_id = 2;
+}
+
+message DescribeBatchOperationResponse {
+    // Batch operation type
+    temporal.api.enums.v1.BatchOperationType operation_type = 1;
+    // Batch job ID
+    string job_id = 2;
+    // Batch operation type
```
dnr, line 1083:
```suggestion
    // Batch operation state
```

temporal/api/workflowservice/v1/batch_operation.proto
```diff
@@ -0,0 +1,86 @@
...
+import "google/protobuf/timestamp.proto";
+
+
+import "temporal/api/common/v1/message.proto";
+import "temporal/api/enums/v1/batch_operation.proto";
+
+message BatchOperationInfo {
+  // Batch job ID
+  string job_id = 1;
+  // Batch operation type
```
dnr, line 44:
```suggestion
  // Batch operation state
```

temporal/api/enums/v1/batch_operation.proto
yux0: Still used in describe response to indicate the operation type

temporal/api/workflowservice/v1/request_response.proto
yux0: Updated

temporal/api/workflowservice/v1/request_response.proto
yux0: Updated

temporal/api/workflowservice/v1/batch_operation.proto
yux0: Updated

temporal/api/enums/v1/batch_operation.proto
dnr: Can the describe response just include the request itself, instead of summarizing it like this? It's not great that we have to keep this in sync with the `oneof` fields

temporal/api/workflowservice/v1/batch_operation.proto
```diff
@@ -0,0 +1,86 @@
+// The MIT License
```
cretz, line 1: Not sure we need a non-service, non-req/resp file in this folder. Any reason to separate the files here? I don't care _that_ much and easy to move around of course, just curious.

cretz (approved):

temporal/api/workflowservice/v1/batch_operation.proto
yux0: The other file named request_response.proto and it should contains requests and responses. I think it is cleaner to have specific protos in different files. Agree this can move around so I don't have a strong opinion on this.

cretz: Yeah, I don't have a strong opinion. This is good. I am quite happy that these only-for-req/resp messages are in this proto package.

alexshtin (approved):

temporal/api/enums/v1/batch_operation.proto
```diff
@@ -0,0 +1,46 @@
+// The MIT License
```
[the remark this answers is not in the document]
yux0: Moved to message.proto.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1040,3 +1042,72 @@ message UpdateWorkflowResponse {
...
+    temporal.api.enums.v1.BatchOperationType operation_type = 1;
+    // Batch job ID
+    string job_id = 2;
+    // Batch operation state
+    temporal.api.enums.v1.BatchOperationState state = 3;
+    // Batch operation start time
+    google.protobuf.Timestamp start_time = 4 [(gogoproto.stdtime) = true];
+    // Batch operation close time
+    google.protobuf.Timestamp close_time = 5 [(gogoproto.stdtime) = true];
```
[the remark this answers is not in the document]
yux0: From impl detail, the list only return basic workflow information from list workflow response. The BatchOperationInfo contains workflow information from describe workflow or query workflow.

temporal/api/workflowservice/v1/batch_operation.proto
yux0: Updated

PR #199, opened 2022-07-12
Add namespace_division field

dnr:
**What changed?**
Add the first set of fields needed for namespace divisions.

**Why?**
Namespace divisions are initially a visibility concept that will let the server keep certain internal workflows separate from user workflows in the same namespace. It will be used for the new scheduler and batch operation workflows, and won't be available to users yet. All workflows will have a namespace division, a string, which will be the empty string by default. Visibility queries will be limited to a single division at a time. Eventually we may extend other namespace features to namespace divisions, such as resource limits and other dynamic config.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -98,6 +98,8 @@ message WorkflowExecutionStartedEventAttributes {
     // It should be used together with parent_initiated_event_id to identify
     // a child initiated event for global namespace
     int64 parent_initiated_event_version = 26;
+    // The namespace division this workflow exists in
+    string namespace_division = 28;
 }
 
 message WorkflowExecutionCompletedEventAttributes {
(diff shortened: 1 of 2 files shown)


bergundy (commented):
Do we want to expose this in sdks at all? Is this a server only feature?
The proto comments could use more clarification.

cretz:
Hrmm, I wonder if we can use search attributes (especially since we can filter w/out ES these days). It just seems like users have the ability to filter workflows based on their visibility/listing needs, and for some reason that ability isn't good enough for us to reuse.

dnr:
> Hrmm, I wonder if we can use search attributes (especially since we can filter w/out ES these days)

We can't filter on search attributes without ES today, but we will soon be able to. I agree it makes sense to reevaluate since we're shuffling around some work. Consider this pending for the moment

dnr:
We did this with a special search attribute for now

PR #204, opened 2022-07-19
Add failure Payloads

bergundy:
During SDK team standup, we'd discussed the need to allow opting into encrypting error messages and stack traces originating in user code.

This change is the first required block to allow this functionality.
Server should not use the new field, it should only be used by failures originating in the SDK.

Once the server repo is updated with these new protos, we should add a new server capability so clients can know they can start encrypting errors.

### DIFF
--- a/temporal/api/failure/v1/message.proto
+++ b/temporal/api/failure/v1/message.proto
@@ -90,14 +90,14 @@ message Failure {
     // message.
     //
     // SDK authors: 
-    // - The SDK should provide an overridable default `encodeFailureAttributes` and `decodeFailureAttributes`
-    //   implementation, those methods may mutate the original failure.
-    // - The default implementation uses a JSON object to represent `{ message, stack_trace }`.
-    //   - The original message is overridden with "Encoded failure" to indicate that more information could be
-    //     extracted
+    // - The SDK should provide a default `encodeFailureAttributes` and `decodeFailureAttributes` implementation that:
+    //   - Uses a JSON object to represent `{ message, stack_trace }`.
+    //   - Overwrites the original message with "Encoded failure" to indicate that more information could be extracted.
+    //   - Overwrites the original stack_trace with an empty string.
     //   - The resulting JSON object is converted to Payload using the default PayloadConverter and should be processed
     //     by the user-provided PayloadCodec
     //
+    // - If there's demand, we could allow overriding the default SDK implementation to encode other opaque Failure attributes.
     // (-- api-linter: core::0203::optional=disabled --)
     temporal.api.common.v1.Payload encoded_attributes = 20;
     Failure cause = 4;
--- a/temporal/api/failure/v1/message.proto
+++ b/temporal/api/failure/v1/message.proto
@@ -80,8 +80,18 @@ message ChildWorkflowExecutionFailureInfo {
 
 message Failure {
     string message = 1;
+    // The source this Failure originated in, e.g. TypeScriptSDK / JavaSDK
+    // In some SDKs this is used to rehydrate the stack trace into an exception object.
     string source = 2;
     string stack_trace = 3;
+    // Alternative way to supply `message` and `stack_trace`, used for encryption of errors originating in user code
+    // which might contain sensitive information. Mutually exclusive with `message` and `stack_trace`.
+    // SDK authors: 
+    // * Use one Payload for `message` and one for `stack_trace`
+    // * If only the one Payload is provided, it is assumed to be the message
+    // * Both Payloads are optional and if omitted the rehydrated error will assume an empty string for both
+    // * These strings MUST be created with the default PayloadConverter and MAY be processed by a PayloadCodec
+    temporal.api.common.v1.Payloads attributes = 20;
     Failure cause = 4;
     oneof failure_info {
         ApplicationFailureInfo application_failure_info = 5;
--- a/temporal/api/failure/v1/message.proto
+++ b/temporal/api/failure/v1/message.proto
@@ -87,10 +87,12 @@ message Failure {
     // Alternative way to supply `message` and `stack_trace`, used for encryption of errors originating in user code
     // which might contain sensitive information. Mutually exclusive with `message` and `stack_trace`.
     // SDK authors: 
-    // * Use one Payload for `message` and one for `stack_trace`
-    // * If only the one Payload is provided, it is assumed to be the message
-    // * Both Payloads are optional and if omitted the rehydrated error will assume an empty string for both
-    // * These strings MUST be created with the default PayloadConverter and MAY be processed by a PayloadCodec
+    // - Use one Payload for `message` and one for `stack_trace`
+    // - If only the one Payload is provided, it is assumed to be the message
+    // - Both Payloads are optional and if omitted the rehydrated error will assume an empty string for both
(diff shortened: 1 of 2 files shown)


temporal/api/failure/v1/message.proto
```diff
@@ -80,8 +80,20 @@ message ChildWorkflowExecutionFailureInfo {
...
+    // Alternative way to supply `message` and `stack_trace`, used for encryption of errors originating in user code
+    // which might contain sensitive information. Mutually exclusive with `message` and `stack_trace`.
+    // SDK authors: 
+    // - Use one Payload for `message` and one for `stack_trace`
+    // - If only the one Payload is provided, it is assumed to be the message
+    // - Both Payloads are optional and if omitted the rehydrated error will assume an empty string for both
+    // - These strings MUST be created with the default PayloadConverter and MAY be processed by a PayloadCodec
+    //
+    // (-- api-linter: core::0203::optional=disabled --)
+    temporal.api.common.v1.Payloads attributes = 20;
```
cretz, line 96: I don't think using positional payloads is the best way. Alternative options:

* Make the payload be a serialized form of `temporal.api.failure.v1.Failure` instead of these rules about positional payloads
* Create a new proto for what this could be
* Specify a JSON object with certain expected field names

All 3 of those options are better structurally than "optional payload 1 is message, optional payload 2 is stack trace".

cretz, line 96: Can you clarify when an SDK might use this? All the time or only opt-in? I think it's worth a discussion if all the time.

lorensr: Opt-in. See: https://github.com/temporalio/sdk-features/issues/74

temporal/api/failure/v1/message.proto
```diff
@@ -80,8 +80,20 @@ message ChildWorkflowExecutionFailureInfo {
 
 message Failure {
     string message = 1;
+    // The source this Failure originated in, e.g. TypeScriptSDK / JavaSDK
+    // In some SDKs this is used to rehydrate the stack trace into an exception object.
     string source = 2;
     string stack_trace = 3;
+    // Alternative way to supply `message` and `stack_trace`, used for encryption of errors originating in user code
```
lorensr, line 87: Any chance someone will want to encrypt other strings in this file?

`ActivityType.name, ApplicationFailureInfo.type, source, namespace, identity`

temporal/api/failure/v1/message.proto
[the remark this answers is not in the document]
bergundy: Yeah that would work

bergundy: The reason I went with positionals is because codecs convert Payloads to Payloads so that multiple payloads could be compressed together but a JSON / proto object with a single payload would also achieve that.
I'll change.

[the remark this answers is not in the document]
bergundy: Not sure about the name `encoded_failure`, then you have `failure.encoded_failure`.
How about `encoded_attributes`?

temporal/api/failure/v1/message.proto
bergundy: Could be...
Let me try to figure out a way to encode arbitrary - possibly nested - attributes.
One approach that comes to mind is to expose a method on the data converter for applying these encoded attributes to failure objects where the default implementation applies only the message and stack trace.

temporal/api/failure/v1/message.proto
Sushisource: I'd say we should prefer a new proto msg.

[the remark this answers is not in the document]
Sushisource: That's fine, or `encoded_details`

bergundy: Updated to us defaulting to using JSON and allowing users to customize this behavior and possible encode any other failure field.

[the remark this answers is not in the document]
bergundy: Changed to `encoded_attributes`.

temporal/api/failure/v1/message.proto
bergundy: This should be addressed now after my last commit.

temporal/api/failure/v1/message.proto
Sushisource: That's a decent reason, but could still be a proto message with `message`, `stack_trace`, and a map for extra fields. Having it defined as a message is just nice in terms of consistency, introducing this less-well-typed concept is a bit more annoying for the SDK internals. I won't hold it up on that though.

temporal/api/failure/v1/message.proto
```diff
@@ -80,8 +80,26 @@ message ChildWorkflowExecutionFailureInfo {
...
     string stack_trace = 3;
+    // Alternative way to supply `message` and `stack_trace` and possibly other attributes, used for encryption of
+    // errors originating in user code which might contain sensitive information.
+    // The `encoded_attributes` Payload could represent any serializable object, e.g. JSON object or a `Failure` proto
+    // message.
+    //
+    // SDK authors: 
+    // - The SDK should provide an overridable default `encodeFailureAttributes` and `decodeFailureAttributes`
+    //   implementation, those methods may mutate the original failure.
+    // - The default implementation uses a JSON object to represent `{ message, stack_trace }`.
```
cretz, line 95: Hrmm. What other implementations are allowed besides the default? If someone wants to have arbitrary payloads, that's what application failure details are for. I think the post-decoding format here should be a strict expectation unless I'm misunderstanding the use case.

bergundy: This is to address a hypothetical need to encode other failure attributes like `ApplicationFailure.type` or any other opaque value.

temporal/api/failure/v1/message.proto
```diff
@@ -80,8 +80,26 @@ message ChildWorkflowExecutionFailureInfo {
...
+    //   - The original message is overridden with "Encoded failure" to indicate that more information could be
+    //     extracted
+    //   - The resulting JSON object is converted to Payload using the default PayloadConverter and should be processed
+    //     by the user provided PayloadCodec
```
lorensr, line 99:
```suggestion
    //     by the user-provided PayloadCodec
```

temporal/api/failure/v1/message.proto
bergundy: Doesn't have to be implemented initially, but thought I'd mention that if needed we could provide this flexibility with the current approach.

cretz: I see. The way it reads, it doesn't come across as a strictly-SDK-controlled value, but rather anything anyone can set. I think we should strictly define what can go in here and we can expand it as SDK needs.

Sushisource: I agree, which would argue for a proto msg, and makes the serialization process a little more straightforward if you do that too.

bergundy: TBH proto payloads aren't that simple to work with in every language, they require the proto message types to be registered somewhere. I'd just keep JSON for simplicity.

@cretz I'll clarify that this comment is a suggestion for future use in case we want to allow further customization.

cretz (approved):

PR #205, opened 2022-07-19
Added FailoverStatus to the replication NamespaceReplicationConfig proto

roselander:
This is the first of a set of changes that will enable replicating the full NamespaceReplicationConfig


bergundy (approved):
Only thing I'd ask for is some comments on the fields, we're trying to get as much of that as possible. After @Sushisource's mega PR that added docs for much of the API surface.

PR #208, opened 2022-07-19
Minor adjustments from #196

bergundy:
Addressed some comments from #196 and a slack conversation with @Sushisource


temporal/api/enums/v1/event_type.proto
```diff
@@ -157,8 +157,9 @@ enum EventType {
     EVENT_TYPE_WORKFLOW_UPDATE_ACCEPTED = 42;
     // Workflow update has been completed
     EVENT_TYPE_WORKFLOW_UPDATE_COMPLETED = 43;
-    // Some property or properties of the workflow as a whole have changed
-    EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED = 44;
-    // Some property or properties of an already-scheduled activity have changed
-    EVENT_TYPE_ACTIVITY_PROPERTIES_MODIFIED = 45;
+    // Some property or properties of the workflow as a whole have changed by non-workflow code
+    // The distnction of external vs. workflow modification is important for the SDK to be able to verify
```
bergundy, line 161: Oops, forgot to finish the sentence

temporal/api/enums/v1/event_type.proto
```diff
@@ -157,8 +157,12 @@ enum EventType {
     EVENT_TYPE_WORKFLOW_UPDATE_ACCEPTED = 42;
     // Workflow update has been completed
     EVENT_TYPE_WORKFLOW_UPDATE_COMPLETED = 43;
-    // Some property or properties of the workflow as a whole have changed
-    EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED = 44;
-    // Some property or properties of an already-scheduled activity have changed
-    EVENT_TYPE_ACTIVITY_PROPERTIES_MODIFIED = 45;
+    // Some property or properties of the workflow as a whole have changed by non-workflow code.
+    // The distinction of external vs. workflow modification is important for the SDK to be able to verify a
+    // corresponding command is generated by the workflow code to ensure deterministic execution.
```
Sushisource, line 162:
```suggestion
    // The distinction of external vs. command-based modification is important so the SDK can
    // maintain determinism when using the command-based approach.
```

temporal/api/enums/v1/event_type.proto
```diff
@@ -157,8 +157,12 @@ enum EventType {
...
+    EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED_EXTERNALLY = 44;
+    // Some property or properties of an already-scheduled activity have changed by non-workflow code.
+    // The distinction of external vs. workflow modification is important for the SDK to be able to verify a
+    // corresponding command is generated by the workflow code to ensure deterministic execution.
```
Sushisource, line 166:
```suggestion
    // The distinction of external vs. command-based modification is important so the SDK can
    // maintain determinism when using the command-based approach.
```

Sushisource (approved):

temporal/api/history/v1/message.proto
```diff
@@ -741,8 +741,8 @@ message HistoryEvent {
         UpdateWorkflowRequestedEventAttributes update_workflow_requested_event_attributes = 46;
         UpdateWorkflowAcceptedEventAttributes update_workflow_accepted_event_attributes = 47;
         UpdateWorkflowCompletedEventAttributes update_workflow_completed_event_attributes = 48;
-        WorkflowPropertiesModifiedEventAttributes workflow_properties_modified_event_attributes = 49;
-        ActivityPropertiesModifiedEventAttributes activity_properties_modified_event_attributes = 50;
+        WorkflowPropertiesModifiedEventAttributes workflow_properties_modified_externally_event_attributes = 49;
```
cretz, line 744: Should the message names change to `WorkflowPropertiesModifiedExternallyEventAttributes` and `ActivityPropertiesModifiedExternallyEventAttributes` or are we ok w/ breaking the consistent approach of each event type having its own message named after the field here? (I still think the event should be more about the value changing and just have "initiator" for command matching)

temporal/api/enums/v1/event_type.proto
```diff
@@ -157,8 +157,12 @@ enum EventType {
...
     // Workflow update has been completed
     EVENT_TYPE_WORKFLOW_UPDATE_COMPLETED = 43;
-    // Some property or properties of the workflow as a whole have changed
-    EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED = 44;
-    // Some property or properties of an already-scheduled activity have changed
-    EVENT_TYPE_ACTIVITY_PROPERTIES_MODIFIED = 45;
+    // Some property or properties of the workflow as a whole have changed by non-workflow code.
+    // The distinction of external vs. command-based modification is important so the SDK can
+    // maintain determinism when using the command-based approach.
+    EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED_EXTERNALLY = 44;
```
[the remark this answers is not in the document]
bergundy: @Sushisource mentioned it'd be easier to switch on the event type in the SDK than to check the initiator field on the union's variant message.
I don't think it matters either way, but I left it up to him to make the call.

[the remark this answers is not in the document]
bergundy: I'll rename the messages to match the event types, missed that yesterday.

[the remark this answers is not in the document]
cretz: > the event proto will have states that should not be representable depending on the originator.

Yeah, I just figured that'd be the tradeoff of type reuse. But I can understand that someone writing a determinism command-to-event matcher would have to make sure all the non-workflow-allowed properties remain unset when originator was "workflow" instead of "external". The concern was whether the cost of more events outweighs that.

I don't have a strong opinion either way.

temporal/api/history/v1/message.proto
```diff
@@ -669,10 +669,10 @@ message WorkflowPropertiesModifiedEventAttributes {
     // If set, update the workflow memo with the provided values. The values will be merged with
     // the existing memo. If the user wants to delete values, a default/empty Payload should be
     // used as the value for the key being deleted.
-    temporal.api.common.v1.Memo memo = 5;
+    temporal.api.common.v1.Memo upserted_memo = 5;
 }
 
-message ActivityPropertiesModifiedEventAttributes {
+message ActivityPropertiesModifiedEventAttributesExternally {
```
cretz, line 675: `Attributes` needs to be the last word. The message name should be the pascal-cased form of the snake-cased field.

temporal/api/history/v1/message.proto
```diff
@@ -656,7 +656,7 @@ message UpdateWorkflowCompletedEventAttributes {
     }
 }
 
-message WorkflowPropertiesModifiedEventAttributes {
+message WorkflowPropertiesModifiedEventAttributesExternally {
```
rodrigozhou, line 659: So, for the event that I'll create later, are we thinking about supporting the exact same attributes, just under a different name?

cretz: I think just memo, right? Isn't that the only one we're allowing to be updated from inside of a workflow at this time?

temporal/api/history/v1/message.proto
bergundy: 🤦 that's what I meant to do but spaced out.

temporal/api/history/v1/message.proto
```diff
@@ -656,7 +656,7 @@ message UpdateWorkflowCompletedEventAttributes {
     }
 }
 
-message WorkflowPropertiesModifiedEventAttributes {
+message WorkflowPropertiesModifiedEventExternallyAttributes {
```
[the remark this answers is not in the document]
bergundy: Oh no! lol

yiminc (approved):

PR #214, opened 2022-07-29
Update cluster APIs

yux0:
**What changed?**
Update cluster APIs

**Why?**
Those APIs are moved from admin service. We need to review those before public them.

**Breaking changes**
Yes. But they are not used anywhere.


temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -112,50 +105,21 @@ message RemoveRemoteClusterRequest {
...
-message ListClusterMembersResponse {
-    repeated temporal.api.cluster.v1.ClusterMember active_members = 1;
-    bytes next_page_token = 2;
+message ClusterMetadata {
+    string cluster_name = 1;
+    string cluster_id = 2;
+    string address = 3;
+    int64 initial_failover_version = 4;
+    int64 history_shard_count = 5;
+    bool is_connection_enabled = 6;
```
bergundy, line 124: Let's add docstrings here

bergundy (approved):
Another option would be to mark them as experimental instead of removing them

yiminc (approved):

PR #216, opened 2022-08-09
Add fields for batch request/response

yux0:
**What changed?**
Add fields for batch request/response

**Why?**
1. Add reason and operator in stop batch operation request
2. Add operator in describe batch operation response.

**Breaking changes**


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1066,6 +1066,10 @@ message StopBatchOperationRequest {
     string namespace = 1;
     // Batch job id
     string job_id = 2;
+    // Reason to stop a batch operation
+    string reason = 3;
+    // Operator of the operation
+    string operator = 4;
```
yiminc, line 1072: will those be required fields?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1095,6 +1099,8 @@ message DescribeBatchOperationResponse {
     int64 complete_operation_count = 7;
     // Failure operation count
     int64 failure_operation_count = 8;
+    // Operator indicates the batch operation starter identity
+    string operator = 9;
```
yiminc, line 1103: This might be confusing. The batch operation start request has the identity field, which as explained is **the identity of the worker/client**, is this operator the same as that identity?

yiminc (approved):

bergundy (approved):

temporal/api/workflowservice/v1/request_response.proto
yux0: Yes

temporal/api/workflowservice/v1/request_response.proto
yux0: Updated to identity

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1111,6 +1117,8 @@ message DescribeBatchOperationResponse {
     int64 complete_operation_count = 7;
     // Failure operation count
     int64 failure_operation_count = 8;
+    // Identity indicates the operator identity
+    string identity = 9;
```
bergundy, line 1121: I'd put the reason in the response too

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1064,6 +1064,8 @@ message StartBatchOperationRequest {
     string namespace = 1;
     // Visibility query defines the the group of workflow to do batch operation
     string visibility_query = 2;
+    // Job ID defines the unique ID for the batch job
+    string job_id = 3;
```
bergundy, line 1068: Is this required? If so, why do we need it in the response? It's a bit confusing IMHO.

bergundy: Should this include a reason too? I know terminate and cancel batch operations include a reason but signal doesn't.

temporal/api/workflowservice/v1/request_response.proto
bergundy: (Only set in stop ATM)

yux0: Good call out. Added.

temporal/api/workflowservice/v1/request_response.proto
yux0: Had a discussion. To avoid duplication, we allow pass in job id. And I will remove the job_id from response. The reason is included in the operation options.

bergundy: No reason for signal though..

yux0: Yes. There are two options termination, cancellation, signal. The first two options contain reason.

bergundy: I'd consider adding a reason here and remove it from the batch operations so there can be a reason for batch signal too.

PR #217, opened 2022-08-10
Add worker build id fields to poll requests

Sushisource:
**What changed?**
Added worker build id field to poll requests.

I've used the id field consistency's sake so workflow and activity polls look similar. Alternatively, the workflow poll could be a boolean flag and lean on binary checksum, but this seems more obvious at the cost of duping a small amount of data in the poll request.

Open to other ideas too.

**Why?**
We need to know if a worker is opting in to the new versioning features when it polls. It does so by specifying its build id. 

**Breaking changes**
No

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -219,7 +219,13 @@ message PollWorkflowTaskQueueRequest {
     // The identity of the worker/client who is polling this task queue
     string identity = 3;
     // Each worker process should provide an ID unique to the specific set of code it is running
+    // "checksum" in this field name isn't very accurate, it should be though of as an id.
     string binary_checksum = 4;
+    // If set, the worker is opting in to build-id based versioning and wishes to only
+    // receive tasks that are considered compatible with the version provided in the string.
+    // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
+    // When set, `binary_checksum` and this field should typically have the same value.
+    string worker_build_id = 5;
 }
 
 message PollWorkflowTaskQueueResponse {
@@ -318,6 +324,10 @@ message PollActivityTaskQueueRequest {
     // The identity of the worker/client
     string identity = 3;
     temporal.api.taskqueue.v1.TaskQueueMetadata task_queue_metadata = 4;
+    // If set, the worker is opting in to build-id based versioning and wishes to only
+    // receive tasks that are considered compatible with the version provided in the string.
+    // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
+    string worker_build_id = 5;
 }
 
 message PollActivityTaskQueueResponse {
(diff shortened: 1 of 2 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -219,7 +219,13 @@ message PollWorkflowTaskQueueRequest {
     // The identity of the worker/client who is polling this task queue
     string identity = 3;
     // Each worker process should provide an ID unique to the specific set of code it is running
+    // "checksum" in this field name isn't very accurate, it should be though of as an id.
     string binary_checksum = 4;
+    // If set, the worker is opting in to build-id based versioning and wishes to only
+    // receive tasks that are considered compatible with the version provided in the string.
+    // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
+    // When set, `binary_checksum` and this field should typically have the same value.
```
bergundy, line 227: Can we record `worker_build_id` as `binary_checksum` in the server to avoid sending duplicate information?
The 2 fields have the same purpose, seems unnecessary to provide both.

Sushisource: Yeah, so as mentioned in the description:

> Alternatively, the workflow poll could be a boolean flag and lean on binary checksum, but this seems more obvious at the cost of duping a small amount of data in the poll request.

Having the two polls look consistent was the main concern. If we use the flag, it might make sense to add the flag to activity poll as well, just so checking whether or not versioning is "on" doesn't look too different in the handlers.

dnr: I think Roey's suggestion is: if the server gets a poll request with `worker_build_id` set and `binary_checksum` empty, just copy `worker_build_id` to `binary_checksum` on the server side. Of course that won't work if you want to connect to older servers, but if you can confirm the server is new enough that seems fine to me.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -219,7 +219,13 @@ message PollWorkflowTaskQueueRequest {
...
+    string worker_build_id = 5;
```
dnr, line 228: It's not clear from just the name that this has to do with the task queue versioning feature. What about `build_id_for_versioning` or `compatible_with_build_id` or something? It's a mouthful but it might reduce confusion.

dnr (approved):

temporal/api/workflowservice/v1/request_response.proto
bergundy: Yes, that's what I meant.
We should add versioned_task_queues to the list of server capabilities so the SDK can determine whether it can use `worker_build_id` or the to-be-deprecated `binary_checksum`.
We need to expose this capability regardless so the worker can fail when configured to work with versioned task queues and it connects to an older server.

Sushisource: Cool, yeah that makes sense we can do that, I'll update the comment.

cretz (approved):
Today web and users use the `DescribeTaskQueue` API to check if they have workers listening. These are returned in `temporal.api.taskqueue.v1.PollerInfo`. Do we need to somehow add this build identifier that is being polled for to the poller info there so the UI can show it?

dnr:
> Today web and users use the `DescribeTaskQueue` API to check if they have workers listening. These are returned in `temporal.api.taskqueue.v1.PollerInfo`. Do we need to somehow add this build identifier that is being polled for to the poller info there so the UI can show it?

That's a good idea.. it should be possible to track and return this. Can you (or Spencer) file an issue in the server repo and assign it to me? You could add it to that message in this PR if you want

PR #218, opened 2022-08-15
Add command and event enums to modify workflow properties

rodrigozhou:
**What changed?**
Adding a command and an event for modifying workflow properties from user workflow code.
It's named generically so it can be re-used for modifying any workflow property, and for now, it only supports modifying memo.

**Why?**
Adding a feature to upsert memo.

**Breaking changes**
No.

### DIFF
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -156,7 +156,8 @@ message UpsertWorkflowSearchAttributesCommandAttributes {
 }
 
 message ModifyWorkflowPropertiesCommandAttributes {
-    temporal.api.common.v1.Memo memo = 1;
+    // If set, replace workflow memo with provided value.
+    temporal.api.common.v1.Memo new_memo = 1;
 }
 
 message RecordMarkerCommandAttributes {
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -155,6 +155,10 @@ message UpsertWorkflowSearchAttributesCommandAttributes {
     temporal.api.common.v1.SearchAttributes search_attributes = 1;
 }
 
+message ModifyWorkflowPropertiesCommandAttributes {
+    temporal.api.common.v1.Memo memo = 1;
+}
+
 message RecordMarkerCommandAttributes {
     string marker_name = 1;
     map<string, temporal.api.common.v1.Payloads> details = 2;
@@ -255,5 +259,6 @@ message Command {
         UpsertWorkflowSearchAttributesCommandAttributes upsert_workflow_search_attributes_command_attributes = 14;
         AcceptWorkflowUpdateCommandAttributes accept_workflow_update_command_attributes = 15;
         CompleteWorkflowUpdateCommandAttributes complete_workflow_update_command_attributes = 16;
+        ModifyWorkflowPropertiesCommandAttributes modify_workflow_properties_command_attributes = 17;
     }
 }
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -156,8 +156,10 @@ message UpsertWorkflowSearchAttributesCommandAttributes {
 }
 
 message ModifyWorkflowPropertiesCommandAttributes {
-    // If set, replace workflow memo with provided value.
-    temporal.api.common.v1.Memo new_memo = 1;
+    // If set, update the workflow memo with the provided values. The values will be merged with
+    // the existing memo. If the user wants to delete values, a default/empty Payload should be
+    // used as the value for the key being deleted.
+    temporal.api.common.v1.Memo upserted_memo = 1;
 }
 
 message RecordMarkerCommandAttributes {
--- a/temporal/api/enums/v1/event_type.proto
+++ b/temporal/api/enums/v1/event_type.proto
@@ -165,4 +165,6 @@ enum EventType {
     // The distinction of external vs. command-based modification is important so the SDK can
     // maintain determinism when using the command-based approach.
     EVENT_TYPE_ACTIVITY_PROPERTIES_MODIFIED_EXTERNALLY = 45;
+    // Workflow properties modified by user workflow code
+    EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED = 46;
 }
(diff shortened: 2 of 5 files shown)


temporal/api/command/v1/message.proto
```diff
@@ -155,6 +155,10 @@ message UpsertWorkflowSearchAttributesCommandAttributes {
     temporal.api.common.v1.SearchAttributes search_attributes = 1;
 }
 
+message ModifyWorkflowPropertiesCommandAttributes {
+    temporal.api.common.v1.Memo memo = 1;
```
lorensr, line 159: Not sure what the modification operation is. Could be different for different fields. Eg `upsertMemo` or `overrideMemo` here, and in future, `appendFoo` or `setFoo`.

temporal/api/enums/v1/event_type.proto
```diff
@@ -165,4 +165,6 @@ enum EventType {
     // The distinction of external vs. command-based modification is important so the SDK can
     // maintain determinism when using the command-based approach.
     EVENT_TYPE_ACTIVITY_PROPERTIES_MODIFIED_EXTERNALLY = 45;
+    // Workflow properties modified by user workflow code
+    EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED = 46;
```
bergundy, line 169: Wondering if there should be a suffix to clarify this was modified from workflow

temporal/api/command/v1/message.proto
rodrigozhou: Hm, my intention is to override the memo. No strong feeling about the naming. I can change if you think `overrideMemo` is better.

temporal/api/enums/v1/event_type.proto
rodrigozhou: Let me know what kind of suffix would be good for this case.

temporal/api/history/v1/message.proto
```diff
@@ -506,6 +506,12 @@ message UpsertWorkflowSearchAttributesEventAttributes {
     temporal.api.common.v1.SearchAttributes search_attributes = 2;
 }
 
+message WorkflowPropertiesModifiedEventAttributes {
+    // The `WORKFLOW_TASK_COMPLETED` event which this command was reported with
+    int64 workflow_task_completed_event_id = 1;
+    temporal.api.common.v1.Memo memo = 2;
```
bergundy, line 512: Should we call this `upserted_memo` to match the external event? https://github.com/temporalio/api/blob/547f3c99ca8b2d2805ce9c46bb032a8d780301e9/temporal/api/history/v1/message.proto#L678

temporal/api/command/v1/message.proto
Sushisource: Override makes sense to me - it's already available in the WF context so if a user wants to append to it, they can do that themselves easily.

The only concern I can think of would be additional wire size, but, this thing should never be all that humongous anyway.

temporal/api/enums/v1/event_type.proto
bergundy: Only thing that comes to mind is `EVENT_TYPE_WORKFLOW_PROPERTIES_SELF_MODIFIED` but maybe someone has a better idea that's more consistent with the related event names.

temporal/api/command/v1/message.proto
dnr: Whatever the operation is, add a comment that clearly describes the semantics, including what happens if this field is not present, if it's present but the map it contains is empty, etc.

If the semantics are _not_ "upsert", then the PR description and other references should be changed to reflect that.

temporal/api/history/v1/message.proto
dnr: Same here, this needs a comment that describes the semantics, what should sdk and server do when this field is present/not present.

temporal/api/enums/v1/event_type.proto
cretz: I don't have a strong opinion here, all I care about is that this is an all-caps snake case name of the attributes message sans `Attributes` like all other event types.

temporal/api/command/v1/message.proto
lorensr: If the server API is override, SDKs will probably provide upsert to users and merge with existing memo before sending command.

temporal/api/enums/v1/event_type.proto
lorensr: `EVENT_TYPE_WORKFLOW_PROPERTIES_MODIFIED_FROM_WORKFLOW`

lorensr: Or default is from workflow and `_EXTERNALLY` is external?

temporal/api/command/v1/message.proto
dnr: Imo, if the sdk api is upsert than this should be too. We already have upsert for search attributes, so it fits. (Unless there's some reason I'm not aware of that upsert was a bad choice?)

rodrigozhou: @dnr We want to create a more generic command/event so if we need to modify other properties of the workflow, we can re-use it and just add new attributes to the struct.

dnr: Yeah, I mean the semantics of this specific field should be upsert (or not, depending on what the sdk api is). The semantics of other fields in this message can be overwrite (and would have to be if they're single scalar values instead of a map-valued field like memo).

I just think it would be weird to have the user interface be upsert, and then force the sdk to merge and pass the whole new value, when we have another similar command with upsert semantics.

temporal/api/enums/v1/event_type.proto
rodrigozhou: @lorensr That's what I thought. Also, there are comments for the "external vs command" above.

temporal/api/command/v1/message.proto
rodrigozhou: So, you mean I should call it `upsert_memo` and make it behave the same as the command upsert search attributes? I have no strong feeling about this. @lorensr @Sushisource ?

Sushisource: It makes sense, and being consistent with the external modification event makes things easy.

Either way the SDK has to do merging. It either merges before creating the command, or it merges when reading history, so that code exists either way. Hence we might as well keep the events small.

cretz: :+1: Also from a visualization POV (i.e. web/tctl), it is easier to show a diff than a whole thing and it's not as easy asking them to diff from the last upsert/start event.

bergundy: +1 for upsert semantics, will make the experience when upserting externallly better since we could use LWW for a specific key and won't have to bother providing a way to do optimistic updates.

bergundy (approved):

yiminc (approved):

PR #219, opened 2022-08-18
Add build_id_based_dispatch capability

bergundy:
Comments added to the capability, SDK will rely on this for using the feature

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -821,6 +821,11 @@ message GetSystemInfoResponse {
 
         // True if server uses protos that include temporal.api.failure.v1.Failure.encoded_attributes
         bool encoded_failure_attributes = 5;
+
+        // True if server supports dispatching Workflow and Activity tasks based on a worker's build_id
+        // (see:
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -821,6 +821,11 @@ message GetSystemInfoResponse {
 
         // True if server uses protos that include temporal.api.failure.v1.Failure.encoded_attributes
         bool encoded_failure_attributes = 5;
+
+        // True if server supports dispatching Workflow and Activity tasks based on a worker's build_id
+        // (see:
+        // https://github.com/temporalio/proposals/blob/a123af3b559f43db16ea6dd31870bfb754c4dc5e/versioning/worker-versions.md)
+        bool build_id_based_dispatch = 6;
```
dnr, line 828: should the name mention "versioning"? we're using sorta different terms in different places to refer to the same thing

dnr (approved):

temporal/api/workflowservice/v1/request_response.proto
bergundy: Couldn't think of a better name, sorry for auto-merging this.

Sushisource: Given the capabilities are all internal it's probably not so big of a deal, but, yeah I should've caught that

PR #224, opened 2022-08-20
Rename bad memo error to be generic bad modify properties

rodrigozhou:
**What changed?**
Just a small change from #218 renaming `WORKFLOW_TASK_FAILED_CAUSE_BAD_MEMO` to `WORKFLOW_TASK_FAILED_CAUSE_BAD_MODIFY_WORKFLOW_PROPERTIES_ATTRIBUTES` so it can be re-used for any error from `ModifyWorkflowProperties` command.

**Why?**
N/A

**Breaking changes**
No, this enum was introduced in #218 and it's not used yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ $(PROTO_OUT):
 	mkdir $(PROTO_OUT)
 
 ##### Compile proto files for go #####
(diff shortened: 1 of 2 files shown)


Sushisource (approved):

yiminc (approved):
The error message should include details about which property is bad

PR #227, opened 2022-08-22
Allow delete proto fields with reserved id and name

yux0:
**What changed?**
Allow delete proto fields with reserved id and name

**Why?**
To allow delete non used fields with reserved id and name.
The new policies is the same as PACKAGE except ENUM_VALUE_NO_DELETE and FIELD_NO_DELETE

**Breaking changes**


rodrigozhou (approved):

Sushisource (approved):

buf.yaml
```diff
@@ -4,7 +4,37 @@ breaking:
...
+    - FILE_SAME_OBJC_CLASS_PREFIX
+    - FILE_SAME_JAVA_STRING_CHECK_UTF8
+    - FILE_SAME_JAVA_PACKAGE
+    - FILE_SAME_JAVA_OUTER_CLASSNAME
+    - FILE_SAME_JAVA_MULTIPLE_FILES
+    - FILE_SAME_GO_PACKAGE
+    - FILE_SAME_CSHARP_NAMESPACE
+    - FIELD_SAME_TYPE
+    - FIELD_SAME_JSTYPE
+    - FIELD_SAME_CTYPE
```
alexshtin, line 37: According to the [doc](https://docs.buf.build/breaking/configuration#except) you can just write:
```suggestion
    - PACKAGE
  except:
    - ENUM_VALUE_NO_DELETE
    - FIELD_NO_DELETE
```

alexshtin (changes requested):
I am against this. I believe we should continue to adhere `PACKAGE` compatibility level and make breaking changes case by case.

PR #228, opened 2022-08-23
Revert jobID on batch request

yux0:
**What changed?**
Revert jobID on batch request

**Why?**
No need to pass in job id from user. The batch will generate deterministic id from request. 

**Breaking changes**

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -3,6 +3,7 @@ breaking:
   ignore:
     - temporal/api/schedule/v1
     - temporal/api/update/v1
+    - temporal/api/workflowservice/v1
   use:
     - PACKAGE
 lint:
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1064,8 +1064,6 @@ message StartBatchOperationRequest {
     string namespace = 1;
     // Visibility query defines the the group of workflow to do batch operation
     string visibility_query = 2;
-    // Job ID defines the unique ID for the batch job
-    string job_id = 3;
     // Operation input
     oneof operation {
         temporal.api.batch.v1.BatchOperationTermination termination_operation = 10;
@@ -1075,6 +1073,7 @@ message StartBatchOperationRequest {
 }
 
 message StartBatchOperationResponse {
+    string job_id = 1;
 }
 
 message StopBatchOperationRequest {


bergundy (commented):
WDYM by deterministic ID?

yiminc:
> WDYM by deterministic ID?

The idea is to hash the request to an ID so we don't ask user to supply a job_id which feels weird.

yiminc (approved):

yux0:
> WDYM by deterministic ID?

Adding the job ID in request was for de-dup retries from clients.(Make sure no duplicate batch job created). The new approach is to hash the request so that it is deterministic from the request parameters. If those parameters are the same, we treats it as the same request and do the de-dup. 

The request parameters will be hashed to a workflow ID. The workflow ID and run ID will be encoded into job ID based on base64. The describe and stop requests need to decode the job ID to get the workflow ID and run ID. The job ID length is fixed to 64.

dnr:
So if I want to batch signal a bunch of my workflows, and then do the same batch signal five minutes later, I need to add a dummy clause to the query so that it won't de-dup? (Let's say the signal itself has no args.)

yux0:
> So if I want to batch signal a bunch of my workflows, and then do the same batch signal five minutes later, I need to add a dummy clause to the query so that it won't de-dup? (Let's say the signal itself has no args.)

If the batch job is finished, it will start a new one with a different run ID. If the batch job is still running, it will not start a new one.

dnr:
> If the batch job is finished, it will start a new one with a different run ID. If the batch job is still running, it will not start a new one.

Then there's a weird discontinuity: say my batch signal job takes exactly 5:00. If I want to signal again in 4:59, it de-dups and doesn't work. But if I do it in 5:01, it signals again. This seems a little surprising

yiminc:
> If the batch job is finished, it will start a new one with a different run ID. If the batch job is still running, it will not start a new one.

It is the same behavior as you try to start same workflow.

bergundy:
When does the batch job list workflows? In the beginning of the job?

bergundy:
I'm asking because if we dedupe based on the input we prevent the same batch from being used to signal / cancel workflows as they're being created unless the batch job takes incoming workflows that match the query into account.

dnr:
> It is the same behavior as you try to start same workflow.

Yeah but when starting a workflow, the caller can choose whether they want deduping or not. Here, the behavior is not easily predictable. I'd be fine with making job id optional and generating it from the input if missing, but there should be a way for a caller to opt out of deduping, e.g. if they want to send a signal multiple times, or cancel new workflows that have started after a previous batch is running before that batch finishes.

yiminc:
> When does the batch job list workflows? In the beginning of the job?

The list could result in many pages and the batch operation progress as it page through the result set. Each page is a new ListWorkflowExecutions call. If you want to use batch to send signal to workflows including workflows started after the batch operation initiated, then there is race condition and even if you could start multiple runs you risk of having some workflows receive it once and others received twice. Better option is to limit the time window of each batch (by filter on startTime to that fixed window), and you can start multiple batches each cover a disjoined time window.

> I'd be fine with making job id optional 
It is weird to ask for job_id as input and it is even more weird to return it as output as well. 

> they want to send a signal multiple times
I think you are making up weird use case here. If the signal is different, they can start different batches. Or they can scope the time range of target workflows.

yux0:
> I'm asking because if we dedupe based on the input we prevent the same batch from being used to signal / cancel workflows as they're being created unless the batch job takes incoming workflows that match the query into account.

As Yimin mentioned, there is race condition on the listing. It means the de-dup won't work if we takes the incoming workflows. 

> I'd be fine with making job id optional
What about take in an optional request id, we can hash the input including the request id?

bergundy:
I’m still not 100% convinced that a user will never want to send a signal multiple times.
What if they send the signal and then realize that they have some workers up running workflow code that doesn’t have the signal handler configured, they might want to restart the batch and not have to wait for the previous batch to complete. We can maybe tell them that they should stop the previous batch first but it doesn’t seem like it’ll be the most obvious behavior, I don’t think people will pay attention to the fact that we give them back the same batch ID.
If we dedupe, there should be an explicit server response saying that a batch with these parameters is already running and that no nee work was started, I’d even include the guidance given in the comments of this PR in the server message.
For example the message might say:

“A batch with these parameters is already running with ID …. To restart this batch do … if you want to target run this batch job on new workflows do …”

We could also provide `restart` and `continue from` helper APIs to help users achieve these things.

yux0:
> I’m still not 100% convinced that a user will never want to send a signal multiple times.

There is definitely cases that a user want to send the same signal multiple times. The rare case is to send the same signal multiple times concurrently. And I agree we need to indicate it to users if a batch job is already running. I think include an optional request id would solve your concern.

PR #229, opened 2022-08-24
Add worker_versioning_id to more messages, use VersionId

dnr:
**What changed?**
Add `worker_versioning_id` to `WorkflowTaskCompletedEventAttributes` and `RespondWorkflowTaskCompletedRequest`.
Also switch string fields that were added recently (unused) to use `VersionId`.

**Why?**
We need to record this in history to ensure consistency on replay and replication.
Use `VersionId` everywhere for consistency.

**Breaking changes**
Deprecate string build id fields and switch to a new field. These were not used yet.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -192,6 +192,9 @@ message WorkflowTaskCompletedEventAttributes {
     string identity = 3;
     // Binary ID of the worker who completed this task
     string binary_checksum = 4;
+    // ID of the worker who picked up this workflow task, or missing if worker
+    // is not using versioning.
+    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
 }
 
 message WorkflowTaskTimedOutEventAttributes {
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -77,7 +77,8 @@ message PollerInfo {
     double rate_per_second = 3;
     // If a worker has specified an ID for use with the worker versioning feature while polling,
     // that id must appear here.
-    string worker_versioning_build_id = 4;
+    string worker_versioning_build_id = 4 [deprecated=true]; // use VersionId field
+    VersionId worker_versioning_id = 5;
 }
 
 message StickyExecutionAttributes {
@@ -98,7 +99,7 @@ message VersionIdNode {
 }
 
 // Used by the worker versioning APIs, represents a specific version of something
-// Currently, that's just a whole-worker id. In the future, if we support 
+// Currently, that's just a whole-worker id. In the future, if we support
 // WASM workflow bundle based versioning, for example, then the inside of this
 // message may become a oneof of different version types.
 message VersionId {
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -77,8 +77,7 @@ message PollerInfo {
     double rate_per_second = 3;
     // If a worker has specified an ID for use with the worker versioning feature while polling,
     // that id must appear here.
-    string worker_versioning_build_id = 4 [deprecated=true]; // use VersionId field
-    VersionId worker_versioning_id = 5;
+    VersionId worker_versioning_id = 4;
 }
 
 message StickyExecutionAttributes {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -226,8 +226,7 @@ message PollWorkflowTaskQueueRequest {
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
     // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
     // set, that value should also be considered as the `binary_checksum`.
-    string worker_versioning_build_id = 5 [deprecated=true]; // use VersionId field
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 6;
+    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
 }
 
 message PollWorkflowTaskQueueResponse {
@@ -334,8 +333,7 @@ message PollActivityTaskQueueRequest {
     // If set, the worker is opting in to build-id based versioning and wishes to only
     // receive tasks that are considered compatible with the version provided.
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
-    string worker_versioning_build_id = 5 [deprecated=true]; // use VersionId field
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 6;
+    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
 }
 
 message PollActivityTaskQueueResponse {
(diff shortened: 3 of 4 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -181,6 +181,9 @@ message WorkflowTaskStartedEventAttributes {
     // continue-as-new regardless of the suggestion. Note that history event count is
     // just the event id of this event, so we don't include it explicitly here.
     int64 history_size_bytes = 5;
+    // Build ID of the worker who picked up this workflow task, or empty string
+    // if worker is not using build id versioning.
+    string worker_versioning_build_id = 6;
```
yiminc, line 186: should we use [VersionId](https://github.com/temporalio/api/blob/03108e6a2dc728c2acd13fbf15963f186e7952ee/temporal/api/taskqueue/v1/message.proto#L104)

Sushisource: Yeah, I think this probably makes sense. If, EX, we add WASM bundle versioning capabilities later, that would supersede the worker build ID for the purposes of choosing the appropriate version for the next task, so you'd want to switch to that (currently nonexistent) variant in the `VersionId` oneof.

dnr:
second try, with more `VersionId`

temporal/api/taskqueue/v1/message.proto
```diff
@@ -77,7 +77,8 @@ message PollerInfo {
     double rate_per_second = 3;
     // If a worker has specified an ID for use with the worker versioning feature while polling,
     // that id must appear here.
-    string worker_versioning_build_id = 4;
+    string worker_versioning_build_id = 4 [deprecated=true]; // use VersionId field
```
yiminc, line 80: Should just remove this field and make it reserved. Worker versioning is not available yet anyway.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -222,11 +222,12 @@ message PollWorkflowTaskQueueRequest {
...
     // If set, the worker is opting in to build-id based versioning and wishes to only
-    // receive tasks that are considered compatible with the version provided in the string.
+    // receive tasks that are considered compatible with the version provided.
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
-    // When set, and `binary_checksum` is not, this value should also be considered as the
-    // `binary_checksum`.
-    string worker_versioning_build_id = 5;
+    // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
+    // set, that value should also be considered as the `binary_checksum`.
+    string worker_versioning_build_id = 5 [deprecated=true]; // use VersionId field
```
yiminc, line 229: same here

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -326,9 +332,10 @@ message PollActivityTaskQueueRequest {
     string identity = 3;
     temporal.api.taskqueue.v1.TaskQueueMetadata task_queue_metadata = 4;
     // If set, the worker is opting in to build-id based versioning and wishes to only
-    // receive tasks that are considered compatible with the version provided in the string.
+    // receive tasks that are considered compatible with the version provided.
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
-    string worker_versioning_build_id = 5;
+    string worker_versioning_build_id = 5 [deprecated=true]; // use VersionId field
```
yiminc, line 337: and here

temporal/api/taskqueue/v1/message.proto
dnr: I did that to avoid the buf breaking check :disappointed: I really don't want to do the disable/re-enable dance just to keep it happy. This part of our workflow is broken (see like half of recent api changes, and a few recent slack conversations too). We need to either use long-lived branches or relax the check.

How about this: I'll remove it if you let me change the compatibility mode to `WIRE_JSON`

cretz: Yeah, we probably shouldn't have merged the API before it was developed.

bergundy: I'm with @yiminc, would rather not pollute the API for something that was never in use.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -222,11 +222,12 @@ message PollWorkflowTaskQueueRequest {
...
-    // receive tasks that are considered compatible with the version provided in the string.
+    // receive tasks that are considered compatible with the version provided.
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
-    // When set, and `binary_checksum` is not, this value should also be considered as the
-    // `binary_checksum`.
-    string worker_versioning_build_id = 5;
+    // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
+    // set, that value should also be considered as the `binary_checksum`.
+    string worker_versioning_build_id = 5 [deprecated=true]; // use VersionId field
+    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 6;
```
bergundy, line 230: Not sure how this will work. If we support multiple "versions" in a single worker, we'd need to pass all of those in the request here. I think we probably want to only pass a single worker version here and have the worker dispatch to the appropriate bundle internally.

temporal/api/taskqueue/v1/message.proto
dnr: alright, I'll make it breaking. agreed that we need to improve our processes here.

temporal/api/workflowservice/v1/request_response.proto
dnr: I don't know what the plan is for these bundles.. how will it know the appropriate bundle? Is this something we can plan for at this point or should we just leave that whole feature until later?

bergundy: Either design it and come back to this PR when we know the API or just use the worker versioning ID for now.
I don't want to block progress on build ID based dispatch so if the design sets us back, I'm leaning towards the latter.

Sushisource: Polling would likely need to have this field be `repeated` (ex: a worker can have some version, and support bundle downloading), but completions can stick with one field, depending on how the task was processed.

Alternatively, we just support only one mode of operation per task queue, which is simple and understandable.

So, I think keeping this how it is right now makes good sense and we can deprecate or revise as needed later

Sushisource (approved):

dnr:
How about this as a process: after I get an approval from server (probably @yiminc ), I'll update my in-progress work to use this and send a draft PR so you can test with it and see if the pieces fit, before merging this

yiminc (approved):

bergundy:
TBH since we're not sure about this change and how bundle based versioning (or dispatching within a single worker) works, I would have left the API as is for now.

We might regret using VersionId without thinking this through, I'd prefer if this PR is reverted and revisited when there's time to design this properly.

Sushisource:
@bergundy 

I don't really see how that can be true - they're essentially equivalent at this point, but way we are able to possibly augment the `VersionId` message later. This makes the client usage of the API mildly more verbose, but actually makes the server internals less verbose since these can be passed straight around in more cases.

As I see it these are the possible future outcomes:

1. We never add new version types. Oh well, this is slightly more verbose than it needed to be
2. We add new version types which are either mutually exclusive or supersede each other (WASM bundles would seem to supersede whole worker ids). This works with this change.
3. We add new version types which it might make sense to include with existing version types. Then we need to make this field `repeated`, or add another field, or something else. That was already true before this change too, one string would not have been sufficient.

So, at worst this is equivalent, at best it prevents needing a future API change.

bergundy:
Sounds good, agree there's not much harm in merging this now.
We'll mostly likely have to revisit and make changes no matter what the bundling design looks like.

PR #233, opened 2022-09-02
Move DeleteWorkflowExecution to workflow service

alexshtin:
**What changed?**
Move `DeleteWorkflowExecution` to workflow service.

**Why?**
After number of discussions we came up to conclusion that all workflow execution APIs must be in `workflowservice`.

**Breaking changes**
Yes, but `DeleteWorkflowExecution` is accesible only using `tctl v2` which is in beta.

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -4,6 +4,8 @@ breaking:
     # TODO: remove these after pr 229
     - temporal/api/taskqueue/v1
     - temporal/api/workflowservice/v1
+    # TODO (alex): remove these after pr 233
+    - temporal/api/operatorservice/v1
   use:
     - PACKAGE
 lint:
--- a/temporal/api/operatorservice/v1/request_response.proto
+++ b/temporal/api/operatorservice/v1/request_response.proto
@@ -36,7 +36,6 @@ import "google/protobuf/timestamp.proto";
 import "google/protobuf/duration.proto";
 
 import "temporal/api/cluster/v1/message.proto";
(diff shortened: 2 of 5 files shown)


buf.yaml
```diff
@@ -4,6 +4,7 @@ breaking:
     - temporal/api/schedule/v1
     - temporal/api/update/v1
     - temporal/api/batch/v1
+    - temporal/api/operatorservice/v1
```
alexshtin, line 7: Subject to revert in next PR.

buf.yaml
```diff
@@ -4,6 +4,8 @@ breaking:
     # TODO: remove these after pr 229
     - temporal/api/taskqueue/v1
     - temporal/api/workflowservice/v1
```
dnr, line 6: please remove these as the friendly comment requests :)

I think we can just get in the habit of having each pr remove the breaking ignore lines from the last one

yiminc: lol

yiminc (approved):

buf.yaml
dnr: I mean, I still kinda think we should use `WIRE/WIRE_JSON` and not even do this. but in the interest of making progress without doubling the number of commits in the repo...

bergundy (approved):

cretz (approved):

buf.yaml
alexshtin: I will do it in that follow up PR.

PR #237, opened 2022-09-14
Refactor versioning API

Sushisource:
**What changed?**
* Simpler representation of version "graph"
* Support id for each compatible set
* Better support for future WASM bundles

**Why?**
We need the ID to support separate physical task queues, which is required to avoid a backup in one version starving tasks on another version.

The graph representation was more complex than necessary.

Better support for future WASM bundles, now that we know we will be doing Rust SDK release soonish.

**Breaking changes**
Haha, yes - to an unreleased API.

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -122,7 +122,7 @@ message MeteringMetadata {
     uint32 nonfirst_local_activity_execution_attempts = 13;
 }
 
-// Identifies the version(s) that a worker used to process a task
+// Identifies the version(s) of a worker that processed a task
 message WorkerVersionStamp {
     // An opaque whole-worker identifier
     string build_id = 1;
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -121,3 +121,20 @@ message MeteringMetadata {
     //     aip.dev/not-precedent: Negative values make no sense to represent. --)
     uint32 nonfirst_local_activity_execution_attempts = 13;
 }
+
+// Identifies the version(s) that a worker used to process a task
+message WorkerVersionStamp {
+    // An opaque whole-worker identifier
+    string build_id = 1;
+    // Set if the worker used a dynamically loadable bundle to process
+    // the task. The bundle could be a WASM blob, JS bundle, etc.
+    string bundle_id = 2;
+}
+
+// Identifies the version(s) that a worker is compatible with when polling or identifying itself
+message WorkerVersionCapabilities {
+    // An opaque whole-worker identifier
+    string build_id = 1;
+
+    // Later, may include info like "I can process WASM and/or JS bundles"
+}
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -196,7 +196,7 @@ message WorkflowTaskCompletedEventAttributes {
     // Version info of the worker who processed this workflow task, or missing if worker is not
     // using versioning. If present, the `build_id` field within supersedes `binary_checksum`, which
     // may be populated with the same value to preserve compatability.
-    temporal.api.common.v1.WorkerVersionStamp worker_versions = 5;
+    temporal.api.common.v1.WorkerVersionStamp worker_version = 5;
     // Data the SDK wishes to record for itself, but server need not interpret, and does not
     // directly impact workflow state.
     temporal.api.sdk.v1.WorkflowTaskCompletedMetadata sdk_metadata = 6;
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -194,8 +194,8 @@ message WorkflowTaskCompletedEventAttributes {
     // Binary ID of the worker who completed this task
     string binary_checksum = 4;
     // Version info of the worker who processed this workflow task, or missing if worker is not
-    // using versioning. If present, the `build_id` field within supersedes `binary_checksum`, which
-    // may be populated with the same value to preserve compatability.
+    // using versioning. If present, the `build_id` field within is also used as `binary_checksum`,
+    // which may be omitted in that case (it may also be populated to preserve compatability).
     temporal.api.common.v1.WorkerVersionStamp worker_version = 5;
     // Data the SDK wishes to record for itself, but server need not interpret, and does not
     // directly impact workflow state.
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -193,9 +193,10 @@ message WorkflowTaskCompletedEventAttributes {
     string identity = 3;
     // Binary ID of the worker who completed this task
     string binary_checksum = 4;
-    // ID of the worker who picked up this workflow task, or missing if worker
-    // is not using versioning.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    // Version info of the worker who processed this workflow task, or missing if worker is not
+    // using versioning. If present, the `build_id` field within supersedes `binary_checksum`, which
+    // may be populated with the same value to preserve compatability.
+    temporal.api.common.v1.WorkerVersionStamp worker_versions = 5;
     // Data the SDK wishes to record for itself, but server need not interpret, and does not
     // directly impact workflow state.
     temporal.api.sdk.v1.WorkflowTaskCompletedMetadata sdk_metadata = 6;
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -38,6 +38,7 @@ import "google/protobuf/wrappers.proto";
 import "dependencies/gogoproto/gogo.proto";
 
 import "temporal/api/enums/v1/task_queue.proto";
+import "temporal/api/common/v1/message.proto";
 
 // See https://docs.temporal.io/docs/concepts/task-queues/
 message TaskQueue {
@@ -71,13 +72,12 @@ message TaskQueuePartitionMetadata {
 }
 
 message PollerInfo {
-    // Unix Nano
     google.protobuf.Timestamp last_access_time = 1 [(gogoproto.stdtime) = true];
     string identity = 2;
     double rate_per_second = 3;
-    // If a worker has specified an ID for use with the worker versioning feature while polling,
-    // that id must appear here.
-    VersionId worker_versioning_id = 4;
+    // If a worker has opted into the worker versioning feature while polling, its capabilities will
+    // appear here.
+    temporal.api.common.v1.WorkerVersionCapabilities worker_version_capabilities = 4;
 }
 
 message StickyExecutionAttributes {
@@ -87,22 +87,11 @@ message StickyExecutionAttributes {
     google.protobuf.Duration schedule_to_start_timeout = 2 [(gogoproto.stdduration) = true];
 }
 
-// Used by the worker versioning APIs, represents a node in the version graph for a particular
-// task queue
-message VersionIdNode {
-    VersionId version = 1;
-    // A pointer to the previous version this version is considered to be compatible with
-    VersionIdNode previous_compatible = 2;
-    // A pointer to the last incompatible version (previous major version)
-    VersionIdNode previous_incompatible = 3;
+// Used by the worker versioning APIs, represents an ordering of one or more versions which are
+// considered to be compatible with each other. Currently the versions are always worker build ids.
+message CompatibleVersionSet {
+    // A unique identifier for this version set
+    string id = 1;
+    // All the compatible versions, ordered from oldest to newest
+    repeated string versions = 2;
 }
-
-// Used by the worker versioning APIs, represents a specific version of something
-// Currently, that's just a whole-worker id. In the future, if we support
-// WASM workflow bundle based versioning, for example, then the inside of this
-// message may become a oneof of different version types.
-message VersionId {
-    // An opaque whole-worker identifier
-    string worker_build_id = 1;
-}
-
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -92,7 +92,7 @@ message StickyExecutionAttributes {
 message CompatibleVersionSet {
     // A unique identifier for this version set. Users don't need to understand or care about this
     // value, but it has value for debugging purposes.
-    string id = 1;
+    string version_set_id = 1;
     // All the compatible versions, ordered from oldest to newest
-    repeated string versions = 2;
+    repeated string build_ids = 2;
 }
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -90,7 +90,8 @@ message StickyExecutionAttributes {
 // Used by the worker versioning APIs, represents an ordering of one or more versions which are
 // considered to be compatible with each other. Currently the versions are always worker build ids.
 message CompatibleVersionSet {
-    // A unique identifier for this version set
+    // A unique identifier for this version set. Users don't need to understand or care about this
+    // value, but it has value for debugging purposes.
     string id = 1;
     // All the compatible versions, ordered from oldest to newest
     repeated string versions = 2;
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -57,6 +57,8 @@ message WorkflowExecutionInfo {
     string task_queue = 13;
     int64 state_transition_count = 14;
     int64 history_size_bytes = 15;
+    // If set, the most recent worker version stamp that appeared in a workflow task completion
+    temporal.api.common.v1.WorkerVersionStamp most_recent_worker_version_stamp = 16;
 }
 
 message WorkflowExecutionConfig {
@@ -100,9 +102,6 @@ message PendingWorkflowTaskInfo {
     google.protobuf.Timestamp original_scheduled_time = 3 [(gogoproto.stdtime) = true];
     google.protobuf.Timestamp started_time = 4 [(gogoproto.stdtime) = true];
     int32 attempt = 5;
-    // If set, this pending workflow task will need to be (or is currently being) handled by a
-    // a worker using this build id.
-    string will_use_build_id = 6;
 }
 
 message ResetPoints {
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -100,6 +100,9 @@ message PendingWorkflowTaskInfo {
     google.protobuf.Timestamp original_scheduled_time = 3 [(gogoproto.stdtime) = true];
     google.protobuf.Timestamp started_time = 4 [(gogoproto.stdtime) = true];
     int32 attempt = 5;
+    // If set, this pending workflow task will need to be (or is currently being) handled by a
+    // a worker using this build id.
+    string will_use_build_id = 6;
 }
 
 message ResetPoints {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -237,12 +237,12 @@ message PollWorkflowTaskQueueRequest {
     // Each worker process should provide an ID unique to the specific set of code it is running
     // "checksum" in this field name isn't very accurate, it should be though of as an id.
     string binary_checksum = 4;
-    // If set, the worker is opting in to build-id based versioning and wishes to only
-    // receive tasks that are considered compatible with the version provided.
+    // If set, the worker is opting in to versioning and wishes to only
+    // receive tasks that are considered compatible with the version capabilities provided.
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
-    // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
+    // When this field has a `worker_build_id`, and `binary_checksum` is not
     // set, that value should also be considered as the `binary_checksum`.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    temporal.api.common.v1.WorkerVersionCapabilities worker_version_capabilities = 5;
 }
 
 message PollWorkflowTaskQueueResponse {
@@ -310,11 +310,11 @@ message RespondWorkflowTaskCompletedRequest {
     // Responses to the `queries` field in the task being responded to
     map<string, temporal.api.query.v1.WorkflowQueryResult> query_results = 8;
     string namespace = 9;
-    // If using versioning, worker should send the same id here that it used to
-    // poll for the workflow task.
-    // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
-    // set, that value should also be considered as the `binary_checksum`.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 10;
+    // If using versioning, the worker uses this field to indicate what version(s) it used to
+    // process the task.
+    // When this field has a `worker_build_id`, and `binary_checksum` is not set, that value should
+    // also be considered as the `binary_checksum`.
+    temporal.api.common.v1.WorkerVersionStamp worker_versions_stamp = 10;
     // Protocol messages piggybacking on a WFT as a transport
     repeated temporal.api.protocol.v1.Message messages = 11;
     // Data the SDK wishes to record for itself, but server need not interpret, and does not
@@ -359,10 +359,10 @@ message PollActivityTaskQueueRequest {
     // The identity of the worker/client
     string identity = 3;
     temporal.api.taskqueue.v1.TaskQueueMetadata task_queue_metadata = 4;
-    // If set, the worker is opting in to build-id based versioning and wishes to only
-    // receive tasks that are considered compatible with the version provided.
+    // If set, the worker is opting in to versioning and wishes to only
+    // receive tasks that are considered compatible with the capabilities provided.
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    temporal.api.common.v1.WorkerVersionCapabilities worker_version_capabilities = 5;
 }
 
 message PollActivityTaskQueueResponse {
@@ -1047,20 +1047,22 @@ message UpdateWorkerBuildIdOrderingRequest {
     // a given task queue must have the same set of workflow & activity
     // implementations, there is no reason to specify a task queue type here.
     string task_queue = 2;
-    // The version id we are targeting.
-    temporal.api.taskqueue.v1.VersionId version_id = 3;
-    // When set, indicates that the `version_id` in this message is compatible
-    // with the one specified in this field. Because compatability should form
-    // a DAG, any build id can only be the "next compatible" version for one
-    // other ID of a certain type at a time, and any setting which would create a cycle is invalid.
-    temporal.api.taskqueue.v1.VersionId previous_compatible = 4;
-    // When set, establishes the specified `version_id` as the default of it's type
-    // for the queue. Workers matching it will begin processing new workflow executions.
-    // The existing default will be marked as a previous incompatible version
-    // to this one, assuming it is not also in `is_compatible_with`.
+    // The version id being added or updated.
+    string version_id = 3;
+    // When set, indicates that the `version_id` in this message is compatible with the version
+    // specified in this field. This version must already exist, and if found, the ID we are
+    // targeting will be added to the compatible set containing this version.
+    string compatible_version = 4;
+    // When set, establishes the specified `version_id` as the default for the queue. Workers
+    // matching it will begin processing new workflow executions. The existing default set will be
+    // marked as a previous incompatible version to this one, assuming it is not also in
+    // `compatible_version`.
     bool become_default = 5;
 }
-message UpdateWorkerBuildIdOrderingResponse {}
+message UpdateWorkerBuildIdOrderingResponse {
+    // The id of the compatible set that the updated version was added to, or exists in.
+    string version_set_id = 1;
+}
 
 // (-- api-linter: core::0134::request-resource-required=disabled
 //     aip.dev/not-precedent: GetWorkerBuildIdOrderingRequest RPC doesn't follow Google API format. --)
@@ -1068,17 +1070,17 @@ message GetWorkerBuildIdOrderingRequest {
     string namespace = 1;
     // Must be set, the task queue to interrogate about worker id ordering
     string task_queue = 2;
-    // Limits how deep the returned DAG will go. 1 will return only the
-    // default build id. A default/0 value will return the entire graph.
-    int32 max_depth = 3;
+    // Limits how many compatible sets will be returned. Specify 1 to only return the current
+    // default major version set. 0 returns all sets.
+    int32 max_sets = 3;
 }
 message GetWorkerBuildIdOrderingResponse {
-    // The currently established default version
-    temporal.api.taskqueue.v1.VersionIdNode current_default = 1;
-    // Other current latest-compatible versions who are not the overall default. These are the
-    // versions that will be used when generating new tasks by following the graph from the
-    // version of the last task out to a leaf.
-    repeated temporal.api.taskqueue.v1.VersionIdNode compatible_leaves = 2;
+    // Major version sets, in order from oldest to newest. The last element of the list will always
+    // be the current default major version. IE: New workflows will target the most recent version
+    // in that version set.
+    //
+    // There may be fewer sets returned than exist, if the request chose to limit this response.
+    repeated temporal.api.taskqueue.v1.CompatibleVersionSet major_version_sets = 1;
 }
 
 // (-- api-linter: core::0134=disabled
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -311,10 +311,9 @@ message RespondWorkflowTaskCompletedRequest {
     map<string, temporal.api.query.v1.WorkflowQueryResult> query_results = 8;
     string namespace = 9;
     // If using versioning, the worker uses this field to indicate what version(s) it used to
-    // process the task.
-    // When this field has a `worker_build_id`, and `binary_checksum` is not set, that value should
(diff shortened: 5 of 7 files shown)


temporal/api/taskqueue/v1/message.proto
```diff
@@ -87,22 +89,16 @@ message StickyExecutionAttributes {
...
-    // A pointer to the previous version this version is considered to be compatible with
-    VersionIdNode previous_compatible = 2;
-    // A pointer to the last incompatible version (previous major version)
-    VersionIdNode previous_incompatible = 3;
+// Used by the worker versioning APIs, represents an ordering of one or more versions which are
+// considered to be compatible with each other
+message CompatibleVersionSet {
+    // A unique identifier for this version set
+    string id = 1;
+    // The type of the versions in this set. All versions in a set are always of the same type.
```
bergundy, line 97: Why would all versions need to be of the same type?

I can definitely see a scenario where we introduce bundle versions for an SDK and want workers using this new SDK version to be compatible with older worker versions.

temporal/api/common/v1/message.proto
```diff
@@ -110,3 +110,19 @@ message RetryPolicy {
     // this is not a substring match, the error *type* (not message) must match exactly.
     repeated string non_retryable_error_types = 5;
 }
+
+// Identifies the version(s) that a worker used to process a task
+message WorkerVersionsStamp {
```
bergundy, line 115: When would this be more than 1 version?

temporal/api/taskqueue/v1/message.proto
bergundy: This is only relevant for the transition period from per-bundle to post-bundle workers so it'd be acceptable since I can see how it simplifies the implementation.

bergundy (approved):
LGTM, had a question and a comment that I pretty much resolved myself.

temporal/api/common/v1/message.proto
bergundy: Spencer answered me on slack

temporal/api/taskqueue/v1/message.proto
Sushisource: Plus it's just supported a different way - it will be supported by different sets referencing each other, rather than stuffing multiple types into the same set which could end up rather duplicative.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -87,22 +89,16 @@ message StickyExecutionAttributes {
...
+    temporal.api.enums.v1.VersionType version_type = 2;
```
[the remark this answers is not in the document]
Sushisource: We already know that we will be doing this, hence the preparation

That design doesn't really work, because all versions in a set must be of the same type. The type tag expresses that, whereas the oneof inside the version allows for heterogeneous types.

temporal/api/common/v1/message.proto
Sushisource: It's added, now. There can be a build id, as well as a bundle ID which processed any given task.

temporal/api/history/v1/message.proto
```diff
@@ -192,9 +192,9 @@ message WorkflowTaskCompletedEventAttributes {
     string identity = 3;
     // Binary ID of the worker who completed this task
     string binary_checksum = 4;
-    // ID of the worker who picked up this workflow task, or missing if worker
-    // is not using versioning.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    // Version info of the worker who processed this workflow task, or missing if worker is not
+    // using versioning.
+    temporal.api.common.v1.WorkerVersionsStamp worker_versions = 5;
```
dnr, line 197: should we describe how this does/doesn't obsolete `binary_checksum`? or leave that until later?

temporal/api/taskqueue/v1/message.proto
```diff
@@ -75,9 +77,9 @@ message PollerInfo {
     google.protobuf.Timestamp last_access_time = 1 [(gogoproto.stdtime) = true];
     string identity = 2;
     double rate_per_second = 3;
-    // If a worker has specified an ID for use with the worker versioning feature while polling,
-    // that id must appear here.
-    VersionId worker_versioning_id = 4;
+    // If a worker has opted into the worker versioning feature while polling, its capabilities must
```
dnr, line 80:
```suggestion
    // If a worker has opted into the worker versioning feature while polling, its capabilities will
```
feels weird to say "must" when it's coming as a reply

temporal/api/taskqueue/v1/message.proto
```diff
@@ -75,9 +77,9 @@ message PollerInfo {
     google.protobuf.Timestamp last_access_time = 1 [(gogoproto.stdtime) = true];
```
dnr, line 77: please delete the `// Unix Nano` above this line

temporal/api/taskqueue/v1/message.proto
dnr: what is this "different sets referencing each other" idea?

temporal/api/taskqueue/v1/message.proto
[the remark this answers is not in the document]
dnr: I might be sorta with Chad here.. what is the point of having `bundle_id` in `WorkerVersionsStamp` if it's not used at all yet? and when we do use it, what will this look like? these `versions` will be bundle ids? then what about the build id, where does that go? is it ignored entirely for matching?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1011,38 +1012,46 @@ message UpdateWorkerBuildIdOrderingRequest {
...
-    temporal.api.taskqueue.v1.VersionId previous_compatible = 4;
-    // When set, establishes the specified `version_id` as the default of it's type
+    // The version id being added or updated.
+    string version_id = 3;
+    // The type of the version being added or updated.
+    temporal.api.enums.v1.VersionType version_type = 4;
+    // When set, indicates that the `version_id` in this message is compatible with the version
+    // specified in this field. This version must already exist, and if found, the ID we are
+    // targeting will be added to the compatible set containing this version.
+    string compatible_version = 5;
```
dnr, line 1022: it has to be the same type?

oh, I see the comment at the end.. well, either put that here or comment here (for now) that it must be the same type

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1011,38 +1012,46 @@ message UpdateWorkerBuildIdOrderingRequest {
...
+    // When set, establishes the specified `version_id` as the default of its type
```
dnr, line 1023: what does "of its type" mean here?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1011,38 +1012,46 @@ message UpdateWorkerBuildIdOrderingRequest {
...
+    // to this one, assuming it is not also in `compatible_version`.
+    bool become_default = 6;
+
+    // It's TBD if we'll actually require the ability to specify different version types as
+    // compatible with one another, but that's can be added compatibly later, so no version type
+    // is specified for `compatible_version` right now.
+}
+message UpdateWorkerBuildIdOrderingResponse {
+    // The id of the compatible set that the updated version was added to, or exists in.
+    string version_set_id = 1;
```
dnr, line 1035: why do we need this? I thought the id could be a server internal thing, as an implementation detail.

actually on that note, I wonder if there's a way to keep it out of the public proto entirely?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1011,38 +1012,46 @@ message UpdateWorkerBuildIdOrderingRequest {
...
 message GetWorkerBuildIdOrderingRequest {
     string namespace = 1;
     // Must be set, the task queue to interrogate about worker id ordering
     string task_queue = 2;
-    // Limits how deep the returned DAG will go. 1 will return only the
-    // default build id. A default/0 value will return the entire graph.
-    int32 max_depth = 3;
+    // Limits how many compatible sets will be returned. Specify 1 to only return the current
+    // default major version set. 0 returns all sets.
+    int32 max_sets = 3;
```
dnr, line 1046: limit on max versions per set also?

temporal/api/taskqueue/v1/message.proto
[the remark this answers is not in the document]
Sushisource: The point is just to be able to not have to cut a new server release when we do decide to add it, which we know we'll do at some point for WASM bundles, at least. I'm fine with leaving it out for now, @bergundy wanted it in. Given the amount of work it'll take to get there, I'm not too worried about adding it when we about to start work on it, and next server release should be out by the time we can use it.

For an MVP where there is no relation between build and bundle IDs (see other comment), build_id would still be set and used as normal. If, for example, WASM bundles do not implement data converters (which is likely, because side effects), then you will need to rev versions to account for incompatible changes to those. So, SDK would handle downloading the appropriate bundle, but you'd still want to lean on the normal versioning mechanism for changes to things like dataconverters.

temporal/api/common/v1/message.proto
Sushisource: I'm fine with singular here. Either interpretation makes sense, just preferential.

temporal/api/taskqueue/v1/message.proto
Sushisource: You would eventually want to specify which build id sets are compatible with which bundle id sets. EX: Your WASM bundle is only compatible with the data converters in workers with build id `1.x`, so the wasm bundle set `w1.x` would be compatible with the build id set `1.x`, etc.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: I dunno if it's worth it. The most we would return now is 10k versions, if every set is maxed out, and that's still not a gigantor response.

This mostly exist to just have a "get only the current default" toggle.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: I dunno if we do. I figured this would come in handy for debugging purposes.

We could remove it from here, and the compat set definition, and wrap that in the persistence layer. Doing so removes the ability to eventually have sets reference one another, though. I dunno that it's doing much harm here, given people are gonna be interacting with this via the client or tctl.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: In a world where there are multiple version types, there would be a latest default per-type

temporal/api/taskqueue/v1/message.proto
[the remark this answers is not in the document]
bergundy: Let’s leave it in, we could with relatively low effort implement multiple bundles in a single worker if we have that field.

Bundled codecs is still not enough to not require buildID since you still have local activities.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Removed

temporal/api/history/v1/message.proto
```diff
@@ -192,9 +192,10 @@ message WorkflowTaskCompletedEventAttributes {
     string identity = 3;
     // Binary ID of the worker who completed this task
     string binary_checksum = 4;
-    // ID of the worker who picked up this workflow task, or missing if worker
-    // is not using versioning.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    // Version info of the worker who processed this workflow task, or missing if worker is not
+    // using versioning. If present, the `build_id` field within supersedes `binary_checksum`, which
+    // may be populated with the same value to preserve compatability.
+    temporal.api.common.v1.WorkerVersionStamp worker_versions = 5;
```
bergundy, line 198:
```suggestion
    temporal.api.common.v1.WorkerVersionStamp worker_version = 5;
```

temporal/api/taskqueue/v1/message.proto
```diff
@@ -87,22 +87,11 @@ message StickyExecutionAttributes {
...
-    VersionId version = 1;
-    // A pointer to the previous version this version is considered to be compatible with
-    VersionIdNode previous_compatible = 2;
-    // A pointer to the last incompatible version (previous major version)
-    VersionIdNode previous_incompatible = 3;
+// Used by the worker versioning APIs, represents an ordering of one or more versions which are
+// considered to be compatible with each other. Currently the versions are always worker build ids.
+message CompatibleVersionSet {
+    // A unique identifier for this version set
+    string id = 1;
```
bergundy, line 94: What's the reason to expose this in the API?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -292,11 +292,11 @@ message RespondWorkflowTaskCompletedRequest {
...
-    // If using versioning, worker should send the same id here that it used to
-    // poll for the workflow task.
-    // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
-    // set, that value should also be considered as the `binary_checksum`.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 10;
+    // If using versioning, the worker uses this field to indicate what version(s) it used to
+    // process the task.
+    // When this field has a `worker_build_id`, and `binary_checksum` is not set, that value should
+    // also be considered as the `binary_checksum`.
+    temporal.api.common.v1.WorkerVersionStamp worker_versions_stamp = 10;
```
bergundy, line 299:
```suggestion
    temporal.api.common.v1.WorkerVersionStamp worker_version_stamp = 10;
```

temporal/api/taskqueue/v1/message.proto
Sushisource: Debugging, mainly. Users almost never see the API directly, but this will help us to have in logs, etc.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1006,43 +1006,60 @@ message ListSchedulesResponse {
...
-    // for the queue. Workers matching it will begin processing new workflow executions.
-    // The existing default will be marked as a previous incompatible version
-    // to this one, assuming it is not also in `is_compatible_with`.
-    bool become_default = 5;
-}
-message UpdateWorkerBuildIdOrderingResponse {}
+    oneof operation {
+        // An id for a new version. This operation will create a new set which will be the new
+        // overall default version for the queue, with this version as its only member.
+        string new_default_major_version_id = 3;
```
bergundy, line 1030: I know we've used the term "major" internally but I don't see it expressed in the API anywhere else.

As a personal preference, I liked the previous flat API better.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -87,22 +87,11 @@ message StickyExecutionAttributes {
...
+    // All the compatible versions, ordered from oldest to newest
```
bergundy, line 95: I find oldest to newest to be confusing since I can use the API to mark a previous version as the default which would not make it the "newest".

Sushisource: It will do that

bergundy: If that's what you consider "newest", it doesn't read that way to me.
I'd just add the the "default" will alway be the last version in the list.

Sushisource: Well, within a set, the order never changes. You only mark overall sets as defaults, and the comment for that is clear about it. There's no reason to have some kind of "inner" default, since all the versions are compatible, we always pick the latest one.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Changed name

temporal/api/taskqueue/v1/message.proto
bergundy: I'd add this to the docstring

temporal/api/taskqueue/v1/message.proto
bergundy: Oh I see, so there's no way to revert marking a build ID as "latest" (or default) in the set?

What if I mark a build as compatible in the set and later change my mind? What if the deployment for that build fails?
How do I tell the server to route requests in that set to a previous build?

Sushisource: I'll add a promote-within-set op

bergundy: I still think that we should have a `default` attribute on this message to mark the default in the set so we can leave the `versions` array sorted by insertion time.
Another thing we might want to consider is to store some more information like when was each version added to the set and by who (using `identity` in the request).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -774,6 +773,9 @@ message QueryWorkflowResponse {
 message DescribeWorkflowExecutionRequest {
     string namespace = 1;
     temporal.api.common.v1.WorkflowExecution execution = 2;
+    // If set, the response will include which build id shall be used to process the pending
+    // workflow task, if any.
+    bool include_pending_workflow_task_build_id = 3;
```
bergundy, line 778: Why not include this in every response?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1010,43 +1012,99 @@ message ListSchedulesResponse {
...
-    // other ID of a certain type at a time, and any setting which would create a cycle is invalid.
-    temporal.api.taskqueue.v1.VersionId previous_compatible = 4;
-    // When set, establishes the specified `version_id` as the default of it's type
-    // for the queue. Workers matching it will begin processing new workflow executions.
-    // The existing default will be marked as a previous incompatible version
-    // to this one, assuming it is not also in `is_compatible_with`.
-    bool become_default = 5;
-}
-message UpdateWorkerBuildIdOrderingResponse {}
+    oneof operation {
```
bergundy, line 1033: All of these are exactly what we need but I feel like I'd have to look at the docs every time to know what each of these operations mean.

This is better IMHO:
```proto
message AddNewVersion {
  string version = 1;
  string compatible_with_version = 2;
  bool default_for_queue = 3;
}

message PromoteExistingVersion {
  string version = 1;
  bool default_for_queue = 2;
}
```

There's one case that we might not want to allow: `AddNewVersion(version, default_for_queue=false)` but I think the readability makes up for that.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1010,43 +1012,99 @@ message ListSchedulesResponse {
...
+    // There may be fewer sets returned than exist, if the request chose to limit this response.
+    repeated temporal.api.taskqueue.v1.CompatibleVersionSet major_version_sets = 1;
+
+    message RetirementCandidate {
+        // The worker build id which is ready for retirement
+        string build_id = 1;
+        // If true, there are no open *or* closed workflows, meaning there is no reason at all
+        // to keep the worker alive, not even to service queries on closed workflows. If not true,
+        // then there are no open workflows, but some closed ones.
+        bool no_closed_workflows = 2;
```
bergundy, line 1089: Prefer `has_closed_workflows` instead of using the negative form, later on in code it avoids conditions like `if not no_closed_workflows` which have more parsing overhead.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1010,43 +1012,99 @@ message ListSchedulesResponse {
...
+    // default major version set. 0 returns all sets.
+    int32 max_sets = 3;
+    // If set, the response will include information about worker versions which are ready to be
+    // retired.
+    bool include_retirement_candidates = 4;
+    // If set, the response will include information about which versions have open workflows, and
+    // whether or not there are currently polling workers who are compatible with those versions.
+    bool include_poller_compatability = 5;
 }
 message GetWorkerBuildIdOrderingResponse {
```
bergundy, line 1075: Do we want to call this concept "Build Id Ordering"?
Maybe "Build Id Compatibility" or "Mapping"?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1010,43 +1012,99 @@ message ListSchedulesResponse {
...
-    int32 max_depth = 3;
+    // Limits how many compatible sets will be returned. Specify 1 to only return the current
+    // default major version set. 0 returns all sets.
+    int32 max_sets = 3;
+    // If set, the response will include information about worker versions which are ready to be
+    // retired.
+    bool include_retirement_candidates = 4;
+    // If set, the response will include information about which versions have open workflows, and
+    // whether or not there are currently polling workers who are compatible with those versions.
+    bool include_poller_compatability = 5;
```
bergundy, line 1073: I'm not sure I follow why this is needed.
It's also way too late for me to be reviewing this..

temporal/api/workflowservice/v1/request_response.proto
Sushisource: It's more expensive than a normal describe op.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: I don't think we should be optimizing our gRPC API for readability over correctness. We make this easy to use / readable in our clients and tctl. The API should optimize for correctness, so that when we implement those things mistakes are easily prevented. Looking at the docstring is quick and easy in every editing environment.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: It'd need to be `is_only_closed_workflows` to be correct, but sure agreed.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: It allows you to understand if there are tasks that can't make progress b/c there aren't compatible pollers

temporal/api/workflowservice/v1/request_response.proto
bergundy: I think we should be optimizing for readability wherever possible TBH.
Notice the consistency in my proposed message names, they start with verbs as one might expect for an operation, they follow the same format, etc.
I'd be down to see an alternative proposal that disallows `AddNewVersion(version, default_for_queue=false)` and is easier to parse.

Sushisource: I'm talking about relative values, and correctness _over_ readability is I think pretty clear. 

But, regardless, I'm really not following what's hard to read about this. I could just make the field names slightly longer and that would seem to do it.

bergundy: I'd say there are 3 issues with your proposed API:
1. These 2 variants sound the same, `existing_version_id_in_set_to_promote` and `promote_version_id_within_set`.
2. Some variant names sound like operations while others don't (the ones that don't start with a verb).
3. The term "default" is overloaded, there's a set default and a queue default, anywhere we mention `default` we should explicitly disambiguate.

Sushisource: I have made the field names a bit more clear to address this

temporal/api/taskqueue/v1/message.proto
```diff
@@ -87,22 +87,12 @@ message StickyExecutionAttributes {
...
-    // A pointer to the previous version this version is considered to be compatible with
-    VersionIdNode previous_compatible = 2;
-    // A pointer to the last incompatible version (previous major version)
-    VersionIdNode previous_incompatible = 3;
+// Used by the worker versioning APIs, represents an ordering of one or more versions which are
+// considered to be compatible with each other. Currently the versions are always worker build ids.
+message CompatibleVersionSet {
+    // A unique identifier for this version set. Users don't need to understand or care about this
+    // value, but it has value for debugging purposes.
+    string id = 1;
```
dnr, line 95: maybe:
```suggestion
    string version_set_id = 1;
```
to be clear?

dnr, line 95: do we want to document anything about this value? I can say that it may not contain `/` or `:`. but maybe it's just distracting to document that since it's assigned by the server and clients should treat it as an opaque string

temporal/api/taskqueue/v1/message.proto
```diff
@@ -87,22 +87,12 @@ message StickyExecutionAttributes {
...
+    // All the compatible versions, ordered from oldest to newest
+    repeated string versions = 2;
```
dnr, line 97: shouldn't we name it
```suggestion
    repeated string build_ids = 2;
```
to use the same name for the same type consistently?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1041,44 +1043,103 @@ message ListSchedulesResponse {
 //     aip.dev/not-precedent: UpdateWorkerBuildIdOrderingRequest doesn't follow Google API format --)
 // (-- api-linter: core::0134::request-resource-required=disabled
 //     aip.dev/not-precedent: UpdateWorkerBuildIdOrderingRequest RPC doesn't follow Google API format. --)
-message UpdateWorkerBuildIdOrderingRequest {
+message UpdateWorkerBuildIdCompatabilityRequest {
+    message AddNewCompatibleVersion {
+        // A new id to be added to an existing compatible set.
+        string new_version_id = 1;
```
dnr, line 1049:
```suggestion
        string new_build_id = 1;
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1041,44 +1043,103 @@ message ListSchedulesResponse {
...
-}
-message UpdateWorkerBuildIdOrderingResponse {}
+    oneof operation {
+        // An id for a new version. This operation will create a new set which will be the new
+        // overall default version for the queue, with this version as its only member. This new
+        // set is incompatible with all previous sets/versions.
+        //
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: In makes perfect sense here. --)
+        string add_new_version_id_in_new_default_set = 3;
```
dnr, line 1072:
```suggestion
        string add_new_build_id_in_new_default_set = 3;
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1041,44 +1043,103 @@ message ListSchedulesResponse {
...
+        // Adds a new id to an existing compatible set, see sub-message definition for more.
+        AddNewCompatibleVersion add_new_compatible_version = 4;
+        // Promote an existing set to be the current default (if it isn't already) by targeting
+        // an existing version id within it. This field's value is the extant version id.
+        //
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: Names are hard. --)
+        string promote_set_by_version_id = 5;
```
dnr, line 1080:
```suggestion
        string promote_set_by_build_id = 5;
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1041,44 +1043,103 @@ message ListSchedulesResponse {
...
+        // Promote an existing version id within some set to be the current default for that set.
+        //
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: Within makes perfect sense here. --)
+        string promote_version_id_within_set = 6;
```
dnr, line 1085:
```suggestion
        string promote_build_id_within_set = 6;
```

temporal/api/workflow/v1/message.proto
```diff
@@ -100,6 +100,9 @@ message PendingWorkflowTaskInfo {
     google.protobuf.Timestamp original_scheduled_time = 3 [(gogoproto.stdtime) = true];
     google.protobuf.Timestamp started_time = 4 [(gogoproto.stdtime) = true];
     int32 attempt = 5;
+    // If set, this pending workflow task will need to be (or is currently being) handled by a
+    // a worker using this build id.
+    string will_use_build_id = 6;
```
dnr, line 105: 1. I don't remember talking about this.. what's the planned use case?
2. what's the plan to implement it? I don't think it's obvious.
4. even if this is needed, what do you think about adding the WorkerVersionStamp for the previous wft in WorkflowExecutionInfo? that's going to be in mutable state so it's free to copy in there

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -310,11 +310,10 @@ message RespondWorkflowTaskCompletedRequest {
...
     string namespace = 9;
-    // If using versioning, worker should send the same id here that it used to
-    // poll for the workflow task.
-    // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
-    // set, that value should also be considered as the `binary_checksum`.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 10;
+    // If using versioning, the worker uses this field to indicate what version(s) it used to
+    // process the task. When this field has a `worker_build_id`, and `binary_checksum` is not set,
+    // that value should also be considered as the `binary_checksum`.
+    temporal.api.common.v1.WorkerVersionStamp worker_version_stamp = 10;
```
dnr, line 316: what happens if this field is missing for a workflow that was previously processed by a versioned worker? should we document here that it's not allowed? (and I can make the server reject the wft?)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -237,12 +237,12 @@ message PollWorkflowTaskQueueRequest {
...
-    // If set, the worker is opting in to build-id based versioning and wishes to only
-    // receive tasks that are considered compatible with the version provided.
+    // If set, the worker is opting in to versioning and wishes to only
+    // receive tasks that are considered compatible with the version capabilities provided.
     // Doing so only makes sense in conjunction with the `UpdateWorkerBuildIdOrdering` API.
-    // When `worker_versioning_id` has a `worker_build_id`, and `binary_checksum` is not
+    // When this field has a `worker_build_id`, and `binary_checksum` is not
     // set, that value should also be considered as the `binary_checksum`.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    temporal.api.common.v1.WorkerVersionCapabilities worker_version_capabilities = 5;
```
dnr, line 245: is there going to be a capability flag for this? how should a versioned worker behave if it connects to a server that doesn't support versioning?

temporal/api/common/v1/message.proto
```diff
@@ -121,3 +121,20 @@ message MeteringMetadata {
...
     uint32 nonfirst_local_activity_execution_attempts = 13;
 }
+
+// Identifies the version(s) that a worker used to process a task
+message WorkerVersionStamp {
+    // An opaque whole-worker identifier
+    string build_id = 1;
+    // Set if the worker used a dynamically loadable bundle to process
+    // the task. The bundle could be a WASM blob, JS bundle, etc.
+    string bundle_id = 2;
```
dnr, line 131: is bundle_id included in build_id, i.e. does build_id definitely change if bundle_id changes? the compatible set stuff is all about build ids, not bundle ids at all, so if one build_id can use multiple bundles then it seems like this won't work

temporal/api/history/v1/message.proto
```diff
@@ -193,9 +193,10 @@ message WorkflowTaskCompletedEventAttributes {
     string identity = 3;
     // Binary ID of the worker who completed this task
     string binary_checksum = 4;
-    // ID of the worker who picked up this workflow task, or missing if worker
-    // is not using versioning.
-    temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    // Version info of the worker who processed this workflow task, or missing if worker is not
+    // using versioning. If present, the `build_id` field within supersedes `binary_checksum`, which
+    // may be populated with the same value to preserve compatability.
+    temporal.api.common.v1.WorkerVersionStamp worker_version = 5;
```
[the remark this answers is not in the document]
Sushisource: I think we'll have to have it around for some deprecation period before we just eliminate it entirely?

temporal/api/taskqueue/v1/message.proto
Sushisource: I think it'd probably just be distracting

temporal/api/taskqueue/v1/message.proto
Sushisource: Ya good call

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Good call out - it shouldn't be the case that such a worker ever got the task, so replying to it with this empty should indeed be a nonsense situation that we reject. Added to comments.

temporal/api/common/v1/message.proto
Sushisource: My intention was that any worker with a given build_id should be able to load up all bundle_ids that any other compatible worker (as defined by build_id compatibility) would also be able to load. Hence by construction it should slot into the idea of build_id compatibility.

To directly answer your question - no, build_id doesn't necessarily change if bundle_id changes. However, from a compat perspective it should be consistent.

temporal/api/workflow/v1/message.proto
Sushisource: 1. IIRC I thought the reason for existence is to make it easier in the UI / CLI to show what the next task will use, and be able to easily do that "no compatible pollers" warning.... but as I write next that seems to be redundant.
2. You tell me :shrug: -- if this is hard to do, we can eliminate it since this info is available via the `active_versions_and_pollers` field in `GetWorkerBuildIdCompatabilityResponse`. This is technically slightly different since it shows last version / current compatible pollers rather than "next version"... but seems like we should probably just get rid of this unless we remember we need it for some reason and add it then.
3. Yes, I think this makes sense to add regardless

temporal/api/workflowservice/v1/request_response.proto
dnr: In a discussion this week we decided we want to be able to turn this on for individual namespaces, so we might need more than capabilities. We could figure that out in another PR

temporal/api/workflowservice/v1/request_response.proto
dnr: Was this to populate `will_use_build_id`? If so this can be removed

Sushisource: Thanks, fixed.

temporal/api/common/v1/message.proto
Sushisource: Ah, because it simply looks at the history which includes the version stamp on the last WFT complete, and downloads that  bundle.

dnr (approved):

PR #239, opened 2022-09-26
Remove unnecessary namespace fields from Batch API

feedmeapples:
**What changed?**

- These namespaces are not used since
- Added Job ID to start batch operation response

**Why?**

- Batch jobs run in Temporal System Local namespace and not in the namespace from requests
- returning Job ID can help with making the Job ID in request optional instead of required 

This came up after working on https://github.com/temporalio/temporal/pull/3428

**Breaking changes**


feedmeapples:
is this something we now can't just remove? should we use placeholders then?

PR #240, opened 2022-09-27
Copy namespacpace APIs to operator service

alexshtin:
**What changed?**
Copy namespacpace APIs to operator service.

**Why?**
Namespace APIs on workflow service will be deprecated and removed later.

**Breaking changes**
No.


alexshtin:
After some discussions we decided not to do it for now, because we want deeper refactoring of namespace APIs.

bergundy:
@alexshtin why not move it already? It’s not making things worse

PR #241, opened 2022-09-27
Unify proto build process with server

alexshtin:
**What changed?**
Unify proto build process with server.

**Why?**
Parity with https://github.com/temporalio/temporal/pull/3377.

**Breaking changes**
No.


go.mod
```diff
@@ -0,0 +1,3 @@
+module api
```
dnr, line 1: maybe add a comment on this file explaining why it exists at all

dnr (approved):

Makefile
```diff
@@ -45,21 +45,21 @@ fix-path:
...
 	printf $(COLOR) "Install/update gRPC plugins..."
-	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2.0
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 
 gogo-protobuf-install: go-protobuf-install
-	GO111MODULE=off go get github.com/temporalio/gogo-protobuf/protoc-gen-gogoslick
+	go install -modfile build/go.mod github.com/temporalio/gogo-protobuf/protoc-gen-gogoslick
 
 go-protobuf-install:
 	go install github.com/golang/protobuf/protoc-gen-go@v1.5.2
```
dnr, line 54: er, wait, why are these not in build/go.mod also? is that planned?

dnr:
> Why is there Go code gen stuff in this repo instead of [temporalio/api-go](https://github.com/temporalio/api-go)? Can we remove all `Makefile` targets that don't start with `api-linter` and `buf`?

It runs everything through protoc (with gogo and grpc extensions, etc.) just as a syntax check. The output is ignored. If we're willing to rely on buf to do a thorough check enough itself, I think maybe we can get rid of them. @alexshtin?

Makefile
alexshtin: Because they can. `go install` doesn't touch `go.mod` if there is a version specified. And yes, I plan to move all tools in all repos under `build/go.mod`.

go.mod
alexshtin: I thought about it. go gives you an error if you remove it, but I agree, I will add it.

alexshtin:
Yes, David is right. When I added it buf wasn't enough. It gets better and may be can do validation better. I will check this later.

Makefile
```diff
@@ -45,21 +45,21 @@ fix-path:
 ##### Plugins & tools #####
 grpc-install: gogo-protobuf-install
 	printf $(COLOR) "Install/update gRPC plugins..."
-	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2.0
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 
 gogo-protobuf-install: go-protobuf-install
-	GO111MODULE=off go get github.com/temporalio/gogo-protobuf/protoc-gen-gogoslick
+	go install -modfile build/go.mod github.com/temporalio/gogo-protobuf/protoc-gen-gogoslick
```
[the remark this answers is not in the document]
alexshtin: To be able to call `go list -modfile build/go.mod -m -f '{{.Dir}}'` (above) module needs to be in `go.mod`. I can add it to the root `go.mod` but on server we have in separate `build` dir and I want to keep parts of `Makefile` the same to simplify future synchronization.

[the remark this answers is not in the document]
cretz: > simplify future synchronization.

Eek. Will approve, but I really hope we get the platform-specific build script and the language-specific files out of this generic repo one day.

cretz (approved):

PR #243, opened 2022-10-11
Add merged search attributes and memo maps to upsert history events

rodrigozhou:
**What changed?**
Add fields to search attributes and memo upsert events to store the merged/final maps.

**Why?**
Adding those fields so they are stored in the history events and they can be removed from mutable state. More info: https://github.com/temporalio/temporal/pull/3471

**Breaking changes**
No.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -507,6 +507,9 @@ message UpsertWorkflowSearchAttributesEventAttributes {
     // The `WORKFLOW_TASK_COMPLETED` event which this command was reported with
     int64 workflow_task_completed_event_id = 1;
     temporal.api.common.v1.SearchAttributes search_attributes = 2;
+    // `merged_search_attributes` contains the result of upserting
+    // search attributes if they are not the same
+    temporal.api.common.v1.SearchAttributes merged_search_attributes = 3;
 }
 
 message WorkflowPropertiesModifiedEventAttributes {
@@ -516,6 +519,9 @@ message WorkflowPropertiesModifiedEventAttributes {
     // the existing memo. If the user wants to delete values, a default/empty Payload should be
     // used as the value for the key being deleted.
     temporal.api.common.v1.Memo upserted_memo = 2;
+    // `merged_memo` contains the result of upserting memo
+    // if `upserted_memo` is not nil and they are not the same
+    temporal.api.common.v1.Memo merged_memo = 3;
 }
 
 message StartChildWorkflowExecutionInitiatedEventAttributes {


temporal/api/history/v1/message.proto
```diff
@@ -507,6 +507,9 @@ message UpsertWorkflowSearchAttributesEventAttributes {
     // The `WORKFLOW_TASK_COMPLETED` event which this command was reported with
     int64 workflow_task_completed_event_id = 1;
     temporal.api.common.v1.SearchAttributes search_attributes = 2;
+    // `merged_search_attributes` contains the result of upserting
+    // search attributes if they are not the same
+    temporal.api.common.v1.SearchAttributes merged_search_attributes = 3;
```
cretz, line 512: Can you provide some more information on the justification for this? Specifically, what about today's solution is not acceptable?

cretz, line 512: To confirm, if I have 100 search attributes in the start event, and I add 1, this is not gonna have 101 values in it?

bergundy: I see @cretz's concern.
It feels like we're potentially moving the problem from one place to another.
Apart from the obvious implication of increasing the required disk space to store the duplicates, we'll incur network overhead for shipping this to the SDK for every replay.
Can we consider not sending the merged attributes back to the SDK?

Just as a note, I would like to have the merged search attributes in history in some cases though assuming the server will ever allow deleting history events.

dnr: First, I'd like to suggest a slight modification of the semantics of this value to reduce space concerns:

This should contain all values in the merged set that are not in `search_attributes`. That is, if I previously had `{a:1, b:2, c:3, d:4}` and I upsert `{b:22, d:44}`, then `search_attributes` would contain `{b:22, d:44}` and this field would contain `{a:1, c:3}`. To get the merged set you'd merge them back again.

The name would then be something like `previous_search_attributes`.

The benefit is that each attribute would only appear once per event, so less duplication.

dnr: Second, it seems more common to update all or most of a workflow's search attributes each time, so I think this "previous" field won't end up being large overall, if we do it that way.

temporal/api/history/v1/message.proto
```diff
@@ -516,6 +519,9 @@ message WorkflowPropertiesModifiedEventAttributes {
     // the existing memo. If the user wants to delete values, a default/empty Payload should be
     // used as the value for the key being deleted.
     temporal.api.common.v1.Memo upserted_memo = 2;
+    // `merged_memo` contains the result of upserting memo
+    // if `upserted_memo` is not nil and they are not the same
+    temporal.api.common.v1.Memo merged_memo = 3;
```
dnr, line 524: I'd suggest the same change of semantics here.

temporal/api/history/v1/message.proto
cretz: Is there a way to alleviate server internal state concerns w/out modifying the user-facing events/history?

rodrigozhou:
I'm abandoning this PR. We're gonna take another approach to reduce the size of the mutable state.

PR #245, opened 2022-10-26
Comment with default page size

lorensr:
Add comment with default value so that users know what they'll get if they don't provide it.


cretz (approved):

mindaugasrukas (changes requested):
I wouldn't say I like these comments, as they might become outdated and difficult to maintain. Is there a way to have a direct association with the default value?

mindaugasrukas:
Also, please add a description/reason for the change.

lorensr:
> I wouldn't say I like these comments, as they might become outdated and difficult to maintain. Is there a way to have a direct association with the default value?

I don't know where they exist or how we'd directly associate

PR #246, opened 2022-10-28
Convert update to interactions

alexshtin:
**What changed?**
Convert Workflow update functionality to interactions.

**Why?**
New design proposal.

**Breaking changes**
Yes, but update is not released yet.

### DIFF
--- a/temporal/api/enums/v1/interaction_type.proto
+++ b/temporal/api/enums/v1/interaction_type.proto
@@ -22,25 +22,18 @@
 
 syntax = "proto3";
 
-package temporal.api.update.v1;
+package temporal.api.enums.v1;
 
-option go_package = "go.temporal.io/api/update/v1;update";
-option java_package = "io.temporal.api.update.v1";
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
 option java_multiple_files = true;
-option java_outer_classname = "MessageProto";
-option ruby_package = "Temporal::Api::Update::V1";
-option csharp_namespace = "Temporal.Api.Update.V1";
+option java_outer_classname = "InteractionTypeProto";
+option ruby_package = "Temporal::Api::Enums::V1";
+option csharp_namespace = "Temporal.Api.Enums.V1";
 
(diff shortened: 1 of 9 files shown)


temporal/api/enums/v1/interaction_type.proto
```diff
@@ -22,25 +22,18 @@
...
-    // Headers that are passed with the update to the processing workflow.
-    // These can include things like auth or tracing tokens.
-    temporal.api.common.v1.Header header = 1;
-
-    // The name of the update function to invoke on the target workflow.
-    string name = 2;
-
-    // The arguments to pass to the named update function.
-    temporal.api.common.v1.Payloads args = 3;
+enum InteractionType {
```
bergundy, line 34: I thought we're calling these messages now.. did that change?

bergundy (commented):
This doesn't match what we discussed yesterday in our design session.

temporal/api/enums/v1/interaction_type.proto
mmcshane: This is an intermediate step. Alex has some code that we want to get working end-to-end but it hasn't adopted the protocol/message paradigm yet. We want to be refactoring from a known-working base state.

mmcshane (approved):

alexshtin:
@bergundy, this is v2 of our design. What we discussed is v3. We might postpone v3 to after upcomming server release (1.19).

yiminc (approved):

bergundy:
Got it, thanks for the update, this came out of the blue for me.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -304,6 +306,8 @@ message RespondWorkflowTaskCompletedResponse {
     PollWorkflowTaskQueueResponse workflow_task = 1;
     // See `ScheduleActivityTaskCommandAttributes::request_start`
     repeated PollActivityTaskQueueResponse activity_tasks = 2;
+
+    int64 reset_history_event_id = 3;
```
Sushisource, line 310: This already got merged, but, this really deserves a docstring.

PR #248, opened 2022-11-03
Add search attribute type for list of keyword

rodrigozhou:
**What changed?**
Adding enum for search attribute type list of keywords.

**Why?**
We want to make distinction between single keyword vs list of keyword. Other types will no longer support list of values.

**Breaking changes**
No.

### DIFF
--- a/temporal/api/enums/v1/common.proto
+++ b/temporal/api/enums/v1/common.proto
@@ -45,6 +45,7 @@ enum IndexedValueType {
     INDEXED_VALUE_TYPE_DOUBLE = 4;
     INDEXED_VALUE_TYPE_BOOL = 5;
     INDEXED_VALUE_TYPE_DATETIME = 6;
+    INDEXED_VALUE_TYPE_KEYWORD_LIST = 7;
 }
 
 enum Severity {


temporal/api/enums/v1/common.proto
```diff
@@ -45,6 +45,7 @@ enum IndexedValueType {
     INDEXED_VALUE_TYPE_DOUBLE = 4;
     INDEXED_VALUE_TYPE_BOOL = 5;
     INDEXED_VALUE_TYPE_DATETIME = 6;
+    INDEXED_VALUE_TYPE_LIST_KEYWORD = 7;
```
bergundy, line 48:
```suggestion
    INDEXED_VALUE_TYPE_KEYWORD_LIST = 7;
```

Or

```
    INDEXED_VALUE_TYPE_KEYWORDS = 7;
```

bergundy (approved):
Thanks!
Had a single suggestion.

yiminc (approved):

temporal/api/enums/v1/common.proto
rodrigozhou: @bergundy 
I thought as "list of keywords" or `list<keyword>`, thus `LIST_KEYWORD`. Similarly to why it's `TYPE_KEYWORD` and not `KEYWORD_TYPE`.

I don't like `KEYWORDS`, just a single `S` to differentiate. I feel like it's prone to typo bugs.

bergundy: Agree that `KEYWORDS` is too hard to differentiate from `KEYWORD`.

`KEYWORD_LIST` reads better to me than `LIST_KEYWORD`.
I totally see where you're coming from but I prefer the value to be readable like an english sentence.

Not blocking this PR though, up to you to decide, it's a matter of personal taste.

Sushisource: +1 to `KEYWORD_LIST`

lorensr: `KEYWORD_LIST` feels more readable to me too. When list is first, I parse it as an adjective, because in English adjectives come before nouns, so I think it's a "list keyword", and I don't know what type of keyword a list keyword is.

rodrigozhou: Updated to `KEYWORD_LIST`.

PR #249, opened 2022-11-09
Add an error code to signify that the number of pending child workflows has been exceeded

MichaelSnowden:
**What changed?**
I added a new error code.

**Why?**
To support child workflow limits.

**Breaking changes**
No.

### DIFF
--- a/temporal/api/enums/v1/failed_cause.proto
+++ b/temporal/api/enums/v1/failed_cause.proto
@@ -67,6 +67,24 @@ enum WorkflowTaskFailedCause {
     // what the workflow code actually did.
     WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR = 24;
     WORKFLOW_TASK_FAILED_CAUSE_BAD_MODIFY_WORKFLOW_PROPERTIES_ATTRIBUTES = 25;
+
+    // We send the below error codes to users when their requests would violate a size constraint
+    // of their workflow. We do this to ensure that the state of their workflow does not become too
+    // large because that can cause severe performance degradation. You can modify the thresholds for
+    // each of these errors within your dynamic config.
+    //
+    // Spawning a new child workflow would cause this workflow to exceed its limit of pending child
+    // workflows.
+    WORKFLOW_TASK_FAILED_CAUSE_PENDING_CHILD_WORKFLOWS_LIMIT_EXCEEDED = 26;
+    // Starting a new activity would cause this workflow to exceed its limit of pending activities
+    // that we track.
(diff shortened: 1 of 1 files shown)


bergundy (approved):

temporal/api/enums/v1/failed_cause.proto
```diff
@@ -67,6 +67,8 @@ enum WorkflowTaskFailedCause {
     // what the workflow code actually did.
     WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR = 24;
     WORKFLOW_TASK_FAILED_CAUSE_BAD_MODIFY_WORKFLOW_PROPERTIES_ATTRIBUTES = 25;
+    WORKFLOW_TASK_FAILED_CAUSE_PENDING_CHILD_WORKFLOWS_LIMIT_EXCEEDED = 26;
+    WORKFLOW_TASK_FAILED_CAUSE_PENDING_ACTIVITIES_LIMIT_EXCEEDED = 27;
```
yycptt, line 71: Just to confirm, we will have one cause for each pending entity map in mutable state? And causes for pending signal, pending cancellation will be added later?

yiminc: @MichaelSnowden I think we should add for pending signals and pending cancellations as well. What about pending timers @yycptt ? Do you think we need that limitation as well?

yycptt: The size for each timer info is quite small, so caller may probably hit the history count limit first. 

But still I think we need to have the capability to limit # of pending timers. It will make things much easier and make system easy to reason about.

MichaelSnowden: I've added pending signals and pending cancellations. If we need timers, we can add them in a follow-up PR.

temporal/api/enums/v1/failed_cause.proto
```diff
@@ -67,6 +67,10 @@ enum WorkflowTaskFailedCause {
...
+    WORKFLOW_TASK_FAILED_CAUSE_PENDING_SIGNALS_LIMIT_EXCEEDED = 28;
```
yiminc, line 72: The name is not clear what it is limiting. 
This is the limit of number of pending signals to be sent to external workflows. 
Please add comments to all the new enums.

yiminc (approved):

temporal/api/enums/v1/failed_cause.proto
MichaelSnowden: Added comments

PR #253, opened 2023-01-04
Convert updates to protocol/message data types

alexshtin:
**What changed?**
Convert updates to `protocol`/`message` data types.

**Why?**
Previous version was very specific to workflow updates. After many discussions between @temporalio/sdk and @temporalio/server teams we came up to the idea to create generic messaging protocol which can be reused for other purposes. 

**Breaking changes**
Yes, but previous version was never released and used.


temporal/api/protocol/v1/message.proto
```diff
@@ -22,18 +22,22 @@
...
 
-enum InteractionType {
-   INTERACTION_TYPE_UNSPECIFIED = 0;
-   INTERACTION_TYPE_WORKFLOW_QUERY = 1;
-   INTERACTION_TYPE_WORKFLOW_UPDATE = 2;
-   INTERACTION_TYPE_WORKFLOW_SIGNAL = 3;
+import "google/protobuf/any.proto";
+
+// (-- api-linter: core::0146::any=disabled
+//     aip.dev/not-precedent: We want runtime extensibility for the body field --)
```
alexshtin, line 37: Is `Any` forbidden by `api-linter`?

temporal/api/protocol/v1/message.proto
```diff
@@ -22,18 +22,22 @@
...
+message Message {
+    string id = 1;
+    string protocol_id = 2;
```
alexshtin, line 40: What are the examples of `protocol_id`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1051,40 +1055,33 @@ message GetWorkerBuildIdOrderingResponse {
...
 }
 
-message UpdateWorkflowResponse {
-    // An opaque token that can be used to retrieve the update result via
-    // polling if it is not returned as part of the gRPC response
-    bytes update_token = 1;
-    // The success or failure status of the update
-    temporal.api.interaction.v1.Output output = 2;
+message UpdateWorkflowExecutionResponse {
+    temporal.api.update.v1.UpdateRef update_ref = 1;
```
alexshtin, line 1083: Why is it here? To unify with future `Poll` response?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -323,6 +325,8 @@ message RespondWorkflowTaskFailedRequest {
     // Worker process' unique binary id
     string binary_checksum = 5;
     string namespace = 6;
+    // Protocol messages piggybacking on a WFT as a transport
+    repeated temporal.api.protocol.v1.Message messages = 7;
```
alexshtin, line 329: What kind of messages are going back to worker here?

temporal/api/update/v1/message.proto
```diff
@@ -0,0 +1,126 @@
...
+option ruby_package = "Temporal::Api::Update::V1";
+option csharp_namespace = "Temporal.Api.Update.V1";
+
+import "temporal/api/common/v1/message.proto";
+import "temporal/api/enums/v1/update.proto";
+import "temporal/api/failure/v1/message.proto";
+
+// Sepcifies to the gRPC server how the client wants the UpdateWorkflowExecution
+// call to block before returning control to the caller.
+message BlockingPolicy {
```
alexshtin, line 40: `WaitingPolicy` (not sure it is better, just a suggestion).

temporal/api/workflowservice/v1/request_response.proto
alexshtin: Actually it would be nice to see examples of all 3 messages.

temporal/api/protocol/v1/message.proto
mmcshane: It would identify the instance of a running protocol. So for any given update, this would identify the update state machine within the WFE amongst the other concurrently executing protocols. It doesn't have to be globally unique but the easiest thing to do in our case is probably re-use the UpdateID.

temporal/api/protocol/v1/message.proto
mmcshane: Yes. https://google.aip.dev/146#any

cretz:
Can we make sure https://github.com/temporalio/api-go/pull/100 and https://github.com/temporalio/sdk-go/pull/974 land first? Otherwise compatibility issues may be caused with our proxy code gen as they did w/ these types of changes before. For example, see how we reference https://github.com/temporalio/sdk-go/blob/06e474c93e936b71dc4afcec973460b22c13986d/converter/grpc_interceptor.go#L192.

Also @alexshtin can you please review https://github.com/temporalio/api-go/pull/100 so we can move forward on it?

temporal/api/protocol/v1/message.proto
alexshtin: Oh.. so it is not "id of protocol" but "id defined by protocol". Update protocol defines it as `UpdateID`, some other protocols might use something else.

alexshtin (commented):
LGTM. Integrating with server code.

temporal/api/update/v1/message.proto
```diff
@@ -0,0 +1,112 @@
...
+
+// Used as a placeholder to indicate that a workflow execution update has not
+// yet completed and thus its outcome is not ready.
+message Incomplete { }
+
+// The outcome of a workflow update - either success or failure.
+message Outcome {
+    oneof value {
+        Incomplete incomplete = 1;
+        temporal.api.common.v1.Payloads success = 2;
```
Sushisource, line 62: Just gonna swoop in here to say `Payloads` is my enemy and if we can do just one `Payload` here to avoid the various footguns that `Payloads` introduces that'd be super. Forces people into better practices from the get-go.

alexshtin: There might be more than one return value, right?

temporal/api/protocol/v1/message.proto
```diff
@@ -22,18 +22,33 @@
 
 syntax = "proto3";
 
-package temporal.api.enums.v1;
+package temporal.api.protocol.v1;
 
-option go_package = "go.temporal.io/api/enums/v1;enums";
-option java_package = "io.temporal.api.enums.v1";
+option go_package = "go.temporal.io/api/protocol/v1;messaging";
```
alexshtin, line 27:
```suggestion
option go_package = "go.temporal.io/api/protocol/v1;protocol";
```

temporal/api/protocol/v1/message.proto
```diff
@@ -22,18 +22,33 @@
...
+    // belongs.
+    string protocol_instance_id = 2;
+
+    // The event ID or command ID after which this message can be delivered. The
+    // effects of history up to and including this event ID should be visible to
+    // the code that handles this message. Set to -1 to opt out of sequencing.
+    int64 sequencing_id = 3;
+    
+    // The opaque data carried by this message. The protocol type can be
+    // extracted from the
```
alexshtin, line 52: "from the..." what?

temporal/api/update/v1/message.proto
Sushisource: What I mean to say by that is just don't allow it. Force people into the best practice of returning one struct which contains everything they want, and is usually safer to change later than a list of un-named values.

temporal/api/history/v1/message.proto
```diff
@@ -686,6 +672,21 @@ message ActivityPropertiesModifiedExternallyEventAttributes {
     temporal.api.common.v1.RetryPolicy new_retry_policy = 2;
 }
 
+message WorkflowExecutionUpdateAcceptedEventAttributes {
+    temporal.api.protocol.v1.Message request_message = 1;
```
alexshtin, line 676: I don't want to save entire message here. Seems like:
```suggestion
    temporal.api.update.v1.Meta meta = 1;
```
should be enough.

temporal/api/update/v1/message.proto
alexshtin: I wish we follow this rule everywhere. Basically same way as gRPC does.

mmcshane: I think we can actually do this because right now the reflection code requires that update functions return at either one or two values and one of those is the error.

That being said, it's extremely satisfying that temporal.api.common.v1.Payloads and temporal.api.failure.v1.Failure have the same number of characters so the lines line up just right so

temporal/api/history/v1/message.proto
mmcshane: I'll concede on not storing an Any here. But the worker is going to need enough information in this event to reconstruct the original input protocolpb.Message

temporal/api/update/v1/message.proto
cretz: FWIW, as we're developing many-to-many payload codecs in .NET we discovered that people may want to encode one to many payloads (though they of course and often do encode many to one). The only place today where we use `Payload` without `Payloads` in non-map situations is `Failure.encoded_attributes` and I think @bergundy regrets that.

IMO We should stay with `Payloads` in every non-map situation.

temporal/api/enums/v1/update.proto
```diff
@@ -31,10 +31,9 @@ option java_outer_classname = "UpdateProto";
...
-    WORKFLOW_UPDATE_RESULT_ACCESS_STYLE_UNSPECIFIED = 0;
-
-    // Indicates that the update response _must_ be included as part of the gRPC
-    // response body
-    WORKFLOW_UPDATE_RESULT_ACCESS_STYLE_REQUIRE_INLINE = 1;
+enum UpdateWorkflowExecutionLifecycleStage {
+    UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_UNSPECIFIED = 0;
+    UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED = 1;
+    UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED = 2;
+    UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED = 3;
```
bergundy, line 38: Some comments on these enum values would be great.

temporal/api/history/v1/message.proto
```diff
@@ -686,6 +671,40 @@ message ActivityPropertiesModifiedExternallyEventAttributes {
...
+    // message as part of replay.
+    string accepted_request_message_id = 2;
+    // The message payload of the original request message that initiated this
+    // update.
+    temporal.api.update.v1.Request accepted_request = 3;
+}
+
+message WorkflowExecutionUpdateCompletedEventAttributes {
+    // The metadata about this update.
+    temporal.api.update.v1.Meta meta = 1;
```
bergundy, line 688: Can we spell this out?

```suggestion
    temporal.api.update.v1.Metadata metadata = 1;
```

temporal/api/history/v1/message.proto
```diff
@@ -686,6 +671,44 @@ message ActivityPropertiesModifiedExternallyEventAttributes {
...
+    // The message payload of the original request message that initiated this
+    // update.
+    temporal.api.update.v1.Request accepted_request = 4;
+}
+
+message WorkflowExecutionUpdateCompletedEventAttributes {
+    // The metadata about this update.
+    temporal.api.update.v1.Meta meta = 1;
+    // The outcome of executing the workflow update function.
+    temporal.api.update.v1.Outcome outcome = 2;
```
bergundy, line 692: For consistency with the other events, I would consider having separate Completed and Failed events instead of this internal outcome field.
I like your version better but would prefer consistency over personal preference.

temporal/api/history/v1/message.proto
alexshtin: Does Mark approve?

temporal/api/update/v1/message.proto
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Used as a placeholder to indicate that a workflow execution update has not
+// yet completed and thus its outcome is not ready.
+message Incomplete { }
+
+// The outcome of a workflow update - success, failure, or not done yet.
+message Outcome {
+    oneof value {
+        Incomplete incomplete = 1;
```
bergundy, line 61: Is `incomplete` a valid outcome though?
When would this be used?

bergundy: I see it used below, nm.

temporal/api/history/v1/message.proto
Sushisource: I'd question why exactly is it that consistency is important here? There are vanishingly few users making use of our gRPC APIs directly. Consistency for its own sake appears less valuable to me than better typing.

I won't belabor anything here, but I do think it's important to question what the _value_ of staying consistent is here - if we can't name that, then just citing it as a de-facto good seems like something to be avoided.

bergundy: This is mostly for clarity when viewing a brief summary of the workflow history. Yes, it can be handled in our UI but it would be a special case.
Another point worth mentioning is that the `Incomplete` outcome is not a valid outcome here.

Sushisource: Cool, those make sense to me.

mmcshane: With respect guys, we did this conversation. We talked about how using oneof is a divergence and why both the server and the sdk teams were ok with it. This has been settled.

I can remove Incomplete (it's only invalid until we support other wait policies) for this pre-release but it's coming back shortly.

temporal/api/history/v1/message.proto
```diff
@@ -686,6 +671,40 @@ message ActivityPropertiesModifiedExternallyEventAttributes {
...
+message WorkflowExecutionUpdateAcceptedEventAttributes {
+    // The instance ID of the update protocol that generated this event.
+    string protocol_instance_id = 1;
+    // The message ID of the original request message that initiated this
+    // update. Needed so that the worker can recreate and deliver that same
+    // message as part of replay.
+    string accepted_request_message_id = 2;
+    // The message payload of the original request message that initiated this
+    // update.
+    temporal.api.update.v1.Request accepted_request = 3;
```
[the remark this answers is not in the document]
mmcshane: It reads a little better in the code with the accepted prefix. Otherwise you have just a "request" type and variable in there and you wonder if it's a request about the Acceptance or what the semantics are. With the prefix it's a  more clear that this is the request-that-was-accepted rather than a request that needs to be acted upon

temporal/api/enums/v1/update.proto
mmcshane: Yup, missed these. Will do

temporal/api/update/v1/message.proto
bergundy: I don't regret using single `Payload` for `Failure.encoded_attributes`, the fact that we use single `Payload` in map values already makes encoding one-to-many impractical and as you've argued @cretz there's not really a _need_ for that.

I would definitely support single `Payload` in this use case. Yes, it breaks consistency with other return values but in practice we don't support multiple return values in the SDK and never will for language interoperability.

cretz: Arguably `encoded_attributes` (plural) should have been `map<string, Payload>` to they could have some alongside ours, but 🤷 

As for here, I'm ok with a single payload for return.

bergundy: Not sure `map<string, Payload>` would've better there, user can still have their own attributes with `Payload` and a single `Payload` is easier to work with and is more flexible than a `map`. 
But what's done is done.

I think we all are okay with single or multi here even though we will always only use single so it's up to Matt to make the call IMHO.

bergundy (approved):
I think everyone that commented is okay with what we have now, it's GTM AFAIC.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1051,40 +1055,35 @@ message GetWorkerBuildIdOrderingResponse {
...
+    string first_execution_run_id = 3;
 
-    // A string identifying the agent that requested this interaction.
-    string identity = 6;
+    // Describes when this request should return - basically whether the
+    // update is synchronous, asynchronous, or somewhere in between.
+    temporal.api.update.v1.WaitPolicy wait_policy = 4;
 
     // The name under which the workflow update function is registered and the
     // arguments to pass to said function.
```
alexshtin, line 1075: This is very wrong comment.

temporal/api/workflowservice/v1/request_response.proto
cretz, line 329: I know it can seem late in the process for this, but after SDK team discussion, we believe for responses (so not necessarily accept/reject), it is much clearer to make this a command instead of being expected to maintain the sequence identifier of the event it should be within. The concept of "all ordered happenings in a workflow are sent as commands in their order" is much clearer than "some ordered happenings in a workflow are sent as commands and some sent as messages to be interleaved with commands".

temporal/api/command/v1/message.proto
```diff
@@ -252,9 +236,7 @@ message Command {
         StartChildWorkflowExecutionCommandAttributes start_child_workflow_execution_command_attributes = 12;
         SignalExternalWorkflowExecutionCommandAttributes signal_external_workflow_execution_command_attributes = 13;
         UpsertWorkflowSearchAttributesCommandAttributes upsert_workflow_search_attributes_command_attributes = 14;
-        AcceptWorkflowUpdateCommandAttributes accept_workflow_update_command_attributes = 15;
-        CompleteWorkflowUpdateCommandAttributes complete_workflow_update_command_attributes = 16;
+        // 15 and 16 are available for use - they were used as part of a prototype that never made it into a release
         ModifyWorkflowPropertiesCommandAttributes modify_workflow_properties_command_attributes = 17;
-        RejectWorkflowUpdateCommandAttributes reject_workflow_update_command_attributes = 18;
```
cretz, line 258: Per the previous comment, I think we may need some kind of `ProtocolMessageCommand` or something.

temporal/api/protocol/v1/message.proto
```diff
@@ -22,18 +22,36 @@
...
+//     aip.dev/not-precedent: We want runtime extensibility for the body field --)
+message Message {
+    // An ID for this specific message.
+    string id = 1;
+
+    // Identifies the specific instance of a protocol to which this message
+    // belongs.
+    string protocol_instance_id = 2;
+
+    // The event ID or command ID after which this message can be delivered. The
```
alexshtin, line 46:
```suggestion
    // The event ID or command ID after which this message can be processed. The
```
?

temporal/api/workflowservice/v1/request_response.proto
cretz: To clarify, if it's too late for this, it's too late for this.

temporal/api/protocol/v1/message.proto
mmcshane: I think delivered is the right word here. It's up to the implementation as to whether or not it gets processed.

Makefile
```diff
@@ -29,15 +29,15 @@ $(PROTO_OUT):
 	mkdir $(PROTO_OUT)
 
 ##### Compile proto files for go #####
-grpc: buf-lint api-linter buf-breaking gogo-grpc fix-path
+grpc: buf-lint api-linter gogo-grpc fix-path
```
alexshtin, line 32: Don't forget to bring it back!

mmcshane (approved):

temporal/api/protocol/v1/message.proto
Sushisource: Alternatively, "may be processed"

temporal/api/protocol/v1/message.proto
```diff
@@ -22,18 +22,36 @@
...
+    // effects of history up to and including this event ID should be visible to
+    // the code that handles this message. Omit to opt out of sequencing.
+    oneof sequencing_id {
+        int64 event_id = 3;
+        int64 command_index = 4;
```
Sushisource, line 51: Do we ever need this, now that the pointer will be in the command list? Can this just be an optional `event_id`?

mmcshane: Definitely want to preserve the flexibility to use different things as a sequence number (e.g. could have a dotted version vector or something causal like that) so I want to keep the oneof. Command index may end up being unused and if so we can remove it but at least for update it might be useful to set as part of message sent to point back to the ProtocolMessageCommand so that we can do bidirectional consistency checking.

PR #255, opened 2023-01-19
Add fields for eager workflow dispatch from StartWorkflowExecution

bergundy:


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -168,10 +168,19 @@ message StartWorkflowExecutionRequest {
...
+    // `workflow_task_timeout`.
+    bool request_eager_execution = 17;
 }
 
 message StartWorkflowExecutionResponse {
     string run_id = 1;
+    // When `request_eager_execution` is set on the `StartWorkflowExecutionRequest`, the server - if supported - will
+    // return the first workflow task to be eagerly executed.
+    // The caller is expected to have a worker available to process the task.
+    PollWorkflowTaskQueueResponse eager_workflow_task = 2;
```
cretz, line 183: I assume the eager workflow task for child workflows (or maybe this workflow if the server is fast enough) on wft complete response is a separate project?

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -168,10 +168,19 @@ message StartWorkflowExecutionRequest {
     temporal.api.common.v1.Memo memo = 14;
     temporal.api.common.v1.SearchAttributes search_attributes = 15;
     temporal.api.common.v1.Header header = 16;
+    // Request to get the first workflow task inline in the response bypassing matching service and worker polling.
+    // If set to `true` the caller is expected to have a worker available to process the task.
```
Sushisource, line 172: Just to make sure it is a worker that can actually handle it - has the right workflow types, etc.
```suggestion
    // If set to `true` the caller is expected to have a worker available and capable of processing the task.
```

bergundy: Noice

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -168,10 +168,19 @@ message StartWorkflowExecutionRequest {
     temporal.api.common.v1.Memo memo = 14;
     temporal.api.common.v1.SearchAttributes search_attributes = 15;
     temporal.api.common.v1.Header header = 16;
+    // Request to get the first workflow task inline in the response bypassing matching service and worker polling.
+    // If set to `true` the caller is expected to have a worker available and capable of processing the task.
+    // The returned task will be marked as started and is expected to be completed by the specified
+    // `workflow_task_timeout`.
+    bool request_eager_execution = 17;
```
rodrigozhou, line 175: nit: `request_eager_execution` -> `return_eager_workflow_task`

rodrigozhou (approved):

temporal/api/workflowservice/v1/request_response.proto
bergundy: I chose this name for consistency with requesting eager activities

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -851,6 +860,9 @@ message GetSystemInfoResponse {
 
         // True if server supports upserting workflow memo
         bool upsert_memo = 7;
+
+        // True if server supports eager workflow task dispatching for the StartWorkflowExecution API
+        bool eager_workflow_start = 8;
```
yycptt, line 865: If server doesn't support this feature, the `eager_workflow_task` field will simply be null? Do we still need this feature flag in that case?

Is there any design doc I can read more about this feature and also the eager start for child workflows?

bergundy: Yes, because there's no better way for the client to know if the server support this.
We don't have a way to eagerly start child workflows but I'll provide a design doc for eager workflow start in slack.

cretz: > Yes, because there's no better way for the client to know if the server support this.

I think it's important to spell out why the client needs to know ahead of time. If it's just for our test cases, ok (though I disagree but don't care much).

bergundy: I'd say this is primarily for testing, not just ours, users could leverage this flag as well.
If my app requires this functionality, it's better to fail fast at startup than to start a defunct app.

bergundy:
I'm merging this PR as the server implementation is almost complete and depends on it.

PR #256, opened 2023-01-19
Add AliasSearchAttributes endpoint to operator service

rodrigozhou:
**What changed?**
Adding `AliasSearchAttributes` endpoint to operator service

**Why?**
Adding command to `tctl` to alias custom search attributes.

**Breaking changes**
No.

### DIFF
--- a/temporal/api/operatorservice/v1/request_response.proto
+++ b/temporal/api/operatorservice/v1/request_response.proto
@@ -38,6 +38,7 @@ import "temporal/api/enums/v1/common.proto";
 message AddSearchAttributesRequest {
     // Mapping between search attribute name and its IndexedValueType.
     map<string, temporal.api.enums.v1.IndexedValueType> search_attributes = 1;
+    string namespace = 2;
 }
 
 message AddSearchAttributesResponse {
@@ -46,12 +47,14 @@ message AddSearchAttributesResponse {
 message RemoveSearchAttributesRequest {
     // Search attribute names to delete.
     repeated string search_attributes = 1;
+    string namespace = 2;
 }
 
 message RemoveSearchAttributesResponse {
 }
 
 message ListSearchAttributesRequest {
+    string namespace = 1;
 }
 
 message ListSearchAttributesResponse {
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -62,6 +62,7 @@ message NamespaceConfig {
     // If unspecified (ARCHIVAL_STATE_UNSPECIFIED) then default server configuration is used.
     temporal.api.enums.v1.ArchivalState visibility_archival_state = 5;
     string visibility_archival_uri = 6;
+    map<string, string> custom_search_attributes_alias = 7;
 }
 
 message BadBinaries {


temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -63,6 +63,14 @@ message ListSearchAttributesResponse {
     map<string, string> storage_schema = 3;
 }
 
+message AliasSearchAttributesRequest {
+    string namespace = 1;
+    map<string, string> field_alias_map = 2;
```
cretz, line 68: Can you add docs or clarify:

* What this does to previous aliases
* How to remove aliases (if supported)
* Whether aliases are used as search attribute keys in other parts of the API
* How to list/know what aliases exist
* Purpose of aliases (e.g. only for querying, or is it a way to support complete renaming? and what value is renaming if the old one is still present everywhere?)

rodrigozhou: - If a search attribute already has alias, it will overwrite it (but no change to data).
- Added a new endpoint to remove alias.
- What other parts of API are using search attributes? I can check if anything needs to be updated.
- Added a new field to `ListSearchAttributesResponse` to show existing aliases.
- Besides better for server side implementation, with SQL databases, we have, for example, 10 keyword search attributes per namespace with aliases, instead of global 10 keyword search attributes without aliases.

Added more details in `service.proto`.

cretz: > What other parts of API are using search attributes? I can check if anything needs to be updated.

Every user can see their search attributes in their workflow. Are those going to use the canonical names in the map, the aliased ones, or both?

> Besides better for server side implementation

Sounds like it complicates the API, though I admit to not knowing enough about to server to understand how going from search attributes to search attributes + aliases is better.

> with SQL databases, we have, for example, 10 keyword search attributes per namespace with aliases, instead of global 10 keyword search attributes without aliases.

Can you help me understand a bit better. Why 10 attributes per namespace + aliases is better than 10 global attributes? I am trying to understand who would use this and why.

bergundy: Joining @cretz's questions here.

At minimum, this feature deserves an explanation of what the behavior is and why it is needed.

rodrigozhou: > One alias per attribute? I'd have to understand the use case to see if too limiting. What reason are aliases needed but that two aren't?

> Can you help me understand a bit better. Why 10 attributes per namespace + aliases is better than 10 global attributes?

The general idea is the same as what we have in Temporal Cloud.
For visibility in SQL databases, the custom search attributes columns in DB will be pre-allocated (ie., the user won't be able to use more than the pre-allocated number of search attributes). So, the underlying names are pre-defined in the server. This is the reason to creating aliases.

> Every user can see their search attributes in their workflow. Are those going to use the canonical names in the map, the aliased ones, or both?

For backwards compatibility, both will work. That is, if the custom search attribute is `KeywordField` and the alias is `KeywordAlias`, both will work.

> Can you help me understand a bit better. Why 10 attributes per namespace + aliases is better than 10 global attributes?

rodrigozhou:
Changing approach, not gonna expose alias concept.

PR #258, opened 2023-01-29
Add namespace option to search attributes operations

rodrigozhou:
**What changed?**
Added namespace option to search attributes operations.
Added `custom_search_attribute_alias` map to namespace config to be able to set alias in the namespace.

**Why?**
Advanced visibility with SQL: search attributes will have to be set by namespace.

**Breaking changes**
No.


bergundy (approved):
Just clarifying that per namespace attributes will also work for elasticsearch backed visibility too

rodrigozhou:
For elasticsearch, it will work as it is currently.
We're not pre-allocating custom search attributes, nor exposing the concept of alias, so no point in having search attributes per namespace. Basically, passing namespace option when calling those commands will no-op when using elasticsearch.

temporal/api/namespace/v1/message.proto
```diff
@@ -62,6 +62,7 @@ message NamespaceConfig {
     // If unspecified (ARCHIVAL_STATE_UNSPECIFIED) then default server configuration is used.
     temporal.api.enums.v1.ArchivalState visibility_archival_state = 5;
     string visibility_archival_uri = 6;
+    map<string, string> custom_search_attributes_alias = 7;
```
alexshtin, line 65: Shouldn't it be:
```suggestion
    map<string, string> custom_search_attributes_aliases = 7;
```
?

alexshtin: Add a comment saying that it is field name to alias name map.

alexshtin (approved):

bergundy:
As discussed during standup, we will eventually make ElasticSearch backed search attributes namespace specific to match the experience for SQL based visibility but that will not be a part of the 1.20 server release.

temporal/api/namespace/v1/message.proto
```diff
@@ -62,6 +62,8 @@ message NamespaceConfig {
     // If unspecified (ARCHIVAL_STATE_UNSPECIFIED) then default server configuration is used.
     temporal.api.enums.v1.ArchivalState visibility_archival_state = 5;
     string visibility_archival_uri = 6;
+    // Map from field name to alias.
+    map<string, string> custom_search_attribute_aliases = 7;
```
bergundy, line 66: I know it's a little late but this technically doesn't add any value since we can list the search attributes per namespace.
I feel like it's going to cause confusion for users seeing this field without any context telling them what aliases are.

rodrigozhou: I've changed to list for namespace. When using Elasticsearch, it works as it is currently, but when using SQL DB, it will list the aliases.

bergundy: I understand that but adding the aliases in the `NamespaceConfig` is confusing for anyone that doesn't know what an alias is (which is anyone other than the few people who have context about this stream of work).

PR #259, opened 2023-02-07
Add SDK-specific data to WFT Completions

Sushisource:
**What changed?**
Added an sdk-specific field which server need not inspect to WFT completions and events.

**Why?**
We felt that the generality of the approach in https://github.com/temporalio/proposals/pull/73/files wasn't exactly necessary (see last comment I've left there).

This accomplishes our current specific need in a more wire-efficient and type safe way.

**Breaking changes**
No

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,5 @@
 # Syntax is here:
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-* @temporalio/server @temporalio/sdk
+*                       @temporalio/server @temporalio/sdk
+api/temporal/api/sdk/*  @temporalio/sdk
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -195,6 +195,9 @@ message WorkflowTaskCompletedEventAttributes {
     // ID of the worker who picked up this workflow task, or missing if worker
     // is not using versioning.
     temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    // See `RespondWorkflowTaskCompletedRequest::metadata`. This field contains either all or
+    // some subset of the information from the WFT completion request.
+    map<string, temporal.api.common.v1.Payload> metadata = 6;
 }
 
 message WorkflowTaskTimedOutEventAttributes {
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -44,7 +44,7 @@ import "temporal/api/failure/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
 import "temporal/api/update/v1/message.proto";
 import "temporal/api/workflow/v1/message.proto";
-import "temporal/api/sdk/v1/wft_complete_metadata.proto";
+import "temporal/api/sdk/v1/task_complete_metadata.proto";
 
 // Always the first event in workflow history
 message WorkflowExecutionStartedEventAttributes {
@@ -198,7 +198,7 @@ message WorkflowTaskCompletedEventAttributes {
     temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
     // Data the SDK wishes to record for itself, but server need not interpret, and does not
     // directly impact workflow state.
-    temporal.api.sdk.v1.WFTCompleteMetadata sdk_data = 6;
+    temporal.api.sdk.v1.WorkflowTaskCompletedMetadata sdk_metadata = 6;
 }
 
 message WorkflowTaskTimedOutEventAttributes {
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -44,6 +44,7 @@ import "temporal/api/failure/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
 import "temporal/api/update/v1/message.proto";
 import "temporal/api/workflow/v1/message.proto";
+import "temporal/api/sdk/v1/wft_complete_metadata.proto";
 
 // Always the first event in workflow history
 message WorkflowExecutionStartedEventAttributes {
@@ -195,9 +196,9 @@ message WorkflowTaskCompletedEventAttributes {
     // ID of the worker who picked up this workflow task, or missing if worker
     // is not using versioning.
     temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
-    // See `RespondWorkflowTaskCompletedRequest::metadata`. This field contains either all or
-    // some subset of the information from the WFT completion request.
-    map<string, temporal.api.common.v1.Payload> metadata = 6;
+    // Data the SDK wishes to record for itself, but server need not interpret, and does not
+    // directly impact workflow state.
+    temporal.api.sdk.v1.WFTCompleteMetadata sdk_data = 6;
 }
 
 message WorkflowTaskTimedOutEventAttributes {
--- a/temporal/api/sdk/v1/task_complete_metadata.proto
+++ b/temporal/api/sdk/v1/task_complete_metadata.proto
@@ -1,3 +1,25 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
 syntax = "proto3";
 
 package temporal.api.sdk.v1;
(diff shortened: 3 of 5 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -195,6 +196,9 @@ message WorkflowTaskCompletedEventAttributes {
     // ID of the worker who picked up this workflow task, or missing if worker
     // is not using versioning.
     temporal.api.taskqueue.v1.VersionId worker_versioning_id = 5;
+    // Data the SDK wishes to record for itself, but server need not interpret, and does not
+    // directly impact workflow state.
+    temporal.api.sdk.v1.WFTCompleteMetadata sdk_data = 6;
```
bergundy, line 201: Let's spell this WFT out to be consistent with the rest of the API.
Maybe `WorkflowTaskCompletedMetadata`?

temporal/api/sdk/v1/wft_complete_metadata.proto
```diff
@@ -0,0 +1,46 @@
...
+  // undefined behavior if they replay workflows which used this mechanism.
+  InternalPatches internal_patches = 1;
+}
+
+message InternalPatches {
+  // Patches used by Core. Non-Core-using languages may error if they see anything in this field.
+  //
+  // (-- api-linter: core::0141::forbidden-types=disabled
+  //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+  repeated uint32 core_used_patches = 1;
```
bergundy, line 37: Can we use enums instead so we can display the value of the patches?

temporal/api/sdk/v1/wft_complete_metadata.proto
```diff
@@ -0,0 +1,46 @@
...
+  // * If a code check for a patch does not find the patch in the set of used patches, it must take
+  //   the branch corresponding to the absence of that patch.
+  //
+  // During non-replay execution of new WFTs:
+  // * The SDK is free to use all patches it knows about. It must record any newly-used (IE: not
+  //   previously recorded) patches when completing the WFT.
+  //
+  // SDKs which are too old to even know about this message at all are considered to produce
+  // undefined behavior if they replay workflows which used this mechanism.
+  InternalPatches internal_patches = 1;
```
cretz, line 29:
```suggestion
  repeated SdkFlag sdk_flags = 1;
```

Suggestions:

* Agree with @bergundy below, an enum I think is clearer (and still gives packed semantics)
* No value in an extra message and no value in separating core vs lang
* Afraid of the connotation of the word "patch" and I like the more generic "flag"

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -310,6 +311,9 @@ message RespondWorkflowTaskCompletedRequest {
     temporal.api.taskqueue.v1.VersionId worker_versioning_id = 10;
     // Protocol messages piggybacking on a WFT as a transport
     repeated temporal.api.protocol.v1.Message messages = 11;
+    // Data the SDK wishes to record for itself, but server need not interpret, and does not
+    // directly impact workflow state.
+    temporal.api.sdk.v1.WFTCompleteMetadata sdk_data = 12;
```
cretz, line 316:
```suggestion
    temporal.api.sdk.v1.WFTCompleteMetadata sdk_metadata = 12;
```

Just to match the message name

temporal/api/sdk/v1/wft_complete_metadata.proto
bergundy: > No value in an extra message and no value in separating core vs lang

I was debating that while reviewing yesterday.
Having this distinction potentially prevents tasks unnecessarily hitting lang, but after thinking about it a bit more, I don't think that is important enough to have this special case in the API.

cretz: > Having this distinction potentially prevents tasks unnecessarily hitting lang

Yes, with the PR in core now that just uses a number to know whether core knows about it, that makes sense. I'd be ok with keeping these separate thinking about it more. So maybe:

```suggestion
  repeated SdkLanguageFlag sdk_language_flags = 1;
  repeated SdkCoreFlag sdk_core_flags = 2;
```

Otherwise you're gonna need a core-enum prefix or lang and core will have to communicate on the latest set of enums they can handle together.

bergundy: I don't think the cost is that high and this situation should be rare enough that hitting lang with unsupported tasks shouldn't be that much of an issue.

> Otherwise you're gonna need a core-enum prefix or lang and core will have to communicate on the latest set of enums they can handle together.

I'd rather keep the capabilities dynamic for example if we want to support multiple workflow bundles in the same Core powered worker.

bergundy: But I'm fine having with a separate flag set for core and lang.

temporal/api/sdk/v1/wft_complete_metadata.proto
Sushisource: Then we have to update this every repo every time we add a new SDK internal patch, which is quite obnoxious, and we'd need different enums for every single language. I think the raw ints serve our sdk-only purposes much better here.

temporal/api/sdk/v1/wft_complete_metadata.proto
Sushisource: Combining them seems much, much easier to screw things up to me for virtually no benefit. I'll keep them separate

temporal/api/history/v1/message.proto
Sushisource: Sure, I'll make it long again.

>Will the server, once it knows about this field, be able to store and give back the content of this field correctly, even if it contains subfields that it doesn't know about?

No. This also explains why I want to use raw ints rather than a defined enum for the patches. Otherwise we constantly have to ensure there is an up to date server to make this all work.

temporal/api/sdk/v1/wft_complete_metadata.proto
mjameswh: > Then we have to update this every repo every time we add a new SDK internal patch, which is quite obnoxious, and we'd need different enums for every single language. I think the raw ints serve our sdk-only purposes much better here.

That's also my feeling, especially wrt core based SDKs. Then, that would mean that to add a new internal patch in lang, we would have to 1) add the enum to api repo, 2) update core to get the new api, 3) update lang to get the newest core...

bergundy: > Plus, we'd then need server to have that updated proto, which couples our versioning strongly. Not a good place to be.

That's not true, the enum values should not be lost even if the server does not know about them.

> That's also my feeling, especially wrt core based SDKs. Then, that would mean that to add a new internal patch in lang, we would have to 1) add the enum to api repo, 2) update core to get the new api, 3) update lang to get the newest core...

Yes that would be the price for this rare event when we need to create a new internal patch but the value would be OOTB UI visibility.

Sushisource: In theory it works with enums, but in practice that may not actually be true. If server deserializes this and then reserializes it, there's a nonzero chance that it at some point the enum value is coerced to the default `0` unknown value (though it shouldn't). Especially with gogoproto's history for us, I'm not big on trusting it. We could verify that.

To repeat myself though, UI visibility is a purely theoretical concern which is trivial to overcome, it doesn't strike me as likely we're going to need to see what internal patches were used on a routine basis in the UI. If that's the only reason, I don't think it's good enough.

Quinn-With-Two-Ns: A proxy could also introduce the same risk.

I agree with spencer the benefit of OOTB UI visibility is not a very valuable benefit to justify having to coordinate a bunch of updates across multiple repos. IF we want visibility in the UI it seems simpler to update one repo (the UI repo) with know patches vs update at least 3 (maybe more) repos to get a new version marker. 

We should have as little friction as possible to fix bugs.

temporal/api/history/v1/message.proto
dnr: > > Will the server, once it knows about this field, be able to store and give back the content of this field correctly, even if it contains subfields that it doesn't know about?
> 
> No. This also explains why I want to use raw ints rather than a defined enum for the patches. Otherwise we constantly have to ensure there is an up to date server to make this all work.

Wait a sec... storing and forward unknown fields is totally a design goal and supported use of protocol buffers. All implementations support it. Now, it turns out we're using one of the implementations that allows you to disable it, and we do. But there's a long-term goal to move to a more supported implementation (e.g. vtproto, which doesn't look it can even disable it). I think maybe we should consider turning back on unknown fields in gogo and relying on that, if we have a good use case for it.

temporal/api/sdk/v1/task_complete_metadata.proto
```diff
@@ -0,0 +1,46 @@
...
+
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "TaskCompleteMetadataProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+message WorkflowTaskCompletedMetadata {
+  // Internal patches used by an SDK. SDKs must comply with the following behavior:
```
cretz, line 13: Pedantic, but can we call this "flags" instead of "patches"?

cretz (approved):

Spikhalskiy (approved):

.github/CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 # Syntax is here:
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-* @temporalio/server @temporalio/sdk
+*                       @temporalio/server @temporalio/sdk
+api/temporal/api/sdk/*  @temporalio/sdk
```
bergundy, line 5: If we're not using enums for our patches we shouldn't change the code owners.
There's nothing SDK specific here anymore.

The whole point of this SDK folder was so we can make quick changes to protos that the server doesn't care about.
That only applies to SDK patches not the entirety of WorkflowTaskCompletedMetadata.

Sushisource: It's in an sdk package, the message could be `SDKWorkflowTaskCompletedMetadata` but it's already a bit of a mouthful.

This allows us to add additional fields in the future, including the set of outstanding awaitables which we know we want to do.

temporal/api/sdk/v1/task_complete_metadata.proto
```diff
@@ -0,0 +1,46 @@
...
+  // During non-replay execution of new WFTs:
+  // * The SDK is free to use all patches it knows about. It must record any newly-used (IE: not
+  //   previously recorded) patches when completing the WFT.
+  //
+  // SDKs which are too old to even know about this message at all are considered to produce
+  // undefined behavior if they replay workflows which used this mechanism.
+  InternalPatches internal_patches = 1;
+}
+
+message InternalPatches {
```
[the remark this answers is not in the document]
cretz: I'd support moving the two fields inside of here just up to that top level. Not sure there is value in being beneath this message unless you planned on adding other fields.

.github/CODEOWNERS
mjameswh: The more it goes, and the more I'm wondering if we might be going the wrong direction...

I mean, based on our previous responses (I'm oversimplifying some points a little bit here, but please bear with me):
- We only need this structure to be meaningful for that specific SDK (ie. it can be opaque to the server);
- We want to be able to alter the structure of that message quickly, without asking the server team;
- We want the server to be able to copy this message without risk that it gets mangled;
- If we need other things that doesn't fit the above criteria, they will be defined outside of this message.

What is the advantage then of defining the fields of that message in API? Why not simply define `SDKWorkflowTaskCompletedMetadata` as `bytes`, and let SDK define a private structure to decode this?

Sushisource: That was more or less original form of my proposal here https://github.com/temporalio/proposals/pull/73/files using `Any`, which we changed to `Payload` and then un-did entirely in favor of just having things be well defined.

I was never a fan of `Payload`. `Any` accomplishes most of what you list here just fine, but isn't terribly bytes efficient. Pure `bytes` is efficient, but less safe.

Well defined is both, but has the downside of updating this repo. Changing the owners for this package at least makes that somewhat less painful. I'm not sure why @bergundy wants to undo that now, which we agreed on earlier when we discussed simply going back to well-defined protos.

Honestly I think we're making a bit of a mountain out of a molehill here. This well-defined message works, the sdk sub-package makes sense, let's just move on with life.

bergundy: > This allows us to add additional fields and/or messages in the future, including the set of outstanding awaitables which we know we want to do.

Server and UI will also want to know about these.

>  I'm not sure why @bergundy wants to undo that now, which we agreed on earlier when we discussed simply going back to well-defined protos.

I thought the whole point of the SDK owned proto definitions are for the enum values, if we're not adding those, I don't think we have any other SDK owned definitions and therefore don't see justification for adding this CODEOWNERS exception and the separate SDK package.

Sushisource: We may add other metadata fields. That's at least one semi-plausible reason to do it, I'm still not sure, given the current state, what the strong reason to avoid it is.

>Server and UI will also want to know about these. 
Not necessarily? UI might render them, but I don't expect any meaningful interpretation. If we do, then great, they go outside of this.

temporal/api/sdk/v1/task_complete_metadata.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
```
dnr, line 1: I think these are supposed to have a copyright/license at the top?

temporal/api/sdk/v1/task_complete_metadata.proto
dnr: I think I like flags over patches also

temporal/api/sdk/v1/task_complete_metadata.proto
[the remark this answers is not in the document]
dnr: Also agree with Chad here

dnr (approved):

temporal/api/sdk/v1/task_complete_metadata.proto
Sushisource: Ah, good point thanks

PR #260, opened 2023-02-09
Move resend workflow API to operator

yux0:
**What changed?**
Move resend workflow API to operator

**Why?**
Move resend workflow API to operator

**Breaking changes**


yycptt (approved):

cretz:
Can we get a bit of background on why? (not that it matters for approval, but it's just a "what" and not a "why" so outsiders can't gather any context)

yux0:
> Can we get a bit of background on why? (not that it matters for approval, but this PR is just a "what" and not a "why" so outsiders can't gather any context)

This is less powerful than the current one. I will close this one and see if we still need this api.

PR #261, opened 2023-02-09
Add field for nonfirst LA attempts count

Sushisource:
**What changed?**
Pursuant to our discussions and conclusions around how we will charge for Local Activities in cloud, I've added the field that contains the information we need to perform those charges accurately.

The field is *not* added to the completed event, because metering does not need it to be there to work properly, and it has no value from a replay perspective.

For discussion:
* Should we embed this field inside a new message? That was my initial inclination, for future proofing purposes, but it's not clear exactly what that name would be. `MeteringMetadata`? Does it seem likely we'll need more info at some point? Arguably not, since we don't even really want to use this as a long term solution anyway.

**Why?**
Needed for accurate metring

**Breaking changes**
No

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -110,3 +110,14 @@ message RetryPolicy {
     // this is not a substring match, the error *type* (not message) must match exactly.
     repeated string non_retryable_error_types = 5;
 }
+
+// Metadata relevant for metering purposes
+message MeteringMetadata {
+    // Count of local activities which have begun an execution attempt during this workflow task,
+    // and whose first attempt occurred in some previous task. This is used for metering
+    // purposes, and does not affect workflow state.
+    //
+    // (-- api-linter: core::0141::forbidden-types=disabled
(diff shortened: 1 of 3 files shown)


Sushisource:
Should we use a `MeteringMetadata` sub-message? There's a bit more reason to now, that it's on the event as well.

:+1: emoji react for yes, :-1: for no

Spikhalskiy:
I'm really in favor of `MeteringMetadata` option, especially if we add this field not just to the command, but also to the event.
I expect more fields with metering-only semantics to be added as time pass. I think putting them right next to fields that are critical for an actual execution and replay mechanics and understanding what happened to the workflow in the history looks very unpolished from the API design perspective. 
There is nothing technically wrong with it putting them on the top level, protobuf does allow us to remove them later if needed, it's mostly a stylistic and "How the public API looks" point of view.

bergundy (approved):

yiminc (approved):

mastermanu (approved):

PR #262, opened 2023-02-10
Add continued_failure and last_completion_result to StartWorkflowExecution

dnr:
**What changed?**
Allow these two fields to be set on new workflows at the frontend api level.

**Why?**
For the schedule feature, we need to be able to set these, but also want the rpc to go through the frontend. They don't have to be exposed anywhere else.

**Breaking changes**
No

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -174,7 +174,10 @@ message StartWorkflowExecutionRequest {
     // The returned task will be marked as started and is expected to be completed by the specified
     // `workflow_task_timeout`.
     bool request_eager_execution = 17;
-    // These two fields are intended for internal use only.
+    // These values will be available as ContinuedFailure and LastCompletionResult in the
+    // WorkflowExecutionStarted event and through SDKs. The are currently only used by the
+    // server itself (for the schedules feature) and are not intended to be exposed in
+    // StartWorkflowExecution.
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
>For the schedule feature, we need to be able to set these, but also want the rpc to go through the frontend. They don't have to be exposed anywhere else.
 I am probably missing some context, but if they are needed how does the current schedules feature work?

dnr:
> > For the schedule feature, we need to be able to set these, but also want the rpc to go through the frontend. They don't have to be exposed anywhere else.
>  I am probably missing some context, but if they are needed how does the current schedules feature work?

It works because it goes direct to history, but we want to change it to go through frontend, so we need to be able to set them in the frontend request

bergundy (approved):
The fields could use a description despite being meant for internal use, otherwise lgtm.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -174,6 +174,9 @@ message StartWorkflowExecutionRequest {
     // The returned task will be marked as started and is expected to be completed by the specified
     // `workflow_task_timeout`.
     bool request_eager_execution = 17;
+    // These two fields are intended for internal use only.
```
yiminc, line 177: Still worth to put some comments here.

yiminc (approved):

PR #264, opened 2023-02-22
Add workflow_start_delay

pdoerner:
**What changed?**
Added ability to specify a delay when starting a workflow. Cannot be used with workflows that have a `cron_schedule`

**Why?**
https://github.com/temporalio/temporal/issues/3378

**Breaking changes**
None

### DIFF
--- a/temporal/api/enums/v1/workflow.proto
+++ b/temporal/api/enums/v1/workflow.proto
@@ -67,6 +67,8 @@ enum ContinueAsNewInitiator {
     CONTINUE_AS_NEW_INITIATOR_RETRY = 2;
     // The workflow continued as new because cron has triggered a new execution
     CONTINUE_AS_NEW_INITIATOR_CRON_SCHEDULE = 3;
+    // The workflow continued as new because the start was delayed
+    CONTINUE_AS_NEW_INITIATOR_DELAY_START = 4;
 }
 
 // (-- api-linter: core::0216::synonyms=disabled
--- a/temporal/api/enums/v1/workflow.proto
+++ b/temporal/api/enums/v1/workflow.proto
@@ -67,8 +67,6 @@ enum ContinueAsNewInitiator {
     CONTINUE_AS_NEW_INITIATOR_RETRY = 2;
     // The workflow continued as new because cron has triggered a new execution
     CONTINUE_AS_NEW_INITIATOR_CRON_SCHEDULE = 3;
-    // The workflow continued as new because the start was delayed
-    CONTINUE_AS_NEW_INITIATOR_DELAY_START = 4;
 }
 
 // (-- api-linter: core::0216::synonyms=disabled
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -180,6 +180,8 @@ message StartWorkflowExecutionRequest {
     // StartWorkflowExecution.
     temporal.api.failure.v1.Failure continued_failure = 18;
     temporal.api.common.v1.Payloads last_completion_result = 19;
+    // Time to wait before starting the first workflow run. Cannot be used with `cron_schedule`.
+    google.protobuf.Duration workflow_start_delay = 20 [(gogoproto.stdduration) = true];
 }
 
 message StartWorkflowExecutionResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -631,8 +631,8 @@ message SignalWithStartWorkflowExecutionRequest {
     // Time to wait before dispatching the first workflow task. Cannot be used with `cron_schedule`.
     // Note that the signal will be delivered with the first workflow task. If the workflow gets
     // another SignalWithStartWorkflow before the delay, a workflow task will be dispatched immediately
-    // and the rest of the delay period will be ignored. Signal via SignalWorkflowExecution will
-    // not unblock the workflow.
+    // and the rest of the delay period will be ignored, even if that request also had a delay. Signal
+    // via SignalWorkflowExecution will not unblock the workflow.
     google.protobuf.Duration workflow_start_delay = 20 [(gogoproto.stdduration) = true];
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -630,7 +630,9 @@ message SignalWithStartWorkflowExecutionRequest {
     temporal.api.common.v1.Header header = 19;
     // Time to wait before dispatching the first workflow task. Cannot be used with `cron_schedule`.
     // Note that the signal will be delivered with the first workflow task. If the workflow gets
-    // another signal before the delay, a workflow task will be dispatched immediately.
+    // another SignalWithStartWorkflow before the delay, a workflow task will be dispatched immediately
+    // and the rest of the delay period will be ignored. Signal via SignalWorkflowExecution will
+    // not unblock the workflow.
     google.protobuf.Duration workflow_start_delay = 20 [(gogoproto.stdduration) = true];
 }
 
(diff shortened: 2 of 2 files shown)


cretz:
Hrmm, this is a user request that predates schedules right? And even if they didn't want to use schedules, IMO it doesn't seem like our public API and then every SDK needs to be updated for this feature just to save that user a sleep step.

yycptt:
This [request](https://github.com/temporalio/temporal/issues/3378) actually comes from @tusharroy25. 
@tusharroy25 Can you elaborate a bit more on the use case? 

Re. Schedule. I am not sure if schedule support this or not, but using Schedule for one-time delayed workflow execution seems weird.

cretz:
> but using Schedule for one-time delayed workflow execution seems weird.

My fear is that that continual one-offs for how to schedule workflows in the future will pile up outside of the actual scheduling. It's kinda silly we can't use the schedule-future-workflow feature to solve our schedule future workflow use case.

Maybe we need to support a fixed time on our schedule trigger? Or maybe we need to accept a set of fixed times in our schedule spec? For now, technically this can be solved with a calendar spec of the exact time you want schedule and just set the action limit to 1. If this is too unwieldy we should consider making it easier.

dnr:
Fwiw, I think using schedules for this is a good idea in theory. We can easily add a list of fixed times option to the spec, and an option to automatically exit if there are no more times.

The main caveat is that it's quite a bit more expensive than this method, since we deliberately traded efficiency for flexibility with the initial schedules implementation. If we reimplement it with lower-level primitives then that objection would go away (that would be something like 1+ years out if it happens)

This has nothing to do with phase, that's for adjusting the phase of a regular interval (from the epoch), this is for one-offs (from call time).

As far as NewWorkflowExecutionInfo: I had a proposal a while ago to consolidate the various proto messages there ([internal link](https://www.notion.so/98b07daf0f864cd49c113747d5d42369)). I think it does belong there, but it's maybe silly since the only user of that message for now (schedules) would probably force it to be zero, since it doesn't make sense to combine with schedules.

cretz:
I think what's here is fine, I am just providing a counter points. Other counter points:

* Why can I not just pause any workflow for an amount of time instead of just at start? (this would also allow delaying of signal and other potentially useful features)
* Why can I not delay start of a child workflow or activity the same way? If it's because you could just sleep, why is that answer good enough there but not here?
* Is a timestamp to start clearer than a delay?

I have dozens of more questions. I think there are potentially simpler more flexible options here, but I understand this has already been discussed and decided.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -624,6 +626,8 @@ message SignalWithStartWorkflowExecutionRequest {
     temporal.api.common.v1.Memo memo = 17;
     temporal.api.common.v1.SearchAttributes search_attributes = 18;
     temporal.api.common.v1.Header header = 19;
+    // Time to wait before starting the first workflow run.
+    google.protobuf.Duration workflow_start_delay = 20 [(gogoproto.stdduration) = true];
```
dnr, line 630: This one doesn't actually make much sense to me. The point is to start a workflow and immediately send it a signal, which usually means that the caller wants it to handle the signal as soon as possible. We already have a bunch of inconsistent/buggy behavior around delayed workflow start + signals, so I'd rather just leave this out and make this impossible to specify.

dnr: Talked internally, this is fine, but please change the comment to include more details

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -180,6 +180,8 @@ message StartWorkflowExecutionRequest {
     // StartWorkflowExecution.
     temporal.api.failure.v1.Failure continued_failure = 18;
     temporal.api.common.v1.Payloads last_completion_result = 19;
+    // Time to wait before starting the first workflow run. Cannot be used with `cron_schedule`.
```
dnr, line 183:
```suggestion
    // Time to wait before dispatching the first workflow task. Cannot be used with `cron_schedule`.
    // If the workflow gets a signal before the delay, a workflow task will be dispatched and the rest
    // of the delay will be ignored.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -624,6 +626,8 @@ message SignalWithStartWorkflowExecutionRequest {
     temporal.api.common.v1.Memo memo = 17;
     temporal.api.common.v1.SearchAttributes search_attributes = 18;
     temporal.api.common.v1.Header header = 19;
+    // Time to wait before starting the first workflow run.
```
dnr, line 629:
```suggestion
    // Time to wait before dispatching the first workflow task. Note that the
    // signal will be delivered with the first workflow task. If the workflow gets
    // another signal before the delay, a workflow task will be dispatched
    // immediately and the rest of the delay will be ignored.
```

yycptt (approved):

temporal/api/workflowservice/v1/request_response.proto
yycptt: gets another signal, via SignalWorkflow, before the delay, a workflow task will be dispatched ...

SignalWithStart won't unblock the workflow.

temporal/api/enums/v1/workflow.proto
```diff
@@ -67,6 +67,8 @@ enum ContinueAsNewInitiator {
     CONTINUE_AS_NEW_INITIATOR_RETRY = 2;
     // The workflow continued as new because cron has triggered a new execution
     CONTINUE_AS_NEW_INITIATOR_CRON_SCHEDULE = 3;
+    // The workflow continued as new because the start was delayed
+    CONTINUE_AS_NEW_INITIATOR_DELAY_START = 4;
```
bergundy, line 71: How is this related to continue as new?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -624,6 +628,12 @@ message SignalWithStartWorkflowExecutionRequest {
     temporal.api.common.v1.Memo memo = 17;
     temporal.api.common.v1.SearchAttributes search_attributes = 18;
     temporal.api.common.v1.Header header = 19;
+    // Time to wait before dispatching the first workflow task. Cannot be used with `cron_schedule`.
+    // Note that the signal will be delivered with the first workflow task. If the workflow gets
+    // another SignalWithStartWorkflow before the delay, a workflow task will be dispatched immediately
```
bergundy, line 633: What if that second `SignalWithStart` also has a start delay?

temporal/api/enums/v1/workflow.proto
pdoerner: We use this on the server side to map to reason for why a workflow was delayed. So this change is to keep the same logic for the delay, but it doesn't really relate to continue as new.

temporal/api/workflowservice/v1/request_response.proto
pdoerner: That delay will not be respected, because a new workflow is not started.

bergundy: Worth noting that the start delay will be ignored in this case.

pdoerner: Updated

temporal/api/enums/v1/workflow.proto
yycptt: actually, is it possible to hide this from public api and only define it in only server repo and make the code work? 

Looks like we are defining it here just to workaround some impl detail in server.

pdoerner: There are probably ways we could work around it in server, but then my question is why do we have `CONTINUE_AS_NEW_INITIATOR_CRON_SCHEDULE` in the public api? AFAICT this value is only ever set by the server, not by users. So I guess it comes down to whether we want to keep the logic consistent or keep the API as clean as possible.

pdoerner: Makes sense. I removed it and will handle on the server side.

yiminc (approved):

PR #266, opened 2023-03-15
Signal without wft

pdoerner:
**What changed?**
Added a flag to `SignalWorkflowExecutionRequest` and `SignalWithStartWorkflowExecutionRequest` to indicate that a new workflow task should not be generated when that signal is received.

**Why?**
Resolves https://github.com/temporalio/temporal/issues/3406

**Breaking changes**
No

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -180,6 +180,8 @@ message StartWorkflowExecutionRequest {
     // StartWorkflowExecution.
     temporal.api.failure.v1.Failure continued_failure = 18;
     temporal.api.common.v1.Payloads last_completion_result = 19;
+    // Time to wait before starting the first workflow run. Cannot be used with `cron_schedule`.
+    google.protobuf.Duration workflow_start_delay = 20 [(gogoproto.stdduration) = true];
 }
 
 message StartWorkflowExecutionResponse {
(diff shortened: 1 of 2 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -592,6 +592,8 @@ message SignalWorkflowExecutionRequest {
     // Headers that are passed with the signal to the processing workflow.
     // These can include things like auth or tracing tokens.
     temporal.api.common.v1.Header header = 8;
+    // Indicates that a new workflow task should not be generated when this signal is received.
+    bool skip_generate_workflow_task = 9;
```
cretz, line 596: I would like to add a similar comment I did on delay and will likely also add on pause when it comes:

People want to control task generation in a more general way. We should consider general solutions.

(having said that, I have no opinion on this PR specifically, LGTM)

pdoerner: Yeah, I think we agree that the workflow pause feature will be the general solution to both these problems. The goal of this change and the delay change is to support a couple specific use cases in the short-term and we don't want to wait until pause is ready.

MichaelSnowden (commented):
This change introduces a new [FlagArgument](https://martinfowler.com/bliki/FlagArgument.html) to our API, and it looks very awkward here. I agree with Chad that we should consider a more general solution here

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -635,6 +637,8 @@ message SignalWithStartWorkflowExecutionRequest {
     // and the rest of the delay period will be ignored, even if that request also had a delay. Signal
```
yycptt, line 637: probably update this comment because if `skip_generate_workflow_task` is set, the rest of the delay still counts.

yycptt (approved):

alexshtin (approved):

bergundy (approved):

PR #267, opened 2023-03-17
Add PollWorkflowExecutionUpdate rpc

mmcshane:
With the associated request and respose message types.

**What changed?**
Add PollWorkflowExecutionUpdate rpc with the associated request and respose message types.

**Why?**
In support of asynchronous update (asynchronous from the remote caller's perspective)

**Breaking changes**
No

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1277,12 +1277,15 @@ message PollWorkflowExecutionUpdateRequest {
     temporal.api.update.v1.UpdateRef update_ref = 2;
     // The identity of the worker/client who is polling this update outcome
     string identity = 3;
+    // Describes when this poll request should return a response
+    temporal.api.update.v1.WaitPolicy wait_policy = 4;
 }
 
 message PollWorkflowExecutionUpdateResponse {
-    // The outcome of the update if and only if the workflow execution update
-    // has completed. If this response is being returned before the update has
-    // completed (e.g. due to a gRPC timeout on the PollWorkflowExecutionUpdate
-    // RPC) then this field will not be set.
+    // The outcome of the update if and only if the update has completed. If
+    // this response is being returned before the update has completed (e.g. due
+    // to the specification of a wait policy that only waits on
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED) then this field will
+    // not be set.
     temporal.api.update.v1.Outcome outcome = 1;
 }
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -400,13 +400,9 @@ service WorkflowService {
     }
 
     // Polls a workflow execution for the outcome of a workflow execution update
-    // previously issued through the UpdateWorkflowExecution RPC. To remain
-    // consistent with other long-poll calls, this call will return a response
-    // with an empty (i.e. nil/null/None/etc.) Outcome field just prior to the
-    // earlier of the client-supplied gRPC timeout or the server's default
-    // long-poll timeout if the update has not completed by that time. In such
-    // cases it is expected that clients will continue to make calls to this RPC
-    // until the outcome is returned.
+    // previously issued through the UpdateWorkflowExecution RPC. The effective
+    // timeout on this call will be shorter of the the caller-supplied gRPC
+    // timeout and the server's configured long-poll timeout.
     // (-- api-linter: core::0134=disabled
     //     aip.dev/not-precedent: UpdateWorkflowExecution doesn't follow Google API format --)
     rpc PollWorkflowExecutionUpdate(PollWorkflowExecutionUpdateRequest) returns (PollWorkflowExecutionUpdateResponse){
--- a/temporal/api/update/v1/message.proto
+++ b/temporal/api/update/v1/message.proto
@@ -35,8 +35,8 @@ import "temporal/api/common/v1/message.proto";
 import "temporal/api/enums/v1/update.proto";
 import "temporal/api/failure/v1/message.proto";
 
-// Sepcifies to the gRPC server how the client wants the UpdateWorkflowExecution
-// call to wait before returning control to the caller.
+// Specifies to the gRPC server how long the client wants the an update-related
+// RPC call to wait before returning control to the caller.
 message WaitPolicy {
 
     // Indicates the update lifecycle stage that the gRPC call should wait for


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1257,3 +1257,21 @@ message ListBatchOperationsResponse {
...
 }
+
+message PollWorkflowExecutionUpdateRequest {
+    // The namespace of the workflow execution to which the update was
+    // originally issued.
+    string namespace = 1;
+    // The update reference returned in the initial
+    // UpdateWorkflowExecutionResponse
+    temporal.api.update.v1.UpdateRef update_ref = 2;
+    // The identity of the worker/client who is polling this task queue
```
Sushisource, line 1268: Is it really polling a "task queue" or should this just read "polling for the update outcome" or something?

mmcshane: oh god. copy paste shame on me.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1257,3 +1257,21 @@ message ListBatchOperationsResponse {
...
+    // UpdateWorkflowExecutionResponse
+    temporal.api.update.v1.UpdateRef update_ref = 2;
+    // The identity of the worker/client who is polling this update outcome
+    string identity = 3;
+}
+
+message PollWorkflowExecutionUpdateResponse {
+    // The outcome of the update if and only if the workflow execution update
+    // has completed. If this response is being returned before the update has
+    // completed then this field will not be set.
```
bergundy, line 1275: I’d mention that the SDK is expected to issue a new poll request in this case.

Also mention how timeout works, is it tied to grpc deadline? Is there a default?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1257,3 +1257,21 @@ message ListBatchOperationsResponse {
...
+
+message PollWorkflowExecutionUpdateRequest {
+    // The namespace of the workflow execution to which the update was
+    // originally issued.
+    string namespace = 1;
+    // The update reference returned in the initial
+    // UpdateWorkflowExecutionResponse
+    temporal.api.update.v1.UpdateRef update_ref = 2;
+    // The identity of the worker/client who is polling this update outcome
+    string identity = 3;
```
bergundy, line 1269: We might want to allow polling on accepted status at some point.
Worth considering here.

cretz: Agreed. As a client (that may be separate from where update sent) I may just want to wait until the update is accepted by the workflow post validation.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1257,3 +1257,21 @@ message ListBatchOperationsResponse {
     repeated temporal.api.batch.v1.BatchOperationInfo operation_info = 1;
     bytes next_page_token = 2;
 }
+
+message PollWorkflowExecutionUpdateRequest {
+    // The namespace of the workflow execution to which the update was
+    // originally issued.
+    string namespace = 1;
+    // The update reference returned in the initial
+    // UpdateWorkflowExecutionResponse
```
cretz, line 1266: Can you clarify/confirm that this can be hand-built with no run ID in the workflow execution and it will still properly work like other run-ID-less calls on client?

temporal/api/workflowservice/v1/request_response.proto
cretz, line 1275: In what situation is the response be returned before update completed?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -399,6 +399,13 @@ service WorkflowService {
     rpc UpdateWorkflowExecution(UpdateWorkflowExecutionRequest) returns (UpdateWorkflowExecutionResponse) {
     }
 
+    // Polls a workflow execution for the outcome of a workflow execution update
+    // previously issued through the UpdateWorkflowExecuiton rpc.
```
cretz, line 403:
```suggestion
    // previously issued through the UpdateWorkflowExecution rpc.
```

temporal/api/workflowservice/v1/request_response.proto
mmcshane: No, the run ID is required here. The server doesn't really anything to scan multiple WF execution IDs to find the update.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: I meant to say that reissuing the request in lieu of a Poll call is confusing. I think the Poll call is at least consistent with the rest of our API and is roughly what a user (developer) would expect. 

I tend not to like raw history APIs - while I acknowledge that the cat is out of the bag, I don't think we want to encourage further coupling at that layer as it really limits our future flexibility in defining/evolving a low-level construct (history events).

temporal/api/workflowservice/v1/request_response.proto
cretz: I think this may be an issue. I am not aware of any workflow interaction that requires run ID. We are ok without run ID to fetch history right? Can we reuse that same approach?

temporal/api/workflowservice/v1/request_response.proto
bergundy: > We have the lifecycle stage enum or the WaitPolicy that we could add to this message. Technically this Poll call isn't necessary - issuing the the update request again with the same update ID (but potentially a different WaitPolicy) is a somewhat confusing of doing the same thing.

I think it makes sense that a user would want to wait for admitted in the update request and then poll on accepted to update some UI.
I think the poll WaitPolicy makes sense since the application might not have the inputs for the update request at that point and a user would likely rather avoid resubmitting the update request.

But seems like we're all in agreement on this point.

> I tend not to like raw history APIs

We have a chance to do better here, I like the API structure proposed in this PR.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: I suppose we _could_ do that same thing as history and default to the most recent RunID but in the event that that ID is wrong (i.e. the update ran in a different run ID) then the NOT_FOUND error response is going to be pretty confusing. It seems somewhat reasonable to assume that the client has a fully filled-out UpdateRef object because it's the exact type that is returned from UpdateWorkflowExecution and in that situation the RunID is always populated.

cretz: > then the NOT_FOUND error response is going to be pretty confusing.

I think this is acceptable. We're trying to not require the run ID and many people never use it in code anywhere (by default workflows don't retry or get a new run ID anyways).

> It seems somewhat reasonable to assume that the client has a fully filled-out UpdateRef object because it's the exact type that is returned from UpdateWorkflowExecution and in that situation the RunID is always populated.

This assumes a specific approach to fetching update results. But it is totally reasonable to use update as, say, pubsub between two unrelated processes. Or use update as a broadcast mechanism to many waiters. E.g. imagine an update call to a workflow that is "wait for event" and then 20 clients call that update and are now hanging. Then a separate client sends an update or signal or whatever, and all of those updates are then resolved inside the workflow therefore broadcasting to their clients.

These are very powerful use cases for update IMO.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: I think we agree on WaitPolicy as well. But since we won't have waiting on admitted for this first round of async development I'm going to omit the WaitPolicy here for now as there is not yet any benefit.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: > and then 20 clients call that update ...

Did you mean "poll that update" here? And if so, how did those clients get the UpdateID and WorkflowID to poll?

temporal/api/workflowservice/v1/request_response.proto
bergundy: Only thing I'm concerned with is that the API will be inconsistent with the update request where AFAICT `WaitPolicy` is required. Once we add WaitPolicy we'll have to treat UNSPECIFIED as COMPLETED.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: Chad and I discussed externally and while we both prefer the more direct thing of returning a DEADLINE_EXCEEDED status error here, remaining consistent with existing Poll APIs and server deployments is also a factor. Will consult with server team on our options re: returning an error directly.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: My intent here was that you could only poll for an update if you had the UpdateRef or could construct one. As we discussed, RunID would be required until we had the right execution scopes on the server side to be able to "find" updates given an execution ID but not the run ID. The UpdateRef would contain everything needed to route the request to a running update. A token could have served the same purpose but the point was always that the issuer of the update could distribute the UpdateRef to other process that may want to poll, similar to how we use tokens in other contexts. You could also think of it as a capability.

I think that for this release we can get by with everything being required since it's clear where all the data comes from (the response to the initial request). We can always make make run ID optional down the road and add first_execution_run_id if we have a way to scan workflow execution runs.

cretz: > until we had the right execution scopes on the server side to be able to "find" updates given an execution ID

I think most recent run ID is acceptable from a user POV like it is with every other workflow interaction

bergundy: > I think defaulting to most recent run ID is acceptable from a user POV like it is with every other workflow interaction

Agree this would be the expected default and is inline with the other workflow APIs.

temporal/api/workflowservice/v1/request_response.proto
mmcshane: That seems like a fair point. I'll add WaitPolicy here

mmcshane:
Updated with feedback from server team that it _is_ acceptable to return a timeout error here so no more docs about returning "empty" on timeout. However with the addition of WaitPolicy to the request it becomes possible to Poll on admitted|accepted, in which case Outcome would be nil/null/None in the same way that it would be for direct UpdateWorkflowExecution calls that return before the result is known.

cretz (approved):
LGTM!

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1257,3 +1257,25 @@ message ListBatchOperationsResponse {
...
+    temporal.api.update.v1.WaitPolicy wait_policy = 4;
+}
+
+message PollWorkflowExecutionUpdateResponse {
+    // The outcome of the update if and only if the update has completed. If
+    // this response is being returned before the update has completed (e.g. due
+    // to the specification of a wait policy that only waits on
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED) then this field will
+    // not be set.
+    temporal.api.update.v1.Outcome outcome = 1;
```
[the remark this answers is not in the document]
cretz: Can you assume that, since the caller has to provide the state to wait on, that it's at least that state when this returns without an error? I think if we want to have something that returns update status, we may need to consider a "describe update" for this use case and other reasons.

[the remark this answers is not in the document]
mmcshane: That was my thinking - if you're getting this response (as opposed to an error) it's because you issued a request with at particular lifecycle state specified.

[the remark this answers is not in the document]
yiminc: 👍

yiminc (approved):

PR #272, opened 2023-04-14
Worker versioning related updates

bergundy:
Decided not to expose (the yet unused) `CompatibleVersionSet.version_set_id`, the set will have multiple IDs internally.

The affected projects have been updated:
- [x] [Server PR](https://github.com/temporalio/temporal/pull/4170)
- [x] [Go SDK PR](https://github.com/temporalio/sdk-go/pull/1089)

Also added a `use_versioning` flag to `RespondWorkflowTaskCompletedRequest` to differentiate between the version stamp being used for versioning (matching) purposes or just as a marker.

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -129,10 +129,6 @@ message WorkerVersionStamp {
     // Set if the worker used a dynamically loadable bundle to process
     // the task. The bundle could be a WASM blob, JS bundle, etc.
     string bundle_id = 2;
-
-    // If set, the worker is opting in to worker versioning. Otherwise, the version stamp is used as a marker for
-    // workflow reset points and the BuildId search attibute.
-    bool uses_versioning = 3;
 }
 
 // Identifies the version(s) that a worker is compatible with when polling or identifying itself
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -129,6 +129,10 @@ message WorkerVersionStamp {
     // Set if the worker used a dynamically loadable bundle to process
     // the task. The bundle could be a WASM blob, JS bundle, etc.
     string bundle_id = 2;
+
+    // If set, the worker is opting in to worker versioning. Otherwise, the version stamp is used as a marker for
+    // workflow reset points and the BuildId search attibute.
+    bool uses_versioning = 3;
 }
 
 // Identifies the version(s) that a worker is compatible with when polling or identifying itself
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -88,11 +88,10 @@ message StickyExecutionAttributes {
 }
 
 // Used by the worker versioning APIs, represents an ordering of one or more versions which are
-// considered to be compatible with each other. Currently the versions are always worker build ids.
+// considered to be compatible with each other. Currently the versions are always worker build IDs.
 message CompatibleVersionSet {
-    // A unique identifier for this version set. Users don't need to understand or care about this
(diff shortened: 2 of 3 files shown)


Sushisource (approved):

temporal/api/taskqueue/v1/message.proto
```diff
@@ -88,11 +88,10 @@ message StickyExecutionAttributes {
...
 message CompatibleVersionSet {
-    // A unique identifier for this version set. Users don't need to understand or care about this
-    // value, but it has value for debugging purposes.
-    string version_set_id = 1;
-    // All the compatible versions, ordered from oldest to newest
-    repeated string build_ids = 2;
+    // All the compatible versions, in insertion order.
+    // The last element is considered the set "default" and may not be the last inserted item in case of promoting an
+    // older build ID to be the set default.
+    repeated string build_ids = 1;
```
rodrigozhou, line 96: Is it safe to change the id from 2 to 1?

bergundy: This API isn't in use yet

dnr: Not in general, but this is not in use anywhere yet

temporal/api/taskqueue/v1/message.proto
```diff
@@ -88,11 +88,10 @@ message StickyExecutionAttributes {
...
+// considered to be compatible with each other. Currently the versions are always worker build IDs.
 message CompatibleVersionSet {
-    // A unique identifier for this version set. Users don't need to understand or care about this
-    // value, but it has value for debugging purposes.
-    string version_set_id = 1;
-    // All the compatible versions, ordered from oldest to newest
-    repeated string build_ids = 2;
+    // All the compatible versions, in insertion order.
+    // The last element is considered the set "default" and may not be the last inserted item in case of promoting an
+    // older build ID to be the set default.
```
dnr, line 95: this is contradictory.. it can't be both "in insertion order" and "the last element ... may not be the last inserted item"

bergundy: This is what we decided though..
If we have keep the timestamps internally, we can preserve the insertion order but we also wanted the default to be the last.

bergundy: What are you suggesting?

dnr: Maybe just remove the first line? They're in arbitrary order except that the last one is special

bergundy: It's nice that we can preserve the insertion order though, why take that away?
I'm almost tempted to add timestamps (wall clock) to each build ID to give users some extra bit of useful information.
Another nice bit of information would be identity and reason for each build ID but I'm hesitant to open this can of worms.

temporal/api/common/v1/message.proto
```diff
@@ -129,6 +129,10 @@ message WorkerVersionStamp {
     // Set if the worker used a dynamically loadable bundle to process
     // the task. The bundle could be a WASM blob, JS bundle, etc.
     string bundle_id = 2;
+
+    // If set, the worker is opting in to worker versioning. Otherwise, the version stamp is used as a marker for
+    // workflow reset points and the BuildId search attibute.
+    bool uses_versioning = 3;
```
dnr, line 135: hmm, I would think this flag should not be part of the stamp itself, but part of the requests that include a stamp

bergundy: Yeah, good point.

bergundy:
Changed the base branch to `worker-versioning` and merged.
Let's keep it all in the feature branch for now.

bergundy:
Comments still welcome of course.

PR #273, opened 2023-04-19
Add flags for using the default build id set

Sushisource:
**What changed?**
Add flags to activity/child/c-a-n to explicitly use the default version set rather than the compatible set when sticking to the same queue.

**Why?**
Flexibility with the versioning feature

**Breaking changes**

### DIFF
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -88,7 +88,7 @@ message ScheduleActivityTaskCommandAttributes {
     // is the default behavior) and instead will use the current overall default for the queue.
     // If this command's `task_queue` field differs from the executing workflow's task queue, then
     // this flag has no effect.
-    bool use_default_build_id_set = 13;
+    bool use_latest_build_id = 13;
 }
 
 message RequestCancelActivityTaskCommandAttributes {
@@ -202,7 +202,7 @@ message ContinueAsNewWorkflowExecutionCommandAttributes {
     // is the default behavior) and instead will use the current overall default for the queue.
     // If this command's `task_queue` field differs from the executing workflow's task queue, then
     // this flag has no effect.
-    bool use_default_build_id_set = 15;
+    bool use_latest_build_id = 15;
 
     // `workflow_execution_timeout` is omitted as it shouldn't be overridden from within a workflow.
 }
@@ -235,7 +235,7 @@ message StartChildWorkflowExecutionCommandAttributes {
     // is the default behavior) and instead will use the current overall default for the queue.
     // If this command's `task_queue` field differs from the executing workflow's task queue, then
     // this flag has no effect.
-    bool use_default_build_id_set = 17;
+    bool use_latest_build_id = 17;
 }
 
 message ProtocolMessageCommandAttributes {
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -83,6 +83,12 @@ message ScheduleActivityTaskCommandAttributes {
     // Request to start the activity directly bypassing matching service and worker polling
     // The slot for executing the activity should be reserved when setting this field to true.
     bool request_eager_execution = 12;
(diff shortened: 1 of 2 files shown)


bergundy (approved):
Overall LGTM.
I'm just not convinced that for continue-as-new (and maybe even child workflow) we should stay within the compatible set by default.
It's of course a tradeoff between safety and being able to retire workers.
I have a feeling some users will be surprised by these defaults but I'm okay with starting out on the safer side.
I do think we should at least discuss this a bit more before making a call because up until recently the understanding was that CAN will go to the default since the expectation is that workflows are external APIs and cannot arbitrarily change their signatures, that's of course case dependent, and some workflows are public while others are private but we don't have this concept in the SDKs.

Sushisource:
@bergundy 
Right, good point about about CAN. I am fine with inverting its flag.

The other two, I have fairly high confidence that's the right choice there.

bergundy:
Oh and you're going to want to propagate this intent to the relevant history events.
For example a workflow will need to know which compatible set to use.

bergundy:
I have this thought that maybe if we see that a workflow was started within a compatible set, when it continues as new, it stays within that set.
It might give an indication that it's a "private" child workflow but I'd need to think about this some more, it's definitely harder to explain to users.

temporal/api/command/v1/message.proto
```diff
@@ -83,6 +83,12 @@ message ScheduleActivityTaskCommandAttributes {
     // Request to start the activity directly bypassing matching service and worker polling
     // The slot for executing the activity should be reserved when setting this field to true.
     bool request_eager_execution = 12;
+    // If this is set and the workflow executing this command is on a task queue using build-id
+    // versioning, then the scheduled activity will not use the same compatible version set (which
+    // is the default behavior) and instead will use the current overall default for the queue.
+    // If this command's `task_queue` field differs from the executing workflow's task queue, then
+    // this flag has no effect.
+    bool use_default_build_id_set = 13;
```
dnr, line 91: not sure about the name.. "build id set" isn't really what users are thinking in terms of, right?
```suggestion
    bool use_latest_version = 13;
```
?
```suggestion
    bool use_latest_build_id = 13;
```
?

dnr (approved):
shouldn't this go to the feature branch instead of master?

I think my leanings are:
- CAN should go to the same set by default if the task queue and workflow type are the same. that feels very "private". the typical use for that kind of CAN is to make an infinite loop (given our limitations). it would be surprising to me if it started running some other code on another binary.
- if the task queue is different I agree it should go to the overall default.
- if the task queue is the same but the workflow type is different... I don't know.
- whatever the behavior, it seems easier to explain/learn if all three of these are consistent

Sushisource:
> I have this thought that maybe if we see that a workflow was started within a compatible set, when it continues as new, it stays within that set. It might give an indication that it's a "private" child workflow but I'd need to think about this some more, it's definitely harder to explain to users.

By that logic, it would always continue within the set. The only case a workflow starts without being pinned to some set is if it started before there were any versions, and then we add one. Otherwise, the workflow's version set is determined once its first task starts. Alternatively, you could interpret that as they never start with some defined set. The only case where it seems clear they are predestined for some specific set is when they're run as child workflows, for example - if that's what you're getting at?

I think there's merit to both approaches for CAN. The above feels much too annoying to explain.

I'm inclined to say they should start on the same compat set for two reasons:

1. Makes all the commands consistent.
2. If you have already been smart enough to make a one-arg input which is wire-compatible, you know this, and probably are comfortable with setting a flag that says "start using the latest default" because you've had the expectation in your head all along that you were going to do that. Contrastingly, users who didn't think ahead as much get behavior that doesn't surprise them with an error. Seems like a fairly strong reason to me.

temporal/api/command/v1/message.proto
Sushisource: `use_latest_build_id` is probably better

PR #275, opened 2023-05-03
Add NewerBuildExistsFailure

dnr:
**What changed?**
Add error details for NewerBuildExistsFailure.

**Why?**
This error (with OutOfRange code) will be returned to pollers which will not receive any tasks because they've been superseded by a newer build.

### DIFF
--- a/temporal/api/errordetails/v1/message.proto
+++ b/temporal/api/errordetails/v1/message.proto
@@ -106,3 +106,8 @@ message SystemWorkflowFailure {
 
 message WorkflowNotReadyFailure {
 }
+
+message NewerBuildExistsFailure {
+    // Build ID of the newer compatible build that will receive tasks.
+    string latest_build_id = 1;
+}


temporal/api/errordetails/v1/message.proto
```diff
@@ -106,3 +106,9 @@ message SystemWorkflowFailure {
 
 message WorkflowNotReadyFailure {
 }
+
+message NewerBuildExistsFailure {
+    string namespace = 1;
+    string task_queue = 2;
+    string latest_build_id = 2;
```
dnr, line 113: question: do we need any of this info? is it helpful at all? presumably a poller knows its own ns and task queue and build id, the only real new bit of information is the latest build id

temporal/api/errordetails/v1/message.proto
```diff
@@ -106,3 +106,12 @@ message SystemWorkflowFailure {
 
 message WorkflowNotReadyFailure {
 }
+
+message NewerBuildExistsFailure {
+    string namespace = 1;
+    string task_queue = 2;
+    // TODO: should this be temporal.api.common.v1.WorkerVersionCapabilities instead of string?
+    string latest_build_id = 3;
```
dnr, line 114: question: do we need any of this info? is it helpful at all? presumably a poller knows its own ns and task queue and build id, the only real new bit of information is the latest build id

temporal/api/errordetails/v1/message.proto
```diff
@@ -106,3 +106,12 @@ message SystemWorkflowFailure {
 
 message WorkflowNotReadyFailure {
 }
+
+message NewerBuildExistsFailure {
```
cretz, line 110: Do you plan to, as we do with others error details, add something to https://pkg.go.dev/go.temporal.io/api/serviceerror? (specifically the type and the logic in the conversion)

temporal/api/errordetails/v1/message.proto
bergundy: We only need the latest_build_id.

temporal/api/errordetails/v1/message.proto
```diff
@@ -106,3 +106,12 @@ message SystemWorkflowFailure {
...
+    string namespace = 1;
+    string task_queue = 2;
+    // TODO: should this be temporal.api.common.v1.WorkerVersionCapabilities instead of string?
```
bergundy, line 113: Sure, let's do that for consistency and rename `NewBuildExistsFailure` to `NewerVersionExistsFailure`.

Sushisource: Hmm, I really don't think it should be. The name and the docstring for that message imbue an intent that this field should _not_ have:
`// Identifies the version(s) that a worker is compatible with when polling or identifying itself`

This field could never be multiple versions, and doesn't represent a particular worker.

If we used this for the other failures, I think those should change to just be build id strings.

temporal/api/errordetails/v1/message.proto
dnr: Of course, that's why I'm sending this

temporal/api/errordetails/v1/message.proto
Sushisource: @dnr Shall we fix the other fields Roey mentioned as well?

dnr: What other fields? This is the first custom serviceerror we've created for this feature

Sushisource: Oh man I don't know how I misread that so badly, I for some reason thought he was referencing two other existing messages. Nevermind.

Sushisource (approved):

PR #276, opened 2023-05-03
Record intent to use versioning in WorkerVersionStamp

bergundy:
We need to store this intent to distinguish between a when a version is used as a binary checksum replacement or for task dispatching.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -206,6 +206,12 @@ message WorkflowTaskCompletedEventAttributes {
     // Data the SDK wishes to record for itself, but server need not interpret, and does not
     // directly impact workflow state.
     temporal.api.sdk.v1.WorkflowTaskCompletedMetadata sdk_metadata = 6;
+
+    // If set, the worker that completed this task opted in to worker versioning. Otherwise, worker_version_stamp is
+    // used as a marker for workflow reset points and the BuildId search attibute.
+    // This flag must only be set if worker_version_stamp is provided.
+    bool use_versioning = 7;
+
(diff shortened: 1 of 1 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -206,6 +206,12 @@ message WorkflowTaskCompletedEventAttributes {
     // Data the SDK wishes to record for itself, but server need not interpret, and does not
     // directly impact workflow state.
     temporal.api.sdk.v1.WorkflowTaskCompletedMetadata sdk_metadata = 6;
+
+    // If set, the worker that completed this task opted in to worker versioning. Otherwise, worker_version_stamp is
+    // used as a marker for workflow reset points and the BuildId search attibute.
+    // This flag must only be set if worker_version_stamp is provided.
+    bool use_versioning = 7;
```
cretz, line 213: This versioning stuff sure is complicated from an API POV. I admit to not knowing what any of this is or why we have something called a "stamp" instead of just the version or what these terms are.

```suggestion
    // If set, the worker that completed this task opted in to worker versioning. Otherwise, worker_version_stamp is
    // used as a marker for workflow reset points and the BuildId search attibute.
    // This flag must only be set if worker_version is provided.
    bool use_versioning = 7;
```

Can we set the flag in the stamp? I am hoping to avoid more fields on task completed. Or is there a way where this is not needed? It may not be obvious, but we don't default exclude null/false on history export, so you just added a new field to every task complete on every history export.

bergundy:
Looks like GH outage screwed up this PR, it doesn't reflect my newer commits.

PR #278, opened 2023-05-09
Task reachability API

bergundy:
- Added concept of task reachability
- Added API to know if a build can be retired from a task queue / namespace
- Simplified the GetWorkerBuildIdCompatibility API

After some internal conversations, it seemed like it would make more sense to put the aggregated retirement information in the same API where we list pollers since we only care about retirement of active workers but reviewing different options showed that it's better to have a single API for reachability that accepts multiple build IDs.

I'm fairly certain that this leaves us with a minimal yet flexible API that serves all types of worker retirement related questions without needing a lot of round trips.


temporal/api/enums/v1/task_reachability.proto
```diff
@@ -0,0 +1,32 @@
...
+// Scope for task reachability query.
+enum TaskReachabilityScope {
+    TASK_REACHABILITY_SCOPE_UNSPECIFIED = 0;
+    // Query task reachability globally in a namespace.
+    TASK_REACHABILITY_SCOPE_NAMESPACE = 1;
+    // Query task reachability for a specific task queue.
+    TASK_REACHABILITY_SCOPE_TASK_QUEUE = 2;
+}
+
+// Task reachability for a specific poller.
```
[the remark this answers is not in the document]
bergundy: I'll put an extra description on this enum, I thought the per variant description was enough.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -830,11 +831,19 @@ message DescribeTaskQueueRequest {
     temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
     temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
     bool include_task_queue_status = 4;
+    temporal.api.enums.v1.TaskReachabilityScope task_reachability_query_scope = 5;
 }
 
 message DescribeTaskQueueResponse {
     repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
     temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2;
+    message BuildIdReachability {
```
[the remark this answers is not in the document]
bergundy: It can't be part of the status because that may be omitted from the response.
I have seen nested messages in this file so not sure if that tradition still applies.
Don't have a strong preference either way.

temporal/api/enums/v1/task_reachability.proto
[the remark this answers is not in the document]
cretz: Sometimes you say "task reachability" sometimes "task queue reachability". Now that I see this inconsistency, the latter does make more sense to me.

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
cretz: > It can't be part of the status because that may be omitted from the response.

In what situations does a task queue not have a status but does have reachability? Can you document these? Very confusing to a reader.

> I have seen nested messages in this file so not sure if that tradition still applies.

It applies in master today, but why not follow what we've done elsewhere? There's no reason for it sure, but there's not a huge reason not to I think. I think I brought this up previously and was dismissed there, so dismissing me here will be likely.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -830,11 +831,19 @@ message DescribeTaskQueueRequest {
...
+        // If empty, reachability correlates to "unverisoned" workers on this task queue.
```
cretz, line 841:
```suggestion
        // If empty, reachability correlates to "unversioned" workers on this task queue.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -830,11 +831,19 @@ message DescribeTaskQueueRequest {
...
+        string build_id = 1;
+        // Reachability per task queue.
+        repeated temporal.api.taskqueue.v1.TaskQueueReachability task_queue_reachability = 2;
+    }
+    BuildIdReachability build_id_reachability = 3;
```
cretz, line 846: Is there value in this being a separate message instead of just two fields here?

temporal/api/enums/v1/task_reachability.proto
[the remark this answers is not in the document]
bergundy: It's task reachability, sometimes put in the context of a single task queue.
A single build may be associated with multiple queues.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Thanks

temporal/api/workflowservice/v1/request_response.proto
bergundy: Should have been repeated, thanks linter Chad.

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
bergundy: I'm actually thinking of moving reachability out of describe task queue, and let the caller specify multiple build IDs in GetWorkerTaskReachabilityRequest.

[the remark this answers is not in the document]
bergundy: Status may be omitted from the response if `bool include_task_queue_status = 4` is false in the request.

temporal/api/enums/v1/task_reachability.proto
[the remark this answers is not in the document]
cretz: > It's task reachability

"Task" reachability like workflow task or activity task? How does whether a workflow is running on a task queue and build ID relate to tasks?

> sometimes put in the context of a single task queue.

When is this enum not put in the context of a single task queue?

> A single build may be associated with multiple queues.

Right, but this isn't an enum for build reachability, it's for task queue reachability right?

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
cretz: > let the caller specify multiple build IDs in GetWorkerTaskReachabilityRequest.

This is not necessary

> Status may be omitted from the response if bool include_task_queue_status = 4 is false in the request.

Do we expect to allow them to exclude this status info too? Why allow them to exclude that one but not this one? Is there a server cost for obtaining the build ID information? Is it separate from the cost of obtaining status? Or can we allow them to exclude both with the same boolean and accept that this is just more task queue status information?

temporal/api/workflowservice/v1/request_response.proto
cretz: Might as well make it a message in taskqueue like the other status message right next to it is

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1128,32 +1122,34 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+            string versioned = 2;
+        }
+    }
+    message ReachabilityScope {
+        oneof variant {
+            // Query task reachability globally in a namespace.
+            bool namespace = 1;
+            // Query task reachability for a specific task queue.
+            string task_queue = 2;
+        }
```
bergundy, line 1140: I might move these so they're not nested.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1128,32 +1122,34 @@ message GetWorkerBuildIdCompatibilityResponse {
...
     }
 
-    // A list of workers who are still live and polling the task queue, but may no longer be needed
-    // to make progress on open workflows.
-    repeated RetirementCandidate retirement_candidates = 2;
+    string namespace = 1;
+    // A build id identifies a worker process, and may be associated with multiple task queues.
+    // Must not be unversioned if query scope is namespace wide. 
+    repeated BuildId build_ids = 2;
```
bergundy, line 1146: Debating if this is worth a oneof, the alternative would be to use empty string for unversioned but I don't love that.
Might be worth making this explicit.

temporal/api/enums/v1/task_reachability.proto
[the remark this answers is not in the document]
bergundy: > "Task" reachability like workflow task or activity task? How does whether a workflow is running on a task queue and build ID relate to tasks? Does it need to relate to tasks instead of task queue?

TBH, I think the distinction doesn't matter much, I was looking at it from the perspective of "which tasks can reach this worker".

> When is this enum not put in the context of a single task queue?

Right now we always put this in the context of a single task queue.

> Right, but this isn't an enum for build reachability, it's for task queue reachability right?

The way I think about it is, tasks can always reach a task queue, a worker can always reach a task queue, but a not all tasks may reach a specific worker.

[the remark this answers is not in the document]
cretz: I think it should be clear from an API POV without too much thinking. "Is task queue + build ID in use and how". If we want to call that "reachability", ok. If we want "build reachability" to be a collection of "task queue reachability" values, ok. But not sure it needs to be deeper than that.

"Reachability" makes me think more of GC roots and such, which while that may have some analogue here, it's limited. I'd consider "Build ID task queues" which is just which task queues are in active use and how. Maybe "Build ID usage" or "Build ID task queue references" or something, unsure.

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
cretz: > It fits better that way, with the recent changes there's only one API for reachability and it can work in a task queue local or namespace global context.

Meh, we're ok w/ describing task queues individually and namespaces and such individually, I think we're ok w/ describing build ID usage individually per build ID. Technically we could have all of these calls accept multiple task queues, workflow IDs, namespace names, etc.

Also, when we get to the HTTP API, we're gonna want it REST-like (though by no means strictly RESTful) which would expect the URL path to be a specific build ID you can get info about.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1128,32 +1122,18 @@ message GetWorkerBuildIdCompatibilityResponse {
...
-        // The latest build id which completed a workflow task on some open workflow
-        string most_recent_build_id = 1;
-        // Currently polling workers who are compatible with `most_recent_build_id`.
-        repeated temporal.api.taskqueue.v1.PollerInfo pollers = 2;
-    }
+message GetWorkerTaskReachabilityRequest {
+    string namespace = 1;
+    // A build id identifies a worker process, and may be associated with multiple task queues.
+    // Must not be unversioned if query scope is namespace wide. 
+    repeated temporal.api.taskqueue.v1.MaybeVersioned build_ids = 2;
```
cretz, line 1131: I think a single build ID is fine.

Can we just have a `DescribeWorkerBuildId` and put reachability on there? (if it's too heavy, they can opt-in for this extra info)

bergundy: I tend to agree, I did this mostly for the UI to not have to issue a bunch of separate requests and match some of the goals of the original API which I replaced.
Let’s see what Spencer thinks too, he seemed to feel strongly about having the UI issue less requests.

cretz: Hrmm, what is the situation where UI needs to get the status of multiple build IDs? If the UI wanted to describe multiple task queues at once, how would that API look? Can we follow similar here? I know we have `ListNamespaces` and `ListWorkflowExecutions`, so a `ListWorkerBuildIds` with whatever filtering would make sense to me.

Sushisource: The scenario is discovering which build IDs you can retire, and having an easy way to get notified about that.

This API is fine at telling you if build IDs you know about _a-priori_ are retire-able or not, but it drops the thing the original API helped you with, which is telling you what workers can be retired when all you know is a task queue.

I suppose you can do that by asking about all the IDs, and then making this request with all of them, but that feels a bit like make-work for the client.

I'm also worried it doesn't compose super well with a future `ListTaskQueues` API, where if you want to learn what workers can be retired across a set of N queues you now have to do this set-union thing of all the build IDs from all the queues and then make this request.

`MaybeVersioned` could be changed to have a `string any_on_task_queue` variant to accomplish this goal I think.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -93,3 +93,39 @@ message CompatibleVersionSet {
...
+        bool unversioned = 1;
+        string versioned = 2;
+    }
+}
+
+// Scope of task reachability for a reachability query.
+message TaskReachabilityScope {
+    oneof variant {
+        // Query task reachability globally in a namespace.
+        bool namespace = 1;
```
Sushisource, line 126: What does it mean for this to be false? Looks like this should be `Empty` rather than `bool`

Same for the `unversioned` variant in `MaybeVersioned`

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1: Stepping back if we had to start over, I think three RPCs should exist:

* `DescribeWorkerBuildId` - for a single build ID (with `include_reachability` bool). By default this does what `GetWorkerBuildIdCompatibilityRequest` does too (no need for a separate call IMO)
* `UpdateWorkerBuildId` - generic form of `UpdateWorkerBuildIdCompatibility` (but it sounds like that's its only use today)
* `ListWorkerBuildIds` - list all build IDs (with a `include_reachability` bool). We can add filtering options later, but it sounds like they're not needed now (so no current value in accepting a collection of build IDs here).

And an update to:

* `DescribeTaskQueue` - add `include_build_id_reachability` bool and update response to optionally include all build IDs and their reachability

And I'd consider changing the term "reachability" to "usage" or something, but can't think of a good word so "reachability" is ok.

But I understand if too late for this or I am missing the idea of the RPCs

temporal/api/taskqueue/v1/message.proto
bergundy: Agree, I forgot that this was an option

temporal/api/workflowservice/v1/request_response.proto
bergundy: `DescribeWorkerBuildId` doesn't make much sense to me, what is there to describe besides the reachability?
`ListWorkerBuildIds` is what we call `GetWorkerBuildIdCompatibility` today.

Since we know we need an API to get the reachability of a single build ID, I say we start with that and expand as we see a need.
@Sushisource I'm going to change the reachability API to work on a single build ID, and if we see we need other APIs, we'll add them.

temporal/api/taskqueue/v1/message.proto
bergundy: It would be annoying in JS and the HTTP API, because the JSON representation of empty is `{}`, but I think it's fine.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1128,32 +1122,21 @@ message GetWorkerBuildIdCompatibilityResponse {
...
-    repeated RetirementCandidate retirement_candidates = 2;
-
-    message VersionsWithCompatiblePollers {
-        // The latest build id which completed a workflow task on some open workflow
-        string most_recent_build_id = 1;
-        // Currently polling workers who are compatible with `most_recent_build_id`.
-        repeated temporal.api.taskqueue.v1.PollerInfo pollers = 2;
+message GetWorkerTaskReachabilityRequest {
+    string namespace = 1;
+    // Either a specific worker version or all workers polling on a specific task queue.
```
Sushisource, line 1129:
```suggestion
    // Specify whether reachability should be returned for a specific worker version or all workers polling on a specific task queue.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1128,32 +1122,21 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+    oneof subject {
```
bergundy, line 1130: Settled on this as a compromise.
I was debating whether to have separate APIs to query by task queue or by build ID, but I think this is fine.
@cretz it won't look as nice as a REST API but @Sushisource and I tend to agree its better to contain this query in a single RPC.

Sushisource (approved):
This version makes sense for me. Thanks for upgrading.

temporal/api/workflowservice/v1/request_response.proto
cretz: > DescribeWorkerBuildId doesn't make much sense to me, what is there to describe besides the reachability?

This can combine reachability with compatibility into one getter which is nice for HTTP and matches our other API uses. Why do we have to have a getter for every little thing you want to get instead of just one?

> ListWorkerBuildIds is what we call GetWorkerBuildIdCompatibility today.

Not exactly. But this is needed for simple HTTP APIs. We need simple getters and listers. There is no reason to make this stuff special.

> Since we know we need an API to get the reachability of a single build ID, I say we start with that and expand as we see a need.

You won't expand, you'll be forced to make new APIs because you made these so specific and unlike our other get/list APIs.

PR #279, opened 2023-05-10
Add a current_execution field to WorkflowExecutionStartedEventAttributes

MichaelSnowden:
**What changed?**
We now have a `WorkflowExecution` field in the `WorkflowExecutionStartedEventAttributes` which tracks the current workflow execution.

**Why?**
To make it easier to replay workflows from history.

**Breaking changes**
No.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -100,6 +100,11 @@ message WorkflowExecutionStartedEventAttributes {
     // It should be used together with parent_initiated_event_id to identify
     // a child initiated event for global namespace
     int64 parent_initiated_event_version = 26;
+    // The workflow and run id of the current workflow execution. This run id may be different from the
+    // original_execution_run_id. For example, if a workflow is reset, a second workflow-started event will be created,
+    // and it will have the same original run id, but a new run id in this field.
+    // This field is new in 1.21.
+    temporal.api.common.v1.WorkflowExecution current_execution = 28;
 }
 
 message WorkflowExecutionCompletedEventAttributes {


temporal/api/history/v1/message.proto
```diff
@@ -100,6 +100,11 @@ message WorkflowExecutionStartedEventAttributes {
     // It should be used together with parent_initiated_event_id to identify
     // a child initiated event for global namespace
     int64 parent_initiated_event_version = 26;
+    // The workflow id and run id of the current workflow execution. The run id here is different from the
+    // original_execution_run_id. If a workflow is reset, a second workflow-started event will be created, and it will
+    // have the same original_execution_run_id, but a new current_execution.run_id.
+    // NOTE: this field is not populated for older server versions.
```
bergundy, line 106: Maybe say it will be populated in server>=1.21

temporal/api/history/v1/message.proto
```diff
@@ -100,6 +100,11 @@ message WorkflowExecutionStartedEventAttributes {
     // It should be used together with parent_initiated_event_id to identify
     // a child initiated event for global namespace
     int64 parent_initiated_event_version = 26;
+    // The workflow id and run id of the current workflow execution. The run id here is different from the
```
bergundy, line 103: Run may be different than original run id

bergundy (approved):

cretz (approved):
A great addition and one we've worked around not having in all SDKs. So the SDK team can't really use it (we work on lots of server versions in the past). How are you thinking it might be used by API users (i.e. not internal to the server)?

PR #282, opened 2023-05-15
Add a new resource exhausted cause

samanbarghi:
**What changed?**
Added workflow busy enum item for resource exhausted. 

**Why?**
When workflow is busy and cannot be locked, we want to return a proper error.

**Breaking changes**
No


cretz (approved):

MichaelSnowden (changes requested):
Looks like buf-breaking is failing here. I recently re-enabled it, so let me know if you'd like help troubleshooting.

cretz:
Probably just needs to update branch w/ latest master

yycptt (approved):

samanbarghi:
Yeah my local branch was outdated, just updated.

PR #285, opened 2023-05-16
Final versioning API changes

Sushisource:
**What changed?**

**Why?**

**Breaking changes**

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -131,19 +131,20 @@ message WorkerVersionStamp {
     // the task. The bundle could be a WASM blob, JS bundle, etc.
     string bundle_id = 2;
 
-    // If set, the worker is opting in to worker versioning. Otherwise, this is used only  as a
+    // If set, the worker is opting in to worker versioning. Otherwise, this is used only as a
     // marker for workflow reset points and the BuildIDs search attribute.
     bool use_versioning = 3;
 }
 
 // Identifies the version(s) that a worker is compatible with when polling or identifying itself,
-// and whether or not this worker is opting into the build-id based versioning feature
+// and whether or not this worker is opting into the build-id based versioning feature. This is
+// used by matching to determine which workers ought to receive what tasks.
 message WorkerVersionCapabilities {
     // An opaque whole-worker identifier
     string build_id = 1;
 
     // If set, the worker is opting in to worker versioning, and wishes to only receive appropriate
-    // tasks
+    // tasks.
     bool use_versioning = 2;
 
     // Later, may include info like "I can process WASM and/or JS bundles"
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -124,21 +124,27 @@ message MeteringMetadata {
 
 // Identifies the version(s) of a worker that processed a task
 message WorkerVersionStamp {
-    // An opaque whole-worker identifier
+    // An opaque whole-worker identifier. Replaces the deprecated `binary_checksum` field when this
+    // message is included in requests which previously used that.
     string build_id = 1;
     // Set if the worker used a dynamically loadable bundle to process
     // the task. The bundle could be a WASM blob, JS bundle, etc.
     string bundle_id = 2;
 
-    // If set, the worker is opting in to worker versioning. Otherwise, this is used as a marker for workflow reset
-    // points and the BuildIDs search attribute.
+    // If set, the worker is opting in to worker versioning. Otherwise, this is used only  as a
+    // marker for workflow reset points and the BuildIDs search attribute.
     bool use_versioning = 3;
 }
 
-// Identifies the version(s) that a worker is compatible with when polling or identifying itself
+// Identifies the version(s) that a worker is compatible with when polling or identifying itself,
+// and whether or not this worker is opting into the build-id based versioning feature
 message WorkerVersionCapabilities {
     // An opaque whole-worker identifier
     string build_id = 1;
+
+    // If set, the worker is opting in to worker versioning, and wishes to only receive appropriate
+    // tasks
+    bool use_versioning = 2;
 
     // Later, may include info like "I can process WASM and/or JS bundles"
 }
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -237,6 +237,10 @@ message WorkflowTaskFailedEventAttributes {
     int64 fork_event_version = 8;
     // If a worker explicitly failed this task, it's binary id
     string binary_checksum = 9;
+    // Version info of the worker who processed this workflow task, or missing if worker is not
+    // using versioning. If present, the `build_id` field within is also used as `binary_checksum`,
+    // which may be omitted in that case (it may also be populated to preserve compatibility).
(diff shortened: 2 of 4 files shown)


dnr:
How does it get there? We need it in `RespondActivityTaskCompletedRequest`?

Sushisource:
> How does it get there? We need it in `RespondActivityTaskCompletedRequest`?

:facepalm: added

cretz:
Can you clarify the reasoning here? A workflow's compatibility is tied to the version it starts with (#284) right? Why do we need to record the version each task ran on? Is it just for user awareness? If so, is that worth adding this to so many history entries?

Sushisource:
> Can you clarify the reasoning here? A workflow's compatibility is tied to the version it starts with (#284) right? Why do we need to record the version each task ran on? Is it just for user awareness? If so, is that worth adding this to so many history entries?

Yes it's for user awareness. And users can choose to have activities run on the default version overall rather than the compatible version for the workflow. In either case, you can't deduce the exact version b/c you need to know what the default (overall or compatible set) was at the time the task was taken, which can't be determined easily.

Given we're gonna limit ids to 255 characters, it seems a minimal cost.

bergundy (approved):
We should do the same for cancellations and failures.
Also consider adding this in the `*ById` APIs.

Sushisource:
> We should do the same for cancellations and failures. Also consider adding this in the `*ById` APIs.

I don't see how that works given the client which completes those isn't tied to a version

bergundy:
> > We should do the same for cancellations and failures. Also consider adding this in the `*ById` APIs.
> 
> I don't see how that works given the client which completes those isn't tied to a version

It responds on behalf of a worker

Sushisource:
> > > We should do the same for cancellations and failures. Also consider adding this in the `*ById` APIs.
> > 
> > 
> > I don't see how that works given the client which completes those isn't tied to a version
> 
> It responds on behalf of a worker

But those are for async completions, which happen outside of a worker.

temporal/api/history/v1/message.proto
```diff
@@ -237,6 +234,10 @@ message WorkflowTaskFailedEventAttributes {
     int64 fork_event_version = 8;
     // If a worker explicitly failed this task, it's binary id
     string binary_checksum = 9;
+    // Version info of the worker who processed this workflow task, or missing if worker is not
+    // using versioning. If present, the `build_id` field within is also used as `binary_checksum`,
+    // which may be omitted in that case (it may also be populated to preserve compatibility).
```
bergundy, line 239:
```suggestion
    // Version info of the worker who processed this workflow task, this should contain at least a `build_id` that is
    // used for visibility and reset purposes. Versioned workers must set `use_versioning` for tasks to be dispatched according to the defined build id compatibility constraints.
    // This field replaces the deprecated `binary_checksum`.
```

temporal/api/history/v1/message.proto
```diff
@@ -237,6 +234,10 @@ message WorkflowTaskFailedEventAttributes {
     int64 fork_event_version = 8;
     // If a worker explicitly failed this task, it's binary id
```
bergundy, line 235: Let's say that this is deprecated since server version 1.21.

temporal/api/common/v1/message.proto
```diff
@@ -124,21 +124,27 @@ message MeteringMetadata {
...
+    // An opaque whole-worker identifier. Replaces the deprecated `binary_checksum` field when this
+    // message is included in requests which previously used that.
     string build_id = 1;
     // Set if the worker used a dynamically loadable bundle to process
     // the task. The bundle could be a WASM blob, JS bundle, etc.
     string bundle_id = 2;
 
-    // If set, the worker is opting in to worker versioning. Otherwise, this is used as a marker for workflow reset
-    // points and the BuildIDs search attribute.
+    // If set, the worker is opting in to worker versioning. Otherwise, this is used only  as a
```
bergundy, line 134:
```suggestion
    // If set, the worker is opting in to worker versioning. Otherwise, this is used only as a
```

temporal/api/common/v1/message.proto
```diff
@@ -124,21 +124,27 @@ message MeteringMetadata {
...
+    // marker for workflow reset points and the BuildIDs search attribute.
     bool use_versioning = 3;
 }
 
-// Identifies the version(s) that a worker is compatible with when polling or identifying itself
+// Identifies the version(s) that a worker is compatible with when polling or identifying itself,
+// and whether or not this worker is opting into the build-id based versioning feature
```
bergundy, line 140:
```suggestion
// and whether or not this worker is opting into the build-id based versioning feature.
```

temporal/api/common/v1/message.proto
```diff
@@ -124,21 +124,27 @@ message MeteringMetadata {
...
 message WorkerVersionCapabilities {
     // An opaque whole-worker identifier
     string build_id = 1;
 
+    // If set, the worker is opting in to worker versioning, and wishes to only receive appropriate
+    // tasks
```
bergundy, line 146:
```suggestion
    // tasks.
```

temporal/api/common/v1/message.proto
```diff
@@ -124,21 +124,27 @@ message MeteringMetadata {
...
+    // If set, the worker is opting in to worker versioning. Otherwise, this is used only  as a
+    // marker for workflow reset points and the BuildIDs search attribute.
     bool use_versioning = 3;
 }
 
-// Identifies the version(s) that a worker is compatible with when polling or identifying itself
+// Identifies the version(s) that a worker is compatible with when polling or identifying itself,
+// and whether or not this worker is opting into the build-id based versioning feature
 message WorkerVersionCapabilities {
     // An opaque whole-worker identifier
```
bergundy, line 142: Could put a comment saying that this is used for matching purposes.

temporal/api/history/v1/message.proto
bergundy, line 235:
```suggestion
    // If a worker explicitly failed this task, its binary id
```

temporal/api/history/v1/message.proto
```diff
@@ -277,12 +278,10 @@ message ActivityTaskScheduledEventAttributes {
...
     temporal.api.common.v1.RetryPolicy retry_policy = 12;
-    // If this is set and the workflow executing this command is on a task queue using build-id
-    // versioning, then the scheduled activity will not use the same compatible version set (which
-    // is the default behavior) and instead will use the current overall default for the queue.
-    // If this command's `task_queue` field differs from the executing workflow's task queue, then
-    // this flag has no effect.
-    bool use_latest_build_id = 13;
+    // If this is set, the workflow executing this command wishes to start the activity using
+    // a version compatible with the version that this workflow most recently ran on, if such
+    // behavior is possible.
```
bergundy, line 283: WDYM by "if such behavior is possible"?

temporal/api/history/v1/message.proto
bergundy: Forgot to deprecate here.

bergundy (approved):

temporal/api/history/v1/message.proto
Sushisource: It's not if, ex, you're starting an activity on a different queue which doesn't share your version

bergundy:
@Sushisource make sure the next PR's title is "final final versioning API changes"

PR #286, opened 2023-05-16
Another take on the task reachability API

bergundy:
After implementing the previous take on this API, I decided to go back to a previous design and remove the ability to query by task queue to reduce the amount of visibility query fanout caused by a single request.
Also documented some of the limitations that will be enforced server side along with a recommendation for working around those limits.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1137,7 +1137,7 @@ message GetWorkerTaskReachabilityResponse {
     // Task reachability, broken down by task queue.
     // This response lists all task queues mapped to the requested build id when the requested query scope is for an
     // entire namespace but the number of task queues that include reachability information is limited.
-    // When reaching the limit, task queues that do not have reachability information will be marked with a single
+    // When reaching the limit, task queues that reachability information could not be retrieved for will be marked with a single
     // TASK_REACHABILITY_UNSPECIFIED entry. The caller may issue separate calls to get the reachability for those task
     // queues.
     // Open source users can adjust this limit by setting the server's dynamic config value for
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1126,17 +1126,23 @@ message GetWorkerBuildIdCompatibilityResponse {
 
 message GetWorkerTaskReachabilityRequest {
     string namespace = 1;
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1126,17 +1126,23 @@ message GetWorkerBuildIdCompatibilityResponse {
...
 }
 
 message GetWorkerTaskReachabilityResponse {
-    // Task reachability, broken down by build id.
-    repeated temporal.api.taskqueue.v1.BuildIdReachability build_id_reachability = 1;
+    // Task reachability, broken down by task queue.
+    // This response lists all task queues mapped to the requested build id when the requested query scope is for an
+    // entire namespace but the number of task queues that include reachability information is limited.
+    // When reaching the limit, task queues that do not have reachability information will be marked with a single
+    // TASK_REACHABILITY_UNSPECIFIED entry. The caller may issue separate calls to get the reachability for those task
```
Sushisource, line 1141: It's not that they don't have it, it's that we couldn't get it. Should apply to both these comments

```suggestion
    // When reaching the limit, task queues that reachability information could not be retrieved for will be marked with a single
    // TASK_REACHABILITY_UNSPECIFIED entry. The caller may issue separate calls to get the reachability for those task
```

bergundy: Thanks

temporal/api/workflowservice/v1/service.proto
```diff
@@ -393,8 +393,14 @@ service WorkflowService {
     // Fetches the worker build id versioning sets for a task queue.
     rpc GetWorkerBuildIdCompatibility (GetWorkerBuildIdCompatibilityRequest) returns (GetWorkerBuildIdCompatibilityResponse) {}
 
-    // Fetches task reachability to determine whether a single worker or all workers polling on a specific task queue may
-    // be retired.
+    // Fetches task reachability to determine whether a worker may be retired.
+    // The response lists all task queues mapped to the requested build id when the requested query scope is for an
+    // entire namespace but the number of task queues that include reachability information is limited.
+    // When reaching the limit, task queues that do not have reachability information will be marked with a single
```
Sushisource, line 399: Need to apply the change to this comment as well

Sushisource (approved):

temporal/api/workflowservice/v1/service.proto
bergundy: thanks

PR #287, opened 2023-05-16
Fix comment for use_latest_build_id and add some advice in the Update…

bergundy:
…WorkerBuildIdCompatibility API docstring

**Why?**

The SDK may choose to set `use_latest_build_id` when a single process hosts multiple workers on different task queues. Deleted the incorrect comment.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -161,8 +161,6 @@ message WorkflowExecutionContinuedAsNewEventAttributes {
     // If this is set and the workflow executing this command is on a task queue using build-id
     // versioning, then the scheduled activity will not use the same compatible version set (which
     // is the default behavior) and instead will use the current overall default for the queue.
-    // If this command's `task_queue` field differs from the executing workflow's task queue, then
-    // this flag has no effect.
     bool use_latest_build_id = 15;
 
     // workflow_execution_timeout is omitted as it shouldn't be overridden from within a workflow.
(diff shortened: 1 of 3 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -161,8 +161,6 @@ message WorkflowExecutionContinuedAsNewEventAttributes {
     // If this is set and the workflow executing this command is on a task queue using build-id
     // versioning, then the scheduled activity will not use the same compatible version set (which
     // is the default behavior) and instead will use the current overall default for the queue.
-    // If this command's `task_queue` field differs from the executing workflow's task queue, then
-    // this flag has no effect.
```
Sushisource, line 165: We still need to explain that it has no effect if the worker isn't also listening on that queue as well. It can just be 

```
 // If this command's `task_queue` field differs from the executing workflow's task queue, and
 // the build id of this worker does not exist in both queue's version sets then this flag has no effect.

bergundy: Ouch, that can be problematic.
We might to validate that the build ID is mapped to the other queue in matching when a WFT is completed for this to be safe to accept server side.
The other option would be to admit and add that build ID to the other task queue's version sets on dispatch like we do for replication but that seems dangerous, I'd rather fail a workflow task and let the user know ASAP.
CC @dnr.

bergundy: Thanks for pointing this out @Sushisource

Sushisource: Why's that a problem? The intent is just if the worker's build id exists in both TQs, then it can use the same set or default (use the flag). If it does not exist in the destination queue, then the flag has no effect, always default.

bergundy: It means we need to check if the build id exists in both TQs on completion though, which we haven't had to do yet.

Sushisource: Don't you only need to check when starting the task on the destination queue? The build ID is attached to the command, therefore you've already "checked" that side, and then you check the destination side when attempting to generate the task.

bergundy: We need to decide that.

bergundy:
Decided not to allow dispatching to a non default build across task queues for now.

PR #288, opened 2023-05-19
Apply more feedback to the reachability API

bergundy:
Allow querying multiple build IDs (yes, reverted back to that).
Let the caller specify the type of reachability they're interested in to avoid needlessly doubling the number of visibility queries on the server side.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1126,23 +1126,31 @@ message GetWorkerBuildIdCompatibilityResponse {
 
 message GetWorkerTaskReachabilityRequest {
     string namespace = 1;
-    // Build id to retrieve reachability for. Leave empty to query reachability of an unversioned worker.
-    string build_id = 2;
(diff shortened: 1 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1126,23 +1126,31 @@ message GetWorkerBuildIdCompatibilityResponse {
 
 message GetWorkerTaskReachabilityRequest {
     string namespace = 1;
-    // Build id to retrieve reachability for. Leave empty to query reachability of an unversioned worker.
-    string build_id = 2;
+    // Build id to retrieve reachability for. An empty string will be interpreted as an unversioned worker.
```
Sushisource, line 1129: Comment needs updating since there can be more than one. Maybe add a caveat that only one is supported for now, if you're going with that?

Sushisource (approved):
Thanks!

temporal/api/workflowservice/v1/request_response.proto
bergundy: I'll say this is limited in dynamic config. Missed this one when the changing from single to multi. Thanks!

PR #289, opened 2023-05-19
Allow querying reachability of multiple specific task queues

bergundy:
Figured that if the advice we give when a user has too many task queues for a given build id, they should (usually) be able to query the rest of the task queues in a followup API call and not have to issue a call per task queue.

This also flattens the API a bit, which is a nice side-effect.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1131,9 +1131,11 @@ message GetWorkerTaskReachabilityRequest {
     // Open source users can adjust this limit by setting the server's dynamic config value for
     // `limit.reachabilityQueryBuildIds` with the caveat that this call can strain the visibility store.
     repeated string build_ids = 2;
-    // Scope of the reachability query (namespace or task queue).
-    // Must specify a task queue if querying for an unversioned worker.
-    temporal.api.taskqueue.v1.TaskReachabilityScope scope = 3;
+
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1131,9 +1131,11 @@ message GetWorkerTaskReachabilityRequest {
...
     // `limit.reachabilityQueryBuildIds` with the caveat that this call can strain the visibility store.
     repeated string build_ids = 2;
-    // Scope of the reachability query (namespace or task queue).
-    // Must specify a task queue if querying for an unversioned worker.
-    temporal.api.taskqueue.v1.TaskReachabilityScope scope = 3;
+
+    // Task queues to retrieve reachability for. Leave this empty to query for all task queues associated with given
+    // build ids in the namespace.
+    // Must specify at least one task queue if querying for an unversioned worker.
+    repeated string task_queues = 3;
```
dnr, line 1138: Is there going to be a limit on this similar to `limit.reachabilityQueryBuildIds`?

dnr (approved):

temporal/api/workflowservice/v1/request_response.proto
bergundy: We already have a limit, it's `limit.reachabilityTaskQueueScan`.
I can clarify.

PR #292, opened 2023-05-24
Add capability for worker versioning

bergundy:


dnr:
It's already there, field 6

bergundy:
Oh wow, missed that, thanks!

PR #294, opened 2023-05-25
Add a flag to enable running when schedule runs out of actions

samanbarghi:
**What changed?**
Added a new flag to schedule policies to keep the schedule running when it runs
out of action. The plan is to finish the schedule by default when it
runs out of actions.

**Why?**
Currently, when a schedule runs out of actions the schedule workflow keeps
running without doing anything unless it receives a signal or the schedule gets
updated.

**Breaking changes**
This PR along with the one in server
(https://github.com/temporalio/temporal/pull/4399) are changing the default
behavior of schedule to finish workflow when it runs out of actions instead of
(current behavior) running. This can affect users that expect the schedule to
run even after we run out of actions.

### DIFF
--- a/temporal/api/schedule/v1/message.proto
+++ b/temporal/api/schedule/v1/message.proto
@@ -242,6 +242,11 @@ message SchedulePolicies {
     // This applies after retry policies: the full chain of retries must fail to
     // trigger a pause here.
     bool pause_on_failure = 3;
+
+    // If true, and the schedule runs out of actions, it will keep running. By default
+    // the schedule workflow finishes execution when it runs out of actions.
+    bool run_with_no_actions = 4;
+
 }
 
 message ScheduleAction {
--- a/temporal/api/schedule/v1/message.proto
+++ b/temporal/api/schedule/v1/message.proto
@@ -243,8 +243,8 @@ message SchedulePolicies {
     // trigger a pause here.
     bool pause_on_failure = 3;
 
-    // If true, and the schedule runs out of actions, it will keep running. By default
-    // the schedule workflow finishes execution when it runs out of actions.
+    // If true and the schedule runs out of actions it will not finish execution. 
+    // By default the schedule workflow finishes execution when it runs out of actions.
     bool remain_after_action_limit = 4;
 
 }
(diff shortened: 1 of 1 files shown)


temporal/api/schedule/v1/message.proto
```diff
@@ -242,6 +242,11 @@ message SchedulePolicies {
     // This applies after retry policies: the full chain of retries must fail to
     // trigger a pause here.
     bool pause_on_failure = 3;
+
+    // If true, and the schedule runs out of actions, it will keep running. By default
+    // the schedule workflow finishes execution when it runs out of actions.
+    bool run_with_no_actions = 4;
```
bergundy, line 248: I'm not sure I understand the comment, why do we need this flag as opposed to increasing the number of actions?

cretz:
Hrmm, one might expect that a schedule with no future actions still exists for me to update calendar on or trigger manually. I am not sure we should ever auto-delete a schedule.

temporal/api/schedule/v1/message.proto
dnr, line 248: Not sure about the name.. maybe this?
```suggestion
    bool remain_after_action_limit = 4;
```

dnr:
> Hrmm, one might expect that a schedule with no future actions still exists for me to update calendar on or trigger manually. I am not sure we should ever auto-delete a schedule. But if auto-delete-when-limited-actions-complete is really a needed feature, I think it should be opt-in.

It should definitely be an option, but the consensus so far was opt-out. See/join the slack discussion

temporal/api/schedule/v1/message.proto
samanbarghi: agreed, I trimmed the name too much and it lost its original meaning.

temporal/api/schedule/v1/message.proto
```diff
@@ -242,6 +242,13 @@ message SchedulePolicies {
     // This applies after retry policies: the full chain of retries must fail to
     // trigger a pause here.
     bool pause_on_failure = 3;
+
+    // If true, idle schedule workflow will keep running. Idle schedule is an schedule
+    // workflow that does not have any more jobs scheduled (e.g., ran out of actions)
+    // By default idle schedule workflow finishes execution when it has not more schedules
+    // to run.
+    bool keep_idle_schedule = 4;
```
cretz, line 250: I think auto-delete-by-default is still being discussed off PR

PR #297, opened 2023-06-06
Add operation for merging sets to build ID api

Sushisource:
**What changed?**
Added an operation for the merging of two compatible sets

**Why?**
It's possible in some edge situations that an incomplete replication could end up making two sets where the user's intention was that there only be one. This allows them to fix that.

Plus, theoretically this is useful if they've just made a mistake and declared something incompat they meant to be compat.

**Breaking changes**
No

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1084,6 +1084,13 @@ message UpdateWorkerBuildIdCompatibilityRequest {
         bool make_set_default = 3;
     }
 
+    message MergeSets {
+        // A build ID in the set whose default will become the merged set default
+        string primary_set_build_id = 1;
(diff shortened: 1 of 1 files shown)


bergundy (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1110,6 +1117,9 @@ message UpdateWorkerBuildIdCompatibilityRequest {
         // (-- api-linter: core::0140::prepositions=disabled
         //     aip.dev/not-precedent: Within makes perfect sense here. --)
         string promote_build_id_within_set = 6;
+        // Merge two existing sets together, thus declaring all build IDs in both sets compatible
```
bergundy, line 1120: I’d put what you wrote in the pr description here

PR #300, opened 2023-06-21
Support for SignalWithStartChildWorkflowExecutionCommandAttributesfr…

edmondop:
Expose https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#LL615C9-L615C48 SignalWithStartWorkflow capability to the client

**Why?**

To avoid the current workaround of using Temporal client from activities

**Breaking changes**

### DIFF
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -231,6 +231,48 @@ message StartChildWorkflowExecutionCommandAttributes {
     bool use_compatible_version = 17;
 }
 
+message SignalWithStartChildWorkflowExecutionCommandAttributes {
+    string namespace = 1;
+    string workflow_id = 2;
+    temporal.api.common.v1.WorkflowType workflow_type = 3;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 4;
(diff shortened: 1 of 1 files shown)


cretz:
I fear this implementation of this on the server is a larger challenge than just adding the API

edmondop:
> I think the implementation of this on the server is a larger challenge than it may seem

Can you share more ?

cretz:
Implementing a new command in the server and in every SDK is a non-trivial task and often it requires an accompanying event in workflow history. It can take a really long time to understand the code, and then a long time to implement it.

edmondop:
> Implementing a new command in the server and in every SDK is a non-trivial task and often it requires an accompanying event in workflow history. It can take a really long time to understand the code, and then a long time to implement it.

I see, would you prefer to have at least the accompanying PR on the server side before merging this? I think SDK can wait longer to implement the feature that the server offers

cretz:
Yes, I think the implementation should be worked on and understood before this part since the implementation will effect this. For example, you'll likely need an event. Warning, implementing a new command/event in the server is not easy.

PR #301, opened 2023-06-26
HTTP API annotations

cretz:
**What changed?**

* Add HTTP annotations to implement https://github.com/temporalio/proposals/pull/79
* Add gRPC gateway build dependency for proto build check
* Update Makefile to support additional Google HTTP annotations
* Altered API linter rules to fit our expectations of HTTP

Related PRs:

* Depended on by https://github.com/temporalio/api-go/pull/112
* Depended on by https://github.com/temporalio/temporal/pull/4543

This is now ready for review.


temporal/api/workflowservice/v1/service.proto
```diff
@@ -54,24 +55,33 @@ service WorkflowService {
     // isolation for all resources within the namespace. All resources belongs to exactly one
     // namespace.
     rpc RegisterNamespace (RegisterNamespaceRequest) returns (RegisterNamespaceResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces"
```
cretz, line 59: We could remove the `/api` or even the `/api/v1` prefixes here if we really wanted

temporal/api/workflowservice/v1/service.proto
```diff
@@ -107,6 +131,9 @@ service WorkflowService {
     // tasks. The worker is expected to call `RespondWorkflowTaskCompleted` when it is done
     // processing the task. The service will create a `WorkflowTaskStarted` event in the history for
     // this task before handing it to the worker.
+    //
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
```
cretz, line 136: Note that while I disabled most strict HTTP-annotation-based API linter rules, I left the must-have-HTTP rule on. I want future RPC calls to default to being available for HTTP and have to explicitly opt-out if they don't want to.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -413,36 +554,56 @@ service WorkflowService {
...
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: We do yet expose versioning API to HTTP. --)
     rpc GetWorkerTaskReachability (GetWorkerTaskReachabilityRequest) returns (GetWorkerTaskReachabilityResponse) {}
 
     // Invokes the specified update function on user workflow code.
-    // (-- api-linter: core::0134=disabled
-    //     aip.dev/not-precedent: UpdateWorkflowExecution doesn't follow Google API format --)
     rpc UpdateWorkflowExecution(UpdateWorkflowExecutionRequest) returns (UpdateWorkflowExecutionResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/workflows/{workflow_execution.workflow_id}/update/{request.input.name}"
```
cretz, line 565: After some self reflection, I decided not to add the additional `POST /api/v1/namespaces/{namespace}/workflows/{workflow_execution.workflow_id}/updates/{request.meta.update_id}` at this time. While I do agree that we need "updates" referenceable as resources, until we have a non-blocking way to get an update's status (i.e. `GET /api/v1/namespaces/{namespace}/workflows/{workflow_execution.workflow_id}/updates/{update_id}`), there is little value in supporting the update-as-resource URL currently. So this just matches signal and query today.

temporal/api/workflowservice/v1/service.proto
bergundy: I wouldn't.
We may want to have a v2 api and if we use this to power the web ui, we may want to server on the same port and thus need a prefix.

temporal/api/workflowservice/v1/service.proto
bergundy: Update ID is a mandatory param though server side (IIRC), might as well put the ID in the URL to be consistent with the GET method when that's added later.

temporal/api/workflowservice/v1/service.proto
cretz: I would discourage reusing the port for other purposes from outside the project, but I agree inside the project there could have other HTTP purposes maybe. I will leave it.

temporal/api/workflowservice/v1/service.proto
cretz: I think it's best to wait for `GET` to show up first, and I think I would use `/updates/{update_id}`. The fact is that name and ID are both required, and `POST /update/{update_name}` matches `POST /signal/{signal_name}` and `POST /query/{query_name}` (but needs ID in the body), whereas `POST /updates/{update_id}` matches `POST /workflows/{workflow_id}` where they both need a "name" in the body too.

Regardless, we really need a non-blocking update describe/get.

bergundy: Starting an update to me seems more like starting a workflow than a query / signal, it's an ALO. That's part of the reason I insisted they get caller assigned IDs, this will be super useful when we expose updates in Nexus handlers.

cretz: It is more like starting a workflow, but that's not what this call does, this call is more like querying. We will support starting an update like starting a workflow as soon as we support describing an update like describing a workflow.

bergundy: Okay, so why not put the update_id in the URL now already? We know where this API is going.

bergundy: An update is not the same as a signal or query, it is addressable.
Putting the update id in the url is not a mistake, it’s imho the right way to go.
When we add the apis to get the update status are you going to change your mind?

cretz: >  Putting the update id in the url is not a mistake, it’s imho the right way to go.

Then we can do it when we get the rest of the addressable pieces in there

>  When we add the apis to get the update status are you going to change your mind?

No, I have always believed update ID in the URL is the right way, see the proposal. I suggested _both_. I am just not wanting to put it in the URL _right now_ until we have other methods exposed with the ID.

bergundy: Won’t it be too late?

cretz: No, definitely not. You can add URLs (or `additional_bindings`) at any time.

MichaelSnowden (commented):
I have a few opinions here, but I think it's probably better that SDK folks drive what the API looks like, so please consider them nits. 

From server perspective, I think we really need some tests with this change. Can you add a test which makes an HTTP client and tries a few of these methods?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -241,57 +317,89 @@ service WorkflowService {
...
     rpc ListWorkflowExecutions (ListWorkflowExecutionsRequest) returns (ListWorkflowExecutionsResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/workflows"
+        };
     }
 
     // ListArchivedWorkflowExecutions is a visibility API to list archived workflow executions in a specific namespace.
     rpc ListArchivedWorkflowExecutions (ListArchivedWorkflowExecutionsRequest) returns (ListArchivedWorkflowExecutionsResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/archived-workflows"
```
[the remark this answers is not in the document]
MichaelSnowden: Hmm actually idk about that because it would conflict with something like workflows/{id}. This is probably fine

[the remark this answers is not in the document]
Sushisource: +1

temporal/api/workflowservice/v1/service.proto
```diff
@@ -241,57 +317,89 @@ service WorkflowService {
...
+        };
     }
 
     // ResetWorkflowExecution will reset an existing workflow execution to a specified
     // `WORKFLOW_TASK_COMPLETED` event (exclusive). It will immediately terminate the current
     // execution instance.
     // TODO: Does exclusive here mean *just* the completed event, or also WFT started? Otherwise the task is doomed to time out?
     rpc ResetWorkflowExecution (ResetWorkflowExecutionRequest) returns (ResetWorkflowExecutionResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/workflows/{workflow_execution.workflow_id}/reset"
```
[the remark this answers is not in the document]
cretz: Can you clarify the question a bit?

If you're asking how we validate the runtime value, how do we ensure that it is correct in gRPC? This is a dumb proto field setter nothing more. The gRPC server should validate input as it always does.

If you're asking how we validate at code gen time whether this field exists in the proto, yes, it grpc-gateway protoc gen will fail if this field is not an accurate proto message field.

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
cretz: Exactly. This was touched on in the proposal at https://github.com/temporalio/proposals/blob/master/api/http-api.md:

> Intentionally did not do `/workflows/archived` due to ambiguity

temporal/api/workflowservice/v1/service.proto
cretz: > I think /updates is better than /update.

It will be `/updates` when it is ID based

> But what will we do if/when we get that method? I imagine we'll wish the POST method was updates/ instead of update/

You will have both available to you. You have `POST /update/my-update-name` that works like `POST /query/my-query-name`, and when we are resource based, you have `POST /updates/my-update-id` that works like `POST /workflows/my-workflow-id`. This is covered a bit in the proposal at https://github.com/temporalio/proposals/blob/master/api/http-api.md. I just chose to wait on the ID one until we have other ID-based calls users can leverage here.

cretz:
> Can you add a test which makes an HTTP client and tries a few of these methods?

@MichaelSnowden - see the server PR

temporal/api/workflowservice/v1/service.proto
```diff
@@ -413,36 +554,56 @@ service WorkflowService {
...
     // Polls a workflow execution for the outcome of a workflow execution update
     // previously issued through the UpdateWorkflowExecution RPC. The effective
     // timeout on this call will be shorter of the the caller-supplied gRPC
     // timeout and the server's configured long-poll timeout.
-    // (-- api-linter: core::0134=disabled
-    //     aip.dev/not-precedent: UpdateWorkflowExecution doesn't follow Google API format --)
-    rpc PollWorkflowExecutionUpdate(PollWorkflowExecutionUpdateRequest) returns (PollWorkflowExecutionUpdateResponse){
+    //
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: We do expose updating polling API to HTTP in favor of a potential future non-blocking form. --)
```
alexshtin, line 576:
```suggestion
    //     aip.dev/not-precedent: We don't expose update polling API to HTTP in favor of a potential future non-blocking form. --)
```

temporal/api/workflowservice/v1/service.proto
alexshtin: `update_id` is not required. If omitted it will be auto-generated. For me, "update" is more like an action than a resource. This action returns "update outcome" and it is something which we can make addressable later with non-blocking API.

Makefile
```diff
@@ -15,29 +15,32 @@ GOPATH := $(shell go env GOPATH)
...
+	-I=$(shell go list -modfile build/go.mod -m -f '{{.Dir}}' github.com/temporalio/gogo-protobuf)/protobuf \
+	-I=$(shell go list -modfile build/go.mod -m -f '{{.Dir}}' github.com/grpc-ecosystem/grpc-gateway)/third_party/googleapis
 
 $(PROTO_OUT):
 	mkdir $(PROTO_OUT)
 
 ##### Compile proto files for go #####
 grpc: buf-lint api-linter buf-breaking gogo-grpc fix-path
 
-go-grpc: clean $(PROTO_OUT)
```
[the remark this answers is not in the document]
cretz: Easy to add back when that day comes

Makefile
```diff
@@ -49,6 +52,7 @@ grpc-install: gogo-protobuf-install
 
 gogo-protobuf-install: go-protobuf-install
 	go install -modfile build/go.mod github.com/temporalio/gogo-protobuf/protoc-gen-gogoslick
+	go install github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway@latest
```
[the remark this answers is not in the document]
cretz: This was mostly taken from api-go where we do install `@latest` for some of these. Having said that, I can pin a version in that build project no problem.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -54,24 +55,33 @@ service WorkflowService {
...
+        option (google.api.http) = {
+            post: "/api/v1/namespaces"
+            body: "*"
+        };
     }
 
     // DescribeNamespace returns the information and configuration for a registered namespace.
     rpc DescribeNamespace (DescribeNamespaceRequest) returns (DescribeNamespaceResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}"
```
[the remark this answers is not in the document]
cretz: Technically other fields, e.g. `id`, are supported via query params but yes, for HTTP, `namespace` is required. There are several calls that I didn't feel were regularly made enough to support HTTP interfaces for, and I think describing namespace by ID is one of them (we made the same decision in the Go SDK in the high-level namespace client).

temporal/api/workflowservice/v1/service.proto
```diff
@@ -80,6 +90,10 @@ service WorkflowService {
     // workflow executions will continue to run on deprecated namespaces.
     // Deprecated.
     rpc DeprecateNamespace (DeprecateNamespaceRequest) returns (DeprecateNamespaceResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/deprecate"
+            body: "*"
+        };
```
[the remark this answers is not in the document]
cretz: :+1: Will happily remove the HTTP exposure of this

temporal/api/workflowservice/v1/service.proto
```diff
@@ -153,6 +189,10 @@ service WorkflowService {
     // the workflow history. Calling `RecordActivityTaskHeartbeat` will fail with `NotFound` in
     // such situations, in that event, the SDK should request cancellation of the activity.
     rpc RecordActivityTaskHeartbeat (RecordActivityTaskHeartbeatRequest) returns (RecordActivityTaskHeartbeatResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/activities/heartbeat"
```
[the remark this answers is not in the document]
cretz: People pass around that token through their external systems, and you can easily set that token in the body of this call

temporal/api/workflowservice/v1/service.proto
```diff
@@ -161,6 +201,10 @@ service WorkflowService {
     // (-- api-linter: core::0136::prepositions=disabled
     //     aip.dev/not-precedent: "By" is used to indicate request type. --)
     rpc RecordActivityTaskHeartbeatById (RecordActivityTaskHeartbeatByIdRequest) returns (RecordActivityTaskHeartbeatByIdResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/activities/heartbeat-by-id"
```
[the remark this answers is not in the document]
cretz: I want the default way to be task token based, `-by-id` is for those using the less common activity ID approach. You have to provide workflow information too when calling the `-by-id` form, so putting ID in the URL makes little sense.

Makefile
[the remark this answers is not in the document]
alexshtin: I think you should pin version in both this and `api-go` repos.

alexshtin (approved):

Makefile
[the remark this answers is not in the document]
cretz: Concur, will do.

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
cretz: Exactly, that's what's here

temporal/api/workflowservice/v1/service.proto
cretz: Exactly. So you have the `POST update` action like `POST signal` or `POST query`. But when we have better ID support for other calls, you'll _also_ have `POST updates/{update_id}` resource creation like `POST workflows/{workflow_id}`.

Makefile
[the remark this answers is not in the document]
cretz: Fixed

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
cretz: Fixed

temporal/api/workflowservice/v1/service.proto
cretz: Fixed

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
MichaelSnowden: > If you're asking how we validate at code gen time whether this field exists in the proto, yes, grpc-gateway protoc gen will fail if this field is not an accurate proto message field.

Yeah, this is what I was looking for

temporal/api/workflowservice/v1/service.proto
MichaelSnowden: Would we have `update/{name}` and `updates/{id}`, but only `namespaces/{name}`?

bergundy: I find `update/{name}` and `updates/{id}` confusing and I'm quite certain users will too.
Strongly prefer a single way of doing things, especially when it's too easy to mistype the url and get the wrong update method.

cretz: > Would we have `update/{name}` and `updates/{id}`, but only `namespaces/{name}`?

Correct. Namespaces is a resource. Update is either a resource or an action.

> Strongly prefer a single way of doing things

But which way? Is update a resource or an action? If we must only have one way to create an update, I think it needs to be the "action" way. Everyone has to provide an update name and many (most?) of them will never provide an ID. By default an update will not return until complete so the default use is like a query.

Maybe we can have different-enough naming for the resource side? So you have `POST /update/{update_name}` today which is simple and works as expected of course. But for resource-based updates (whenever we get there), you can have `POST /workflow-updates/{update_id}` + `GET /workflow-updates/{update_id}`? Suggestions?

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
cretz: Add a comment here copying over justifications from the proposal? The comments in proto are essentially user docs, we don't really justify our design decisions inline (lest we need to do it in a lot of places).

temporal/api/workflowservice/v1/service.proto
```diff
@@ -88,17 +101,27 @@ service WorkflowService {
...
+            post: "/api/v1/namespaces/{namespace}/workflows/{workflow_id}"
+            body: "*"
+        };
     }
 
     // GetWorkflowExecutionHistory returns the history of specified workflow execution. Fails with
     // `NotFound` if the specified workflow execution is unknown to the service.
     rpc GetWorkflowExecutionHistory (GetWorkflowExecutionHistoryRequest) returns (GetWorkflowExecutionHistoryResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/workflows/{execution.workflow_id}/history"
```
feedmeapples, line 114: what about run id?

feedmeapples: that's what we did in UI `/namespaces/${parameters?.namespace}/workflows/${parameters?.workflowId}/runs/${parameters?.runId}/events`

temporal/api/workflowservice/v1/service.proto
```diff
@@ -88,17 +101,27 @@ service WorkflowService {
...
+        };
     }
     
     // GetWorkflowExecutionHistoryReverse returns the history of specified workflow execution in reverse 
     // order (starting from last event). Fails with`NotFound` if the specified workflow execution is 
     // unknown to the service.
     rpc GetWorkflowExecutionHistoryReverse (GetWorkflowExecutionHistoryReverseRequest) returns (GetWorkflowExecutionHistoryReverseResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/workflows/{execution.workflow_id}/history-reverse"
```
feedmeapples, line 123: same regarding the Run ID

temporal/api/workflowservice/v1/service.proto
cretz: This was discussed in the proposal PR: https://github.com/temporalio/proposals/pull/79/files#r1199151569. Basically all other programmatic interfaces to workflows allow run ID to be optional, so this programmatic interface does too

temporal/api/workflowservice/v1/service.proto
cretz: Same answer

temporal/api/workflowservice/v1/service.proto
bergundy: > `/workflow-updates/{update_id}`

I'm assuming you meant for it to be prefixed with `/namespaces/.../workflows/...`, it seems less than ideal to have to repeat `workflow` in the URL.

How about this?
```
POST /send-update/{update_name}
GET | POST /updates/{update_id}
```

cretz: Yes, it is assumed per workflow prefixed (I just left off in my explanation). Our API for working with updates in SDKs also require workflow information. Update IDs are not required to be globally unique, just workflow unique I believe. So update resources are specific to a workflow, not global.

Quinn-With-Two-Ns (approved):

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
yiminc: I feel that we should not expose all the by-id version of activity APIs. I doubt anyone are using them. We could always expose if we know there is strong demand.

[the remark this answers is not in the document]
cretz: Every SDK offers these valls via a high-level programmatic wrapper because we believe these are used. I think HTTP should at least offer the calls that a high-level SDK offers.

PR #303, opened 2023-07-12
Add group by support to count API

rodrigozhou:
**What changed?**
Add fields for count group by response.

The `CountWorkflowExecutions` will support the `GROUP BY` clause in the query.
The returned response will contain an additional field that will contain the groups and their counts.
The group type (`AggregationGroup`) contains two fields:
- `GroupValues`: this is tuple of values that defines the group.
- `Count`: the number workflows in the group.

### Example
If you have a query like `GROUP BY ExecutionStatus`, a possible response could be:
```
&CountWorkflowExecutionsResponse{
  Count: 10,
  Groups: []*AggregationGroup{
    {
      GroupValues: &Payloads{
        Payloads: []*Payload{"Completed"},
      },
      Count: 8,
    },
    {
      GroupValues: &Payloads{
        Payloads: []*Payload{"Running"},
      },
      Count: 2,
    },
  },
}
```

**Why?**

**Breaking changes**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -781,7 +781,21 @@ message CountWorkflowExecutionsRequest {
 }
 
 message CountWorkflowExecutionsResponse {
+    // If `query` is not grouping by any field, the count is an approximate number
+    // of workflows that matches the query.
+    // If `query` is grouping by a field, the count is simply the sum of the counts
+    // of the groups returned in the response. This number can be smaller than the
+    // total number of workflows matching the query.
     int64 count = 1;
+
+    // `groups` contains the groups if the request is grouping by a field.
+    // The list might not be complete, and the counts of each group is approximate.
+    repeated AggregationGroup groups = 2;
+
+    message AggregationGroup {
+        repeated temporal.api.common.v1.Payload group_values = 1;
+        int64 count = 2;
+    }
 }
 
 message GetSearchAttributesRequest {


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -782,6 +782,7 @@ message CountWorkflowExecutionsRequest {
 
 message CountWorkflowExecutionsResponse {
     int64 count = 1;
+    temporal.api.common.v1.AggregationGroup groups = 2;
```
cretz, line 785: Can you help me understand how this works (and maybe add details as comments)?

So will the query start accepting a `GROUP BY`? How does that work with`ListWorkflowExecutions`? If it doesn't, maybe what should be grouped by should not be in the query but instead accepted as a collection of strings?

temporal/api/common/v1/message.proto
```diff
@@ -66,6 +66,11 @@ message Memo {
     map<string, Payload> fields = 1;
 }
 
+message AggregationGroup {
+    Payloads name = 1;
```
bergundy, line 70: Why is this `Payloads`? Can group names be something other than strings?

bergundy (commented):
As @cretz mentioned, this requires some documentation for the reasoning and expected usage.

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: The Count API will support `GROUP BY` clause in the query.

`ListWorkflowExecutions` will not accept.

temporal/api/common/v1/message.proto
rodrigozhou: Here's the definition of `Payloads`: https://github.com/temporalio/api/blob/e29a98e37e9cc39673f4c224f0cdf93ea0bb8899/temporal/api/common/v1/message.proto#L46-L48.

It's just a wrapper around `Payload` to store a list of this type. `Payload` is the wrapper for search attribute values.

temporal/api/workflowservice/v1/request_response.proto
cretz: I wonder then if what should be grouped by should not be in the query but instead accepted as a collection of strings? That way we don't have to do special parsing in this special case

temporal/api/common/v1/message.proto
cretz: What @bergundy is asking is why isn't this `string name`?

cretz: > Can group names be something other than strings?

@rodrigozhou - there is confusion here. You are using `name` to mean distinct `value`. You should probably change that field name to be `values`. The value stored in search attributes that you may group by is not always a "name".

temporal/api/common/v1/message.proto
```diff
@@ -66,6 +66,11 @@ message Memo {
     map<string, Payload> fields = 1;
 }
 
+message AggregationGroup {
```
cretz, line 69: Is this really a common message if it's only used in one place? If only we supported nested messages.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -782,6 +782,7 @@ message CountWorkflowExecutionsRequest {
 
 message CountWorkflowExecutionsResponse {
     int64 count = 1;
+    repeated temporal.api.common.v1.AggregationGroup groups = 2;
```
cretz, line 785: Is this really a "common" message if it's only used in one place? Some other more recent API additions use nested messages, can we also do so here? It is a backwards compatible change to move it later if we need to.

temporal/api/common/v1/message.proto
rodrigozhou: I was thinking in the sense of the group name: after you group by the keys, the tuple of values is the name of the group (copied from Python pandas).
I had initially named `values` and then changed, couldn't quite think what would be the ideal name.

cretz: Yeah, I can't think of the best word. Maybe just `group` or `group_values` or `distinct_values` or something. I do think `name` is a bit confusing because it's not really a "name" in that sense.

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: If I nest, the proto will be compiled to something like `CountWorkflowExecutionsResponse_AggregationGroup`, which I'm not sure it will be nice since I'm using this type in the visibility manager when building the response there.

temporal/api/common/v1/message.proto
Sushisource: Also, side note: IMO the `Payloads` message is useless. In all the time it's existed we've never added anything to it and as it stands it's just pointless indirection around `repeated Payload`. I'd vote you just use that directly.

rodrigozhou: I had the same thought. Just used because it was there...

temporal/api/workflowservice/v1/request_response.proto
cretz: I mean from a user POV you now have different supported query formats for different calls

temporal/api/workflowservice/v1/request_response.proto
cretz: I would not concern yourself w/ internal code struggles when building this API. Rather think of the caller. I think as a principle we shouldn't put things in "common" that aren't, well, common.

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: That's a valid point. We had talked about this before, and went with supporting via clause as it wouldn't impact any of the existing tools. If we add as collection of search attribute, we'd have to update all tools to accept an additional param.

cretz: For tools to show these groups in their response they have to be updated anyways, right? So using tool update as the reason doesn't seem valid. But I don't have a strong opinion here, if y'all want in the string instead of as a field, ok, though I slightly disagree.

Lets just make sure we add some comments to the proto fields explaining what's happening here.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -781,7 +781,18 @@ message CountWorkflowExecutionsRequest {
 }
 
 message CountWorkflowExecutionsResponse {
+    // Approximate total count of workflows that matches the filter.
+    // It can be larger than the sum of the counts in `groups`.
```
cretz, line 785: I am a bit curious, why is this? Why not, in group by queries, set this value as just the sum of distinct group counts?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -781,7 +781,18 @@ message CountWorkflowExecutionsRequest {
...
     int64 count = 1;
+
+    // `groups` contains the groups if the request is grouping by fields.
+    // The list might not be complete, and the counts of each group is approximate.
+    repeated AggregationGroup groups = 2;
+
+    message AggregationGroup {
+        repeated temporal.api.common.v1.Payload group_values = 1;
+        int64 count = 2;
```
cretz, line 794: Can you document these fields? Specifically, does `count` represent the same "approximate" as `count` in the parent message does?

temporal/api/workflowservice/v1/request_response.proto
cretz: Per offline discussion and link to internal discussion where others already decided, we are going to continue using our existing stringly-typed approach.

But I would _really_ appreciate a publicly documented grammar for the query string.

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: I'm actually ok with either option. Being the sum of the groups will make the query faster. Let me change it to be the sum of the groups here first.

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -781,7 +781,21 @@ message CountWorkflowExecutionsRequest {
 }
 
 message CountWorkflowExecutionsResponse {
+    // If `query` is not grouping by any field, the count is an approximate number
+    // of documents that matches the query.
```
yiminc, line 785: documents is ES term.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -781,7 +781,21 @@ message CountWorkflowExecutionsRequest {
...
+    // If `query` is grouping by fields, the count is simply the sum of the counts
```
yiminc, line 786: I think we only support grouping by one field?

yiminc (approved):

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: Yes, fixed comments.

PR #304, opened 2023-08-03
Add capability flag for count group by status

rodrigozhou:
**What changed?**
Capability flag for count group by `ExecutionStatus`.

**Why?**

**Breaking changes**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -926,6 +926,10 @@ message GetSystemInfoResponse {
         // True if the server knows about the sdk metadata field on WFT completions and will record
         // it in history
         bool sdk_metadata = 9;
+
+        // True if the server supports count group by execution status
+        // (-- api-linter: core::0140::prepositions=disabled --)
+        bool count_group_by_execution_status = 10;
     }
 }
 


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -926,6 +926,10 @@ message GetSystemInfoResponse {
         // True if the server knows about the sdk metadata field on WFT completions and will record
         // it in history
         bool sdk_metadata = 9;
+
+        // True if the server supports count group by execution status
+        // (-- api-linter: core::0140::prepositions=disabled --)
+        bool count_group_by_execution_status = 10;
```
cretz, line 932: How is this expected to be used?

rodrigozhou: Oh, that's a good point. Initially, I was thinking of using in our canary-go tests. But now I'm thinking if I should just ignore "NotSupportedOperation" error to skip the test.
cc: @yiminc

Sushisource: There's one possible reason it could make sense: If it's known count is supported, it's possible some logic could be optimized by using the count api where previously list was used. But... I don't think we do that anywhere and if we ever did, it'd probably be fine to just try the API first.

cretz: > Initially, I was thinking of using in our canary-go tests. But now I'm thinking if I should just ignore "NotSupportedOperation" error to skip the test.

This can remain here as is IMO. I just wanted to see if it had any value to anyone. If anyone is going to use it in anyway, I'd much rather be safe than sorry and keep it here (I totally support loading up this capabilities object w/ as many booleans as needed over and over, I just was curious how it was gonna be used).

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
yiminc: I think the web UI will need to based on this flag decide if they are going to use single CountWorkflowExecution call to get count per each ExecutionStatus, or making individual calls one for each ExecutionStatus.

yiminc (approved):

PR #310, opened 2023-08-30
Bad build ids api changes

dnr:
**What changed?**
- Add `MarkBadBuild` option to `UpdateWorkerBuildIdCompatibilityRequest` and related messages
- Add build-id-based option to reset, including `RESET_TYPE_BUILD_ID` to `ResetType` in `BatchOperationReset`

**Why?**
Help recover from a bad deployment situation

**Breaking changes**
no

### DIFF
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -93,9 +93,13 @@ message BatchOperationDeletion {
 // Keep the parameter in sync with temporal.api.workflowservice.v1.ResetWorkflowExecutionRequest.
 message BatchOperationReset {
   // Reset type.
-  temporal.api.enums.v1.ResetType reset_type = 1; 
+  temporal.api.enums.v1.ResetType reset_type = 1;
   // History event reapply options.
-  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2; 
+  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
   // The identity of the worker/client.
   string identity = 3;
+
+  // For reset_type == RESET_TYPE_BUILD_ID only:
+  // Resets to before the first workflow task processed by the specified build id.
+  string reset_build_id = 4;
 }
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -100,6 +100,5 @@ message BatchOperationReset {
   string identity = 3;
 
   // For reset_type == RESET_TYPE_BUILD_ID only:
-  // Resets to before the first workflow task processed by the specified build id.
-  string reset_build_id = 4;
+  temporal.api.common.v1.BuildIdResetOptions reset_build_id = 4;
 }
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -92,13 +92,16 @@ message BatchOperationDeletion {
 // BatchOperationReset sends reset requests to batch workflows.
 // Keep the parameter in sync with temporal.api.workflowservice.v1.ResetWorkflowExecutionRequest.
 message BatchOperationReset {
+  // The identity of the worker/client.
+  string identity = 3;
+
+  // New API (exclusive with `reset_type` and `reset_reapply_type`):
+  temporal.api.common.v1.ResetOptions options = 4;
+
+  // Old API (exclusive with `options`):
   // Reset type.
   temporal.api.enums.v1.ResetType reset_type = 1;
   // History event reapply options.
   temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
-  // The identity of the worker/client.
-  string identity = 3;
 
-  // For reset_type == RESET_TYPE_BUILD_ID only:
-  temporal.api.common.v1.BuildIdResetOptions reset_build_id = 4;
 }
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -149,3 +149,12 @@ message WorkerVersionCapabilities {
 
     // Later, may include info like "I can process WASM and/or JS bundles"
 }
+
+// Common options for performing a workflow reset based on build id.
+message BuildIdResetOptions {
+    // Resets to before the first workflow task processed by this build id. Note that by default,
+    // this reset is allowed to be to a prior run in a chain of continue-as-new.
+    string build_id = 1;
+    // If true, limit the reset to only within the current run.
+    bool current_run_only = 2;
+}
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -32,10 +32,12 @@ option ruby_package = "Temporalio::Api::Common::V1";
 option csharp_namespace = "Temporalio.Api.Common.V1";
 
 import "google/protobuf/duration.proto";
+import "google/protobuf/empty.proto";
 
 import "dependencies/gogoproto/gogo.proto";
 
 import "temporal/api/enums/v1/common.proto";
+import "temporal/api/enums/v1/reset.proto";
 
 message DataBlob {
     temporal.api.enums.v1.EncodingType encoding_type = 1;
@@ -150,11 +152,28 @@ message WorkerVersionCapabilities {
     // Later, may include info like "I can process WASM and/or JS bundles"
 }
 
-// Common options for performing a workflow reset based on build id.
-message BuildIdResetOptions {
-    // Resets to before the first workflow task processed by this build id. Note that by default,
-    // this reset is allowed to be to a prior run in a chain of continue-as-new.
-    string build_id = 1;
-    // If true, limit the reset to only within the current run.
-    bool current_run_only = 2;
+message ResetOptions {
+    // Which workflow task to reset to.
+    oneof target {
+        // Resets to event of the first workflow task completed, or if it does not exist, the event after task scheduled.
+        google.protobuf.Empty first_workflow_task = 1;
+        // Resets to event of the last workflow task completed, or if it does not exist, the event after task scheduled.
+        google.protobuf.Empty last_workflow_task = 2;
+        // The id of a specific `WORKFLOW_TASK_COMPLETED`,`WORKFLOW_TASK_TIMED_OUT`, `WORKFLOW_TASK_FAILED`, or
+        // `WORKFLOW_TASK_STARTED` event to reset to.
+        int64 workflow_task_id = 3;
+        // Resets to before the first workflow task processed by this build id.
+        // If the workflow was not processed by the build id, or the workflow task can't be
+        // determined, no reset will be performed.
+        // Note that by default, this reset is allowed to be to a prior run in a chain of
+        // continue-as-new.
+        string build_id = 4;
+    }
+
+    // History event reapply options.
+    temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 10;
+
+    // If true, limit the reset to only within the current run. (Applies to build_id targets and
+    // possibly others in the future.)
+    bool current_run_only = 11;
 }
--- a/temporal/api/errordetails/v1/message.proto
+++ b/temporal/api/errordetails/v1/message.proto
@@ -110,4 +110,6 @@ message WorkflowNotReadyFailure {
 message NewerBuildExistsFailure {
     // The current default compatible build ID which will receive tasks
     string default_build_id = 1;
+    // If this failure is due to the polling build id being marked as a bad build.
+    bool is_bad_build = 2;
 }
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -95,6 +95,10 @@ message StickyExecutionAttributes {
 message CompatibleVersionSet {
     // All the compatible versions, unordered, except for the last element, which is considered the set "default".
     repeated string build_ids = 1;
+    // Build ids that have been marked as bad. If the last element in build_ids is also in
+    // bad_build_ids, then no tasks will be dispatched on this version set until the default is
+    // changed.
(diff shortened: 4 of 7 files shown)


temporal/api/batch/v1/message.proto
```diff
@@ -93,9 +93,13 @@ message BatchOperationDeletion {
...
-  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2; 
+  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
   // The identity of the worker/client.
   string identity = 3;
-}
\ No newline at end of file
+
+  // For reset_type == RESET_TYPE_BUILD_ID only:
+  // Resets to before the first workflow task processed by the specified build id.
+  string reset_build_id = 4;
```
bergundy, line 104: Might be worth wrapping in a message in case we want to allow more options here. I'm already thinking that some users may not want to reset past the current run ID.

temporal/api/errordetails/v1/message.proto
```diff
@@ -110,4 +110,6 @@ message WorkflowNotReadyFailure {
 message NewerBuildExistsFailure {
     // The current default compatible build ID which will receive tasks
     string default_build_id = 1;
+    // If this failure is due to the polling build id being marked as a bad build.
+    bool is_bad_build = 2;
```
bergundy, line 114: Maybe change to a reason enum to support other reasons in the future?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -663,12 +663,15 @@ message ResetWorkflowExecutionRequest {
...
     temporal.api.common.v1.WorkflowExecution workflow_execution = 2;
     string reason = 3;
-    // The id of a `WORKFLOW_TASK_COMPLETED`,`WORKFLOW_TASK_TIMED_OUT`, `WORKFLOW_TASK_FAILED`, or
-    // `WORKFLOW_TASK_STARTED` event to reset to.
-    int64 workflow_task_finish_event_id = 4;
+    oneof reset_type {
+        // The id of a `WORKFLOW_TASK_COMPLETED`,`WORKFLOW_TASK_TIMED_OUT`, `WORKFLOW_TASK_FAILED`, or
+        // `WORKFLOW_TASK_STARTED` event to reset to.
+        int64 workflow_task_finish_event_id = 4;
+        string reset_build_id = 7;
```
bergundy, line 670: same, consider making this a message instead of a string.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -95,6 +95,10 @@ message StickyExecutionAttributes {
 message CompatibleVersionSet {
     // All the compatible versions, unordered, except for the last element, which is considered the set "default".
     repeated string build_ids = 1;
+    // Build ids that have been marked as bad. If the last element in build_ids is also in
+    // bad_build_ids, then no tasks will be dispatched on this version set until the default is
+    // changed.
+    repeated string bad_build_ids = 2;
```
bergundy, line 101: In the internal representation, you can add a new state in the enum, this is fine here though.

temporal/api/batch/v1/message.proto
Sushisource: +1

dnr: good idea

temporal/api/errordetails/v1/message.proto
dnr: can you imagine anything else (that would fit generally under `NewerBuildExistsFailure` instead of some other error type)?

temporal/api/taskqueue/v1/message.proto
dnr: yup, that was the plan. we could try to transition the string list to a message but... seems more trouble than it's worth. I don't like the repetition but it's the simplest thing here. I guess it could be a parallel array of bools but that's worse

temporal/api/batch/v1/message.proto
```diff
@@ -93,9 +93,13 @@ message BatchOperationDeletion {
...
+  temporal.api.enums.v1.ResetType reset_type = 1;
   // History event reapply options.
-  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2; 
+  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
   // The identity of the worker/client.
   string identity = 3;
-}
\ No newline at end of file
+
+  // For reset_type == RESET_TYPE_BUILD_ID only:
```
dnr, line 102: I wish this was a oneof in the first place :(

temporal/api/taskqueue/v1/message.proto
bergundy: Yup, this is the best option given the current API, let's keep it and avoid breaking existing clients.

temporal/api/errordetails/v1/message.proto
bergundy: Not at the moment, but we didn't see this one coming either.
As a general guideline, I prefer enums even for dual states, you can't predict how this evolve so early on.

dnr: Hmm. Except without another example it's not clear if these should be exclusive (like an enum) or orthogonal (like other fields). E.g. a build can be both not-default and also a bad build.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -663,12 +663,16 @@ message ResetWorkflowExecutionRequest {
...
     string reason = 3;
-    // The id of a `WORKFLOW_TASK_COMPLETED`,`WORKFLOW_TASK_TIMED_OUT`, `WORKFLOW_TASK_FAILED`, or
-    // `WORKFLOW_TASK_STARTED` event to reset to.
-    int64 workflow_task_finish_event_id = 4;
+    oneof reset_type {
+        // The id of a `WORKFLOW_TASK_COMPLETED`,`WORKFLOW_TASK_TIMED_OUT`, `WORKFLOW_TASK_FAILED`, or
+        // `WORKFLOW_TASK_STARTED` event to reset to.
+        int64 workflow_task_finish_event_id = 4;
+        // Reset to before a specific build id.
+        temporal.api.common.v1.BuildIdResetOptions reset_build_id = 7;
```
yiminc, line 671: Currently, we have reset option to reset to first or last workflow task. And the caller (batch workflow) has to figure out which workflow task event_id that option maps to. That is different from this new option which history service has to figure out which event_id it maps to.
I like this new approach, but it make it inconsistent with existing ones. Probably should (in future) add the first / last workflow task as option here and move those logic from batch workflow to history service.

yiminc: Just realize this is already requested as https://github.com/temporalio/temporal/issues/1338

dnr: I see. Do you think we should extract all this into a whole new message and migrate to that here? Otherwise we have a oneof here, and then an enum for the batch request forever... instead we could make a new message and use it for both, and deprecate the old fields.

Or we could have the batch workflow look up the reset point, but that seems silly since history owns it.

temporal/api/errordetails/v1/message.proto
dnr: I'm still not sure about this one. Note that these two fields are not exclusive and actually maybe we wouldn't set `default_build_id` if the default was marked bad without a replacement. I.e. the options are:
- there is a new default build id, but the one you polled with is not marked bad
- there is a new default build id, also btw the one you polled with is bad
- there is no default since there was one and it got marked bad, so now nobody gets any tasks

At this point it's really the name of the message/error type that's not accurate, but I don't want to change that now

temporal/api/batch/v1/message.proto
dnr: I just went and made it a oneof. I think it's the cleanest way to deal with the new reset options.

temporal/api/workflowservice/v1/request_response.proto
dnr: I just made a new message and used it in both places. It'll require some compatibility code to translate but it shouldn't be too bad.

temporal/api/batch/v1/message.proto
dnr: I made a new message for overall ResetOptions and put that flag (only current run) at the top level, since it could apply to other reset types. So then build_id didn't need a message again. I'm not opposed to a message there but are there really reset options that apply to build ids only?

temporal/api/errordetails/v1/message.proto
dnr: We already have the custom error type, I'm just adding a field to it

bergundy: I know, this is more of a lesson for next time.

dnr:
I changed this to use the new ResetOptions only for the batch request, not the individual request. Using it for the individual request could come later, although maybe there are good reasons to have them all go through batch (so that we do the digging through history on the worker instead of frontend or history).

I also added a build id to ResetPointInfo so it has both a build id and binary checksum now, and we can remove binary checksum later.

There are a few situations that may require some more changes here, but this should be good to go for phase 1.

temporal/api/batch/v1/message.proto
```diff
@@ -92,10 +92,16 @@ message BatchOperationDeletion {
...
-  // Reset type.
-  temporal.api.enums.v1.ResetType reset_type = 1; 
-  // History event reapply options.
-  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2; 
   // The identity of the worker/client.
   string identity = 3;
-}
\ No newline at end of file
+
+  // New API (exclusive with `reset_type` and `reset_reapply_type`):
```
Sushisource, line 98: IMO this comment should just say what this is for, and the comments for the older things can call out that they are old/deprecated.

Maybe keep the exclusivity mention, but don't need to say it's "New" I think.

temporal/api/common/v1/message.proto
```diff
@@ -149,3 +151,29 @@ message WorkerVersionCapabilities {
 
     // Later, may include info like "I can process WASM and/or JS bundles"
 }
+
+message ResetOptions {
+    // Which workflow task to reset to.
+    oneof target {
+        // Resets to event of the first workflow task completed, or if it does not exist, the event after task scheduled.
```
Sushisource, line 158: What if there's no event after task scheduled either?

I think this wording is slightly more clear:
```suggestion
        // Resets to the first `WORKFLOW_TASK_COMPLETED` event, or if it does not exist, the first `WORKFLOW_TASK_SCHEDULED` event.
```

temporal/api/common/v1/message.proto
```diff
@@ -149,3 +151,29 @@ message WorkerVersionCapabilities {
...
+        google.protobuf.Empty first_workflow_task = 1;
+        // Resets to event of the last workflow task completed, or if it does not exist, the event after task scheduled.
```
Sushisource, line 160: Same comment as w/ first

temporal/api/common/v1/message.proto
```diff
@@ -149,3 +151,29 @@ message WorkerVersionCapabilities {
...
+        google.protobuf.Empty last_workflow_task = 2;
+        // The id of a specific `WORKFLOW_TASK_COMPLETED`,`WORKFLOW_TASK_TIMED_OUT`, `WORKFLOW_TASK_FAILED`, or
+        // `WORKFLOW_TASK_STARTED` event to reset to.
+        int64 workflow_task_id = 3;
+        // Resets to before the first workflow task processed by this build id.
```
Sushisource, line 165: Which event, specifically, before the first WFT processed by the build ID?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1110,10 +1110,26 @@ message UpdateWorkerBuildIdCompatibilityRequest {
         string secondary_set_build_id = 2;
     }
 
+    message MarkBadBuild {
+        // The build id to mark as bad.
+        string bad_build_id = 1;
+        // If promote_build_id is set, then also promote it to be the default of the set.
+        // If it's not set, then no tasks will be dispatched on the task queue set until another
+        // build id is made the default by another call to UpdateWorkerBuildIdCompatibility that
+        // sets/adds a new default for the set.
```
Sushisource, line 1119: This comment implies that `bad_build_id` is the current default for it's set, but, that might not be the case. Presumably you really only care to do this if it is, but, it's worth mentioning that.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1142,6 +1158,14 @@ message UpdateWorkerBuildIdCompatibilityRequest {
         // declare as compatible. The unusual case of incomplete replication during failover could
         // also result in a split set, which this operation can repair.
         MergeSets merge_sets = 7;
+        // Mark a build as "bad" for a particular task queue. This means that:
+        // 1. the build will not receive any more tasks (workflow or activity)
+        // 2. it may not be promoted to the set default anymore
+        // See the MarkBadBuild for more information.
```
Sushisource, line 1164:
```suggestion
        // See MarkBadBuild for more information.
```

Sushisource (commented):
Structurally makes sense to me. Some stuff to fix in the docstrings.

temporal/api/common/v1/message.proto
dnr: I'll just say "first workflow task completed or started event", I think that's enough for this purpose.

temporal/api/common/v1/message.proto
dnr: As I understand it, it doesn't matter whether it's the scheduled, started, or completed/failed/timedout event, the effect is the same for all. I'll just say "first workflow task processed by this build id"

dnr (commented):
Splitting this into a separate PR for reset, and the rest can come later

dnr:
See #333

PR #311, opened 2023-08-31
Add buffer size and dropped actions fields to ScheduleInfo

ShahabT:
**What changed?**
Adding two fields to ScheduleInfo proto

**Why?**
This is for visibility and completeness. Buffer overruns are already exposed in a metric, but we have similar fields in ScheduleInfo for number of runs skipped, etc. so I think it should go there too.

Buffer size is also interesting to monitor, but it doesn't make sense as a metric, it's more useful as a per-schedule thing to check with DescribeSchedule. I could imagine:

- a user does a large backfill with BufferAll and wants to know when they're all done
- a user runs a schedule periodically with BufferAll and wants to be alerted when there are > 5 runs backed up, but doesn't care before that

**How did you test it?**
No test

**Potential risks**
None

**Is hotfix candidate?**
No

### DIFF
--- a/temporal/api/schedule/v1/message.proto
+++ b/temporal/api/schedule/v1/message.proto
@@ -322,6 +322,12 @@ message ScheduleInfo {
     // Number of skipped actions due to overlap.
     int64 overlap_skipped = 3;
 
+    // Number of actions in the buffer.
+    int64 buffer_size = 10;
+
+    // Number of actions dropped due to the buffer limit.
+    int64 buffer_dropped = 11;
+
     // Currently-running workflows started by this schedule. (There might be
     // more than one if the overlap policy allows overlaps.)
     // Note that the run_ids in here are the original execution run ids as
--- a/temporal/api/schedule/v1/message.proto
+++ b/temporal/api/schedule/v1/message.proto
@@ -325,7 +325,9 @@ message ScheduleInfo {
     // Number of dropped actions due to buffer limit.
     int64 buffer_dropped = 10;
 
-    // Number of actions in the buffer.
+    // Number of actions in the buffer. The buffer holds the actions that cannot
+    // be immediately triggered (due to the overlap policy). These actions can be a result of
+    // the normal schedule or a backfill.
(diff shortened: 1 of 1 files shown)


cretz:
Can you add a bit more to the "why", as in specifically how users may use it?

ShahabT:
> Can you add a bit more to the "why", as in specifically how users may use it?

Great point, @cretz. This is a warm up task for me and I do not have all the context about how users benefit from this. @dnr, could you explain the use case please? (I'll add your reply to the "Why" section)

dnr:
It's basically just for visibility and completeness. Buffer overruns are already exposed in a metric, but we have similar fields in ScheduleInfo for number of runs skipped, etc. so I think it should go there too.

Buffer size is also interesting to monitor, but it doesn't make sense as a metric, it's more useful as a per-schedule thing to check with DescribeSchedule. I could imagine:
- a user does a large backfill with BufferAll and wants to know when they're all done
- a user runs a schedule periodically with BufferAll and wants to be alerted when there are > 5 runs backed up, but doesn't care before that

temporal/api/schedule/v1/message.proto
```diff
@@ -322,6 +322,12 @@ message ScheduleInfo {
     // Number of skipped actions due to overlap.
     int64 overlap_skipped = 3;
 
+    // Number of actions in the buffer.
+    int64 buffer_size = 10;
+
+    // Number of actions dropped due to the buffer limit.
+    int64 buffer_dropped = 11;
```
dnr, line 329: I'd put this above in a group with the other counters. buffer_size is a gauge and more related to running_workflows.

cretz:
Makes sense. I may request a bit more detail into what the "buffer" is on the proto comment. For example, I don't know whether buffer is used for future calendar/interval or just backlog of backfill or otherwise pending actions from the past.

dnr (approved):
both backfills and normal runs (and trigger immediately) go in the buffer together

cretz (approved):

PR #313, opened 2023-09-12
Adding namespace_id in DeleteNamespaceRequest

prathyushpv:
**What changed?**
Adding namespace_id in DeleteNamespace request.

**Why?**
DeleteNamespace API currently takes namespace name as input. Namespace name will be changed during the deletion process.
If the client wants to retry deleting the namespace again, it will have to use the new namespace name in next request.
By supporting namespace_id in the request, the client can retry DeleteNamespace with the same namespace_id again.

**Breaking changes**
None

### DIFF
--- a/temporal/api/operatorservice/v1/request_response.proto
+++ b/temporal/api/operatorservice/v1/request_response.proto
@@ -72,6 +72,7 @@ message ListSearchAttributesResponse {
 //     aip.dev/not-precedent: DeleteNamespace RPC doesn't follow Google API format. --)
 message DeleteNamespaceRequest {
     string namespace = 1;
+    string namespace_id = 2;
 }
 
 message DeleteNamespaceResponse {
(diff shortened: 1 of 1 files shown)


temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -72,6 +72,7 @@ message ListSearchAttributesResponse {
 //     aip.dev/not-precedent: DeleteNamespace RPC doesn't follow Google API format. --)
 message DeleteNamespaceRequest {
     string namespace = 1;
+    string namespace_id = 2;
```
tdeebswihart, line 75: Are both fields required for the request? What happens on a retry if both are specified but the old namespace name is sent?

prathyushpv: Only one of the fields should be set. the request will fail if both fields are specified.  Let me add that as comment here.

alexshtin (approved):

cretz (approved):
> Namespace name will be changed during the deletion process.

It is unfortunate that our internal logic is causing users to have to account for it. The better solution is probably not to change the namespace name on a user but have another mechanism to mark it deleted. But I am guessing this (like other similar things) is just too difficult to change so we'd rather burden our users.

PR #314, opened 2023-09-12
Add Nexus worker definitions

bergundy:
This PR targets the `nexus` feature branch, which will not be merged until the server implements this API to avoid going back and forth on these definitions.

Future PRs will add APIs for the Nexus service registry.

See https://github.com/nexus-rpc/api/blob/initial-http-api/SPEC.md and https://github.com/bergundy/proposals/blob/nexus-handler/nexus/go-handler.md for reference.

**What changed?**

Added API definitions for the Nexus worker API endpoints.
The server will implement the Nexus HTTP API and translate HTTP requests into Nexus tasks, which the worker can poll on.

**Why?**

To enable Temporal workers to handle incoming Nexus requests.

**Breaking changes**

Nope.


temporal/api/nexus/v1/message.proto
```diff
@@ -0,0 +1,72 @@
...
+    // A request ID that can be used as an idempotentency key.
+    string request_id = 2;
+    // Callback URL to call upon completion if the started operation is async.
+    string callback = 3;
+    // Full request body from the incoming HTTP request.
+    bytes body = 4;
+}
+
+// A request to cancel an operation.
+message CancelOperationRequest {
```
cretz, line 49: Does this need a request ID? `RequestCancelWorkflowExecutionRequest` has one. Maybe `request_id` should be on `Request` itself, to be ignored by already-idempotent calls

temporal/api/nexus/v1/message.proto
```diff
@@ -0,0 +1,72 @@
...
+option csharp_namespace = "Temporalio.Api.Nexus.V1";
+
+message HeaderValues {
+    repeated string elements = 1;
+}
+
+// A request to start an operation.
+message StartOperationRequest {
+    // Type of operation to start.
+    string operation = 1;
```
cretz, line 39: If this applies to all Nexus requests, we can lift it to the `Request` object

bergundy: I'd rather leave it on a per request basis in case we support non operation related requests in the future.
This also maps nicely to the Nexus HTTP API.

PR #315, opened 2023-09-27
Add Nexus incoming registry operator APIs

bergundy:
Added OperatorService definitions for the Nexus incoming service registry CRUD.

Cloud is likely to have a separate set of APIs with more built-in functionality such as authorization, self-hosted users will have to use provide metadata on service registration which can be used by the server's customizable authorizer.

Note that the base branch of this PR is `nexus`, this and the previous PR (#314) will be merged into `master` once implementation is done and we feel like the APIs are stable enough.


temporal/api/operatorservice/v1/service.proto
```diff
@@ -77,4 +77,23 @@ service OperatorService {
     // ListClusters returns information about Temporal clusters.
     rpc ListClusters(ListClustersRequest) returns (ListClustersResponse) {
     }
+
+    // Get a registered incoming Nexus service. The returned version can be used for optimistic updates.
+    rpc GetNexusIncomingService(GetNexusIncomingServiceRequest) returns (GetNexusIncomingServiceResponse) {
+    }
+
+    // Optimistically create or update a Nexus service based on provided version.
+    // Set version to 0 to create a new service.
```
tdeebswihart, line 86: What happens if we try to upsert an existing service with version 0? Will the request be rejected?

bergundy: Yes, I can clarify if you think it's worth it.

tdeebswihart: Nah, that’s fine by me

tdeebswihart (approved):

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -122,3 +123,59 @@ message ClusterMetadata {
     // A flag to indicate if a connection is active.
     bool is_connection_enabled = 6;
 }
+
+// A binding from a base URL to namespace, task queue, and metadata for dispatching incoming Nexus requests.
+message NexusIncomingService {
+    // Storage version for this service. Used for optimistically updating the record.
```
cretz, line 129:
```suggestion
    // Data version for this service. Must match on update or be 0 on first update. Altered by server on each update.
```

Was confused by "storage" at first

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -122,3 +123,59 @@ message ClusterMetadata {
...
+message GetNexusIncomingServiceResponse {
+    NexusIncomingService service = 1;
+}
+
+message UpsertNexusIncomingServiceRequest {
+    NexusIncomingService service = 1;
+}
+
+message UpsertNexusIncomingServiceResponse {
+}
```
cretz, line 155:
```suggestion
message UpsertNexusIncomingServiceResponse {
    NexusIncomingService service = 1;
}
```

If you're doing version-based updating, you owe it to the user to return it on each update

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -122,3 +123,59 @@ message ClusterMetadata {
...
+message GetNexusIncomingServiceRequest {
+    // Base URL of service to retrieve.
+    string base_url = 1;
+}
+
+message GetNexusIncomingServiceResponse {
+    NexusIncomingService service = 1;
+}
+
+message UpsertNexusIncomingServiceRequest {
```
cretz, line 150:
```suggestion
message UpdateNexusIncomingServiceRequest {
```

I know we use the "upsert" verb a bit in our stuff, but it's a weird one for RPCs IMO. Could also use "set" or "put". I don't feel strongly about this, so no need to change.

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -122,3 +123,59 @@ message ClusterMetadata {
     // A flag to indicate if a connection is active.
     bool is_connection_enabled = 6;
 }
+
+// A binding from a base URL to namespace, task queue, and metadata for dispatching incoming Nexus requests.
+message NexusIncomingService {
+    // Storage version for this service. Used for optimistically updating the record.
+    int64 version = 1;
+    // Base URL to match requests by.
+    string base_url = 2;
```
cretz, line 132: * Can you give more clarity on this? This is expected to be unique? Is this the full base URL, so what if this is behind a proxy, is it the caller's URL or the literal HTTP URL of the HTTP request? So I have to have DNS setup or I have to have this as an IP? Also, shouldn't we have a unique user-defined identifier that can be used to look this up instead of a may-be-changed piece of the data as the "primary key" for this?

temporal/api/operatorservice/v1/request_response.proto
bergundy: The version is part of the service message.

temporal/api/operatorservice/v1/request_response.proto
bergundy: I just don't want users wondering how to create a new service.
put and set don't feel right to me, I can expand to `CreateOrUpdate` or have a separate API to create if that feels easier to reason about.

temporal/api/operatorservice/v1/request_response.proto
bergundy: The URL is the primary key here and expected to be unique.

I was thinking that it's be either just the path if you don't care about the host or a full URL so we could match different hosts in case a single server is backing multiple APIs, e.g. `prod.myservice.com` vs. `dev.myservice.com`.

temporal/api/operatorservice/v1/request_response.proto
cretz: I should be able to chain updates or do an update after create or whatever. But right now, I have no clue what version my completed update is at. Why make me make another get call when you can just return me the current state of the object after I updated it? That's a normal thing.

temporal/api/operatorservice/v1/request_response.proto
cretz: I prefer `CreateOrUpdate` (or "set" or "put"), but I have no strong opinion

temporal/api/operatorservice/v1/request_response.proto
bergundy: You do know, because you set the next version to previous version + 1.

temporal/api/operatorservice/v1/request_response.proto
cretz: Can you spell this out on what the format is expected and how it works? I can comment on that more specifically. I would have expected just service name (and if that's part of the URL, ok).

temporal/api/operatorservice/v1/request_response.proto
bergundy: Changed.

temporal/api/operatorservice/v1/request_response.proto
bergundy: I did

temporal/api/operatorservice/v1/request_response.proto
cretz: It's normal to return back what was updated and I think you should (there may be other defaulted or cleaned up fields later). And if you expect users to know that version is + 1, it should be documented, but I would encourage not requiring the user to know that, they should just know it's unique and use it as given without assumptions about how it's updated internally.

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -122,3 +123,67 @@ message ClusterMetadata {
...
+    int64 version = 1;
+    // Base URL to match requests by.
+    // This may be just a path if to match any host or a full URL to match an exact host.
+    // The base_url is used as a unique key and can be used to get information about the service.
+    //
+    // Examples:
+    // /dev/my-service
+    // /prod/my-service/v1
+    // prod.my-service.com
+    // dev.my-service.com/v1
```
cretz, line 139: To confirm, you are expecting to use the host header? What if I have `/foo` and `my-service.com/foo` registered, who wins? What if I have `/foo` and `/foo/also-an-op-name` registered, who wins? What if I put `/foo/` with trailing slash? Why is it important to have host as part of the service description instead of just path? Why do we need to support multi-level pathing?

There are many more ambiguities and concerns I can bring up. What is the harm in accepting just a simple `name` here and requiring it be the first part of the URL path regardless of host header without all this extra confusion?

temporal/api/operatorservice/v1/request_response.proto
cretz: I'll carry on the conversation there then

temporal/api/operatorservice/v1/request_response.proto
bergundy: In Cloud, I'm fairly certain we'll need at least another level of nesting in the URLs.
For example we may have a `nexus.api/{account}/{service}`. We could change that to `{account}.nexus.api/{service}`, that can work.

I'm also thinking of users that might want to have versions in the URL, and `v1/service` is a bit nicer than forcing concatenation such as `service-v1`.

For the domain specific URLs, it's also a nice to have feature IMHO, but that can be added later, can remove for now.

I'd like to at least support a base path and figure out the rules for disambiguation.
For trailing slashes, we can trim those, the URL is always a base URL.
For prefixes, we could disallow those at registration time, or always pick the more specific prefix.
And so on...

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -124,3 +127,79 @@ message ClusterMetadata {
     // A flag to indicate if a connection is active.
     bool is_connection_enabled = 6;
 }
+
+// A binding from a base URL to namespace, task queue, and metadata for dispatching incoming Nexus requests.
+message NexusIncomingService {
```
cretz, line 132: This should probably be in the Nexus package. We don't usually have top-level messages besides request/response in the request/response file

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -124,3 +127,79 @@ message ClusterMetadata {
...
+    string namespace = 3;
+    // Task queue to route requests to.
+    string task_queue = 4;
+    // Generic service metadata that is available to the server's authorizer.
+    map<string, google.protobuf.Any> metadata = 5;
+}
+
+message GetNexusIncomingServiceRequest {
+    // Name of service to retrieve.
+    string name = 1 [
```
cretz, line 155: I don't believe this PR is where we should start using these new Google annotations. I think it can be discussed for general API use and applicable to all.

temporal/api/operatorservice/v1/request_response.proto
cretz, line 132: Why "IncomingService" instead of just "Service"?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -124,3 +127,79 @@ message ClusterMetadata {
...
+//     aip.dev/not-precedent: This resource has no parent. --)
+message CreateOrUpdateNexusIncomingServiceRequest {
+    NexusIncomingService nexus_incoming_service = 1 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+}
+
+message CreateOrUpdateNexusIncomingServiceResponse {
+    // Data version post acceptance. Can be used to issue additional updates to this record.
+    int64 version = 1;
```
cretz, line 175: Arguably for clarity you should just return the full service. Not only will it make a `PUT` reasonable for REST (if we ever do that), but you'll be happy you did when you start adding server-set fields to the service model.

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -124,3 +127,79 @@ message ClusterMetadata {
...
+    // To get the next page, pass in `ListNexusIncomingServicesResponse.next_page_token` from the previous page's
+    // response, the token will be empty if there's no other page.
+    // Note: the last page may be empty if the total number of services registered is a multiple of the page size.
+    bytes next_page_token = 2;
+}
+
+message ListNexusIncomingServicesResponse {
+    // Token for getting the next page.
+    bytes next_page_token = 1;
+    repeated string service_names = 2;
```
cretz, line 204:
```suggestion
    repeated NexusIncomingService services = 2;
```

Why hide the model and require our UI to do a GET for every one individually to list it in the table?

temporal/api/operatorservice/v1/request_response.proto
bergundy: sgtm

bergundy: There's also going to be an outgoing service which is used to configure how the history service routes nexus operations from a workflow.

temporal/api/operatorservice/v1/request_response.proto
bergundy: The linter is complaining because the `name` field has a special meaning in the google guidelines.
Would rather I disable the lint rules?

temporal/api/operatorservice/v1/request_response.proto
bergundy: SGTM

temporal/api/operatorservice/v1/request_response.proto
bergundy: Mostly because I suspect we won't need *all* of the info in the list view but let's do what you're suggesting and later if the need arises we could let the client specify the fields it needs in the request.

temporal/api/operatorservice/v1/request_response.proto
bergundy: Hmm... the Google style guide doesn't like this separation.

```
  - message: Resource type "temporal.api.nexus.v1/IncomingService" should be declared
      in the same package as it is referenced
...
https://linter.aip.dev/133/request-resource-field
```

We need to consider how closely we want to follow those aip rules.

cretz: Hrmm, how come this is an issue now and not with all other APIs we have that follow a similar pattern?

temporal/api/operatorservice/v1/request_response.proto
cretz: How come `temporal.api.common.v1.WorkflowType.name` is ok but this is not? Is it because it's a request call? Yes, I vote for disabling this rule in the rule file if it's not one we believe in.

temporal/api/operatorservice/v1/request_response.proto
cretz: :+1: Hopefully this `nexus` branch can collect these calls so we can know before merging `master`

google/api/resource.proto
cretz, line 1: Can this file be removed?

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -124,3 +125,43 @@ message ClusterMetadata {
...
+    // Name of service to retrieve.
+    string name = 1;
+}
+
+message GetNexusIncomingServiceResponse {
+    temporal.api.nexus.v1.IncomingService service = 1;
+}
+
+message CreateOrUpdateNexusIncomingServiceRequest {
+    temporal.api.nexus.v1.IncomingService nexus_incoming_service = 1;
```
cretz, line 139:
```suggestion
    temporal.api.nexus.v1.IncomingService service = 1;
```

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -124,3 +125,43 @@ message ClusterMetadata {
...
+}
+
+message CreateOrUpdateNexusIncomingServiceResponse {
+    // Data post acceptance. Can be used to issue additional updates to this record.
+    temporal.api.nexus.v1.IncomingService nexus_incoming_service = 1;
```
cretz, line 144:
```suggestion
    temporal.api.nexus.v1.IncomingService service = 1;
```

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -124,3 +125,43 @@ message ClusterMetadata {
...
+}
+
+message DeleteNexusIncomingServiceRequest {
+    // Name of service to delete.
+    string name = 1;
```
cretz, line 149:
```suggestion
    string name = 1;
    
    // If non-zero, this checks whether the version equals this before trying to delete
    int64 version = 2;
```

temporal/api/nexus/v1/message.proto
```diff
@@ -117,3 +118,22 @@ message Response {
         CancelOperationResponse cancel_operation = 2;
     }
 }
+
+// A binding from a service name to namespace, task queue, and metadata for dispatching incoming Nexus requests.
+message IncomingService {
+    // Data version for this service. Must match current version on update or be 0 on first update.
```
cretz, line 124:
```suggestion
    // Data version for this service. Must match current version on update or be 0 on create.
```

To confirm, this will never be zero when retrieving right?

google/api/resource.proto
bergundy: Yeah, I missed that one, thanks.

temporal/api/nexus/v1/message.proto
bergundy: Yup.

temporal/api/operatorservice/v1/request_response.proto
bergundy: I thought about adding this too as some users may want that for extra safety although I doubt it'll be used and can be added later.

cretz (approved):

PR #318, opened 2023-10-06
Batch operation rate limit

stephanos:
**What changed?**

Adding (optional) `max_operations_per_second` for batch operations to limit their rate.

**Why?**

To fix https://github.com/temporalio/temporal/issues/4926.

**Breaking changes**

None

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1242,15 +1242,17 @@ message StartBatchOperationRequest {
     // Namespace that contains the batch operation
     string namespace = 1;
     // Visibility query defines the the group of workflow to apply the batch operation
-    // This field and Executions are mutually exclusive
+    // This field and `executions` are mutually exclusive
     string visibility_query = 2;
     // Job ID defines the unique ID for the batch job
     string job_id = 3;
     // Reason to perform the batch operation
     string reason = 4;
     // Executions to apply the batch operation
-    // This field and VisibilityQuery are mutually exclusive
+    // This field and `visibility_query` are mutually exclusive
     repeated temporal.api.common.v1.WorkflowExecution executions = 5;
+    // RequestsPerSecond limits the number of requests per second for the batch.
+    int32 requests_per_second = 6;
     // Operation input
     oneof operation {
         temporal.api.batch.v1.BatchOperationTermination termination_operation = 10;
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1251,7 +1251,11 @@ message StartBatchOperationRequest {
     // Executions to apply the batch operation
     // This field and `visibility_query` are mutually exclusive
     repeated temporal.api.common.v1.WorkflowExecution executions = 5;
-    // Limit the number of operations per second for the batch.
+    // Optional parameter that limits the number of operations processed per second within this batch.
+    // Its purpose is to alleviate the server load arising from batch operations and
+    // minimize their impact on other workloads managed by the server's workers. 
+    // When no explicit limit is provided, the server will operate according to its default limit.
+    // If the specified limit exceeds the server's configured limit, the server's limit will take precedence.
     int32 max_operations_per_second = 6;
     // Operation input
     oneof operation {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1252,8 +1252,8 @@ message StartBatchOperationRequest {
     // This field and `visibility_query` are mutually exclusive
     repeated temporal.api.common.v1.WorkflowExecution executions = 5;
     // Limit for the number of operations processed per second within this batch.
-    // Its purpose is to alleviate the server load arising from batch operations and
-    // minimize their impact on other workloads managed by the server's workers. 
+    // Its purpose is to reduce the stress on the system caused by batch operations, which helps to prevent system
+    // overloads and minimize potential delays in executing ongoing tasks for both server workers and user workers.
     // When no explicit limit is provided, the server will operate according to its configured limit.
     // If the specified limit exceeds the server's configured limit, the server's limit will take precedence.
     int32 max_operations_per_second = 6;
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1253,7 +1253,7 @@ message StartBatchOperationRequest {
     repeated temporal.api.common.v1.WorkflowExecution executions = 5;
     // Limit for the number of operations processed per second within this batch.
     // Its purpose is to reduce the stress on the system caused by batch operations, which helps to prevent system
-    // overloads and minimize potential delays in executing ongoing tasks for both server workers and user workers.
+    // overload and minimize potential delays in executing ongoing tasks for user workers.
     // When no explicit limit is provided, the server will operate according to its configured limit.
     // If the specified limit exceeds the server's configured limit, the server's limit will take precedence.
     float max_operations_per_second = 6;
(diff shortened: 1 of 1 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1242,15 +1242,17 @@ message StartBatchOperationRequest {
...
     // Job ID defines the unique ID for the batch job
     string job_id = 3;
     // Reason to perform the batch operation
     string reason = 4;
     // Executions to apply the batch operation
-    // This field and VisibilityQuery are mutually exclusive
+    // This field and `visibility_query` are mutually exclusive
     repeated temporal.api.common.v1.WorkflowExecution executions = 5;
+    // RequestsPerSecond limits the number of requests per second for the batch.
+    int32 requests_per_second = 6;
```
stephanos, line 1255: @bergundy regarding https://github.com/temporalio/temporal/issues/4926, do you have any input on the naming for the field?

I've gotten the feedback that `requests_per_second` might not quite capture the intent. Other ideas include `rps_cap`, `max_rps`, `rps_limit` ...

bergundy, line 1255: I'd call this operations to match the `oneof` below.
```suggestion
    // Limit the number of operations per second for the batch.
    int32 operation_per_second = 6;
```

stephanos: Excellent suggestion, thank you!

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1242,15 +1242,21 @@ message StartBatchOperationRequest {
...
     string job_id = 3;
     // Reason to perform the batch operation
     string reason = 4;
     // Executions to apply the batch operation
-    // This field and VisibilityQuery are mutually exclusive
+    // This field and `visibility_query` are mutually exclusive
     repeated temporal.api.common.v1.WorkflowExecution executions = 5;
+    // Limit for the number of operations processed per second within this batch.
+    // Its purpose is to alleviate the server load arising from batch operations and
+    // minimize their impact on other workloads managed by the server's workers.
```
bergundy, line 1256: I'd also mention that it's to avoid overloading the user's workers, e.g. when a signal comes in it may produce a lot of tasks that will then need to be processed.
This may also block ongoing "normal" load from being processed in a timely fashion.

stephanos: Good point! I didn't yet realize how it would affect _both_ user and server workers.

cretz:
(just a heads up, I have intentionally not reviewed this because it is in draft, but if review wanted even in draft, happy to do so)

stephanos:
@cretz appreciate it! I was bouncing around on this because the API change process seems/seemed unclear to me. At this point I'm happy to receive opinions!

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1242,15 +1242,21 @@ message StartBatchOperationRequest {
...
     string job_id = 3;
     // Reason to perform the batch operation
     string reason = 4;
     // Executions to apply the batch operation
-    // This field and VisibilityQuery are mutually exclusive
+    // This field and `visibility_query` are mutually exclusive
     repeated temporal.api.common.v1.WorkflowExecution executions = 5;
+    // Limit for the number of operations processed per second within this batch.
+    // Its purpose is to reduce the stress on the system caused by batch operations, which helps to prevent system
+    // overloads and minimize potential delays in executing ongoing tasks for both server workers and user workers.
```
bergundy, line 1256: This shouldn't affect the server workers AFAICT.

```suggestion
    // overload and minimize potential delays in executing ongoing tasks for user workers.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1242,15 +1242,21 @@ message StartBatchOperationRequest {
...
+    // When no explicit limit is provided, the server will operate according to its configured limit.
```
bergundy, line 1257:
```suggestion
    // When no explicit limit is provided, the server will operate according to its configured default limit (TODO: what is it?).
```

temporal/api/workflowservice/v1/request_response.proto
stephanos: Oh, I naively thought it would compete for a presumably limited set of workers on the Server side (for the operations that don't go to the user workers, such as Reset or Cancel). But I have to admit, I've only assumed that.

temporal/api/workflowservice/v1/request_response.proto
stephanos: I was hesitant to put the limit in here because it might change in the Server and not be updated here. But I see how not mentioning it isn't helpful either.

Right now, there are two ways for the Server to determine a limit:
(1) the user sets `worker.batcherRPS` explicitly in the dynamic config
(2) if not set, uses the default of `50`

temporal/api/workflowservice/v1/request_response.proto
stephanos: FWIW, I'm seeing `worker.perNamespaceWorkerCount` which seems to imply there is a max number of workers per namespace.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1242,15 +1242,21 @@ message StartBatchOperationRequest {
...
+    // If the specified limit exceeds the server's configured limit, the server's limit will take precedence.
+    int32 max_operations_per_second = 6;
```
cretz, line 1259: Since this is not a float and is only expressed in terms of seconds, can you confirm users will never want less than one operation a second (e.g. one operation every 2 seconds)?

cretz: > The only tradeoff I can see is that (I assume) most users will have to provide a float here even if they have an integer in mind. The CLI could most likely hide that fact by supporting either as input, I hope. The SDK would enforce the float, of course.

Most places that accept float literals also accept integer literals. The SDK shouldn't be exposing this batching API at a high level anyways, so it'd be however that language deals with proto floats which often allows integer literals. I don't think this tradeoff really even is one...quite the opposite really in that float makes clear what is possible.

stephanos: Done! By the way, we're breaking new ground here: this is the first `float` in the entire codebase.

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
cretz: Floats are usually bad as they can't always represent exact values. Having said that, this is a rare case where approximation is acceptable.

temporal/api/workflowservice/v1/request_response.proto
stephanos: @bergundy do you think we should add (1), (2), both or neither to the comment?

temporal/api/workflowservice/v1/request_response.proto
stephanos: @bergundy given this, do you think I should leave "server workers" in the comment or not?

bergundy: I mean technically this could affect server workers to some degree in terms of CPU consumption but I don't think it's worth mentioning specifically. We already mention "system overload".

temporal/api/workflowservice/v1/request_response.proto
bergundy: If we mention it in the docs and link to those here I think it's good enough.

yiminc: The main purpose of this RPS limit is to give user control if they wish to run the batch slower than the default RPS limit.

bergundy (approved):

PR #320, opened 2023-10-19
Ensure docs can build

tdeebswihart:
**What changed?**
I vendored the protos necessary to generate docs _and_ added docs to the normal build process.

**Why?**
So we can ensure our docs build _before_ we merge to main. We'll need these for the gogo->google protobuf transition anyways.

**Breaking changes**


tdeebswihart:
I'm not happy about placing the google protos in the root but buf won't function with them elsewhere. If we wanted to do this properly we'd need to lay things out as per https://buf.build/docs/build/usage#2.-modules-and-workspaces, which won't work with the `temporal` folder at the top level.

We can fix that if we care to

buf.yaml
```diff
@@ -1,6 +1,4 @@
 version: v1
-deps:
-  - buf.build/googleapis/googleapis
```
stephanos, line 3: I saw you added a comment on this PR to explain why this was vendored; could we add this somewhere around here for future reference?

stephanos (approved):

google/api/annotations.proto
stephanos, line 1: The "googleapis" don't have a version, do they? If they do, the file path could reflect that. If they don't, then ignore my comment :)

buf.yaml
tdeebswihart: We need to remove the dep because otherwise buf will use this _and_ our vendored version and break.

google/api/annotations.proto
tdeebswihart: They don't. It's irritating

buf.yaml
tdeebswihart: Since we don't use buf to build we need to vendor it ourselves

yiminc (approved):

buf.yaml
stephanos: Sorry, what I meant to say is: can we put a comment inside of buf.yaml or the README to explain why we _vendored_ these protos? It won't be obvious to anyone without finding your comment on this PR.

stephanos: @tdeebswihart could we still add a little note to that effect somewhere?

stephanos: @tdeebswihart ping

PR #324, opened 2023-11-07
Nexus worker API

bergundy:
Recreating this after applying some of the recent lessons from the Nexus POC.

PR's base branch is the `nexus` branch has been reset back to `master`.

Old PR for reference: #314


temporal/api/nexus/v1/message.proto
```diff
@@ -0,0 +1,119 @@
...
+
+import "google/protobuf/struct.proto";
+import "temporal/api/common/v1/message.proto";
+
+// A general purpose failure message.
+// See: https://github.com/nexus-rpc/api/blob/main/SPEC.md#failure
+message Failure {
+    string message = 1;
+    map<string, string> metadata = 2;
+    google.protobuf.Struct details = 3;
```
GiantRobots, line 40: Can this have a unique ID as well? I'm happy if it's just a string and we enforce it through like `api-name.whatever-error` so we can eventually do i18n and have easy lookups as to what is erroring.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1351,3 +1352,43 @@ message PollWorkflowExecutionUpdateResponse {
...
+}
+
+message RespondNexusTaskCompletedRequest {
+    string namespace = 1;
+    // The identity of the client who initiated this request.
+    string identity = 2;
+    // A unique identifier for this task.
+    bytes task_token = 3;
+    // Embedded response to be translated into a frontend response.
+    temporal.api.nexus.v1.Response response = 4;
```
cretz, line 1377: I wouldn't object if you wanted to cut the RPC count down to just one completion call. There's no great reason I can find for the existing separation, but your choice.

```suggestion
    oneof completion {
        // Embedded response to be translated into a frontend response.
        temporal.api.nexus.v1.Response response = 4;
        // The error the handler failed with.
        temporal.api.nexus.v1.HandlerError error = 4;
    }
```

temporal/api/nexus/v1/message.proto
```diff
@@ -0,0 +1,119 @@
...
+message CancelOperationRequest {
+    // Type of operation to cancel.
+    string operation = 1;
+    // Operation ID as originally generated by a Handler.
+    string operation_id = 2;
+}
+
+// A Nexus request.
+message Request {
+    // Headers extracted from the original request in the Temporal frontend.
```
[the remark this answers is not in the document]
bergundy: This is just for the handler methods. Would you rather I call it `HandlerRequest` and `HandlerResponse`?

temporal/api/workflowservice/v1/request_response.proto
bergundy: I think it's nice to be able to get handler errors just by inspecting the rpc methods.

temporal/api/nexus/v1/message.proto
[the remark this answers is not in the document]
cretz: Maybe, but are those start and cancel and unsuccessful and such protos only for handlers too? I'd only be able to know what needs to be qualified if I could see how else this `nexus` proto package will be used.

temporal/api/workflowservice/v1/request_response.proto
cretz: But you could always say things like "I think it's nice to be able to get sync errors just by inspecting the rpc methods" or "I think it's nice to be able to get start errors separate from cancel errors". Why is it only acceptable to see async handler errors and not sync handler errors? You're ok with combining success and failure in some oneofs and not others.

What you have is fine, I'm just mostly playing devil's advocate here in that you don't have to break out RPCs to capture errors (and you're clearly not for all errors).

temporal/api/nexus/v1/message.proto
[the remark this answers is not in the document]
bergundy: You can see the `nexus-poc` branch for reference. I'm only using Request and Response types internally in the SDK layer that translates from proto to the Nexus handler API.

temporal/api/workflowservice/v1/request_response.proto
bergundy: `HandlerError` means the handler failed to handle this request while `OperationUnsuccessfulError` means that handling was successful but the operation itself wasn't.

Of course you could argue that we should capture OperationUnsuccessful as an API failure, I just don't see it this way (Nexus handling is successful, but not the upstream system that's backing the operation).

temporal/api/nexus/v1/message.proto
```diff
@@ -0,0 +1,119 @@
...
+
+import "google/protobuf/struct.proto";
+import "temporal/api/common/v1/message.proto";
+
+// A general purpose failure message.
+// See: https://github.com/nexus-rpc/api/blob/main/SPEC.md#failure
+message Failure {
+    string message = 1;
+    map<string, string> metadata = 2;
+    google.protobuf.Value details = 3;
```
bergundy, line 40: Changed to `Value` to capture any JSONable.

temporal/api/nexus/v1/message.proto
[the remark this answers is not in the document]
cretz: So to confirm, you aren't going to use the `nexus` package anywhere but workflowservice?

cretz (commented):
LGTM, but hesitant on approving as I don't want to see this merged until its dependent branches in other projects are also ready to merge

temporal/api/nexus/v1/message.proto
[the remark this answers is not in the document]
bergundy: In the PoC I had a Header proto since it supported multiple values but I no longer need this.
The `nexus` package is only going to be used for worker->frontend communication and is Temporal specific since it uses Temporal Payloads.

We may have a non-temporal Nexus grpc API that frontend implements but I can't use Temporal Payloads there so I can't reuse these definitions.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I see why you feel it's inconsistent but I think it's okay to consider operation failure successful as handler invocation which is what this API now models.

bergundy:
> LGTM, but hesitant on approving as I don't want to see this merged until its dependent branches in other projects are also ready to merge

I want to merge this into the `nexus` branch.
We can re-review before we merge into `main`.

bergundy:
I'm merging into the `nexus` feature branch. @cretz LMK if there's any further comments and I'll issue a follow up PR.

PR #325, opened 2023-11-07
Add Update lifecycle stage field to UpdateWorkflow and PollUpdate response protos

dandavison:
### What changed?
- Add a new field `stage` to `UpdateWorkflowExecutionResponse` and `PollWorkflowExecutionUpdateResponse` 

### Why?
- See https://github.com/temporalio/temporal/pull/5084
- Both these endpoints attempt to wait until a requested Update lifecycle stage is reached, but they lack a way to explicitly communicate what was learned about the Update stage during this long poll. Instead, they imply this via the presence or absence of the outcome.
- In particular, `outcome=null` means `Accepted` but not `Completed` is known to have been reached.
- In addition to this being unclear, we have found that we want to use `outcome=null` to mean something different in another context (https://github.com/temporalio/temporal/issues/4742).

### Breaking changes
- No: addition of a field only

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -668,7 +668,7 @@ message ResetWorkflowExecutionRequest {
     int64 workflow_task_finish_event_id = 4;
     // Used to de-dupe reset requests
     string request_id = 5;
-    // Reset reapplay(replay) options.
+    // Reset reapply (replay) options.
     temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 6;
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1352,7 +1352,8 @@ message PollWorkflowExecutionUpdateRequest {
     temporal.api.update.v1.UpdateRef update_ref = 2;
     // The identity of the worker/client who is polling this update outcome
     string identity = 3;
-    // Describes when this poll request should return a response
+    // Describes when this poll request should return a response.
+    // Omit to request a non-blocking poll.
     temporal.api.update.v1.WaitPolicy wait_policy = 4;
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1244,6 +1244,10 @@ message UpdateWorkflowExecutionResponse {
     // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED <
     // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED <
     // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED.
+    // UNSPECIFIED will be returned if and only if the server's maximum wait
+    // time was reached before the Update reached the stage specified in the
+    // request WaitPolicy, and before the context deadline expired; clients may
+    // may then retry the call as needed.
     temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage stage = 3;
 }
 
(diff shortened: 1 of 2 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1236,6 +1237,14 @@ message UpdateWorkflowExecutionResponse {
     // has completed. If this response is being returned before the update has
     // completed then this field will not be set.
     temporal.api.update.v1.Outcome outcome = 2;
+
+    // The most advanced lifecycle stage that the Update is known to have
+    // reached, where lifecycle stages are ordered
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_UNSPECIFIED <
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED <
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED <
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED.
```
cretz, line 1246:
```suggestion
    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED.
    // This will be UNSPECIFIED when and only when the server reached its max amount of
    // time waiting. Callers can then perform PollWorkflowExecutionUpdate as needed.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1350,4 +1359,11 @@ message PollWorkflowExecutionUpdateResponse {
     // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED) then this field will
     // not be set.
     temporal.api.update.v1.Outcome outcome = 1;
+    // The most advanced lifecycle stage that the Update is known to have
+    // reached, where lifecycle stages are ordered
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_UNSPECIFIED <
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED <
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED <
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED.
```
cretz, line 1367:
```suggestion
    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED.
    // This will be UNSPECIFIED when and only when the server reached its max amount of
    // time waiting. Callers can then call again as needed.
```

temporal/api/workflowservice/v1/request_response.proto
cretz, line 1: Can you add a small comment above poll request wait policy that a user can leave wait policy as null to perform a non-blocking call?

bergundy (commented):
Nothing to add on top of what Chad already wrote.
Note that you're blocked on the gogo refactor in the server repo for now and won't be able to use this, you might want to also open a PR targeting a branch before this refactor if you're eager to use these protos already.

temporal/api/workflowservice/v1/request_response.proto
dandavison: Yep, done. (Noting that non-blocking poll is PR-in-progress but not available yet and so in fact null wait policy will be an error right now.)

dandavison:
Ready for final review.

> Note that you're blocked on the gogo refactor in the server repo for now and won't be able to use this, you might want to also open a PR targeting a branch before this refactor if you're eager to use these protos already.

https://github.com/temporalio/api/pull/326

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1236,6 +1237,18 @@ message UpdateWorkflowExecutionResponse {
     // has completed. If this response is being returned before the update has
     // completed then this field will not be set.
     temporal.api.update.v1.Outcome outcome = 2;
+
+    // The most advanced lifecycle stage that the Update is known to have
+    // reached, where lifecycle stages are ordered
+    // UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_UNSPECIFIED <
```
Sushisource, line 1243: This is a super lame nit but `->` makes marginally more sense to me than `<`

Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
dandavison: I think I can see the argument for `->` since these are states of a state machine, but OTOH chaining `<` to describe the ordering of multiple quantities is standard syntax in math, and it was simply the mathematical ordering concept I was going for here -- giving a definition for "most advanced".

PR #327, opened 2023-11-13
Add UpdateRef to poll endpoint

dandavison:
Follow-on from https://github.com/temporalio/api/pull/325

Adds `UpdateRef` to the poll endpoint response.

**Summary of proposed server-side changes to compute this value for the response:**

- The server `PollWorkflowExecutionUpdate` API handler receives an `UpdateRef` from the client, but this is allowed to omit `RunId`.
- It then attempts to locate the Update given the information supplied by the client.
- If successful, it will construct the *returned* `UpdateRef` using the `RunId` associated with the successfully-found Update.
- Thus, the returned `RunId` may be non-empty even though the supplied `RunId` was empty.

### DIFF
--- a/temporal/api/update/v1/message.proto
+++ b/temporal/api/update/v1/message.proto
@@ -44,7 +44,7 @@ message WaitPolicy {
     temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage lifecycle_stage = 1;
 }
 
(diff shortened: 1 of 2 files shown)


cretz (approved):

alexshtin (approved):

dandavison:
Thanks @cretz @alexshtin!

PR #329, opened 2023-11-16
Create export workflow execution history proto

alice-yin:
**What changed?**
Add workflow collection proto. 

**Why?**
If user has a workflow history file that contains multiple workflows, he/she could use the following proto to deserialize:
- WorkflowCollection represents multiple workflows
- Workflow represents  1 workflow and contains all its histories. 

**Breaking changes**

### DIFF
--- a/temporal/api/export/v1/message.proto
+++ b/temporal/api/export/v1/message.proto
@@ -33,7 +33,7 @@ option csharp_namespace = "Temporalio.Api.Export.V1";
 
 import "temporal/api/history/v1/message.proto";
 
-// ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
+// WorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
 // the exported file. It encapsulates a collection of workflow execution histories.
 message WorkflowExecutionHistories {
     repeated WorkflowExecutionHistory items = 1;
--- a/temporal/api/export/v1/message.proto
+++ b/temporal/api/export/v1/message.proto
@@ -33,14 +33,13 @@ option csharp_namespace = "Temporalio.Api.Export.V1";
 
 import "temporal/api/history/v1/message.proto";
 
-
-message WorkflowExecutionHistory {
+message WorkflowExecution {
     temporal.api.history.v1.History history = 1;
 }
 
-// WorkflowExecutionHistories is used by the Cloud Export feature to deserialize 
-// the exported file. It encapsulates a collection of workflow execution histories.
-message WorkflowExecutionHistories {
-    repeated WorkflowExecutionHistory items = 1;
+// WorkflowExecutions is used by the Cloud Export feature to deserialize 
+// the exported file. It encapsulates a collection of workflow execution information.
+message WorkflowExecutions {
+    repeated WorkflowExecution items = 1;
 }
 
--- a/temporal/api/export/v1/message.proto
+++ b/temporal/api/export/v1/message.proto
@@ -33,12 +33,14 @@ option csharp_namespace = "Temporalio.Api.Export.V1";
 
 import "temporal/api/history/v1/message.proto";
 
-// WorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
+
+message WorkflowExecutionHistory {
+    temporal.api.history.v1.History history = 1;
+}
+
+// WorkflowExecutionHistories is used by the Cloud Export feature to deserialize 
 // the exported file. It encapsulates a collection of workflow execution histories.
 message WorkflowExecutionHistories {
     repeated WorkflowExecutionHistory items = 1;
 }
 
-message WorkflowExecutionHistory {
-    temporal.api.history.v1.History history = 1;
-}
--- a/temporal/api/export/v1/message.proto
+++ b/temporal/api/export/v1/message.proto
@@ -22,23 +22,23 @@
 
 syntax = "proto3";
 
-package temporal.api.workflow.v1;
+package temporal.api.export.v1;
 
-option go_package = "go.temporal.io/api/workflow/v1;workflow";
-option java_package = "io.temporal.api.workflow.v1";
+option go_package = "go.temporal.io/api/export/v1;workflow";
+option java_package = "io.temporal.api.export.v1";
 option java_multiple_files = true;
-option java_outer_classname = "WorkflowExecutionHistoryCollectionProto";
-option ruby_package = "Temporalio::Api::Workflow::V1";
-option csharp_namespace = "Temporalio.Api.Workflow.V1";
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Export::V1";
+option csharp_namespace = "Temporalio.Api.Export.V1";
 
 import "temporal/api/history/v1/message.proto";
 
-// WorkflowExecutionHistoryCollection is utilized by the Cloud Export feature to deserialize 
+// ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
 // the exported file. It encapsulates a collection of workflow execution histories.
-message WorkflowExecutionHistoryCollection {
-    repeated WorkflowExecutionHistoryCollection workflow_execution_histories = 1;
+message ExportWorkflowExecutionHistories {
+    repeated ExportWorkflowExecutionHistory export_workflow_execution_histories = 1;
 }
 
-message WorkflowExecutionHistory {
+message ExportWorkflowExecutionHistory {
     temporal.api.history.v1.History history = 1;
 }
--- a/temporal/api/export/v1/message.proto
+++ b/temporal/api/export/v1/message.proto
@@ -36,7 +36,7 @@ import "temporal/api/history/v1/message.proto";
 // ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
 // the exported file. It encapsulates a collection of workflow execution histories.
 message WorkflowExecutionHistories {
-    repeated ExportWorkflowExecutionHistory items = 1;
+    repeated WorkflowExecutionHistory items = 1;
 }
 
 message WorkflowExecutionHistory {
--- a/temporal/api/export/v1/message.proto
+++ b/temporal/api/export/v1/message.proto
@@ -35,10 +35,10 @@ import "temporal/api/history/v1/message.proto";
 
 // ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
 // the exported file. It encapsulates a collection of workflow execution histories.
-message ExportWorkflowExecutionHistories {
-    repeated ExportWorkflowExecutionHistory export_workflow_execution_histories = 1;
+message WorkflowExecutionHistories {
+    repeated ExportWorkflowExecutionHistory items = 1;
 }
 
-message ExportWorkflowExecutionHistory {
+message WorkflowExecutionHistory {
     temporal.api.history.v1.History history = 1;
 }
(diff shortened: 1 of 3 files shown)


temporal/api/workflow/v1/workflow_collection.proto
cretz, line 1: For those reading, this is in a separate file instead of `message.proto` to avoid circular dependency issues. If you have a better place for this, please suggest! (name may change once we get the message names right)

temporal/api/workflow/v1/workflow_collection.proto
```diff
@@ -0,0 +1,42 @@
...
+
+import "temporal/api/history/v1/message.proto";
+
+message WorkflowCollection {
+    repeated Workflow workflows = 1;
+}
+
+message Workflow {
+    temporal.api.history.v1.History history = 1;
+}
```
cretz, line 42: Ok, so we need names for:

* Collection of workflows
* Workflow that just has history

I propose `WorkflowExecutionHistoryCollection` and `WorkflowExecutionHistory`. Anyone else have ideas?

dandavison, line 42: Possible alternative names:
```
WorkflowHistories
WorkflowHistory
```

dandavison: Maybe:
```
WorkflowHistories
WorkflowHistory
```

yux0:
dumb question: if we just need it to decode data to proto, why do we need it in this api proto definition? would it locate in tctl or tdbg works?

cretz:
> dumb question: if we just need it to decode data to proto, why do we need it in this api proto definition? would it locate in tctl or tdbg works?

Everyone needs to be able to decode this data. This format represents the workflow cloud export format we dump onto S3 for users to use. It's not related to CLI tooling.

temporal/api/workflow/v1/workflow_collection.proto
cretz: I could also see:
```
WorkflowExecutionDetailCollection
WorkflowExecutionDetail
```
(again, playing off the `WorkflowExecutionInfo` that already exists)

alice-yin: I usually prefer short name over longer one. 

`WorkflowHistories
WorkflowHistory`

works pretty good to me.

temporal/api/workflow/v1/workflow_collection.proto
```diff
@@ -0,0 +1,42 @@
...
+option go_package = "go.temporal.io/api/workflow/v1;workflow";
+option java_package = "io.temporal.api.workflow.v1";
+option java_multiple_files = true;
+option java_outer_classname = "WorkflowCollectionProto";
+option ruby_package = "Temporalio::Api::Workflow::V1";
+option csharp_namespace = "Temporalio.Api.Workflow.V1";
+
+import "temporal/api/history/v1/message.proto";
+
+message WorkflowCollection {
```
cretz, line 36: I think you can document via a proto comment that among other potential uses, this is the type used by the Cloud Export feature

temporal/api/workflow/v1/workflow_collection.proto
cretz: Not using `WorkflowExecution` as the prefix introduces inconsistencies here. I think a "workflow" is the definition, workflow type, etc and a "workflow execution" is a single invocation. See https://docs.temporal.io/glossary.

alice-yin: I see. Makes sense. The doc mostly referred to "Workflow Execution" to use together.

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Export::V1";
+option csharp_namespace = "Temporalio.Api.Export.V1";
+
+import "temporal/api/history/v1/message.proto";
+
+// ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
+// the exported file. It encapsulates a collection of workflow execution histories.
+message ExportWorkflowExecutionHistories {
+    repeated ExportWorkflowExecutionHistory export_workflow_execution_histories = 1;
```
bergundy, line 39:
```suggestion
    repeated ExportWorkflowExecutionHistory items = 1;
```

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Export::V1";
+option csharp_namespace = "Temporalio.Api.Export.V1";
+
+import "temporal/api/history/v1/message.proto";
+
+// ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
+// the exported file. It encapsulates a collection of workflow execution histories.
+message ExportWorkflowExecutionHistories {
```
bergundy, line 38: No need to stutter here now that it's in the `export` package.
```suggestion
message WorkflowExecutionHistories {
```

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+    repeated ExportWorkflowExecutionHistory export_workflow_execution_histories = 1;
+}
+
+message ExportWorkflowExecutionHistory {
```
bergundy, line 42:
```suggestion
message WorkflowExecutionHistory {
```

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Export::V1";
+option csharp_namespace = "Temporalio.Api.Export.V1";
+
+import "temporal/api/history/v1/message.proto";
+
+// ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
+// the exported file. It encapsulates a collection of workflow execution histories.
+message WorkflowExecutionHistories {
+    repeated ExportWorkflowExecutionHistory items = 1;
```
alice-yin, line 39:
```suggestion
    repeated WorkflowExecutionHistory items = 1;
```

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+option go_package = "go.temporal.io/api/export/v1;workflow";
+option java_package = "io.temporal.api.export.v1";
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Export::V1";
+option csharp_namespace = "Temporalio.Api.Export.V1";
+
+import "temporal/api/history/v1/message.proto";
+
+// ExportWorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize
```
alice-yin, line 36:
```suggestion
// WorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
```

bergundy (approved):
LGTM, thanks!

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+
+import "temporal/api/history/v1/message.proto";
+
+// WorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize 
+// the exported file. It encapsulates a collection of workflow execution histories.
+message WorkflowExecutionHistories {
+    repeated WorkflowExecutionHistory items = 1;
+}
+
+message WorkflowExecutionHistory {
```
alexshtin, line 42: If you think about adding more field here (like `WorkflowId` or `PendingActivitiesInfo`) I would just call it `Workflow`. (Technically `ExportedWorkflow` but word "export" is already in package name).
```suggestion
message Workflow {
```

alexshtin: And then above one will be `Workflows` with `items` in it. And btw, swap these two messages in this file please.

Sushisource: I agree. If this is for export, we could name it closer to that purpose. `ExportedWorkflowExecution` and `ExportedWorkflowExecutions`. Seems highly likely to me we'll want more than just history in here.

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+option go_package = "go.temporal.io/api/export/v1;workflow";
+option java_package = "io.temporal.api.export.v1";
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Export::V1";
+option csharp_namespace = "Temporalio.Api.Export.V1";
+
+import "temporal/api/history/v1/message.proto";
+
+// WorkflowExecutionHistories is utilized by the Cloud Export feature to deserialize
```
Sushisource, line 36: Nit: Sorry, I can't help myself, this is a pet peeve of mine. [If you're curious](https://www.merriam-webster.com/grammar/is-utilize-a-word-worth-using).
```suggestion
// WorkflowExecutionHistories is used by the Cloud Export feature to deserialize 
```

temporal/api/export/v1/message.proto
```diff
@@ -0,0 +1,44 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
+
+package temporal.api.export.v1;
```
cretz, line 25: Not the biggest fan for making this an entire new tiny package, but at this point I don't have a strong enough opinion to disagree w/ other reviewers' approvals if/when they come

temporal/api/export/v1/message.proto
cretz: I don't have an opinion on whether export should be in the message name (makes sense to me), but I definitely support using keeping the words `WorkflowExecution` together and not using just `Workflow`. While you can find a few inconsistencies, in this repo we've done a decent job with specifically calling out https://docs.temporal.io/glossary#workflow-execution over the ambiguous https://docs.temporal.io/glossary#workflow.

alice-yin: So we're only export closed workflow history. In the near term, we don't have plans to add any in flight information as well. Thus, I would like to keep the word history in the name for now.

cretz: I figured the product was workflow export, not just workflow history export. It's not just in-flight things, that was an example, it's anything that can come in the future. It is very possible you may want to add something to the workflow export that is not part of this history.

alice-yin: The product is actually Workflow History Export. But I think I could make the naming more generic, in case any future usage.

temporal/api/export/v1/message.proto
alexshtin: With renames @alice-yin did bellow, having separate package name for export is important.

alexshtin (approved):

temporal/api/export/v1/message.proto
alexshtin: Even closed workflow might have pending activities. Not sure if there is any value to export them but just as example.

PR #331, opened 2023-11-20
Introduce support for error specific retry policy overrides

ast2023:
**What changed?**
error specific backoff added to RetryPolicy

**Why?**
answering OSS-361: Different retry options based on failure type

**Breaking changes**
No

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -92,6 +92,24 @@ message ActivityType {
     string name = 1;
 }
 
+// Backoff setting for a specific error. This message is designed to be used as a value
+// of error backoff overrides in the RetryPolicy message.
+message BackoffSettings {
+    // Interval of the first retry. If retryBackoffCoefficient is 1.0 then it is used for all retries.
+    google.protobuf.Duration initial_interval = 1 [(gogoproto.stdduration) = true];
+    // Coefficient used to calculate the next retry interval.
+    // The next retry interval is previous interval multiplied by the coefficient.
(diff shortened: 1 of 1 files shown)


temporal/api/common/v1/message.proto
```diff
@@ -109,6 +127,9 @@ message RetryPolicy {
     // Non-Retryable errors types. Will stop retrying if the error type matches this list. Note that
     // this is not a substring match, the error *type* (not message) must match exactly.
     repeated string non_retryable_error_types = 5;
+    // Customize backoff settings for specific errors.
+    // Similar to 'Non-Retryable', the key should precisely match the error *type*.
+    map<string, BackoffSettings> error_backoff_overrides = 6;
```
Sushisource, line 132:
```suggestion
    // Customize backoff settings for specific errors.
    // Similar to 'Non-Retryable', the key should match the error *type* 
    // in a manner appropriate for the SDK language. EX: `ErrorTypeName` for Go.
    map<string, BackoffSettings> error_backoff_overrides = 6;
```

The problem here is we really aren't "precise", at least not in every language, and it's language dependent. In Go for example non-retryable errors are specified using only the type name rather than a fully qualified type name which would be much more precise. So I'm not sure the word "precisely" belongs here, at least not in all cases.

We'll have to make sure in the SDKs that we're clear in the docstring about what is actually being matched.

cretz:
Per off-PR discussion, I think we need to discuss use cases and how we can more generically allow retry policy customization instead of specific retry policy overrides for error types which is a very specific customization.

PR #333, opened 2023-11-28
Add build-id-based reset

dnr:
**What changed?**
- Add `ResetOptions` to encapsulate options for resetting one or more workflows
- The `target` for `ResetOptions` uses a `oneof` to allow a reset type to have an argument.
- `BatchOperationReset` uses `ResetOptions`
- `ResetPointInfo` contains a `build_id` in addition to `binary_checksum` (which is now marked deprecated).

**Why?**
Help recover from a bad deployment situation

**Breaking changes**
no, the old reset options will continue to be supported

### DIFF
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -90,10 +90,16 @@ message BatchOperationDeletion {
 // BatchOperationReset sends reset requests to batch workflows.
 // Keep the parameter in sync with temporal.api.workflowservice.v1.ResetWorkflowExecutionRequest.
 message BatchOperationReset {
-  // Reset type.
-  temporal.api.enums.v1.ResetType reset_type = 1; 
-  // History event reapply options.
-  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2; 
   // The identity of the worker/client.
   string identity = 3;
+
+  // Describes what to reset (exclusive with `reset_type` and `reset_reapply_type`)
+  temporal.api.common.v1.ResetOptions options = 4;
+
+  // Deprecated (use `options`):
+  // Reset type.
+  temporal.api.enums.v1.ResetType reset_type = 1;
+  // History event reapply options.
+  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
+
 }
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -93,13 +93,12 @@ message BatchOperationReset {
   // The identity of the worker/client.
   string identity = 3;
 
-  // Describes what to reset (exclusive with `reset_type` and `reset_reapply_type`)
+  // Describes what to reset to and how (exclusive with `reset_type` and `reset_reapply_type`)
   temporal.api.common.v1.ResetOptions options = 4;
 
-  // Deprecated (use `options`):
-  // Reset type.
+  // Reset type (deprecated, use `options`).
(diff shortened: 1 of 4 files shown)


temporal/api/batch/v1/message.proto
```diff
@@ -90,10 +90,16 @@ message BatchOperationDeletion {
...
   // The identity of the worker/client.
   string identity = 3;
+
+  // Describes what to reset (exclusive with `reset_type` and `reset_reapply_type`)
+  temporal.api.common.v1.ResetOptions options = 4;
+
+  // Deprecated (use `options`):
+  // Reset type.
+  temporal.api.enums.v1.ResetType reset_type = 1;
+  // History event reapply options.
```
cretz, line 102: I assume this is also deprecated?

cretz (approved):

temporal/api/batch/v1/message.proto
dnr: Yeah, it's included in `options`. I guess I should comment each individually

temporal/api/batch/v1/message.proto
```diff
@@ -90,10 +90,15 @@ message BatchOperationDeletion {
...
   // The identity of the worker/client.
   string identity = 3;
+
+  // Describes what to reset to and how (exclusive with `reset_type` and `reset_reapply_type`)
+  temporal.api.common.v1.ResetOptions options = 4;
+
+  // Reset type (deprecated, use `options`).
+  temporal.api.enums.v1.ResetType reset_type = 1;
+  // History event reapply options (deprecated, use `options`).
+  temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
```
Sushisource, line 102: Why not use a oneof if these are exclusive?

temporal/api/common/v1/message.proto
```diff
@@ -147,3 +149,30 @@ message WorkerVersionCapabilities {
...
+message ResetOptions {
+    // Which workflow task to reset to.
+    oneof target {
+        // Resets to the first workflow task completed or started event.
+        google.protobuf.Empty first_workflow_task = 1;
+        // Resets to the last workflow task completed or started event.
+        google.protobuf.Empty last_workflow_task = 2;
+        // The id of a specific `WORKFLOW_TASK_COMPLETED`,`WORKFLOW_TASK_TIMED_OUT`, `WORKFLOW_TASK_FAILED`, or
+        // `WORKFLOW_TASK_STARTED` event to reset to.
+        int64 workflow_task_id = 3;
```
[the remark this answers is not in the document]
dnr: They're not, this is so we can use the same message for single-workflow reset in the future. I'll add a comment

temporal/api/common/v1/message.proto
```diff
@@ -147,3 +149,32 @@ message WorkerVersionCapabilities {
...
+        // continue-as-new.
+        string build_id = 4;
+    }
+
+    // History event reapply options.
+    temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 10;
+
+    // If true, limit the reset to only within the current run. (Applies to build_id targets and
+    // possibly others in the future.)
+    bool current_run_only = 11;
```
bergundy, line 179: I'm super late to this review but if this doesn't apply to all targets, maybe it would have been better to make `build_id` a message and put this in the message.
This would also make `build_id` extensible in the future.

```
message BuildId {
  string build_id = 1;
  bool current_run_only = 2;
}

...

BuildId build_id = 4;
```

PR #334, opened 2023-12-05
Allow client to suggest delay until next retry

ast2023:
**What changed?**
`retry_after` field was added to `RespondActivityTaskFailedRequest` message.

**Why?**
Activity should be able to specify retry interval.

**Breaking changes**
No.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -506,6 +506,11 @@ message RespondActivityTaskFailedRequest {
     // always be set by SDKs. Workers opting into versioning will also set the `use_versioning`
     // field to true. See message docstrings for more.
     temporal.api.common.v1.WorkerVersionStamp worker_version = 6;
+    // next_retry_delay can be used by the client to override the activity
+    // retry interval calculated by the retry policy. Retry attempts will
+    // still be subject to the maximum retries limit and total time limit
+    // defined by the policy.
+    google.protobuf.Duration next_retry_delay = 7;
 }
 
 message RespondActivityTaskFailedResponse {


temporal/api/failure/v1/message.proto
```diff
@@ -31,13 +31,16 @@ option java_outer_classname = "MessageProto";
...
+import "google/protobuf/duration.proto";
+
 import "temporal/api/common/v1/message.proto";
 import "temporal/api/enums/v1/workflow.proto";
 
 message ApplicationFailureInfo {
     string type = 1;
     bool non_retryable = 2;
     temporal.api.common.v1.Payloads details = 3;
+    google.protobuf.Duration retry_after = 4;
```
bergundy, line 43: Please document how / when this should be used.

(e.g. we'll only support this for activity retries initially).

bergundy (commented):
This LGTM, left a single comment.

temporal/api/failure/v1/message.proto
```diff
@@ -31,13 +31,20 @@ option java_outer_classname = "MessageProto";
...
 
 message ApplicationFailureInfo {
     string type = 1;
     bool non_retryable = 2;
     temporal.api.common.v1.Payloads details = 3;
+    // next_retry_delay can be used by the client to override the activity
+    // retry interval calculated by the retry policy. Retry attempts will
+    // still be subject to the maximum retries limit and total time limit
+    // defined by the policy.
+    google.protobuf.Duration next_retry_delay = 4;
```
cretz, line 47: I think this is a case where we should implement this in the server before we merge this as we may learn some things about it.

Why just delay and not override the whole retry policy? Why can an activity modify their own retry delay but not their own timeouts? We have many cases where we expect users to disable retry policy and manually handle retry if they need more complicated retry logic, are we sure this deserves special treatment? What happens when an activity start-to-close timeout happens after this is set, does it go back to regular retry? Does the count for backoff coefficient calculation still get incremented?

I could probably think of many many more questions/suggestions. And I could probably think of ways to make these kinds of things generic if this is something we want to implement. I think this is worth discussion before API change.

Sushisource: +1

I'm also not sure it belongs inside `ApplicationFailureInfo`. It may make more sense to expose this on the activity task failure request, as app failures can be thrown from workflows too.

ast2023: @cretz thank you for your insights. The following is some clarifying information: 

- The reason activity can only modify its retry delay is that it is a marketing requirement. It would not make sense for activity to modify the timeouts. I agree that the activity should be able to set its own retry policy, and to this point my proposal was exactly that. However, it was determined that the expense of development was not justified. Furthermore, this function will be provided by the Nexus project. 

- The retry attempt will still be subject to the maximum retries limit and total time limit specified by the policy. This means that attempts will be counted and that if the start-to-close timeout expired, the activity will not be retried.

Please do not hesitate to share your thoughts/suggestions.

ast2023: @Sushisource thank you for your suggestion. Would you be so kind to point me to the code inside go-sdk where the error returned by activity converted into a failure returned to the server?

cretz: > The reason activity can only modify its retry delay is that it is a marketing requirement. It would not make sense for activity to modify the timeouts.
I may need to understand this marketing requirement a bit more. If I am modifying how long to wait until next request, I sure may want to modify my schedule to close timeout.

Can you share the use case a bit more? What is the reasoning behind us specifically adding "activity setting it's own single attempt backoff from the worker" now and where will it be used? We have the server determine backoff for each attempt completeness (not every activity failure comes from a worker). Also, there are many users of ApplicationFailureInfo, activity is just one. If this is something we need to support, you should consider adding it to `RespondActivityTaskFailedRequest` (and the `ById` variant) instead of the generic failure message.

bergundy: > Why just delay and not override the whole retry policy? 

While technically not required, it's more flexible to accept the delay. But one of the alternatives we discussed is returning a whole new policy.

> Why can an activity modify their own retry delay but not their own timeouts?

I think it should be able to do that but that wouldn't be done here, as you wouldn't necessarily want to fail an activity to modify the timeout. I think it would be a separate API.

> What happens when an activity start-to-close timeout happens after this is set, does it go back to regular retry?

I would expect it to go back to the policy at schedule time.

> I'm also not sure it belongs inside ApplicationFailureInfo.

I think it belongs here because that gives control to the activity implementation.

ast2023: @cretz let's assume you are writing a workflow. As a workflow author, you need some activity to get executed. It should not be your concern in how many times activity retries, if at all, how long it takes to perform each retry, how long to wait between retries. The only thing which concerns you, will it be done by your deadline or not. Due to historical reasons, activity retry policies are installed by workflows. It is not what developers need, but it is what we have. Furthermore, policies installed by workflows might be incorrect. For example, activity suppose to post information to an HTTP server but gets back TOO_MANY_RETRIES with Retry_After. In this case, it would be better to wait for the interval specified by Retry_After and not by the retry policy.

Sushisource: > I think it belongs here because that gives control to the activity implementation.
There's no reason the activity implementation can't still control it - it could even still be attached to what looks like the app failure inside the SDK, or it could be it's own type of error, whatever - any of those things can be converted into the task failure request when actually replying to server.

The important thing is adding this to the app failure at the _protobuf_ level implies it's going to work anywhere one of these failures might appear, which isn't the case, and I think is important to avoid.

bergundy: I don't see a reason why we wouldn't want to later support this in workflows too.

Sushisource: > @Sushisource thank you for your suggestion. Would you be so kind to point me to the code inside go-sdk where the error returned by activity converted into a failure returned to the server?

https://github.com/temporalio/sdk-go/blob/76377c79cbf8030daffba5d04202da409e0f0dff/internal/internal_task_pollers.go#L1009

Sushisource: > I don't see a reason why we wouldn't want to later support this in workflows too.

Great, then have it be on the workflow task failure too. The issue I have with putting it on `Failure` is that message appears in about a billion places and already has many states that shouldn't be representable in the different situations it appears in. IE: This delay (or more detailed backoff, whatever way it goes) field wouldn't make sense in the vast majority of situations where `Failure` can appear.

ast2023: @Sushisource Allowing the workflow task to specify a retry interval will result in an inconsistent user experience. Workflow tasks should not fail by definition. Any failure produced by the task should be treated as terminal. In this context, the notion of 'produced by the task' holds significance, as we retry timeout failures. Timeouts are not seen as workflow task failures, but as internal server-sdk communication failures. I say internal communication failures because even when a workflow task fails for some reason and no response is sent back to the server, the server has no way of distinguishing this situation from one when a request to execute a workflow task was never received by the worker. I will look into moving the retry interval from the failure message into RespondActivityTaskFailed.

cretz: > let's assume you are writing a workflow

I was looking for a user use case. I was wondering what's bringing this about now.

> Due to historical reasons, activity retry policies are installed by workflows

This is not just due to historical reasons, it's also because on the server can be the arbiter of retries

> For example, activity suppose to post information to an HTTP server but gets back TOO_MANY_RETRIES with Retry_After. In this case, it would be better to wait for the interval specified by Retry_After and not by the retry policy.

We have traditionally encouraged people who need advanced retry logic to do so in the workflow. We can always make up hypotheticals. For example, what if my HTTP server gives back information saying that I have only 5 retries left? Why can't I update my max attempts? That we can derive scenarios is not usually enough.

> I think it belongs here because that gives control to the activity implementation.

If you want to give control to the activity implementation, that is done through the SDK not the API. Where the SDK then puts this on the API is a matter of internal discussion and is unrelated to whether it's on the error or not. In fact, putting it on the error gives the user _less_ flexibility because they can only update this information at error return/throw/raise time.

If we must do this, we should support the following code:

```go
policy := activity.GetInfo(ctx).RetryPolicy
policy.MaxAttempts = activity.GetInfo(ctx).Attempt + 1
activity.UpdateRetryPolicy(ctx, policy)
```

Or:

```go
activity.UpdateScheduleToCloseTimeout(activity.GetInfo(ctx).ScheduleToCloseTimeout + 10 * time.Minute)
```

> > Why can an activity modify their own retry delay but not their own timeouts?

> I think it should be able to do that but that wouldn't be done here, as you wouldn't necessarily want to fail an activity to modify the timeout. I think it would be a separate API.

Exactly, so why are we hacking in single things that we know are too inflexible in a way that we know is too inflexible? Why don't you just make a flexible way to update activity options and be consistent? Please don't make a "separate API" over and over for this stuff? (I still am not convinced you even need this or have ever needed this)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -492,6 +492,14 @@ message RespondActivityTaskCompletedByIdRequest {
 message RespondActivityTaskCompletedByIdResponse {
 }
 
+message ActivityOverride {
```
cretz, line 495: We don't usually put non-request/response messages at the top level of this file. I wonder if there is somewhere else this could be? You'll also never be able to put this object on history anywhere if it's in here. Also, I'm not sure this should be named "overrides" because it prevents us from using it for data that isn't technically an override.

Maybe `temporal.api.common.v1.UpdatedActivityAttributes` or something? No strong opinion on the exact name.

temporal/api/common/v1/message.proto
```diff
@@ -178,3 +178,14 @@ message ResetOptions {
     // possibly others in the future.)
     bool current_run_only = 11;
 }
+
+// Allow activity to influence decisions made by the server when processing this request
+// with regards to this activity only
```
Sushisource, line 183:
```suggestion
// Allow caller to influence decisions made by the server when processing this request
// with regards to this activity only
```
Since since this is also in the "by id" response, in some sense it's not the "activity" any more and we should be slightly less specific.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -529,6 +532,9 @@ message RespondActivityTaskFailedByIdRequest {
     string identity = 6;
     // Additional details to be stored as last activity heartbeat
     temporal.api.common.v1.Payloads last_heartbeat_details = 7;
+    // Allow activity to influence decisions made by the server when processing this request
+    // with regards to this activity only
```
Sushisource, line 536:
```suggestion
    // Allow caller to influence decisions made by the server when processing this request
    // with regards to this activity only
```

same reasoning

Sushisource (approved):
Makes sense to me

temporal/api/common/v1/message.proto
```diff
@@ -178,3 +178,14 @@ message ResetOptions {
...
+message ActivityRequests {
```
cretz, line 184: There is a concept of "request" in our API, so I think this is a bit confusing. Sorry to continually get so pedantic on naming, but in the API I think it matters. How about `UpdatedActivityAttributes`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -529,6 +532,9 @@ message RespondActivityTaskFailedByIdRequest {
...
+    temporal.api.common.v1.ActivityRequests activity_requests = 8;
 }
 
 message RespondActivityTaskFailedByIdResponse {
```
alexshtin, line 540: You need to add it here too.

temporal/api/common/v1/message.proto
alexshtin: `ActivityHints`!

alexshtin (approved):
After number of different attempts I saw here, and bunch of discussions, I believe this is the best way to do it.

PR #337, opened 2023-12-12
New Worker Versioning API

ShahabT:
**What changed?**
- Deprecating `UpdateWorkerBuildIdCompatibility` and `GetWorkerBuildIdCompatibility` and replacing them with `UpdateWorkerVersioningRules` and `ListWorkerVersioningRules`.
- Deprecating the concept of Version Set and "default Version Set", replacing them with more flexible `BuildIdAssignmentRule` and `CompatibleBuildIdRedirectRule` concepts.
- The assignment rules assign a Build ID to new (Unassigned) Workflow Executions and Activities.
- The redirect rules are only used in context of compatible Build IDs. These rules apply to tasks assigned to a particular Build ID (`source_build_id`) to redirect them to another compatible Build ID (`target_build_id`).

**Why?**
- Be able to support **Gradual rollout** for Canary, Blue/Green, Rolling deployments (and alike)
    - Rolling:  ability to send new and/or existing executions to the new build based on the worker count of new vs old build.
    - Blue/Green: ability to send a particular percentage of the new executions to the new build
    - Canary: ability to filter new executions based on “VersioningHint” Search Attribute set by Starter
- Version Set is not the right abstraction for the Build ID compatibility relationships, replaced it with `CompatibleBuildIdRedirectRule` which:
    - Do not loose the directionality: a redirect is from a source Build ID to a target Build ID
    - Are removable (useful in cases that the compatibility was claimed by mistake)
    - Are more powerful: redirect rules support worker-ratio-based ramps. Other ramp types, dry run mode, and filter can be added in the future.
- Making the API Incompatible-first: The primary beneficiary of WV are use cases with presumed-incompatible build IDs because WV frees them from caring about compatibility. They should be able to use WV without having to learn or think about Version Sets which currently takes the center stage of the API. Now these use cases can use `BuildIdAssignmentRule` with independent Build IDs.

**Breaking changes**
The `UpdateWorkerBuildIdCompatibility` and `GetWorkerBuildIdCompatibility` become deprecated, but supported for one version beside the new API.

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -92,12 +92,6 @@ message ActivityType {
     string name = 1;
 }
 
-message KeywordFilter {
-    // It's possible to either allow-list or block-list some hints.
-    repeated string include = 1;
-    repeated string exclude = 2;
-}
-
 // How retries ought to be handled, usable by both workflows and activities
 message RetryPolicy {
     // Interval of the first retry. If retryBackoffCoefficient is 1.0 then it is used for all retries.
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -92,6 +92,12 @@ message ActivityType {
     string name = 1;
 }
 
+message KeywordFilter {
+    // It's possible to either allow-list or block-list some hints.
+    repeated string include = 1;
+    repeated string exclude = 2;
+}
+
 // How retries ought to be handled, usable by both workflows and activities
 message RetryPolicy {
     // Interval of the first retry. If retryBackoffCoefficient is 1.0 then it is used for all retries.
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -166,9 +166,9 @@ message BuildIdReachability {
 message BuildIdAssignmentRule {
     string target_build_id = 1;
 
-    // Filter executions based on the "VersioningHint" Search Attribute set by the Starter
-    // or parent Workflow.
-    temporal.api.common.v1.KeywordFilter hint_filter = 2;
+    // Filter executions based on the "VersioningHint" Search Attribute set by
+    // the Starter or parent Workflow.
+    string hint_filter = 2;
 
     // Among the tasks that match the hint filter, only this percentage of them
     // to be sent to the target Build ID. The sampling decisions is
@@ -220,12 +220,4 @@ message CompatibleBuildIdRedirectRule {
     // This option can be used only on "terminal" Build IDs (the ones not used
     // as source in any redirect rules).
     bool contingent_on_worker_availability = 3;
-
-    // (For the future)
-    // In this mode, the Workflow Tasks will be sent to both source and target
-    // Build IDs, but the tasks sent to the target Build ID are marked as
-    // "dry run". Workers will process dry run tasks only for the purpose of
-    // verifying compatibility, without causing any mutation to the execution.
-    // Dry run redirect rules do not apply to Activity Tasks.
-    // bool dry_run_mode = 4;
 }
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -112,6 +112,14 @@ message BuildIdReachability {
     repeated TaskQueueReachability task_queue_reachability = 2;
 }
 
+message RampByPercentage {
+    // Acceptable range is [0,100].
+    float ramp_percentage = 1;
+}
+
+message RampByWorkerRatio {
+}
+
 // These rules assign a Build ID to Unassigned Workflow Executions and
 // Activities.
 //
@@ -145,14 +153,15 @@ message BuildIdReachability {
 //      Attribute. This is useful for Canary deployments, for example, when
 //      only the executions with `VersioningHint="canary"` should be sent to
 //      the new Build ID.
-//    - And/or setting a `ramp_percentage` value. This is useful for gradual
-//      shift of the traffic To the new Build ID in Blue/Green deployments and
-//      alike.
-//
-// Additionally, the rule can be contingent upon the availability of target
-// Workers. This is useful for rolling deployments, when the gradual shift of
-// traffic to the new Build ID is driven by the gradual upgrade of the Workers,
-// rather than a `ramp_percentage` value.
+//    - And/or setting a "ramp". This is useful for gradual shift of the
+//      traffic To the new Build ID. An assignment rule can have one of the
+//      following ramps:
+//      - Ramp by percentage: suitable for Blue/Green and similar deployments,
+//        when you want to send a certain percentage of traffic to the new
+//        Build ID.
+//      - Ramp by worker ratio: suitable for rolling deployments, when you want
+//        to shift more traffic to the new Build ID as the workers being
+//        gradually upgraded.
 //
 // When there are multiple assignment rules for a Task Queue, the rules are
 // evaluated in order, starting from index 0. The first applicable rule will be
@@ -170,17 +179,21 @@ message BuildIdAssignmentRule {
     // the Starter or parent Workflow.
     string hint_filter = 2;
 
-    // Among the tasks that match the hint filter, only this percentage of them
-    // to be sent to the target Build ID.
-    // The default ramp value is 100. The valid range is [0-100].
-    float ramp_percentage = 3;
-
-    // This option is intended for rolling deployments. When enabled, we try
-    // immediate dispatch to the target Build ID. If no poller is available, we
-    // skip this rule for the task at hand.
+    // If a ramp is provided, the tasks that match the hint filter will be
+    // subject to the given ramp.
     // This option can be used only on "terminal" Build IDs (the ones not used
     // as source in any redirect rules).
-    bool contingent_on_worker_availability = 4;
+    oneof ramp {
+        // This ramp is useful for gradual Blue/Green deployments (and similar)
+        // where you want to send a certain portion of the traffic to the target
+        // Build ID.
+        RampByPercentage percentage_ramp = 3;
+        // This option is intended for rolling deployments. The ramp percentage
+        // is automatically set as the ratio of the target Build ID's active
+        // workers over all active workers for any Build ID used in the
+        // assignment rules.
+        RampByWorkerRatio worker_ratio_ramp = 4;
+    }
 }
 
 // These rules apply to tasks assigned to a particular Build ID
@@ -196,8 +209,8 @@ message BuildIdAssignmentRule {
 //    newer one.
 //  - Need to hotfix some broken or stuck Workflow Executions.
 //
-// Similar to the assignment rules, redirect rules can be contingent upon the
-// availability of target Workers, which is useful for rolling deployments.
+// Similar to assignment rule, redirect rules support ramp by worker ratio.
+// They do not currently support ramp by percentage.
 //
 // In steady state, redirect rules are beneficial when dealing with old
 // Executions ran on now-decommissioned Build IDs:
@@ -206,15 +219,16 @@ message BuildIdAssignmentRule {
 //    (compatible) Build ID.
 //
 // Redirect rules can be chained, but only the last rule in the chain can have
-// the `contingent_on_worker_availability` options enabled.
+// a ramp.
 message CompatibleBuildIdRedirectRule {
     string source_build_id = 1;
     string target_build_id = 2;
 
-    // This option is intended for rolling deployments. When enabled, we try
-    // immediate dispatch to the target Build ID. If it has no available
-    // pollers, we fallback to the source Build ID.
-    // This option can be used only on "terminal" Build IDs (the ones not used
-    // as source in any redirect rules).
-    bool contingent_on_worker_availability = 3;
+    // Only the last redirect rule in a chain can have a ramp.
+    oneof ramp {
+        // This option is intended for rolling deployments. When this field is
+        // provided, we redirect a fraction of tasks proportional to the ratio
+        // of the target Build ID's active workers over the .
+        RampByWorkerRatio worker_ratio_ramp = 3;
+    }
 }
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -111,3 +111,121 @@ message BuildIdReachability {
     // Reachability per task queue.
     repeated TaskQueueReachability task_queue_reachability = 2;
 }
+
+// These rules assign a Build ID to Unassigned Workflow Executions and
+// Activities.
+//
+// Specifically, assignment rules are applied to the following Executions or
+// Activities when they are scheduled in a Task Queue:
+//    - Generally, any new Workflow Execution, except:
+//      - Child Workflows started without overriding their Task Queue or
+//        assignment status.
+//      - Continue-As-New Executions started without overriding assignment
+//        status.
+//      - Workflow Executions started Eagerly.
+//    - An Activity that is scheduled on a Task Queue different from the one
+//      their Workflow runs on, without overriding assignment status.
+//
+// In absence of (applicable) redirect rules (`CompatibleBuildIdRedirectRule`s)
+// the task will be dispatched to Workers of the Build ID determined by the
+// assignment rules. Otherwise, the final Build ID will be determined by the
+// redirect rules.
+//
+// When using Worker Versioning, in the steady state, for a given Task Queue,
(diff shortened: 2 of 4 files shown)


temporal/api/common/v1/message.proto
```diff
@@ -92,6 +92,12 @@ message ActivityType {
     string name = 1;
 }
 
+message KeywordFilter {
```
bergundy, line 95: I would strongly recommend we use search attributes for this and would also prefer if we don't introduce this in the initial version of this API to minimize the scope.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
     repeated temporal.api.taskqueue.v1.CompatibleVersionSet major_version_sets = 1;
 }
 
+
+// (-- api-linter: core::0134::request-mask-required=disabled
+//     aip.dev/not-precedent: UpdateNamespace RPC doesn't follow Google API format. --)
+// (-- api-linter: core::0134::request-resource-required=disabled
+//     aip.dev/not-precedent: GetWorkerBuildIdCompatibilityRequest RPC doesn't follow Google API format. --)
+message UpdateWorkerVersioningRulesRequest {
```
bergundy, line 1177: If you're using indices here, you'll want to protect from concurrent operations.

You should include the "current" version of the rules in this request.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+}
+
+message ListWorkerVersioningRulesRequest {
+    string namespace = 1;
+    string task_queue = 2;
+
+    // Use this option to list rules only related to particular Build IDs.
+    repeated string build_ids = 3;
+}
+message ListWorkerVersioningRulesResponse {
```
bergundy, line 1280: Return the version here so it can be used in the update API.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: yes, I was planning to add that in a future PR, but I'll add it in this one from the beginning.

temporal/api/common/v1/message.proto
ShahabT: we'll use search attributes. this is to be able to filter based on those search attributes. Given that the search attribute eliminates the extra SDK side work for this feature, I though we can add it to initial version to have a stronger message for supporting canary deployments. but I'm open to commenting it out rn and add it later.

cretz, line 95: This is a very generic message name in a very generic proto package for a very specific use. Could you consider just making it a sub-message in the only place it is used? It is a compatible move for us to extract later if needed. Alternatively could just accept a visibility query and require it only specify the SA you allow.

temporal/api/common/v1/message.proto
```diff
@@ -92,6 +92,12 @@ message ActivityType {
...
+    // It's possible to either allow-list or block-list some hints.
```
cretz, line 96: In filters that provide includes and excludes, documentation should clarify which takes precedence (or that a value cannot be in both)

temporal/api/taskqueue/v1/message.proto
```diff
@@ -111,3 +111,121 @@ message BuildIdReachability {
...
+    string target_build_id = 2;
+
+    // This option is intended for rolling deployments. When enabled, we try
+    // immediate dispatch to the target Build ID. If it has no available
+    // pollers, we fallback to the source Build ID.
+    // This option can be used only on "terminal" Build IDs (the ones not used
+    // as source in any redirect rules).
+    bool contingent_on_worker_availability = 3;
+
+    // (For the future)
```
cretz, line 224: Can we leave this out of the repo and track it elsewhere instead of commented out fields?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+message UpdateWorkerVersioningRulesRequest {
+
+    // Inserts the rule to the list of assignment rules for this Task Queue.
+    // The rules are evaluated in order, starting from index 0. The first
+    // applicable rule will be applied and the rest will be ignored.
+    message InsertBuildIdAssignmentRule {
+        // Use this option to insert the rule in a particular index. By
+        // default, the new rule is inserted at the beginning of the list
+        // (index 0). If the given index is too larger the rule will be
+        // inserted at the end of the list.
```
cretz, line 1186: If I am expected to be able to mutate a server-maintained collection by index that I may have inspected via a "list" call, I need to be able to provide a some kind of conflict key or resource version or whatever to ensure it wasn't changed in the meantime

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+    //  3. Removes any *unconditional* assignment rule for other Build IDs.
+    //  4. Removes the `contingent_on_worker_availability` flag of any incoming
+    //     redirect rules to this Build ID.
+    message CommitBuildId {
+        string target_build_id = 1;
+
+        // To prevent committing invalid Build IDs, we reject the request if no
+        // pollers has been seen recently for this Build ID. Use the `force`
+        // option to disable this validation.
+        string force = 2;
```
cretz, line 1252: Why not a bool?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+        DeleteBuildIdAssignmentRule delete_assignment_rule = 5;
+        AddCompatibleBuildIdRedirectRule insert_compatible_redirect_rule = 6;
+        ReplaceCompatibleBuildIdRedirectRule replace_compatible_redirect_rule = 7;
+        DeleteCompatibleBuildIdRedirectRule delete_compatible_redirect_rule = 8;
+        CommitBuildId commit_build_id = 9;
+    }
+}
+message UpdateWorkerVersioningRulesResponse {
+    reserved 1;
+    reserved "rule_index";
```
cretz, line 1270: Would prefer we did not commit reservations except for deprecations. Can track work elsewhere if this will be done in a future PR.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+        InsertBuildIdAssignmentRule insert_assignment_rule = 3;
+        ReplaceBuildIdAssignmentRule replace_assignment_rule = 4;
+        DeleteBuildIdAssignmentRule delete_assignment_rule = 5;
+        AddCompatibleBuildIdRedirectRule insert_compatible_redirect_rule = 6;
+        ReplaceCompatibleBuildIdRedirectRule replace_compatible_redirect_rule = 7;
+        DeleteCompatibleBuildIdRedirectRule delete_compatible_redirect_rule = 8;
+        CommitBuildId commit_build_id = 9;
+    }
+}
+message UpdateWorkerVersioningRulesResponse {
```
cretz, line 1268:
```suggestion

message UpdateWorkerVersioningRulesResponse {
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
     repeated temporal.api.taskqueue.v1.CompatibleVersionSet major_version_sets = 1;
 }
 
+
```
cretz, line 1172:
```suggestion
```

temporal/api/workflowservice/v1/request_response.proto
cretz, line 1280:
```suggestion

message ListWorkerVersioningRulesResponse {
```

temporal/api/taskqueue/v1/message.proto
```diff
@@ -111,3 +111,121 @@ message BuildIdReachability {
...
+
+    // Filter executions based on the "VersioningHint" Search Attribute set by the Starter
+    // or parent Workflow.
+    temporal.api.common.v1.KeywordFilter hint_filter = 2;
+
+    // Among the tasks that match the hint filter, only this percentage of them
+    // to be sent to the target Build ID. The sampling decisions is
+    // deterministic for a given Workflow Execution (i.e. given a fixed ramp
+    // value, the same decision will be made for a particular Workflow
+    // Execution no matter how many time it is tested against the ramp value).
```
dnr, line 177: Do we need to say this at all? Each execution is only "assigned" once, so it doesn't matter if it's deterministic or not.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+        // the delete operation will be rejected. Set `force` to true to
+        // bypass this validation. An unconditional assignment rule:
+        //   - Has no hint filter
+        //   - Has 100% ramp
+        //   - Is not contingent on worker availability
+        bool force = 2;
+    }
+
+    // Adds the rule to the list of redirect rules for this Task Queue. There
+    // can be at most one redirect rule from each Build ID.
```
dnr, line 1218:
```suggestion
    // can be at most one redirect rule for each distinct Source Build ID.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1169,6 +1169,120 @@ message GetWorkerBuildIdCompatibilityResponse {
...
+    reserved 1;
+    reserved "rule_index";
+}
+
+message ListWorkerVersioningRulesRequest {
+    string namespace = 1;
+    string task_queue = 2;
+
+    // Use this option to list rules only related to particular Build IDs.
+    repeated string build_ids = 3;
```
dnr, line 1278: Why do we need this? Filtering the result will confuse things if we're using indexes. I'd rather always return all

temporal/api/taskqueue/v1/message.proto
ShahabT: the only time that the user would care about this seems in cases of two rules with different ramps but otherwise identical. In those cases, the lower ramp will be effectively ignored (one may think that the ramps will add, or may think the second ramp applies on the remainder of the first rule)

do you think we should clarify this or it's too stretched of a situation?

dnr: I think we should do it in the way we talked about, but don't have to document it here. (This comment doesn't even describe how separate ramps interact anyway.)

temporal/api/workflowservice/v1/request_response.proto
ShahabT: added `conflict_token`

temporal/api/workflowservice/v1/request_response.proto
ShahabT: added `conflict_token`

temporal/api/workflowservice/v1/request_response.proto
ShahabT: got rid of it. We can add it later if we want to.

temporal/api/common/v1/message.proto
ShahabT: removed this altogether

temporal/api/common/v1/message.proto
ShahabT: removed this message

cretz (approved):
NOTE: My approval is purely on syntax. Would appreciate not using it as the "SDK team approve" and get another, more-versioning-familiar SDK team member to provide logical approval.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -111,3 +111,110 @@ message BuildIdReachability {
...
+//
+// In the event that no assignment rule is applicable on a task (or the Task
+// Queue is simply not versioned), the tasks will be sent to unversioned
+// workers, if available. Otherwise, they remain Unassigned, and will be
+// retried for assignment, or dispatch to unversioned workers, at a later time
+// depending on the availability of workers.
+message BuildIdAssignmentRule {
+    string target_build_id = 1;
+
+    // Filter executions based on the "VersioningHint" Search Attribute set by
```
bergundy, line 169: I thought we want to filter on any search attribute. Not on a specific one.

bergundy: Also a note against search attributes, we can use this to assign activity tasks.

bergundy: I think this feature requires more thought and I would remove it from the API entirely for now.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -531,13 +533,26 @@ service WorkflowService {
...
         };
     }
 
+    // Allows updating the Build ID assignment and redirect rules for a given Task Queue.
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: We do yet expose versioning API to HTTP. --)
+    rpc UpdateWorkerVersioningRules (UpdateWorkerVersioningRulesRequest) returns (UpdateWorkerVersioningRulesResponse) {}
+
+    // Fetches the Build ID assignment and redirect rules for a Task Queue
+    rpc ListWorkerVersioningRules (ListWorkerVersioningRulesRequest) returns (ListWorkerVersioningRulesResponse) {
```
bergundy, line 550: I wonder if `Get` is a more appropriate verb than list here.
Our other list APIs imply pagination of identifiable objects.

bergundy:
Not sure if mentioned before, please do not merge this until the feature is implemented.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -111,3 +111,124 @@ message BuildIdReachability {
...
+    oneof ramp {
+        // This ramp is useful for gradual Blue/Green deployments (and similar)
+        // where you want to send a certain portion of the traffic to the target
+        // Build ID.
+        RampByPercentage percentage_ramp = 3;
+        // This option is intended for rolling deployments. The ramp percentage
+        // is automatically set as the ratio of the target Build ID's active
+        // workers over all active workers for any Build ID used in the
+        // assignment rules.
+        RampByWorkerRatio worker_ratio_ramp = 4;
```
ShahabT, line 195: Changed this part to support a ramp based on worker count instead of the `contingent` mode. Feeding target build ID as much as it can could have undesired consequences. The ramp based on ratio of new vs old workers is better in the following ways:
    ◦ does not overwhelm new workers when they are not tuned well
    ◦ when operating with a lot of headroom, ramp % still remains close to upgrade %
    ◦ you don’t end up in a situation that more wfs are moved to the new build than it can really handle. For example, if wf pollers are plenty but activity pollers are not (which may often be the case) you’d still end up moving a lot of workflows without having capacity to process their activities.

temporal/api/workflowservice/v1/service.proto
ShahabT: I don't have an opinion. will wait to see what others have to say.

temporal/api/taskqueue/v1/message.proto
ShahabT: We'll do this at the very end. in the meanwhile we can finalize the design. the current thought is to use a pre-defined SA specific for this purpose.

For activities, maybe we can send the wf's hint to matching automatically. not sure if there is any need to set a different hint for the activity when it comes to deployment use cases.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -111,3 +111,110 @@ message BuildIdReachability {
...
+// Queue is simply not versioned), the tasks will be sent to unversioned
+// workers, if available. Otherwise, they remain Unassigned, and will be
+// retried for assignment, or dispatch to unversioned workers, at a later time
+// depending on the availability of workers.
+message BuildIdAssignmentRule {
+    string target_build_id = 1;
+
+    // Filter executions based on the "VersioningHint" Search Attribute set by
+    // the Starter or parent Workflow.
+    string hint_filter = 2;
```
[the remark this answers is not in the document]
ShahabT: the idea is to have a pre-defined SA such as "VersioningHint" that user can use. History sends only that SA to matching (if present). If we support any SA, then History has to send all the SA's to matching.

ShahabT:
API Changes merged from versioning-2 feature branch to main in #393

PR #339, opened 2023-12-19
Expose CancelRequested boolean in WorkflowExecutionInfo

carlydf:
**What changed?**
Add CancelRequested boolean to WorkflowExecutionInfo message

Because there was previously no way to see whether a cancellation was requested, even though the server already tracks that internally. You could only see whether the cancel had finished via the WorkflowExecutionStatus, but that left the case where the cancel was requested but not completed ambiguous.

Fixes https://github.com/temporalio/temporal/issues/1018

I don't think so, since it's a new field, not a removed field. If it's blank, the value will just be false.


temporal/api/workflow/v1/message.proto
```diff
@@ -57,6 +57,7 @@ message WorkflowExecutionInfo {
     int64 history_size_bytes = 15;
     // If set, the most recent worker version stamp that appeared in a workflow task completion
     temporal.api.common.v1.WorkerVersionStamp most_recent_worker_version_stamp = 16;
+    bool cancel_requested = 17;
```
yycptt, line 60: This message is also used by List workflow APIs, will we populate this field for List APIs as well from visibility records (probably need a new search attribute for that)?

cretz: Agreed ,this should come back from list as well (not sure a new search attribute is needed though, can just persist with the rest of the info I think)

carlydf: would i have to add this as a field in the `executions_visibility` table? I'm working on adding the field to the visibility record structs in the temporal repo, but honestly I'm kind of lost

alexshtin: Do we want to index it or just store for consistency? I can't find good scenario for "give me all workflows that got cancel request but haven't processed it yet".

PR #340, opened 2023-12-27
Add Nexus callback related protos

bergundy:
> NOTE: PR base branch is the `nexus` branch.

**What changed?**

Add protos relevant for Nexus callbacks and consider future support for a generic Temporal callback mechanism.

**Why?**

This is useful in so many ways and critical for backing asynchronous Nexus operations.

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -178,3 +178,18 @@ message ResetOptions {
     // possibly others in the future.)
     bool current_run_only = 11;
 }
+
+// Callback to attach to various events in the system, e.g. workflow run completion.
+message Callback {
+    message Nexus {
+        // Callback URL.
+        // (-- api-linter: core::0140::uri=disabled
+        //     aip.dev/not-precedent: Not following this guideline. --)
+        string url = 1;
+    }
+
+    reserved 1; // For a generic callback mechanism to be added later.
+    oneof variant {
+        Nexus nexus = 2;
+    }
+}
--- a/temporal/api/enums/v1/common.proto
+++ b/temporal/api/enums/v1/common.proto
@@ -59,6 +59,9 @@ enum Severity {
 // stateDiagram-v2
 //    Standby --> Ready
 //    Ready --> Scheduled
+//    Standby --> Blocked
+//    Ready --> Blocked
+//    Blocked --> Scheduled
 //    Scheduled --> Failed
 //    Scheduled --> Succeeded
 //    Failed --> BackingOff
@@ -68,18 +71,19 @@ enum CallbackState {
     CALLBACK_STATE_UNSPECIFIED = 0;
     // Callback is standing by, waiting to be triggered.
     CALLBACK_STATE_STANDBY = 1;
-    // Callback was triggered and is ready to be dispatched.
-    // Callbacks transition directly from STANDBY to SCHEDULED unless:
-    // 1. The callback is waiting for a previous callback to be dispatched.
-    // 2. The callback queue for this (source namespace, destination) is full in a given history shard.
-    // When the queue is full, callbacks can be rescheduled by an operator (TODO: mention which API when implemented).
+    // Callback was triggered and is ready and is waiting for a previous callback to be dispatched.
+    // This may happen if i.e. an Update callback completes before a Run callback for the same destination to ensure
+    // delivery order.
     CALLBACK_STATE_READY = 2;
-    // Callback is in the queue pending execution.
-    CALLBACK_STATE_SCHEDULED = 3;
+    // Callback cannot be scheduled because the queue for this source namespace and destination is full in a given history shard.
+    // Callbacks can be unblocked by an operator (TODO: mention which API when implemented).
+    CALLBACK_STATE_BLOCKED = 3;
+    // Callback is in the queue waiting to be executed or is currently executing.
+    CALLBACK_STATE_SCHEDULED = 4;
     // Callback has failed with a retryable error and is backing off before the next attempt.
-    CALLBACK_STATE_BACKING_OFF = 4;
+    CALLBACK_STATE_BACKING_OFF = 5;
     // Callback has failed.
-    CALLBACK_STATE_FAILED = 5;
+    CALLBACK_STATE_FAILED = 6;
     // Callback has succeeded.
-    CALLBACK_STATE_SUCCEEDED = 6;
+    CALLBACK_STATE_SUCCEEDED = 7;
 }
--- a/temporal/api/enums/v1/common.proto
+++ b/temporal/api/enums/v1/common.proto
@@ -54,3 +54,32 @@ enum Severity {
     SEVERITY_MEDIUM = 2;
     SEVERITY_LOW = 3;
 }
+
+// State of the callback.
+// stateDiagram-v2
+//    Standby --> Ready
+//    Ready --> Scheduled
+//    Scheduled --> Failed
+//    Scheduled --> Succeeded
+//    Failed --> BackingOff
+//    BackingOff --> Scheduled
+enum CallbackState {
+    // Default value, unspecified state.
+    CALLBACK_STATE_UNSPECIFIED = 0;
+    // Callback is standing by, waiting to be triggered.
+    CALLBACK_STATE_STANDBY = 1;
+    // Callback was triggered and is ready to be dispatched.
+    // Callbacks transition directly from STANDBY to SCHEDULED unless:
+    // 1. The callback is waiting for a previous callback to be dispatched.
+    // 2. The callback queue for this (source namespace, destination) is full in a given history shard.
+    // When the queue is full, callbacks can be rescheduled by an operator (TODO: mention which API when implemented).
+    CALLBACK_STATE_READY = 2;
+    // Callback is in the queue pending execution.
+    CALLBACK_STATE_SCHEDULED = 3;
+    // Callback has failed with a retryable error and is backing off before the next attempt.
+    CALLBACK_STATE_BACKING_OFF = 4;
+    // Callback has failed.
+    CALLBACK_STATE_FAILED = 5;
+    // Callback has succeeded.
+    CALLBACK_STATE_SUCCEEDED = 6;
+}
--- a/temporal/api/enums/v1/common.proto
+++ b/temporal/api/enums/v1/common.proto
@@ -56,16 +56,6 @@ enum Severity {
 }
 
 // State of the callback.
-// stateDiagram-v2
-//    Standby --> Ready
-//    Ready --> Scheduled
-//    Standby --> Blocked
-//    Ready --> Blocked
-//    Blocked --> Scheduled
-//    Scheduled --> Failed
-//    Scheduled --> Succeeded
-//    Failed --> BackingOff
-//    BackingOff --> Scheduled
 enum CallbackState {
     // Default value, unspecified state.
     CALLBACK_STATE_UNSPECIFIED = 0;
--- a/temporal/api/enums/v1/common.proto
+++ b/temporal/api/enums/v1/common.proto
@@ -61,19 +61,12 @@ enum CallbackState {
     CALLBACK_STATE_UNSPECIFIED = 0;
     // Callback is standing by, waiting to be triggered.
     CALLBACK_STATE_STANDBY = 1;
-    // Callback was triggered and is ready and is waiting for a previous callback to be dispatched.
-    // This may happen if i.e. an Update callback completes before a Run callback for the same destination to ensure
-    // delivery order.
-    CALLBACK_STATE_READY = 2;
-    // Callback cannot be scheduled because the queue for this source namespace and destination is full in a given history shard.
-    // Callbacks can be unblocked by an operator (TODO: mention which API when implemented).
-    CALLBACK_STATE_BLOCKED = 3;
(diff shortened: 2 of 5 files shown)


temporal/api/common/v1/message.proto
```diff
@@ -178,3 +178,18 @@ message ResetOptions {
...
+// Callback to attach to various events in the system, e.g. workflow run completion.
+message Callback {
+    message Nexus {
+        // Callback URL.
+        // (-- api-linter: core::0140::uri=disabled
+        //     aip.dev/not-precedent: Not following this guideline. --)
+        string url = 1;
+    }
+
+    reserved 1; // For a generic callback mechanism to be added later.
```
MichaelSnowden, line 191: smart

MichaelSnowden (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -852,6 +857,7 @@ message DescribeWorkflowExecutionResponse {
     repeated temporal.api.workflow.v1.PendingActivityInfo pending_activities = 3;
     repeated temporal.api.workflow.v1.PendingChildExecutionInfo pending_children = 4;
     temporal.api.workflow.v1.PendingWorkflowTaskInfo pending_workflow_task = 5;
+    repeated temporal.api.workflow.v1.CallbackInfo callbacks = 6;
```
ast2023, line 860: either CallbackInfo type is misleading or callbacks field name.

temporal/api/workflow/v1/message.proto
```diff
@@ -152,3 +153,32 @@ message NewWorkflowExecutionInfo {
...
+    // Information on how this callback should be invoked (e.g. its URL and type).
+    temporal.api.common.v1.Callback callback = 1;
+    // Trigger for this callback.
+    Trigger trigger = 2;
+    // The time when the callback was registered.
+    google.protobuf.Timestamp registration_time = 3;
+
+    temporal.api.enums.v1.CallbackState state = 4;
+    // The number of attempts made to deliver the callback.
+    // This number represents a minimum bound since the attempt is incremented after the callback request completes.
```
ast2023, line 175: I am a little confused by the comment. Will we try to deliver a callback before it was complete?

temporal/api/workflow/v1/message.proto
```diff
@@ -152,3 +153,32 @@ message NewWorkflowExecutionInfo {
...
+    message WorkflowClosed {}
+
+    message Trigger {
+        oneof variant {
+            WorkflowClosed workflow_closed = 1;
+        }
+    }
+
+    // Information on how this callback should be invoked (e.g. its URL and type).
+    temporal.api.common.v1.Callback callback = 1;
```
ast2023, line 167: I suggest we call it 'invocationMethod' - according to the comment.

temporal/api/history/v1/message.proto
```diff
@@ -103,6 +103,9 @@ message WorkflowExecutionStartedEventAttributes {
     // If this workflow intends to use anything other than the current overall default version for
     // the queue, then we include it here.
     temporal.api.common.v1.WorkerVersionStamp source_version_stamp = 29;
+
+    // Callbacks attached when this workflow was started.
+    repeated temporal.api.common.v1.Callback callbacks = 30;
```
ast2023, line 108:
```suggestion
    repeated temporal.api.common.v1.Callback attached_callbacks = 30;
```

temporal/api/workflow/v1/message.proto
```diff
@@ -152,3 +153,32 @@ message NewWorkflowExecutionInfo {
     temporal.api.common.v1.Header header = 13;
 }
 
+message CallbackInfo {
```
ast2023, line 156: Can it be just Callback?
```suggestion
message Callback {
```

bergundy: I considered this but we already have a `Callback` message (in a different package) and I would rather disambiguate.
This also follows the pattern we have in other `DescribeWorkflowExecutionResponse` fields.

temporal/api/workflow/v1/message.proto
bergundy: We're reusing this type for attaching the callback and returning information about it.

temporal/api/workflow/v1/message.proto
bergundy: We mark increment the attempt only after the request completes.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I feel like this is pretty much inline with the other fields on this message.

temporal/api/enums/v1/common.proto
```diff
@@ -54,3 +54,32 @@ enum Severity {
...
+//    Failed --> BackingOff
+//    BackingOff --> Scheduled
+enum CallbackState {
+    // Default value, unspecified state.
+    CALLBACK_STATE_UNSPECIFIED = 0;
+    // Callback is standing by, waiting to be triggered.
+    CALLBACK_STATE_STANDBY = 1;
+    // Callback was triggered and is ready to be dispatched.
+    // Callbacks transition directly from STANDBY to SCHEDULED unless:
+    // 1. The callback is waiting for a previous callback to be dispatched.
```
yycptt, line 73: if the callback needs to wait another callback to complete first before it can be dispatched, it feels like it's still in standby state to me.

temporal/api/enums/v1/common.proto
```diff
@@ -54,3 +54,32 @@ enum Severity {
...
+    // 2. The callback queue for this (source namespace, destination) is full in a given history shard.
```
yycptt, line 74: Is this the case where queue backlog is too big and task is not generated for the callback? 
Kinda feel we need a name that makes it clear that there's problem here, like Blocked or something. READY makes me think everything is working fine.

temporal/api/workflowservice/v1/request_response.proto
yycptt: +1. callbacks_info aligns better with the type I think.

yycptt (approved):
Some minor comments on the field naming.

temporal/api/enums/v1/common.proto
bergundy: That's what the comment says, it transitions to SCHEDULED **unless**...

temporal/api/workflowservice/v1/request_response.proto
bergundy: We don't call the other fields here `XXX_info`, I think this works.

temporal/api/enums/v1/common.proto
bergundy: Yeah, that makes sense.

I was thinking it's technically still in READY state but the system isn't in a state where it can process it but I like your suggestion, it gives better visibility when there's problems.

temporal/api/workflow/v1/message.proto
cretz, line 156: I think this is overthinking at this early stage. Can we decide later to expose this information? I think to start with, logs and metrics can provide callback information.

temporal/api/workflowservice/v1/request_response.proto
cretz, line 860: Not convinced this needs to be exposed at this time

temporal/api/workflow/v1/message.proto
```diff
@@ -152,3 +153,32 @@ message NewWorkflowExecutionInfo {
...
+    message Trigger {
+        oneof variant {
+            WorkflowClosed workflow_closed = 1;
+        }
+    }
+
+    // Information on how this callback should be invoked (e.g. its URL and type).
+    temporal.api.common.v1.Callback callback = 1;
+    // Trigger for this callback.
+    Trigger trigger = 2;
```
cretz, line 169: This can be left off until there is another trigger and therefore justification for its use

cretz (commented):
I think this PR is trying to predict the future too much. Can we take an extreme minimalist approach and add things later if they are really needed?

temporal/api/enums/v1/common.proto
```diff
@@ -54,3 +54,32 @@ enum Severity {
...
+
+// State of the callback.
+// stateDiagram-v2
+//    Standby --> Ready
+//    Ready --> Scheduled
+//    Scheduled --> Failed
+//    Scheduled --> Succeeded
+//    Failed --> BackingOff
+//    BackingOff --> Scheduled
+enum CallbackState {
```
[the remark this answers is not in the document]
bergundy: Users are going to want to know the state of their callbacks, just as they want to know the state of activities.

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
bergundy: I'm fine removing this until we realize we need it if it's in the describe response.

temporal/api/workflow/v1/message.proto
bergundy: I disagree, it would be extremely useful information for an operator.

temporal/api/workflow/v1/message.proto
bergundy: We will want this in the mutable state implementation, not much of a reason to omit it at this point.
We already know that an Update trigger is coming.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I think it should be exposed, just as we expose info on activities and child workflows.

temporal/api/enums/v1/common.proto
```diff
@@ -54,3 +54,32 @@ enum Severity {
...
+    // Callback is standing by, waiting to be triggered.
+    CALLBACK_STATE_STANDBY = 1;
+    // Callback was triggered and is ready to be dispatched.
+    // Callbacks transition directly from STANDBY to SCHEDULED unless:
+    // 1. The callback is waiting for a previous callback to be dispatched.
+    // 2. The callback queue for this (source namespace, destination) is full in a given history shard.
+    // When the queue is full, callbacks can be rescheduled by an operator (TODO: mention which API when implemented).
+    CALLBACK_STATE_READY = 2;
+    // Callback is in the queue pending execution.
+    CALLBACK_STATE_SCHEDULED = 3;
```
Sushisource, line 78: So this means it's about to be called, right? It seems like you might want an "executing" state for while it's actually being invoked. Either that, or we ought to be explicitly clear somewhere that callbacks have a relatively low maximum timeout for invocation

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -184,6 +184,11 @@ message StartWorkflowExecutionRequest {
     // If the workflow gets a signal before the delay, a workflow task will be dispatched and the rest
     // of the delay will be ignored.
     google.protobuf.Duration workflow_start_delay = 20;
+
+    // Callbacks to be called by the server when this workflow reaches a terminal state.
+    // If the workflow continues-as-new, these callbacks will be carried over to the new execution.
+    // Callback addresses must be whitelisted in the server's dynamic configuration.
+    repeated temporal.api.common.v1.Callback callbacks = 21;
```
Sushisource, line 191:
```suggestion
    repeated temporal.api.common.v1.Callback completion_callbacks = 21;
```

Just to make it clear outside of the comment when these happen

temporal/api/enums/v1/common.proto
bergundy: We intentionally don't have an "executing" state because we don't want to incur the cost of another DB update for every attempt.

Callbacks are more like local activities in that sense.

A callback attempt is limited to 10 seconds (configurable), it's documented in an internal document, we don't have public docs for Nexus yet.

temporal/api/workflowservice/v1/request_response.proto
bergundy: The callback is not necessarily tied to workflow completion.
There will be other uses for callbacks, e.g. update completion callbacks and possibly others.

temporal/api/enums/v1/common.proto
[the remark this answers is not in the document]
cretz: I don't think that's necessarily true any more than they need to know other internal state of the system (e.g. whether it failed sync match, etc). At the least I think we should wait until that's needed.

temporal/api/workflow/v1/message.proto
cretz: It's not whether it would be useful, it's whether it deserves to be in the API. I would like a much more minimal API

temporal/api/workflow/v1/message.proto
cretz: To confirm, our internal implementation needs affect the API?

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
cretz: It is incredibly unfortunate to see internal code behavior affect the user-facing API

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Then the comment needs to get changed, since it's only talking about workflows ending.

temporal/api/enums/v1/common.proto
Sushisource: I would just mention somewhere in these docstrings that they are default limited to take a fairly short amount of time then

PR #341, opened 2024-01-09
Add namespace level capabilities

Quinn-With-Two-Ns:
Add a message to `NamespaceInfo` to include namespace level capabilities. Different then the `capabilities` returned by `GetSystemInfoResponse` because those are scoped to a cluster where these are scoped to a namespace. I included the field in `NamespaceInfo` since that is exposed for all users already and is already been used for similar feature flags in a less general way. This is scoped to server configs, not to include cloud account level configs.

### Is this intended to work in OS and Cloud
Yes

### Who is intended to use this

Applications like UI (Not the SDKs internally)

### DIFF
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -50,8 +50,6 @@ message NamespaceInfo {
 
     // Namespace capability details. Should contain what features are enabled in a namespace.
     message Capabilities {
-        // Server to always set to true. If false the client should disregard Capabilities.
-        bool valid_capabilities = 1;
     }
 
     // Whether scheduled workflows are supported on this namespace. This is only needed
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -45,6 +45,14 @@ message NamespaceInfo {
     // A key-value map for any customized purpose.
     map<string, string> data = 5;
     string id = 6;
+    // All capabilities the namespace supports.
+    Capabilities capabilities = 7;
+
+    // Namespace capability details. Should contain what features are enabled in a namespace.
+    message Capabilities {
+        // Server to always set to true. If false the client should disregard Capabilities.
+        bool valid_capabilities = 1;
+    }
 
     // Whether scheduled workflows are supported on this namespace. This is only needed
     // temporarily while the feature is experimental, so we can give it a high tag.
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -50,6 +50,12 @@ message NamespaceInfo {
 
     // Namespace capability details. Should contain what features are enabled in a namespace.
     message Capabilities {
+        // True if the namespace supports eager workflow start.
+        bool eager_workflow_start = 1;
+        // True if the namespace supports sync update
+        bool sync_update = 2;
+        // True if the namespace supports async update
+        bool async_update = 3;
     }
 
     // Whether scheduled workflows are supported on this namespace. This is only needed


temporal/api/namespace/v1/message.proto
```diff
@@ -45,6 +45,14 @@ message NamespaceInfo {
     // A key-value map for any customized purpose.
     map<string, string> data = 5;
     string id = 6;
+    // All capabilities the namespace supports.
+    Capabilities capabilities = 7;
+
+    // Namespace capability details. Should contain what features are enabled in a namespace.
+    message Capabilities {
+        // Server to always set to true. If false the client should disregard Capabilities.
```
cretz, line 53: Can't the absence of the `capabilities` field be enough to signify this?

temporal/api/namespace/v1/message.proto
```diff
@@ -45,6 +45,14 @@ message NamespaceInfo {
     // A key-value map for any customized purpose.
     map<string, string> data = 5;
     string id = 6;
+    // All capabilities the namespace supports.
+    Capabilities capabilities = 7;
```
cretz, line 49: Can you confirm whether cloud supports `DescribeNamespace` and therefore can also add capabilities here? If not, is this intentionally OSS only?

cretz, line 49: Do all SDKs make a describe namespace call so they can get this kind of information? What if there are client capabilities? Even in the Go SDK doesn't call this until workers are created correct? I wonder if there are better ways to relay namespace-specific capabilities (maybe accepting a `namespace` on `GetSystemInfo`?)

Quinn-With-Two-Ns: Cloud supports `DescribeNamespace` . This feature should work in OSS and Cloud.

temporal/api/namespace/v1/message.proto
Quinn-With-Two-Ns: Oh yeah i'll remove this field, I mixed up how unset message values appeared.

temporal/api/namespace/v1/message.proto
Quinn-With-Two-Ns: Currently `GetSystemInfo` is not a namespace call. I think we'd need server and cloud to chime in on the scope of making it conditionally authorized by namespace.

cretz: My fear is that not all SDKs make this call, and the ones that do only do it before worker start not at client dial time (because we allow people to use another namespace on the same client connection without any network calls in between). I think we can call this `WorkerCapabilities` and feel a bit safer if whatever idea you have for putting in this field are worker specific.

Quinn-With-Two-Ns: So I was not expecting the SDKs to be calling this at client dial time or using it to make any SDK internal decisions. @bergundy you opened the issue [here](https://github.com/temporalio/features/issues/381) What was your expectation of the use of this API?

bergundy: I don't expect the SDK to invoke this call.
I expect applications to use it though.

One of the use cases I had in mind is our features repo or other test frameworks that are designed to work with arbitrary server to disable certain tests when features are not implemented by the targeted server.

The UI is another example, it can use these capabilities to as feature flags.

temporal/api/namespace/v1/message.proto
```diff
@@ -45,6 +45,12 @@ message NamespaceInfo {
...
+
+    // Namespace capability details. Should contain what features are enabled in a namespace.
+    message Capabilities {
+    }
```
bergundy, line 53: You can already add eager_workflow_start, sync_update and async_update as the first set of capabilities.

temporal/api/namespace/v1/message.proto
Quinn-With-Two-Ns: OK that was my expectation as well

cretz: :+1: If we add those initial three capabilities listed later, that works for me.

temporal/api/namespace/v1/message.proto
cretz: To confirm, it is ok if the SDK never calls this?

bergundy: Yeah, the SDK will get errors if it tries to use these APIs, the capabilities are meant to be used at the application layer.

temporal/api/namespace/v1/message.proto
cretz: Can you confirm with cloud team before merging that this is acceptable to them?

temporal/api/namespace/v1/message.proto
```diff
@@ -45,6 +45,18 @@ message NamespaceInfo {
...
+        // True if the namespace supports eager workflow start.
+        bool eager_workflow_start = 1;
+        // True if the namespace supports sync update
+        bool sync_update = 2;
+        // True if the namespace supports async update
+        bool async_update = 3;
+    }
 
     // Whether scheduled workflows are supported on this namespace. This is only needed
     // temporarily while the feature is experimental, so we can give it a high tag.
```
dnr, line 62: This one was an attempt at starting a list of capabilities. Maybe it should be moved into the new structure?

Also do we want to include versioning apis also? (we can have two fields for "v1" and "v2" since we're changing them)

Quinn-With-Two-Ns: I don't think it is worth moving it at this point

I can include versioning APIs, but the items I have in this PR are just initial features we can always add more

dnr: Okay. Looks good for now then

dnr (approved):

temporal/api/namespace/v1/message.proto
mastermanu: cloud does not censor the namespaceInfo object. Don't see a reason to censor capabilities at this point in time, so LGTM

PR #343, opened 2024-01-11
Add namespace_delete_delay option to DeleteNamespaceRequest

hehaifengcn:
**What changed?**

**Why?**
So we can override DeleteNamespaceNamespaceDeleteDelay dynamic config setting for deleting namespace info. 

**Breaking changes**

### DIFF
--- a/temporal/api/operatorservice/v1/request_response.proto
+++ b/temporal/api/operatorservice/v1/request_response.proto
@@ -75,6 +75,8 @@ message DeleteNamespaceRequest {
     // Only one of namespace or namespace_id must be specified to identify namespace.
     string namespace = 1;
     string namespace_id = 2;
+    // If provided, the deletion of namespace info will be delayed for given duration (0 means no delay).
+    // If not provided, use default delay configured in the cluster.
     google.protobuf.Duration namespace_delete_delay = 3;
 }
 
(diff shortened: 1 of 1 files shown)


temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -74,6 +75,9 @@ message DeleteNamespaceRequest {
     // Only one of namespace or namespace_id must be specified to identify namespace.
     string namespace = 1;
     string namespace_id = 2;
+    // If provided, the deletion of namespace info will be delayed for given duration (0 means no delay).
+    // If not provided, use default delay configured in the cluster.
+    google.protobuf.Duration namespace_delete_delay = 3;
```
alexshtin, line 80: It is already in `DeleteNamespaceRequest`. I would call it just `delay`.

hehaifengcn: delay may cause confusion that the whole deletion was delayed, which isn't the case (WF will be still deleted immediately).

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -74,6 +75,9 @@ message DeleteNamespaceRequest {
     // Only one of namespace or namespace_id must be specified to identify namespace.
     string namespace = 1;
     string namespace_id = 2;
+    // If provided, the deletion of namespace info will be delayed for given duration (0 means no delay).
+    // If not provided, use default delay configured in the cluster.
```
Sushisource, line 79:
```suggestion
    // If provided, the deletion of namespace info will be delayed for given duration (0 means no delay).
    // If not provided, the default delay configured in the cluster will be used.
```

Sushisource (approved):

mjameswh (approved):
LGTM

PR #344, opened 2024-01-12
Moving retry interval to ApplicationFailure per Yimins request.

ast2023:
**What changed?**
retry delay interval was moved from ActivityFailed message to ApplicationFailure 

**Why?**
Requested by Yimin

No

### DIFF
--- a/temporal/api/failure/v1/message.proto
+++ b/temporal/api/failure/v1/message.proto
@@ -34,10 +34,19 @@ option csharp_namespace = "Temporalio.Api.Failure.V1";
 import "temporal/api/common/v1/message.proto";
 import "temporal/api/enums/v1/workflow.proto";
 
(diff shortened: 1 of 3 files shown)


temporal/api/failure/v1/message.proto
```diff
@@ -1,5 +1,3 @@
-// The MIT License
-//
```
cretz, line 2: Accidentally removed?

ast2023: Thank you!

cretz (approved):
Since you're technically making a backwards incompatible change to our API, you'll have to disable the buf compatibility checker in this PR and re-enable it in another PR after you've merged this one. Right now CI will fail.

(marking approved as agreed, though I don't personally agree w/ this being on the error)

yiminc (approved):

PR #346, opened 2024-01-16
Add create_time to rules list output

ShahabT:
**What changed?**
Added create_time field for the rules returned by ListWorkerVersioningRules API.

**Why?**
Creation time can be very useful for users when they what to consider cleaning up the rules.

**Breaking changes**
no

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1280,13 +1280,23 @@ message ListWorkerVersioningRulesRequest {
 }
 
 message ListWorkerVersioningRulesResponse {
-    repeated temporal.api.taskqueue.v1.BuildIdAssignmentRule assignment_rules = 1;
-    repeated temporal.api.taskqueue.v1.CompatibleBuildIdRedirectRule compatible_redirect_rules = 2;
+    repeated TimestampedBuildIdAssignmentRule assignment_rules = 1;
+    repeated TimestampedCompatibleBuildIdRedirectRule compatible_redirect_rules = 2;
 
     // This value can be passed back to UpdateWorkerVersioningRulesRequest to
     // ensure that the rules were not modified between this List and the Update,
     // which could lead to lost updates and other confusion.
     bytes conflict_token = 3;
+}
+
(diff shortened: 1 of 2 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1285,15 +1285,25 @@ message ListWorkerVersioningRulesRequest {
...
+    repeated TimestampedBuildIdAssignmentRule assignment_rules = 1;
+    repeated TimestampedCompatibleBuildIdRedirectRule compatible_redirect_rules = 2;
 
     // This value can be passed back to UpdateWorkerVersioningRulesRequest to
     // ensure that the rules were not modified between this List and the Update,
     // which could lead to lost updates and other confusion.
     bytes conflict_token = 3;
 }
 
+message TimestampedBuildIdAssignmentRule {
```
cretz, line 1297: I don't think we should add top-level messages to this file that are not request/response messages

carlydf (approved):
looks good

dnr (approved):
Should we just put a create_time in BuildIdAssignmentRule and CompatibleBuildIdRedirectRule? It's okay if it's not filled in sometimes. Either way, this is fine too

carlydf:
> Should we just put a create_time in BuildIdAssignmentRule and CompatibleBuildIdRedirectRule? It's okay if it's not filled in sometimes. Either way, this is fine too

I also feel like the idea of adding `create_time` to the existing structs is cleaner since it will lead to fewer lines of generated code, but up to you Shahab.

ShahabT:
I feel making it a single message reduces the generated code but makes our own code dirty because we need to verify more things, like what to do if the incoming API call has set the timestamp.
Also, it's likely that we add more server-generated values down the road that is not part of the input but is part of the output.

dnr:
> I feel making it a single message reduces the generated code but makes our own code dirty because we need to verify more things, like what to do if the incoming API call has set the timestamp.

There's going to be validation anyway, but fair.

> Also, it's likely that we add more server-generated values down the road that is not part of the input but is part of the output.

In this case should the name be something other than `TimestampedX`?

PR #348, opened 2024-01-23
ResetReapplyExcludeType (preparation for Update reapply)

dandavison:
These are proposed proto changes in preparation for Update reapply, based on a design by @yycptt.

- Currently, the Reset request allows a user to send `RESET_REAPPLY_TYPE_SIGNAL` or `RESET_REAPPLY_TYPE_NONE`. The server defaults to `RESET_REAPPLY_TYPE_SIGNAL`.

- We are about to start reapplying Updates and, in the future, we may reapply other events such as Nexus-related events.

- We do not want to expand the `ResetReapplyType` enum to encode arbitrary subsets of event types to be reapplied.

- Instead, this PR adds to the Reset request the ability to send a list of event types to be _excluded_ from reapply.

- I have also added one final value to the legacy enum: `RESET_REAPPLY_TYPE_ALL_ELIGIBLE`. I propose to make this the new default for `ResetReapplyType`. In practice, today, that has the same consequences (i.e. "reapply signals"), just under a different name.

- Other than that final addition, `ResetReapplyType` now becomes deprecated.

Thus what I'm proposing is that, in the future, the server's implementation will be to:

1. Take the set of events to reapply implied by the value of `ResetReapplyType`
2. Remove from this set all `ResetReapplyExcludeType` events

The result is the set of event types that will be reapplied during Reset.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -669,8 +669,10 @@ message ResetWorkflowExecutionRequest {
     int64 workflow_task_finish_event_id = 4;
     // Used to de-dupe reset requests
     string request_id = 5;
-    // Reset reapply (replay) options.
+    // Event types to be reapplied (deprecated)
     temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 6;
+    // Event types not to be reapplied
+    repeated temporal.api.enums.v1.ResetReapplyExcludeType reset_reapply_exclude_types = 7;
 }
 
 message ResetWorkflowExecutionResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -669,10 +669,11 @@ message ResetWorkflowExecutionRequest {
     int64 workflow_task_finish_event_id = 4;
     // Used to de-dupe reset requests
     string request_id = 5;
-    // Event type to be reapplied (deprecated; may not be sent together with reset_reapply_operations)
+    // Event types to be reapplied (deprecated)
+    // Default: RESET_REAPPLY_TYPE_ALL_ELIGIBLE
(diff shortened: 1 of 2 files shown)


bergundy (commented):
Why use exclusion in the api instead of inclusion with a default to apply all?

dandavison:
> Why use exclusion in the api instead of inclusion with a default to apply all?
  
Are you suggesting changing the existing field to be `repeated`, but having absence mean ALL rather than NONE? I think that sounds good.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -669,8 +669,10 @@ message ResetWorkflowExecutionRequest {
     int64 workflow_task_finish_event_id = 4;
     // Used to de-dupe reset requests
     string request_id = 5;
-    // Reset reapply (replay) options.
+    // Event types to be reapplied (deprecated)
     temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 6;
+    // Event types not to be reapplied
+    repeated temporal.api.enums.v1.ResetReapplyExcludeType reset_reapply_exclude_types = 7;
```
cretz, line 675:
```suggestion
    repeated temporal.api.enums.v1.ResetReapplyType reset_reapply_types = 7;
```

One wonders if this makes more sense, and it can be validated as mutually exclusive with the singular one. I fear we're going to later regret not letting you choose the types to reapply.

dandavison:
> One wonders if this makes more sense, and it can be validated as mutually exclusive with the singular one. I fear we're going to later regret not letting you choose the types to reapply.

> Why use exclusion in the api instead of inclusion with a default to apply all?

I agree that a pure inclusion-based API would be preferable. I'll push a second commit along those lines shortly.

cretz (approved):
After discussion, due to apply-all being default, exclusion list makes more sense.

PR #351, opened 2024-01-28
New event: update requested

dandavison:
Add new event `EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_REQUESTED` for update reapply.

As explained in comments, this event will initially be restricted to updates reapplied during reset or replication. In the future we plan to use it for updates that are durable on admission.

### DIFF
--- a/temporal/api/enums/v1/update.proto
+++ b/temporal/api/enums/v1/update.proto
@@ -55,9 +55,8 @@ enum UpdateWorkflowExecutionLifecycleStage {
     UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED = 3;
 }
 
-// UpdateRequestedEventOrigin records why an
-// WorkflowExecutionUpdateRequestedEvent was written to history. Note that not
-// all update requests result in a WorkflowExecutionUpdateRequestedEvent.
+// Records why a WorkflowExecutionUpdateRequestedEvent was written to history.
+// Note that not all update requests result in this event.
 enum UpdateRequestedEventOrigin {
     UPDATE_REQUESTED_EVENT_ORIGIN_UNSPECIFIED = 0;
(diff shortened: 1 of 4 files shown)


cretz (approved):
This looks great to me and I hope in the near future, `UpdateRequestedEventOrigin` gets a `UPDATE_REQUESTED_EVENT_ORIGIN_USER_REQUESTED` or something to support admitted or general durability if the user wants it. To clarify, the SDKs should just ignore this event correct? If so, we'll want to remember server-side to set `worker_may_ignore` as `true` for this (though that's often more for backwards compat).

temporal/api/enums/v1/update.proto
```diff
@@ -54,3 +54,14 @@ enum UpdateWorkflowExecutionLifecycleStage {
     // on a worker and has either been rejected or returned a value or an error.
     UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED = 3;
 }
+
+// UpdateRequestedEventOrigin records why an
+// WorkflowExecutionUpdateRequestedEvent was written to history. Note that not
+// all update requests result in a WorkflowExecutionUpdateRequestedEvent.
```
Sushisource, line 60:
```suggestion
// Records why a
// WorkflowExecutionUpdateRequestedEvent was written to history. Note that not
// all update requests result in a WorkflowExecutionUpdateRequestedEvent.
```

You're free of Go! You don't have to start the comment with the message name! 

Lol, doesn't really matter, that rule just always drives me nuts :laughing: . The an->a change is a real typo though.

Sushisource (approved):

temporal/api/enums/v1/update.proto
dandavison: :) fixed.

yycptt (approved):

PR #353, opened 2024-02-06
Generate OpenAPI documentation

tdeebswihart:
**What changed?**

We now generate openapi [v2 (swagger)](https://swagger.io/specification/v2/) and [v3](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md) specs as a part of proto compilation.

**Why?**

So that our customers can understand how to use our HTTP API

**How did you test it?**

Build and run https://github.com/temporalio/temporal/pull/5393 (make start-dependencies + make start)
```
$ curl localhost:7243/api/v1/swagger.json
$ curl localhost:7243/api/v1/openapi.yaml
```


openapi/openapi.yaml
cretz, line 1: Not that it matters that much, but I wonder if the filenames should be a bit more consistent, e.g. `openapiv3.yaml` and `openapiv2.json` instead of their defaults

Makefile
```diff
@@ -46,23 +46,43 @@ grpc: buf-lint api-linter buf-breaking clean go-grpc fix-path
...
 grpc-install:
 	@printf $(COLOR) "Install/update protoc and plugins..."
+	@go install go.temporal.io/api/cmd/protogen@latest
 	@go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 	@go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 	@go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway@latest
 	@go install github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc@latest
+	@go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@latest
+	@go install github.com/google/gnostic/cmd/protoc-gen-openapi@latest
+	@go install github.com/mikefarah/yq/v4@latest
```
cretz, line 85: Not necessarily in this PR, but I wonder if we want to use fixed versions here (or in a go.mod somewhere). I fear a newer OpenAPI gen will make slight alterations to the output that will cause CI to fail for someone that isn't even working on these. But not blocking or anything.

cretz (approved):
LGTM. Only minor note worth looking at is about filenames

Makefile
tdeebswihart: I sure hope not. If we do we should discuss whether we want to pin the version of our protoc plugins as well. We already do for our server repo

cretz: I could see some slight `protoc-gen-openapi` improvement making some tiny inconsequential output change resulting in CI break. But I am not worried (we can cross that road when we get there), but yes I do think we've learned we prefer fixed versions even if we're not the most mindful of upgrading.

stephanos (approved):

PR #354, opened 2024-02-07
Replace buildkite with GHA

tdeebswihart:
**What changed?**

Our Buildkite CI config was replaced with GitHub actions

**Why?**

We're trying to deprecate buildkite in general; the server repo has already done so

**Breaking changes**

None.


.buildkite/pipeline.yml
```diff
@@ -1,10 +0,0 @@
-steps:
-  - label: "build"
-    agents:
-      queue: "default"
-      docker: "*"
-    command: "make ci-build"
-    plugins:
-      - docker-compose#v3.1.0:
```
Quinn-With-Two-Ns, line 8: Am I missing something or did we remove the docker and docker compose file? Was it not needed?

Quinn-With-Two-Ns (approved):

PR #357, opened 2024-02-21
Adding support for as-is for scheduled workflow id

ast2023:
**What changed?**
Added support for as-is for scheduled workflow ID.

**Why?**
Users asked for the ability to run scheduled workflow without adding a timestamp to the scheduled workflow ID.

**Breaking changes**
No

### DIFF
--- a/temporal/api/schedule/v1/message.proto
+++ b/temporal/api/schedule/v1/message.proto
@@ -243,6 +243,10 @@ message SchedulePolicies {
     // This applies after retry policies: the full chain of retries must fail to
     // trigger a pause here.
     bool pause_on_failure = 3;
+
+    // If true, and the action would start a workflow, a timestamp will not be
+    // appended to the scheduled workflow id.
+    bool keep_original_workflow_id = 4;
 }
 
 message ScheduleAction {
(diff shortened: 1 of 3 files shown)


temporal/api/schedule/v1/message.proto
```diff
@@ -254,6 +254,8 @@ message ScheduleAction {
         // it may have a timestamp appended for uniqueness.
         temporal.api.workflow.v1.NewWorkflowExecutionInfo start_workflow = 1;
     }
+    // If true, schedule will not append timestamp to the scheduled workflow id.
```
Sushisource, line 257:
```suggestion
    // If true, and the action would start a workflow, a timestamp will not be appended to the scheduled workflow id.
```

Sushisource (approved):

temporal/api/schedule/v1/message.proto
```diff
@@ -254,6 +254,8 @@ message ScheduleAction {
...
+    bool keep_original_workflow_id = 2;
```
dnr, line 258: Maybe this should go in SchedulePolicies instead? It is sort of a property of the action, but only start_workflow action, not necessarily other future ones. (Similarly, overlap_policy may not apply to actions besides start_workflow either.)

ast2023: moved

dnr (approved):

PR #358, opened 2024-02-26
Clean up versioning protos with the narrowed scope

ShahabT:
**What changed?**
- Removed the hint filter and worker count based ramp from the Versioning protos for now.
- Renamed `use_compatible_version` flag to `use_workflow_build_id` for activities and to `inherit_build_id` for child and CaN executions. The new names capture the essence of the flag better because they are about whether to use the same build ID or to choose a fresh one based on the assignment rules.
- Removed `bundle_id` from the stamp as it is not used in any SDK or in the server. The user's view is polluted right now with this field. We'll add it later when there is a complete design for how to use it.

**Why?**

**Breaking changes**

### DIFF
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -81,11 +81,8 @@ message ScheduleActivityTaskCommandAttributes {
     // Request to start the activity directly bypassing matching service and worker polling
     // The slot for executing the activity should be reserved when setting this field to true.
     bool request_eager_execution = 12;
-    // Only applicable for activities executing on a Task Queue different from their workflow's.
     // If this is set, the activity would be assigned to the Build ID of the workflow. Otherwise,
     // Assignment rules of the activity's Task Queue will be used to determine the Build ID.
-    // Activities that run on their workflow's Task Queue always use workflow's Build ID.
-    // All activities are subject to redirect rules of their Task Queue.
     bool use_workflow_build_id = 13;
 }
 
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -81,10 +81,12 @@ message ScheduleActivityTaskCommandAttributes {
     // Request to start the activity directly bypassing matching service and worker polling
     // The slot for executing the activity should be reserved when setting this field to true.
     bool request_eager_execution = 12;
-    // If this is set, the workflow executing this command wishes to start the activity using
-    // a version compatible with the version that this workflow most recently ran on, if such
-    // behavior is possible.
-    bool use_compatible_version = 13;
+    // Only applicable for activities executing on a Task Queue different from their workflow's.
+    // If this is set, the activity would be assigned to the Build ID of the workflow. Otherwise,
+    // Assignment rules of the activity's Task Queue will be used to determine the Build ID.
+    // Activities that run on their workflow's Task Queue always use workflow's Build ID.
+    // All activities are subject to redirect rules of their Task Queue.
+    bool use_workflow_build_id = 13;
 }
 
 message RequestCancelActivityTaskCommandAttributes {
@@ -193,9 +195,9 @@ message ContinueAsNewWorkflowExecutionCommandAttributes {
     temporal.api.common.v1.Header header = 12;
     temporal.api.common.v1.Memo memo = 13;
     temporal.api.common.v1.SearchAttributes search_attributes = 14;
-    // If this is set, the workflow executing this command wishes to continue as new using a version
-    // compatible with the version that this workflow most recently ran on.
-    bool use_compatible_version = 15;
+    // If this is set, the new execution inherits the Build ID of the current execution. Otherwise,
+    // the assignment rules will be used to independently assign a Build ID to the new execution.
+    bool inherit_build_id = 15;
 
     // `workflow_execution_timeout` is omitted as it shouldn't be overridden from within a workflow.
 }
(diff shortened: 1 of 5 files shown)


cretz:
To save me the trouble of checking, can you confirm all alterations are to non-main pieces of the API (i.e. just the staged stuff being worked on)? Or even better, I guess we can wait until the PR for `versioning-2` -> `main` is opened.

temporal/api/command/v1/message.proto
```diff
@@ -81,10 +81,12 @@ message ScheduleActivityTaskCommandAttributes {
...
-    // If this is set, the workflow executing this command wishes to start the activity using
-    // a version compatible with the version that this workflow most recently ran on, if such
-    // behavior is possible.
-    bool use_compatible_version = 13;
+    // Only applicable for activities executing on a Task Queue different from their workflow's.
+    // If this is set, the activity would be assigned to the Build ID of the workflow. Otherwise,
+    // Assignment rules of the activity's Task Queue will be used to determine the Build ID.
+    // Activities that run on their workflow's Task Queue always use workflow's Build ID.
+    // All activities are subject to redirect rules of their Task Queue.
+    bool use_workflow_build_id = 13;
```
Sushisource, line 89: This (comment) does away with the possibility of running an activity on the same queue but with whatever id would be assigned to it by default. Do we have to get rid of that? By default, certainly, it should use the workflow's Build ID - but a user should still be able to have it use the latest and run on the same queue if they want

ShahabT:
> To save me the trouble of checking, can you confirm all alterations are to non-main pieces of the API (i.e. just the staged stuff being worked on)? Or even better, I guess we can wait until the PR for `versioning-2` -> `main` is opened.

@cretz, that is not the case actually. I renamed the `use_compatible_version` flag which is used in current API. My thinking was it's just a wire-compatible rename and saves us later cleanup work. But now I'm not sure if we want to do that way if it means all the SDKs will fail to compile after versioning-2 is merged to main. If you think that's gonna be a problem, I can hold off on renaming `use_compatible_version` for now. cc @Sushisource

temporal/api/command/v1/message.proto
ShahabT: Hmm... I thought about it and couldn't think of a user why users want to do that, because the old worker has to be around until the workflow itself is done even thought the activities might be running on newer build ID.

But now you mentioned it, I can see a little value on running the activity always on latest so it can benefit from latest bug-fixes etc... although thinking that way hurts my mind because now the activity output has to be backwards compatible always which is a very easy thing to miss if the activity is just a function in the same code (which has to be because the task queue is the same).

All that said, I'm fine with getting rid of the restriction assuming SDK will make sure to send the right default value (true) for same-tq activities.

Sushisource: I would get rid of the restriction yeah. The main reason you got at exactly:

> running the activity always on latest so it can benefit from latest bug-fixes 

And in such situations, it's expected your output is staying wire-compatible.

Sushisource:
> > To save me the trouble of checking, can you confirm all alterations are to non-main pieces of the API (i.e. just the staged stuff being worked on)? Or even better, I guess we can wait until the PR for `versioning-2` -> `main` is opened.
> 
> @cretz, that is not the case actually. I renamed the `use_compatible_version` flag which is used in current API. My thinking was it's just a wire-compatible rename and saves us later cleanup work. But now I'm not sure if we want to do that way if it means all the SDKs will fail to compile after versioning-2 is merged to main. If you think that's gonna be a problem, I can hold off on renaming `use_compatible_version` for now. cc @Sushisource

I'm fine with renaming it now. It is going to cause SDKs to have to deal with the compile changes but that's not exactly a huge problem.

temporal/api/command/v1/message.proto
ShahabT: Sounds good, just removed the restriction.

> And in such situations, it's expected your output is staying wire-compatible.
This is definitely true for bug fixes. But the problem is if you want to benefit it for bug fixes, you have to use this mode all the time because you don't know which build will have a bug beforehand. So you need to make sure that ANY change that is made to the activity remains wire-compatible.

Sushisource (approved):

ShahabT:
> > @cretz, that is not the case actually. I renamed the use_compatible_version flag which is used in current API. My thinking was it's just a wire-compatible rename and saves us later cleanup work. But now I'm not sure if we want to do that way if it means all the SDKs will fail to compile after versioning-2 is merged to main. If you think that's gonna be a problem, I can hold off on renaming use_compatible_version for now. cc @Sushisource
> > > I'm fine with renaming it now. It is going to cause SDKs to have to deal with the compile changes but that's not exactly a huge problem.
> 
> Unless it basically means the same thing, I think we should consider using a different field number too. Assuming we want to remove and not deprecate, removing the old field makes sense (will cause compile error), but silently accepting old bool as new bool is a bit rough. I personally would prefer if the old field remained present but no longer supported and would fail server side if set, but I understand we don't have a deprecated-field-removal policy. Proto will happily accept unknown fields which is why many times invalid fields remain and fail at runtime.

The field is basically the same thing, just the new naming is more consistent with the new API concepts. Given you guys' feedback it seems best to go with the renaming right now and quickly fix the compile errors in the SDKs when this is merged in main.

cretz (approved):

cretz:
Works for me since it's basically the same thing

PR #359, opened 2024-02-29
Add WorkflowIdConflictPolicy

stephanos:
**What changed?**

Introducing the `WorkflowIdConflictPolicy` enum to define the dedup behavior of Start Workflow for a _running_ workflow - and solidifying the `WorkflowIdReusePolicy` as the dedup behavior for _closed_ workflows.

_NOTE: Terminate-If-Running will be deprecated once all API, Server and SDK changes have been merged since otherwise users wouldn't be able to use the new enum but see a deprecation notice already._

Implementations:
 - Server: https://github.com/temporalio/temporal/pull/5507
 - SDK Go: https://github.com/temporalio/sdk-go/pull/1418 

**Why?**

This is in preparation for Update-With-Start; so that the user can indicate that an already-running Workflow is _not_ a failure. By adding this conflict resolution mechanism directly to Start Workflow itself, it allows to combine arbitrary operations, such as Update and Signal, with Start - instead of requiring hard-coded top-level APIs like Signal-With-Start.

Without the new enum, the Server wouldn't know whether the user requested a "Maybe Start" (ie continue when Start failed because of another running Workflow with the same ID) or a "Require Start" (ie stop when Start failed because of another running Workflow with the same ID).

**Breaking changes**


temporal/api/enums/v1/workflow.proto
```diff
@@ -44,10 +44,20 @@ enum WorkflowIdReusePolicy {
...
     WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE = 3;
     // If a workflow is running using the same workflow ID, terminate it and start a new one.
-    // If no running workflow, then the behavior is the same as ALLOW_DUPLICATE
+    // If no running workflow, then the behavior is the same as ALLOW_DUPLICATE.
+    // It is *not* allowed to use this option with a non-default WorkflowIdCollisionPolicy.
     WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING = 4;
 }
 
+// Defines what to do when starting a new run with the same ID as a running workflow.
+enum WorkflowIdCollisionPolicy {
```
stephanos, line 53: Some feedback I've received was worried that the difference between `WorkflowIdReusePolicy` and `WorkflowIdCollisionPolicy` wouldn't be immediately clear. I welcome naming suggestions!

Alternatives I've consodered:

- `WorkflowIdRunCollisionPolicy`
- `WorkflowIdRunCollisionPolicy`

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -184,11 +185,13 @@ message StartWorkflowExecutionRequest {
...
     // of the delay will be ignored.
     google.protobuf.Duration workflow_start_delay = 20;
-
     // Callbacks to be called by the server when this workflow reaches a terminal state.
     // If the workflow continues-as-new, these callbacks will be carried over to the new execution.
     // Callback addresses must be whitelisted in the server's dynamic configuration.
     repeated temporal.api.common.v1.Callback completion_callbacks = 21;
+    // Defines how to resolve an ID conflict with a running workflow.
+    // Default: WORKFLOW_ID_COLLISION_POLICY_FAIL.
+    temporal.api.enums.v1.WorkflowIdCollisionPolicy collision_policy = 22;
```
dnr, line 194: I'm usually on board with the "enums instead of bools" idea, but I'm wondering if this one actually makes more sense as a bool. My thinking is that if it's an enum, there's a temptation to add new _behavior_ there, as in doing something different with workflow state, as opposed to just affecting the return value. But all _behavior_ changes should be specified in the reuse policy. This is really just "do we return an error or success if it's already running and the request id matches". And I don't think there can ever be any other options there.

Note that if the request id doesn't match I think we should always return an error, right?

cretz:
(I have avoided this PR because it is not marked ready for review, but let me know if we should review anyways)

stephanos:
@cretz I was going to open the PR for review once I implemented the Server implementation, as I understand that's minimum requirement to get API changes merged. Having said that, I don't mind any feedback.

temporal/api/workflowservice/v1/request_response.proto
stephanos: I should have provided more context in the PR; I've been discussing this with some folks for a while.

The ultimate idea is to also move the Terminate-if-Running from workflowIdReusePolicy to the workflowIdCollisionPolicy. Since the former is meant for _closed_ workflows only, and the latter is meant for dealing with _running_ workflows. The current behavior is just the first immediate step needed for Update-With-Start.

> Note that if the request id doesn't match I think we should always return an error, right?

The idea is to make this codify the "-with-Start" behavior into Start itself, I call this "Maybe Start". So like Signal-with-Start, when there are multiple Signal+Start requests with different request IDs, they won't fail. That behavior is hard-coded into that particular operation - so by adding this new enum, it'll allow a "Maybe Start" with _any operation_.

stephanos: I'm resolving this conversation for now; I've added more context to the PR description.

cretz (approved):

tdeebswihart (approved):

temporal/api/enums/v1/workflow.proto
```diff
@@ -44,10 +46,24 @@ enum WorkflowIdReusePolicy {
     // could potentially change the policy, allowing re-use of the workflow id.
     WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE = 3;
     // If a workflow is running using the same workflow ID, terminate it and start a new one.
```
dnr, line 48: This header comment for the enum talks about closed workflows but this comment talks about open workflows, which is confusing.. Of course this is because we want to migrate this one to conflict policy. We should explain that to reduce the confusion, though. Maybe:
```suggestion
    // This option belongs in WorkflowIdConflictPolicy but is here for backwards compatibility.
    // If specified, it acts like ALLOW_DUPLICATE, but also the WorkflowId*Conflict*Policy on
    // the request is treated as WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -184,11 +187,15 @@ message StartWorkflowExecutionRequest {
...
-
     // Callbacks to be called by the server when this workflow reaches a terminal state.
     // If the workflow continues-as-new, these callbacks will be carried over to the new execution.
     // Callback addresses must be whitelisted in the server's dynamic configuration.
     repeated temporal.api.common.v1.Callback completion_callbacks = 21;
+    // Defines how to resolve an ID conflict with a *running* workflow.
+    // The default policy is WORKFLOW_ID_CONFLICT_POLICY_FAIL.
+    //
+    // See `workflow_id_reuse_policy` for handling ID duplication with a *closed* workflow.
+    temporal.api.enums.v1.WorkflowIdConflictPolicy workflow_id_conflict_policy = 22;
```
dnr, line 198: Move this up to right after workflow_id_reuse_policy. Fields should be ordered/grouped semantically, regardless of tag numbers.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -660,6 +671,12 @@ message SignalWithStartWorkflowExecutionRequest {
     google.protobuf.Duration workflow_start_delay = 20;
     // Indicates that a new workflow task should not be generated when this signal is received.
     bool skip_generate_workflow_task = 21;
+    // Defines how to resolve an ID conflict with a *running* workflow.
+    // The default policy is WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING.
+    // Note that WORKFLOW_ID_CONFLICT_POLICY_FAIL is an invalid option.
+    //
+    // See `workflow_id_reuse_policy` for handling ID duplication with a *closed* workflow.
+    temporal.api.enums.v1.WorkflowIdConflictPolicy workflow_id_conflict_policy = 22;
```
dnr, line 679: same here

temporal/api/workflowservice/v1/request_response.proto
stephanos: Oh, I wasn't aware of that best practice. Done 👍

temporal/api/enums/v1/workflow.proto
```diff
@@ -44,10 +46,24 @@ enum WorkflowIdReusePolicy {
...
-    // If no running workflow, then the behavior is the same as ALLOW_DUPLICATE
+    // If no running workflow, then the behavior is the same as ALLOW_DUPLICATE.
     WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING = 4;
 }
 
+// Defines what to do when trying to start a workflow with the same ID as a *running* workflow.
+// Note that it is *never* valid to have two actively running instances of the same workflow id.
+//
+// See `WorkflowIdReusePolicy` for handling ID duplication with a *closed* workflow.
```
alexshtin, line 58: I don't want to start this `ID`, `Id`, `id` war again but can you at least be consistent? Two lines above it says "id".

Also I would always explicitly WorkflowId or RunId in comments because it might be obvious for fresh users (who is the target audience for his comments).

alexshtin (approved):

temporal/api/enums/v1/workflow.proto
stephanos: Good point! I'll make it `id` everywhere as this is what's been done before (there are only ~3 places, though).

I can also qualify all "ids" with "workflow", that's a good call out.

dnr (approved):

stephanos:
I just pushed another set of changes to `StartWorkflowExecutionResponse` and `SignalWithStartWorkflowExecutionResponse`. It now includes details about what action(s) were actually taken by the operation.

I'd appreciate another review of that new addition.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -184,15 +192,20 @@ message StartWorkflowExecutionRequest {
...
 }
 
 message StartWorkflowExecutionResponse {
+    // The run id of the workflow that was started - or used (via WorkflowIdConflictPolicy USE_EXISTING).
     string run_id = 1;
+    // If true, a new workflow was started.
+    bool started = 3;
+    // If set, the run id of the workflow that had to be terminated before the workflow could be started,
+    // as per WorkflowIdConflictPolicy (or WorkflowIdReusePolicy).
+    string terminated_run_id = 4;
```
cretz, line 208: While I agree this is generally valuable API information, there is no reasonable way to surface this in SDKs at the moment. Though I understand the reasoning behind this is for certain metering use, so no need to change anything.

cretz, line 208: While I agree these two fields are valuable from an API perspective, there is no reasonable way for SDKs to surface this information at this time. But I understand these fields also have value for certain metering uses, so no need to change.

stephanos: That's good feedback. I didn't realize that. In that case, I'm going to remove the `terminated_run_id` field again for now since it's not strictly needed for the metering use case.

PR #360, opened 2024-03-02
DescribeTaskQueue improvements

ShahabT:
**What changed?**
- Added `improved_mode` to `DescribeTaskQueue` to support build Ids and reachability status.
  - Deprecated some old fields.
- Added `DescribeTaskQueuePartition` to replace the deprecated fields of `DescribeTaskQueue`. 

**Why?**
Existing DescribeTaskQueue does not have provisioning for versioned queues and reports internal info not needed by the end user.

**Breaking changes**
Changes to `DescribeTaskQueue` are wire-compatible so requests coming from old GRPC clients should be fine. Some fields are renamed though, not sure about the implication of that when it comes to the HTTP server with presumably JSON request/responses(?)

### DIFF
--- a/temporal/api/enums/v1/task_queue.proto
+++ b/temporal/api/enums/v1/task_queue.proto
@@ -60,6 +60,7 @@ enum TaskQueueType {
 
 // Specifies which category of tasks may reach a worker on a versioned task queue.
 // Used both in a reachability query and its response.
+// Deprecated.
 enum TaskReachability {
     TASK_REACHABILITY_UNSPECIFIED = 0;
     // There's a possiblity for a worker to receive new workflow tasks. Workers should *not* be retired.
@@ -76,3 +77,24 @@ enum TaskReachability {
     TASK_REACHABILITY_CLOSED_WORKFLOWS = 4;
 }
 
+enum WorkflowReachability {
+    // workflow reachability is not reported
+    WORKFLOW_REACHABILITY_UNSPECIFIED = 0;
+    // May be used for new workflows (base on versioning rules), or there are
+    // open workflows assigned to this Build ID
+    WORKFLOW_REACHABILITY_REACHABLE = 1;
+    // It does not have open workflows and is not reachable by new workflows,
+    // but MAY have closed workflows within the namespace retention period.
+    WORKFLOW_REACHABILITY_CLOSED_WORKFLOWS_ONLY = 2;
+    // It is not used for new executions, nor there is any existing execution
+    // within the retention period.
+    WORKFLOW_REACHABILITY_UNREACHABLE = 3;
+}
+
+enum TaskQueueReport {
+    TASK_QUEUE_REPORT_UNSPECIFIED = 0;
+    TASK_QUEUE_REPORT_POLLERS = 1;
+    TASK_QUEUE_REPORT_WORKFLOW_REACHABILITY = 2;
+    TASK_QUEUE_REPORT_WORKFLOW_BACKLOG = 3;
+    TASK_QUEUE_REPORT_ACTIVITY_BACKLOG = 4;
+}
--- a/temporal/api/enums/v1/task_queue.proto
+++ b/temporal/api/enums/v1/task_queue.proto
@@ -91,10 +91,12 @@ enum WorkflowReachability {
     WORKFLOW_REACHABILITY_UNREACHABLE = 3;
 }
 
-enum TaskQueueReport {
+enum TaskQueueInfo {
     TASK_QUEUE_REPORT_UNSPECIFIED = 0;
-    TASK_QUEUE_REPORT_POLLERS = 1;
-    TASK_QUEUE_REPORT_WORKFLOW_REACHABILITY = 2;
-    TASK_QUEUE_REPORT_WORKFLOW_BACKLOG = 3;
-    TASK_QUEUE_REPORT_ACTIVITY_BACKLOG = 4;
+    // Report backlog info for the requested task queue types and versions
+    TASK_QUEUE_REPORT_BACKLOG = 1;
+    // Report list of pollers for requested task queue types and versions
+    TASK_QUEUE_REPORT_POLLERS = 2;
+    // Report workflow reachability for the requested versions
+    TASK_QUEUE_REPORT_REACHABILITY = 3;
 }
--- a/temporal/api/enums/v1/task_queue.proto
+++ b/temporal/api/enums/v1/task_queue.proto
@@ -99,14 +99,3 @@ enum DescribeTaskQueueMode {
     // Enhanced mode reports aggregated results for all partitions, supports Build IDs, and reports richer info.
     DESCRIBE_TASK_QUEUE_MODE_ENHANCED = 1;
 }
-
-enum DesiredInfo {
-    // Unspecified means legacy behavior.
-    DESIRED_INFO_UNSPECIFIED = 0;
-    // Report backlog info for the requested task queue types and versions.
-    DESIRED_INFO_BACKLOG = 1;
-    // Report list of pollers for requested task queue types and versions.
-    DESIRED_INFO_POLLERS = 2;
-    // Report task reachability for the requested versions.
-    DESIRED_INFO_REACHABILITY = 3;
-}
--- a/temporal/api/enums/v1/task_queue.proto
+++ b/temporal/api/enums/v1/task_queue.proto
@@ -77,16 +77,36 @@ enum TaskReachability {
     TASK_REACHABILITY_CLOSED_WORKFLOWS = 4;
 }
 
-enum WorkflowReachability {
-    // workflow reachability is not reported
-    WORKFLOW_REACHABILITY_UNSPECIFIED = 0;
-    // May be used for new workflows (base on versioning rules), or there are
-    // open workflows assigned to this Build ID
-    WORKFLOW_REACHABILITY_REACHABLE = 1;
-    // It does not have open workflows and is not reachable by new workflows,
+// Specifies which category of tasks may reach a versioned worker of a certain Build ID.
+enum BuildIdTaskReachability {
+    // Task reachability is not reported
+    BUILD_ID_TASK_REACHABILITY_UNSPECIFIED = 0;
+    // Build ID may be used by new workflows or activities (base on versioning rules), or there are
+    // open workflows or backlogged activities assigned to it.
+    BUILD_ID_TASK_REACHABILITY_REACHABLE = 1;
+    // Build ID does not have open workflows and is not reachable by new workflows,
     // but MAY have closed workflows within the namespace retention period.
-    WORKFLOW_REACHABILITY_CLOSED_WORKFLOWS_ONLY = 2;
-    // It is not used for new executions, nor there is any existing execution
+    // Not applicable to activity-only task queues.
+    BUILD_ID_TASK_REACHABILITY_CLOSED_WORKFLOWS_ONLY = 2;
+    // Build ID is not used for new executions, nor it has been used by any existing execution
     // within the retention period.
-    WORKFLOW_REACHABILITY_UNREACHABLE = 3;
+    BUILD_ID_TASK_REACHABILITY_UNREACHABLE = 3;
+}
+
+enum DescribeTaskQueueMode {
+    // Unspecified means legacy behavior.
+    DESCRIBE_TASK_QUEUE_MODE_UNSPECIFIED = 0;
+    // Enhanced mode reports aggregated results for all partitions, supports Build IDs, and reports richer info.
+    DESCRIBE_TASK_QUEUE_MODE_ENHANCED = 1;
+}
+
+enum DesiredInfo {
+    // Unspecified means legacy behavior.
+    DESIRED_INFO_UNSPECIFIED = 0;
+    // Report backlog info for the requested task queue types and versions.
+    DESIRED_INFO_BACKLOG = 1;
+    // Report list of pollers for requested task queue types and versions.
+    DESIRED_INFO_POLLERS = 2;
+    // Report task reachability for the requested versions.
+    DESIRED_INFO_REACHABILITY = 3;
 }
--- a/temporal/api/enums/v1/task_queue.proto
+++ b/temporal/api/enums/v1/task_queue.proto
@@ -90,13 +90,3 @@ enum WorkflowReachability {
     // within the retention period.
     WORKFLOW_REACHABILITY_UNREACHABLE = 3;
 }
-
-enum TaskQueueInfo {
-    TASK_QUEUE_REPORT_UNSPECIFIED = 0;
-    // Report backlog info for the requested task queue types and versions
-    TASK_QUEUE_REPORT_BACKLOG = 1;
-    // Report list of pollers for requested task queue types and versions
-    TASK_QUEUE_REPORT_POLLERS = 2;
-    // Report workflow reachability for the requested versions
-    TASK_QUEUE_REPORT_REACHABILITY = 3;
-}
--- a/temporal/api/operatorservice/v1/service.proto
+++ b/temporal/api/operatorservice/v1/service.proto
@@ -81,11 +81,4 @@ service OperatorService {
     // ListClusters returns information about Temporal clusters.
     rpc ListClusters(ListClustersRequest) returns (ListClustersResponse) {
     }
-
-    // DescribeTaskQueuePartition returns information about a task queue partition helpful for debugging.
-    rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {
-        option (google.api.http) = {
-            get: "/api/v1/namespaces/{namespace}/task-queues/{task_queue.name}/partition"
-        };
-    }
 }
--- a/temporal/api/operatorservice/v1/service.proto
+++ b/temporal/api/operatorservice/v1/service.proto
@@ -81,4 +81,11 @@ service OperatorService {
     // ListClusters returns information about Temporal clusters.
     rpc ListClusters(ListClustersRequest) returns (ListClustersResponse) {
     }
+
+    // DescribeTaskQueuePartition returns information about a task queue partition helpful for debugging.
+    rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/task-queues/{task_queue.name}/partition"
+        };
+    }
 }
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -54,24 +54,34 @@ message TaskQueueMetadata {
     google.protobuf.DoubleValue max_tasks_per_second = 1;
 }
 
-message QueueStatus {
-    // "_unversioned_" represents the unversioned queue
+message TaskQueueVersionInfo {
+    // Empty means unversioned
     string build_id = 1;
-    repeated PollerInfo pollers = 2;
-    temporal.api.enums.v1.WorkflowReachability workflow_reachability = 3;
-    BacklogInfo workflow_backlog_info = 4;
-    BacklogInfo activity_backlog_info = 5;
+    repeated TaskQueueTypeInfo task_queue_types_info = 3;
+    temporal.api.enums.v1.WorkflowReachability workflow_reachability = 4;
 }
 
+message TaskQueueTypeInfo {
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 1;
+    repeated PollerInfo pollers = 2;
+    BacklogInfo backlog_info = 3;
+}
+
+// For workflow task queues, we only report the normal queue metrics, not sticky queues. This means the stats
+// reported here do not count all workflow tasks. However, because the tasks queued in sticky queues only remain
+// valid for a few seconds, the inaccuracy becomes less significant as the backlog age grows.
 message BacklogInfo {
+    // The approximate number of tasks backlogged in this task queue. May count expired tasks but eventually converges
+    // to the right value.
     int64 approximate_backlog_count = 1;
+    // Based on the timestamp of the task at the head of the queue.
     google.protobuf.Duration approximate_backlog_age = 2;
-    // approximate tasks per second added and dispatched based on recent activity
+    // Approximate tasks per second added and dispatched based on recent activity
     float tasks_add_rate = 3;
     float tasks_dispatch_rate = 4;
 }
 
-message InternalQueueStatus {
+message InternalTaskQueueStatus {
     int64 backlog_count_hint = 1;
     int64 read_level = 2;
     int64 ack_level = 3;
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -48,6 +48,18 @@ message TaskQueue {
     string normal_name = 3;
 }
 
+// Represents a normal or sticky partition of a task queue.
+message TaskQueuePartition {
+    // Normal task queue name.
+    string task_queue_name = 1;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 2;
+    // Absent means normal root partition (normal_partition_id=0)
+    oneof partition_id {
+        int32 normal_partition_id = 3;
+        int32 sticky_name = 4;
+    }
+}
+
 // Only applies to activity task queues
 message TaskQueueMetadata {
     // Allows throttling dispatch of tasks from this queue
@@ -57,12 +69,12 @@ message TaskQueueMetadata {
 message TaskQueueVersionInfo {
     // Empty means unversioned
     string build_id = 1;
-    repeated TaskQueueTypeInfo task_queue_types_info = 3;
+    repeated TaskQueueTypeInfo types_info = 3;
     temporal.api.enums.v1.WorkflowReachability workflow_reachability = 4;
 }
 
 message TaskQueueTypeInfo {
-    temporal.api.enums.v1.TaskQueueType task_queue_type = 1;
+    temporal.api.enums.v1.TaskQueueType type = 1;
     repeated PollerInfo pollers = 2;
     BacklogInfo backlog_info = 3;
 }
@@ -83,6 +95,15 @@ message BacklogInfo {
 }
 
 message InternalTaskQueueStatus {
+    int64 backlog_count_hint = 1;
+    int64 read_level = 2;
+    int64 ack_level = 3;
+    double rate_per_second = 4;
+    TaskIdBlock task_id_block = 5;
+}
+
+// Deprecated. Use `InternalTaskQueueStatus`. This is kept until `DescribeTaskQueue` supports legacy behavior.
+message TaskQueueStatus {
     int64 backlog_count_hint = 1;
     int64 read_level = 2;
     int64 ack_level = 3;
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -74,10 +74,11 @@ message BacklogInfo {
     // The approximate number of tasks backlogged in this task queue. May count expired tasks but eventually converges
     // to the right value.
     int64 approximate_backlog_count = 1;
-    // Based on the timestamp of the task at the head of the queue.
+    // Approximate age of the backlog based on the create timestamp of the task at the head of the queue.
     google.protobuf.Duration approximate_backlog_age = 2;
-    // Approximate tasks per second added and dispatched based on recent activity
+    // Approximate tasks per second added to the queue based on recent activity
     float tasks_add_rate = 3;
+    // Approximate tasks per second dispatched to workers based on recent activity
     float tasks_dispatch_rate = 4;
 }
 
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -67,14 +67,16 @@ message TaskQueueMetadata {
 }
 
 message TaskQueueVersionInfo {
-    // Empty means unversioned
+    // Empty means unversioned.
     string build_id = 1;
     repeated TaskQueueTypeInfo types_info = 3;
-    temporal.api.enums.v1.WorkflowReachability workflow_reachability = 4;
+    // Only applicable to versioned workers (workers with `useVersioning=true`).
+    temporal.api.enums.v1.BuildIdTaskReachability task_reachability = 4;
 }
 
 message TaskQueueTypeInfo {
     temporal.api.enums.v1.TaskQueueType type = 1;
+    // Unversioned workers (with `useVersioning=false`) are reported in unversioned result even if they set a Build ID.
     repeated PollerInfo pollers = 2;
     BacklogInfo backlog_info = 3;
 }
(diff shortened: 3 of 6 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,58 @@ message DescribeWorkflowExecutionResponse {
...
+
+// (-- api-linter: core::0203::optional=disabled
+//     aip.dev/not-precedent: field_behavior annotation not available in our gogo fork --)
+message DescribeTaskQueueRequest {
+    string namespace = 1;
+    string task_queue = 5;
+    // Optional. If not provided, the result for the default Build ID will be returned. Default Build ID is the one
+    // mentioned in the first unconditional Assignment Rule. If there is no default Build ID, the result for the
+    // unversioned queue will be returned.
+    // Special values:
```
ShahabT, line 881: Down the road we can add special values such as `_all_`, `_with_backlog_`, `_reachable_`, etc. but those need more bookkeeping that we don't have right away.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,58 @@ message DescribeWorkflowExecutionResponse {
...
+    //  - Use "_unversioned_" to explicitly ask for the unversioned queue
+    //  - Use "_with_worker_" to ask for Build IDs (including unversioned) with active workers. This is particularly
+    //    useful for poller info to list all active pollers regardless of Build ID.
+    repeated string build_ids = 7;
+    repeated temporal.api.enums.v1.TaskQueueReport requested_reports = 8;
```
cretz, line 886: Enumerate may be too limiting here. I think this should be explicit fields to request things. Sure many may seem like booleans, but with the current approach, you can't have `repeated TaskQueueType include_backlog_info_for` or something and now you've missed "Nexus" or any other future one.

Similarly, the per-task-queue-type info should be in its own message on queue status IMO.

temporal/api/workflowservice/v1/request_response.proto
cretz, line 881: When you get into special values, I think we should consider oneof instead of magic strings

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,58 @@ message DescribeWorkflowExecutionResponse {
...
+message DescribeInternalTaskQueueResponse {
+    // Deprecated. Use `DescribeTaskQueueRequest.queues_status.pollers`.
     repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
-    temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2;
+    temporal.api.taskqueue.v1.InternalQueueStatus task_queue_status = 2;
+}
+
+// (-- api-linter: core::0203::optional=disabled
+//     aip.dev/not-precedent: field_behavior annotation not available in our gogo fork --)
+message DescribeTaskQueueRequest {
```
cretz, line 875: I wonder if it can be something like:

```proto

message DescribeTaskQueueRequest {
    string namespace = 1;
    string task_queue = 5;

    repeated TaskQueueInfoRequest requests = 6;

    // Legacy internal info, if any set, no requests can be set (and vice-versa)
    temporal.api.taskqueue.v1.TaskQueue internal_task_queue = 2;
    temporal.api.enums.v1.TaskQueueType internal_task_queue_type = 3;
    bool internal_include_task_queue_status = 4;
}

    // 1:1 with requests at the same index
    repeated TaskQueueInfo responses = 3;

    // Legacy internal info
    repeated temporal.api.taskqueue.v1.PollerInfo internal_pollers = 1;
    temporal.api.taskqueue.v1.InternalQueueStatus internal_queue_status = 2;
}
```
and
```proto
message TaskQueueInfoRequest {
    // All unset means default
    oneof build_id {
        string specific_id = 1;
        bool unversioned = 2;
        bool with_workers = 3;
    }
    temporal.api.enums.v1.TaskQueueType include_backlog_info_for = 4;
    temporal.api.enums.v1.TaskQueueType include_pollers_for = 5;
    bool include_workflow_reachability = 6;
}

message TaskQueueInfo {
    oneof build_id {
        string id = 1;
        bool unversioned = 2;
    }
    BacklogInfo backlog_info = 2;
    repeated PollerInfo pollers = 3;
    WorkflowReachability workflow_reachability = 4;
}
```

Benefits:

* Clearer ability to tie requests to responses
* Automatically get all new task queue types like Nexus 
* Keeps "legacy"/"internal" as not deprecated, just internal

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,7 +54,24 @@ message TaskQueueMetadata {
     google.protobuf.DoubleValue max_tasks_per_second = 1;
 }
 
-message TaskQueueStatus {
+message QueueStatus {
+    // "_unversioned_" represents the unversioned queue
+    string build_id = 1;
```
cretz, line 59: Can you clarify the cases where this would be different than the passed in one? For instance if I expect default, does this get populated with the default?

temporal/api/workflowservice/v1/request_response.proto
bergundy: Consider also the semantics for the future use case. Is `_with_backlog_` and `_reachable_` an OR or AND query?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,58 @@ message DescribeWorkflowExecutionResponse {
...
+    //  - Use "_with_worker_" to ask for Build IDs (including unversioned) with active workers. This is particularly
+    //    useful for poller info to list all active pollers regardless of Build ID.
+    repeated string build_ids = 7;
+    repeated temporal.api.enums.v1.TaskQueueReport requested_reports = 8;
+
+    // The following fields are related to `DescribeInternalTaskQueue`, kept here for compatibility until
+    // the old clients reach EOF. Presence of `internal_task_queue` triggers the old behavior.
+    //
+    // Deprecated. use `DescribeInternalTaskQueue`.
+    temporal.api.taskqueue.v1.TaskQueue internal_task_queue = 2;
```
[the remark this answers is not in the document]
bergundy: This should be the default IMHO and not break backwards compatibility.

Use a version field on the RPC to opt-into the new API.

bergundy (commented):
I'm mostly worried about maintaining the current behavior by default and making the new (and better) API opt-in.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,7 +54,24 @@ message TaskQueueMetadata {
     google.protobuf.DoubleValue max_tasks_per_second = 1;
 }
 
-message TaskQueueStatus {
+message QueueStatus {
```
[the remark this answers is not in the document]
cretz: It's backwards compatible on the wire (e.g. w/ CLI and UI), it's only incompatible with programmatic users, but I think that's an acceptable break (and should break at compile time). I don't have a strong opinion here.

temporal/api/workflowservice/v1/request_response.proto
cretz, line 892: Note, while changing message names is one thing, changing field names is another and you would have to coordinate with the UI team so they can update their gRPC-JSON layer to expect this incompatible JSON change.

ShahabT (commented):
Made changes based on the feedback. Please review again.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: > Is _with_backlog_ and _reachable_ an OR or AND query?
Made it oneof so two of these can't be passed together anymore.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Made changes inspired by this. Please not that the mapping between request and response is not 1:1.

temporal/api/enums/v1/task_queue.proto
```diff
@@ -76,3 +77,26 @@ enum TaskReachability {
...
+    WORKFLOW_REACHABILITY_REACHABLE = 1;
+    // It does not have open workflows and is not reachable by new workflows,
+    // but MAY have closed workflows within the namespace retention period.
+    WORKFLOW_REACHABILITY_CLOSED_WORKFLOWS_ONLY = 2;
+    // It is not used for new executions, nor there is any existing execution
+    // within the retention period.
+    WORKFLOW_REACHABILITY_UNREACHABLE = 3;
+}
+
+enum TaskQueueInfo {
```
Sushisource, line 94: This is almost more like `DesiredTaskQueueInfo` or something that signals it's saying what's wanted rather than what's returned

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,7 +54,34 @@ message TaskQueueMetadata {
...
+// For workflow task queues, we only report the normal queue metrics, not sticky queues. This means the stats
+// reported here do not count all workflow tasks. However, because the tasks queued in sticky queues only remain
+// valid for a few seconds, the inaccuracy becomes less significant as the backlog age grows.
+message BacklogInfo {
+    // The approximate number of tasks backlogged in this task queue. May count expired tasks but eventually converges
+    // to the right value.
+    int64 approximate_backlog_count = 1;
+    // Based on the timestamp of the task at the head of the queue.
+    google.protobuf.Duration approximate_backlog_age = 2;
+    // Approximate tasks per second added and dispatched based on recent activity
```
Sushisource, line 79: Break this into two comments - these end up as docstrings on the fields on code and each one should have it's own appropriate comment

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,7 +54,34 @@ message TaskQueueMetadata {
...
+}
+
+// For workflow task queues, we only report the normal queue metrics, not sticky queues. This means the stats
+// reported here do not count all workflow tasks. However, because the tasks queued in sticky queues only remain
+// valid for a few seconds, the inaccuracy becomes less significant as the backlog age grows.
+message BacklogInfo {
+    // The approximate number of tasks backlogged in this task queue. May count expired tasks but eventually converges
+    // to the right value.
+    int64 approximate_backlog_count = 1;
+    // Based on the timestamp of the task at the head of the queue.
```
Sushisource, line 77: It's worth making this a complete sentence

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,49 @@ message DescribeWorkflowExecutionResponse {
...
+
     string namespace = 1;
-    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
-    // If unspecified (TASK_QUEUE_TYPE_UNSPECIFIED), then default value (TASK_QUEUE_TYPE_WORKFLOW) will be used.
-    temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
-    bool include_task_queue_status = 4;
+    string task_queue = 5;
+    // Optional. If not provided, the result for the default Build ID will be returned. Default Build ID is the one
+    // mentioned in the first unconditional Assignment Rule. If there is no default Build ID, the result for the
+    // unversioned queue will be returned.
```
Sushisource, line 870:
```suggestion
    // Optional. If not provided, the result for the default Build ID will be returned. The default Build ID is the one
    // mentioned in the first unconditional Assignment Rule. If there is no default Build ID, the result for the
    // unversioned queue will be returned.
```

Is it possible for there to be no default build ID, but the queue is still versioned?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,49 @@ message DescribeWorkflowExecutionResponse {
...
+        BuildIdList build_ids = 6;
+        // Explicitly ask for the unversioned queue
+        bool unversioned = 7;
+        // Ask for Build IDs (including unversioned) with active workers. This is particularly useful to list
+        // all active pollers regardless of Build ID.
+        // (-- api-linter: core::0140::prepositions --)
+        bool with_workers = 8;
+    }
+
+    // Task queue types to report info about
```
Sushisource, line 881:
```suggestion
    // Task queue types to report info about. If not specified, all types are considered.
```

Assuming that's accurate

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,49 @@ message DescribeWorkflowExecutionResponse {
...
+    temporal.api.taskqueue.v1.TaskQueue internal_task_queue = 2;
+    temporal.api.enums.v1.TaskQueueType internal_task_queue_type = 3;
+    bool include_internal_task_queue_status = 4;
 }
 
 message DescribeTaskQueueResponse {
-    repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
-    temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2;
+    // The number of items in this list depends on the requested Build IDs. The `with_workers` option can result into
+    // zero or more Build IDs.
```
Sushisource, line 895:
```suggestion
    // The number of items in this list depends on the requested Build IDs. The `with_workers` option may result in
    // zero or more Build IDs.
```

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
cretz: What do we do today server side when UNSPECIFIED?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,58 @@ message DescribeWorkflowExecutionResponse {
...
+    temporal.api.taskqueue.v1.TaskQueue internal_task_queue = 2;
+    // If unspecified (TASK_QUEUE_TYPE_UNSPECIFIED), then default value (TASK_QUEUE_TYPE_WORKFLOW) will be used.
+    // Deprecated. use `DescribeInternalTaskQueue`.
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
+    // Deprecated. use `DescribeInternalTaskQueue`.
+    bool include_internal_task_queue_status = 4;
+}
+
+message DescribeTaskQueueResponse {
+    repeated temporal.api.taskqueue.v1.QueueStatus queues_status = 3;
```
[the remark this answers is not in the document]
cretz: I fear it can be hard to know which response is for which requested piece, especially if I mix special values and non-special values

temporal/api/workflowservice/v1/request_response.proto
cretz: Can we make the mapping 1:1 so you can tie what you asked for to what you got?

temporal/api/enums/v1/task_queue.proto
cretz: This is not "info" as we usually use the term (and prefixes of enums are wrong even if we do want to change this). I recommend we use explicit fields for users to ask for specific things instead of enums.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,49 @@ message DescribeWorkflowExecutionResponse {
...
+        // Ask for Build IDs (including unversioned) with active workers. This is particularly useful to list
+        // all active pollers regardless of Build ID.
+        // (-- api-linter: core::0140::prepositions --)
+        bool with_workers = 8;
+    }
+
+    // Task queue types to report info about
+    repeated temporal.api.enums.v1.TaskQueueType task_queue_types = 9;
+    // Information requested for the mentioned versions and task queue types
+    repeated temporal.api.enums.v1.TaskQueueInfo requested_info = 10;
```
cretz, line 884: I think it's clearer to use more explicit fields to request information instead of enumerates. (I personally like the idea of entire request messages that explicitly say what to return like in my comment at https://github.com/temporalio/api/pull/360/files#r1511255630)

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Yes, one can have only conditional rules (e.g. imagine ramping from unversioned to versioned).

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
cretz: I see, then yeah users are going to have to opt-in to this newer form somehow.

> So it seems we can trigger new/old behavior based on this value.

Users may be calling this directly with older SDKs which will all of the sudden stop working

temporal/api/workflowservice/v1/request_response.proto
cretz: This works for me now that you cannot provide some specific build IDs and some special ones in the same request.

temporal/api/workflowservice/v1/request_response.proto
cretz: This works for me now that you cannot provide some specific build IDs and some special ones in the same request.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,53 @@ message DescribeWorkflowExecutionResponse {
...
+    }
+
+    // Task queue types to report info about. If not specified, all types are considered.
+    repeated temporal.api.enums.v1.TaskQueueType task_queue_types = 9;
+    // Report backlog info for the requested task queue types and versions
+    bool report_backlog_info = 10;
+    // Report list of pollers for requested task queue types and versions
+    bool report_pollers = 11;
+    // Report workflow reachability for the requested versions
+    bool report_workflow_reachability = 12;
```
bergundy, line 888: Still prefer an enum here.
Just in case any of these bools want to allow more options.

E.g. `report_pollers` could later be split up to report pollers of a specific type and the reachability may be customizable.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,53 @@ message DescribeWorkflowExecutionResponse {
     temporal.api.workflow.v1.PendingWorkflowTaskInfo pending_workflow_task = 5;
 }
 
+// (-- api-linter: core::0203::optional=disabled
+//     aip.dev/not-precedent: field_behavior annotation not available in our gogo fork --)
 message DescribeTaskQueueRequest {
+    message BuildIdList {
```
bergundy, line 862: Unused?

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
bergundy: WDYM not popular?
Both the CLI and UI use it, UI is a heavy user of the current API.

Would be worth verifying how breaking this change is.

temporal/api/workflowservice/v1/request_response.proto
cretz: > Still prefer an enum here. Just in case any of these bools want to allow more options.

You mean a single-value enum for each one of these three bools, or three enums and one field? If the latter, enum doesn't allow more options either, it's either set or not just like a bool. Can you show an example of what you mean? I fear overly future-proofing some of this lest we just make each of these a full/empty object preparing for some kind of "report poller config".

temporal/api/workflowservice/v1/request_response.proto
ShahabT: it is used in `versions`.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: I don't have strong preference either way. Not sure enum vs bools is really different for future-proof-ness. We can always add additional config options as new fields later.

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
ShahabT: @bergundy we're not breaking anything with CLI and UI. The existing versions of UI and CLI will work. I have @Alex-Tideman's confirmation that we can make sure new UI versions (OSS and Cloud) will work as expected with new and old server. For the new CLI it's our call to keep the old command params supported in the first GA version of it or not. But if we want to we can definitely do so.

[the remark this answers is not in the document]
ShahabT: hmm.. `versions` is an optional field though. thinking to change the booleans back to enum list like `repeated DesiredTaskQueueInfo desired_info` so that we can base new behavior on presence of desired_info value. (this would be cleaner than basing on OR of a bunch of bools)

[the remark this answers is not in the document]
cretz: I am afraid for backwards compatible wire behavior (not to be confused with backwards compatible generated code), you have to have all existing calls work as they do which means there must be some new thing a user sets.

I still don't think it should be enums though. Enums are inflexible if, say, you wanted to add another thing they could request and wanted more than just a "yes" or "no". If you wanted to use the presence of any of the bools, you could. Also, if you'd like, you can move "requested info" into a separate message and if that is non-null it can be the opt-in.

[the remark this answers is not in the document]
ShahabT: @bergundy plz take a look now. I added explicit `enhanced_mode` flag for opting in to the new mode. old fields remain the same, except when it comes to internal task queue status which I think we should rename still to clarify that it has internal info and also reduce confusion now that we add new info. (I doubt if any user use this, almost everybody should be using the poller list which is unchanged).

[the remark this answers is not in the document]
ShahabT: Thanks @cretz. Inspired by this, I added a bool to explicitly opt in to the new mode. I like it better than a nested message because later we can get rid of old fields and the bool flag (once old clients reach EOL).

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,7 +54,35 @@ message TaskQueueMetadata {
     google.protobuf.DoubleValue max_tasks_per_second = 1;
 }
 
-message TaskQueueStatus {
+message TaskQueueVersionInfo {
+    // Empty means unversioned
+    string build_id = 1;
+    repeated TaskQueueTypeInfo task_queue_types_info = 3;
```
cretz, line 60:
```suggestion
    repeated TaskQueueTypeInfo types_info = 3;
```

Probably don't need the prefix

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,58 @@ message DescribeWorkflowExecutionResponse {
...
+    // Deprecated. Use `enhanced_mode` with `task_queue_types`. Ignored when `enhanced_mode` is set.
     // If unspecified (TASK_QUEUE_TYPE_UNSPECIFIED), then default value (TASK_QUEUE_TYPE_WORKFLOW) will be used.
     temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
-    bool include_task_queue_status = 4;
+    // Ignored when `enhanced_mode` is set.
+    bool include_internal_task_queue_status = 4;
+
+    // This enables enhanced behavior. All options except `task_queue_type` and `include_internal_task_queue_status`
+    // are only available when `enhanced_mode` is set.
+    bool enhanced_mode = 6;
```
cretz, line 872:
```suggestion
    bool enhanced_mode = 5;
```

And decrement other field nums, unless this was intentional

cretz (approved):
I like this approach, marking approved, but some quick notes:

* With this "enhanced" approach, we could just not rename any of the old fields/messages and keep it code compat too instead of just injecting "internal" to break without lots of benefit
* This is wire compatible which is important to me
* I would like others to give their approval instead of taking mine by itself

ShahabT:
@cretz @bergundy the protos message and field names are kept all the same now. Only added some fields and deprecated some others. Added a new `DescribeTaskQueuePartition` API to replace the functionality of the deprecated fields which was really about one partition.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: I'd do repeated enum where the enum consists of these flags, simply because it seems nice to have one field for the reporting options. So, one enum type, one `repeated InfoRequest requested_info` field. Which I think is pretty close to what you had originally?

But I agree it's not tremendously important either way.

Sushisource (approved):

temporal/api/taskqueue/v1/message.proto
```diff
@@ -48,12 +48,61 @@ message TaskQueue {
     string normal_name = 3;
 }
 
+// Represents a normal or sticky partition of a task queue.
```
ShahabT, line 51: FYI, we now consider sticky queues also "partitions" of normal queues in the server code: https://github.com/temporalio/temporal/blob/684c170cd42d4489aa1950911da0c986cc81207f/common/tqid/task_queue_id.go#L100

temporal/api/workflowservice/v1/request_response.proto
cretz: Yeah, this is what it was originally, but I found explicit fields clearer (that they happen to be only bool on/off today doesn't mean the next one won't be a message w/ options). My opinion here isn't that strong though if enough disagree.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,74 @@ message DescribeWorkflowExecutionResponse {
...
 }
 
 message DescribeTaskQueueResponse {
+    // Deprecated. Use `versions_info.types_info.pollers` with `enhanced_mode` instead.
+    // Not set in `enhanced_mode`.
     repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
-    temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2;
+    // Deprecated. Use `DescribeTaskQueuePartition`.
+    // Not set in `enhanced_mode`.
+    temporal.api.taskqueue.v1.TaskQueueStatus queue_status = 2;
```
cretz, line 923: Can we change this back to `task_queue_status` so we can get full code compat?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -436,13 +436,23 @@ service WorkflowService {
...
+    //   - Workflow Reachability status
+    //   - Backlog info for Workflow and/or Activity tasks
     rpc DescribeTaskQueue (DescribeTaskQueueRequest) returns (DescribeTaskQueueResponse) {
         option (google.api.http) = {
             get: "/api/v1/namespaces/{namespace}/task-queues/{task_queue.name}"
         };
     }
 
+    // DescribeTaskQueuePartition returns information about task queue partition helpful for debugging.
+    rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {
```
cretz, line 450: Hrmm, just saw this RPC, is this a new, needed feature/RPC? Does this even need to be in workflow service? Will this work for cloud users?

temporal/api/enums/v1/task_queue.proto
```diff
@@ -76,3 +77,16 @@ enum TaskReachability {
     TASK_REACHABILITY_CLOSED_WORKFLOWS = 4;
 }
 
+enum WorkflowReachability {
```
bergundy, line 80: We discussed renaming this back to TaskReachability to also have it apply to activity and nexus tasks.

You can mark `EXISTING` as deprecated.

bergundy: And mention that CLOSED_WORKFLOWS only applies to workflow reachability.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -856,17 +856,74 @@ message DescribeWorkflowExecutionResponse {
...
+    // Deprecated. Use `enhanced_mode` with `task_queue_types`. Ignored when `enhanced_mode` is set.
     // If unspecified (TASK_QUEUE_TYPE_UNSPECIFIED), then default value (TASK_QUEUE_TYPE_WORKFLOW) will be used.
     temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
+    // Ignored when `enhanced_mode` is set.
+    // Deprecated. Use `DescribeTaskQueuePartition`.
     bool include_task_queue_status = 4;
+
+    // This enables enhanced behavior. All options except `task_queue_type` and `include_task_queue_status`
+    // are only available when `enhanced_mode` is set.
+    bool enhanced_mode = 5;
```
bergundy, line 887: Still think this should be an enum or int to allow later adding a "super enhanced mode" (v3).

temporal/api/workflowservice/v1/request_response.proto
bergundy: I still prefer a repeated enum or maybe even a repeated oneof where you can add options to each request that is only relevant to that request but I'm not blocking the PR for this.

temporal/api/workflowservice/v1/service.proto
bergundy, line 450: We discussed putting this in operator service, since this isn't something most users will be concerned with.

ShahabT: moved it to operatorservice. it is a useful API for debugging. Not a new feature, just replacement for the deprecated `task_queue_status` field.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Went with the majority vote and made it and enum.

ShahabT:
I'll keep it open until tomorrow morning in case more comments come. but it seems all concerns should be resolved now.

temporal/api/operatorservice/v1/service.proto
```diff
@@ -81,4 +81,11 @@ service OperatorService {
     // ListClusters returns information about Temporal clusters.
     rpc ListClusters(ListClustersRequest) returns (ListClustersResponse) {
     }
+
+    // DescribeTaskQueuePartition returns information about a task queue partition helpful for debugging.
+    rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/task-queues/{task_queue.name}/partition"
+        };
```
cretz, line 89:
```suggestion
```

No need to expose to HTTP IMO

temporal/api/workflowservice/v1/service.proto
cretz: Hrmm, so could just undeprecate the existing one. Seems to have a whole new task queue partition thing and feels like a new feature.

temporal/api/enums/v1/task_queue.proto
```diff
@@ -76,3 +77,36 @@ enum TaskReachability {
...
+}
+
+enum DescribeTaskQueueMode {
+    // Unspecified means legacy behavior.
+    DESCRIBE_TASK_QUEUE_MODE_UNSPECIFIED = 0;
+    // Enhanced mode reports aggregated results for all partitions, supports Build IDs, and reports richer info.
+    DESCRIBE_TASK_QUEUE_MODE_ENHANCED = 1;
+}
+
+enum DesiredInfo {
```
cretz, line 103:
```suggestion
enum TaskQueueDesiredInfo {
```

This is a general purpose enumerate package, you have to qualify these IMO

cretz (commented):
Don't really understand the new task queue partition model/feature and not a fan of the concept of a desired info enum, but no biggie. I would like to see it have a more qualified name since it's in the shared enum package and I would like to see the HTTP annotations removed from the operator service call unless we can confirm UI wants to still use this call.

temporal/api/workflowservice/v1/service.proto
ShahabT: Got rid of the new RPC call. We'll add it to `tdbg` later. Not even an operator API.

PR #361, opened 2024-03-05
Change Namespace data map to contain bytes

tdeebswihart:
**What changed?**
The namespace data map now has `bytes` values rather than `string`.

**Why?**

We embed binary data (cast to strings) here in a few places, and that can cause google/protobuf to yell at us for having string data that's invalid UTF-8. Protobuf strings and bytes are wire-compatible

**Breaking changes**

This will change all code generated by protoc plugins to use the language's type for raw bytes. I have paired Go API and server PRs that will go along with this change

### DIFF
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -43,7 +43,7 @@ message NamespaceInfo {
     string description = 3;
     string owner_email = 4;
     // A key-value map for any customized purpose.
-    map<string, string> data = 5;
+    map<string, bytes> data = 5;
     string id = 6;
     // All capabilities the namespace supports.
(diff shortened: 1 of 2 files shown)


stephanos (approved):

cretz (approved):
Not the biggest fan of breaking API compat even for these rarely used APIs, but I understand the reasoning.

You'll have to disable the buf compat detector.

yycptt:
we are ok with the breaking change for customer when they upgrade sdk?

yycptt:
Is there a reason we are just fixing this specific field? I think the same thing could happen to, for example, the `memo` field.

tdeebswihart:
We've decided to axe this and migrate the relevant fields internally

PR #362, opened 2024-03-06
Do not reuse NexusIncomingService for CreateOrUpdate request

bergundy:
**What changed?**

In description

**Why?**

Will allow us to add more fields in the response that are server set and should not be accepted in create or update requests.

**Breaking changes**

It's breaking but was never used.

### DIFF
--- a/temporal/api/nexus/v1/message.proto
+++ b/temporal/api/nexus/v1/message.proto
@@ -120,13 +120,13 @@ message Response {
 
 // A binding from a service name to namespace, task queue, and metadata for dispatching incoming Nexus requests.
 message IncomingService {
-    // Data version for this service. Must match current version on update or set to 0 to create a new service.
+    // Data version for this service, incremented for every mutation.
     int64 version = 1;
     // Service name, unique for this cluster.
     // The service name is used to address this service.
     // By default, when using Nexus over HTTP, the service name is matched against the base URL path.
     // E.g. the URL /my-service would match a service named "my-service".
-    // The name can contain any characters and is escaped when matched against a URL.
+    // The name must match `[a-zA-Z_][a-zA-Z0-9_]*`and is escaped when matched against a URL.
     string name = 2;
     // Namespace to route requests to.
     string namespace = 3;
--- a/temporal/api/operatorservice/v1/request_response.proto
+++ b/temporal/api/operatorservice/v1/request_response.proto
@@ -33,6 +33,7 @@ option csharp_namespace = "Temporalio.Api.OperatorService.V1";
 
 import "temporal/api/enums/v1/common.proto";
 import "temporal/api/nexus/v1/message.proto";
+import "google/protobuf/any.proto";
 import "google/protobuf/duration.proto";
 
 // (-- Search Attribute --)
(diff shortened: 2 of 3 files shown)


pdoerner (approved):

openapi/openapiv2.json
```diff
@@ -4711,11 +4711,11 @@
         "version": {
           "type": "string",
           "format": "int64",
-          "description": "Data version for this service. Must match current version on update or set to 0 to create a new service."
+          "description": "Data version for this service, incremented for every mutation."
         },
         "name": {
           "type": "string",
-          "description": "Service name, unique for this cluster.\nThe service name is used to address this service.\nBy default, when using Nexus over HTTP, the service name is matched against the base URL path.\nE.g. the URL /my-service would match a service named \"my-service\".\nThe name can contain any characters and is escaped when matched against a URL."
+          "description": "Service name, unique for this cluster.\nThe service name is used to address this service.\nBy default, when using Nexus over HTTP, the service name is matched against the base URL path.\nE.g. the URL /my-service would match a service named \"my-service\".\nThe name must match `[a-zA-Z_][a-zA-Z0-9_]*`and is escaped when matched against a URL."
```
antlai-temporal, line 4718: "And is escaped when matched ..."

Do you still need to escape after restricting to these alpha/digit characters?

temporal/api/nexus/v1/message.proto
```diff
@@ -120,13 +120,13 @@ message Response {
 
 // A binding from a service name to namespace, task queue, and metadata for dispatching incoming Nexus requests.
 message IncomingService {
-    // Data version for this service. Must match current version on update or set to 0 to create a new service.
+    // Data version for this service, incremented for every mutation.
```
antlai-temporal, line 123: Can you clarify that "mutation" is using the CreateOrUpdate... api

temporal/api/operatorservice/v1/request_response.proto
```diff
@@ -139,8 +140,23 @@ message GetNexusIncomingServiceResponse {
     temporal.api.nexus.v1.IncomingService service = 1;
 }
 
+// (-- api-linter: core::0133::request-unknown-fields=disabled
+//     aip.dev/not-precedent: CreateOrUpdateNexusIncomingServiceResponse RPC doesn't follow Google API format. --)
 message CreateOrUpdateNexusIncomingServiceRequest {
-    temporal.api.nexus.v1.IncomingService service = 1;
+    // Data version for this service. Must match current version on update or set to 0 to create a new service.
+    int64 version = 1;
```
antlai-temporal, line 147: This is auto incremented, right ? Add a comment?

antlai-temporal (approved):

temporal/api/operatorservice/v1/request_response.proto
bergundy: There's a comment on the return type.

openapi/openapiv2.json
bergundy: Yeah, there's no need to escape any more. Thanks.

temporal/api/nexus/v1/message.proto
bergundy: I'll rephrase a bit.

cretz (approved):
Not the biggest fan of making incompatible changes to API even if unused. Ideally this would be in some kind of feature branch while you work out the final form, but I understand if that's difficult.

bergundy:
> Not the biggest fan of making incompatible changes to API even if unused. Ideally this would be in some kind of feature branch while you work out the final form, but I understand if that's difficult.

Agree. I made this change based on your comment on another document for the outgoing service registry. I figured we'd apply it to incoming too while it's still possible.

PR #363, opened 2024-03-06
Rest of Nexus API definitions

bergundy:
**Note for reviewers**

**We shouldn't merge this PR until we have server support** for it but I do want to get an initial review before relying on these protos in the implementation.
There are also a couple of TODOs I left in here that I still need to figure out if they're required.
Even after we merge this there'll like be very small changes, but hopefully only additions, I'll try my best not to require any more breaking changes.

**What changed?**

- Added commands and events for nexus operations
- Added NexusOperationInfo definition and attached it to DescribeWorkflowExecutionResponse
- Added NexusOutgoingService operator CRUD APIs


temporal/api/command/v1/message.proto
```diff
@@ -234,6 +234,27 @@ message ProtocolMessageCommandAttributes {
     string message_id = 1;
 }
 
+message ScheduleNexusOperationCommandAttributes {
```
cretz, line 237: Can I set headers like I can via Nexus calls directly?

temporal/api/enums/v1/command_type.proto
```diff
@@ -49,4 +49,6 @@ enum CommandType {
     COMMAND_TYPE_UPSERT_WORKFLOW_SEARCH_ATTRIBUTES = 13;
     COMMAND_TYPE_PROTOCOL_MESSAGE = 14;
     COMMAND_TYPE_MODIFY_WORKFLOW_PROPERTIES = 16;
+    COMMAND_TYPE_SCHEDULE_NEXUS_OPERATION = 17;
+    COMMAND_TYPE_CANCEL_NEXUS_OPERATION = 18;
```
cretz, line 53:
```suggestion
    COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION = 18;
```

temporal/api/enums/v1/common.proto
```diff
@@ -70,3 +70,28 @@ enum CallbackState {
...
+    // Operation was started and will complete asynchronously.
+    NEXUS_OPERATION_STATE_STARTED = 4;
+    // Operation succeeded.
+    NEXUS_OPERATION_STATE_SUCCEEDED = 5;
+    // Operation failed.
+    NEXUS_OPERATION_STATE_FAILED = 6;
+    // Operation completed as canceled (may have not ever been delivered).
+    NEXUS_OPERATION_STATE_CANCELED = 7;
+    // Operation timed out.
+    NEXUS_OPERATION_STATE_TIMEDOUT = 8;
```
cretz, line 94:
```suggestion
    NEXUS_OPERATION_STATE_TIMED_OUT = 8;
```

temporal/api/enums/v1/common.proto
```diff
@@ -70,3 +70,28 @@ enum CallbackState {
...
+    // This is a transient state which may never be visible to a client.
+    NEXUS_OPERATION_STATE_INITIALIZED = 1;
+    // Operation is in the queue waiting to be executed or is currently executing.
+    NEXUS_OPERATION_STATE_SCHEDULED = 2;
+    // Operation has failed with a retryable error and is backing off before the next attempt.
+    NEXUS_OPERATION_STATE_BACKING_OFF = 3;
+    // Operation was started and will complete asynchronously.
+    NEXUS_OPERATION_STATE_STARTED = 4;
+    // Operation succeeded.
+    NEXUS_OPERATION_STATE_SUCCEEDED = 5;
```
cretz, line 88:
```suggestion
    NEXUS_OPERATION_STATE_COMPLETED = 5;
```

With events and others, we seem to be calling this state "completed" not "succeeded"

temporal/api/enums/v1/common.proto
```diff
@@ -70,3 +70,28 @@ enum CallbackState {
...
     CALLBACK_STATE_SUCCEEDED = 5;
 }
+
+// State of a Nexus operation.
+enum NexusOperationState {
+    // Default value, unspecified state.
+    NEXUS_OPERATION_STATE_UNSPECIFIED = 0;
+    // Operation has just been initialized, waiting to be scheduled.
+    // This is a transient state which may never be visible to a client.
+    NEXUS_OPERATION_STATE_INITIALIZED = 1;
```
cretz, line 80: To confirm though, this is a state it may be in? I would think that any Nexus operation goes straight to "scheduled".

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -861,6 +861,7 @@ message DescribeWorkflowExecutionResponse {
     repeated temporal.api.workflow.v1.PendingChildExecutionInfo pending_children = 4;
     temporal.api.workflow.v1.PendingWorkflowTaskInfo pending_workflow_task = 5;
     repeated temporal.api.workflow.v1.CallbackInfo callbacks = 6;
+    repeated temporal.api.workflow.v1.NexusOperationInfo pending_nexus_operations = 7;
```
cretz, line 864: I note this doesn't match the other "pending" whose info objects are specific to pending and don't include state enumerates for completion. Are you sure that you shouldn't call the object `PendingNexusOperationInfo` and the enum `PendingNexusOperationState` and get rid of the terminal states? If you _think_ you may want a more robust info, I think you should go ahead and map out the RPC that uses it. Otherwise, you have multiple unused enum values from an API POV.

temporal/api/enums/v1/event_type.proto
```diff
@@ -171,4 +171,19 @@ enum EventType {
...
+    // An asynchronous Nexus operation was started by a Nexus handler.
+    EVENT_TYPE_NEXUS_OPERATION_STARTED = 49;
+    // A Nexus operation completed successfully.
+    EVENT_TYPE_NEXUS_OPERATION_COMPLETED = 50;
+    // A Nexus operation failed.
+    EVENT_TYPE_NEXUS_OPERATION_FAILED = 51;
+    // A Nexus operation completed as canceled.
+    EVENT_TYPE_NEXUS_OPERATION_CANCELED = 52;
+    // A Nexus operation timed out.
+    EVENT_TYPE_NEXUS_OPERATION_TIMEDOUT = 53;
```
cretz, line 186:
```suggestion
    EVENT_TYPE_NEXUS_OPERATION_TIMED_OUT = 53;
```

temporal/api/history/v1/message.proto
```diff
@@ -825,6 +908,13 @@ message HistoryEvent {
         ActivityPropertiesModifiedExternallyEventAttributes activity_properties_modified_externally_event_attributes = 50;
         WorkflowPropertiesModifiedEventAttributes workflow_properties_modified_event_attributes = 51;
         WorkflowExecutionUpdateRequestedEventAttributes workflow_execution_update_requested_event_attributes = 52;
+        NexusOperationScheduledEventAttributes nexus_operation_scheduled_event_attributes = 53;
+        NexusOperationStartedEventAttributes nexus_operation_started_event_attributes = 54;
+        NexusOperationCompletedEventAttributes nexus_operation_completed_event_attributes = 55;
+        NexusOperationFailedEventAttributes nexus_operation_failed_event_attributes = 56;
+        NexusOperationCanceledEventAttributes nexus_operation_canceld_event_attributes = 57;
```
cretz, line 915:
```suggestion
        NexusOperationCanceledEventAttributes nexus_operation_canceled_event_attributes = 57;
```

temporal/api/history/v1/message.proto
```diff
@@ -825,6 +908,13 @@ message HistoryEvent {
...
+        NexusOperationTimedOutEventAttributes nexus_operation_timedout_event_attributes = 58;
```
cretz, line 916:
```suggestion
        NexusOperationTimedOutEventAttributes nexus_operation_timed_out_event_attributes = 58;
```

temporal/api/workflow/v1/message.proto
```diff
@@ -183,3 +183,28 @@ message CallbackInfo {
     // The time when the next attempt is scheduled.
     google.protobuf.Timestamp next_attempt_schedule_time = 8;
 }
+
+message NexusOperationInfo {
```
cretz, line 187: I am wondering if we should have a concept of worker identity reported upon poll with Nexus task handlers. It is useful information for activities and such, so it's probably useful here too.

temporal/api/command/v1/message.proto
bergundy: Not yet, but I suspect we'll need this later for tracing. Might as well just add it now.

temporal/api/enums/v1/command_type.proto
bergundy: Thanks.

temporal/api/enums/v1/common.proto
bergundy: Today it's not a state an operation will ever be in because it immediately transitions to the scheduled state.
In the future we may block operations from being scheduled if the queue per source namespace and destination service is backed up.

I was debating whether this should added here or not, still not sure if I want to initialize the state machine on the server in the UNSPECIFIED state or have this explicit state before transitioning to scheduled.

I'll have to get deeper into the implementation to decide.

temporal/api/enums/v1/common.proto
bergundy: Yeah, the Nexus protocol defines this as SUCCEEDED. I find it clearer to make this distinction but maybe it's better to take a consistency position.

temporal/api/enums/v1/common.proto
bergundy: Makes sense, thanks.

temporal/api/enums/v1/event_type.proto
bergundy: Thanks.

temporal/api/history/v1/message.proto
bergundy: 😬

temporal/api/history/v1/message.proto
bergundy: 👍

temporal/api/workflow/v1/message.proto
bergundy: This is information for the caller. The handler may not even be a worker and I'm not sure the handler side will want to expose worker identities.

I think I'll probably add the operation ID at minimum here.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Yeah, maybe these states should be internally defined in the server repo.
I'll probably want to delete the operation state machines from mutable state after completion.

bergundy (commented):
Thanks @cretz for the great feedback.

temporal/api/enums/v1/common.proto
cretz: :+1: I am fine with the difference here. It can be argued the event deserves to be named succeeded too, but you'll have to make the consistency break somewhere since Nexus and Temporal mismatch, so I'm fine with what's here currently.

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1: To me from an API POV a pending-specific info with a pending-specific state enum matches what we have and doesn't cause any problems with attempted inadvertent reuse of this info for anything else. (for sure actual server model may differ)

temporal/api/workflow/v1/message.proto
cretz: I see, due to the cross-namespace break, we should not report this worker information to the caller. However it does bring up the question, as a caller I can see outstanding Nexus operations, but can I on the incoming side? Maybe it's not a big deal for now since it's usually 1:1 with workflows.

bergundy: Not a big deal now, it's 1:1 with workflows and you'll be able to see that there are nexus callbacks attached to your workflow.

Later on we'll want better tracing abilities but I haven't gotten to that yet.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I'll remove the non-pending states.

cretz (commented):
This looks good to me (not marking a draft "approved" though)

temporal/api/command/v1/message.proto
bergundy: Done.

temporal/api/nexus/v1/message.proto
```diff
@@ -39,7 +38,7 @@ import "temporal/api/common/v1/message.proto";
 message Failure {
     string message = 1;
     map<string, string> metadata = 2;
-    google.protobuf.Value details = 3;
+    bytes details = 3;
```
bergundy, line 41: Note this is a backwards incompatible change but I realized it's simpler to just use JSON serialized to bytes here.

temporal/api/command/v1/message.proto
```diff
@@ -232,6 +232,37 @@ message ProtocolMessageCommandAttributes {
...
+    // Service name, resolved to a URL and a call config via the namespace's outgoing service registry.
+    string service = 1;
```
pdoerner, line 237: So services can only be referenced by name, not ID? Will an operation fail if a service is renamed between Scheduled and Completed?

pdoerner (approved):

rodrigozhou (approved):
LGTM, please address @pdoerner comments.

temporal/api/command/v1/message.proto
bergundy: Outgoing services are referenced only by name and we don't support renaming because of that.
Services can be deleted and in that case all operations referencing these services will fail.

temporal/api/enums/v1/common.proto
```diff
@@ -70,3 +70,31 @@ enum CallbackState {
...
+    // Operation is in the queue waiting to be executed or is currently executing.
+    PENDING_NEXUS_OPERATION_STATE_SCHEDULED = 1;
+    // Operation has failed with a retryable error and is backing off before the next attempt.
+    PENDING_NEXUS_OPERATION_STATE_BACKING_OFF = 2;
+    // Operation was started and will complete asynchronously.
+    PENDING_NEXUS_OPERATION_STATE_STARTED = 3;
+}
+
+// State of a Nexus operation cancelation.
+enum NexusOperationCancelationState {
```
cretz, line 87: While we are a bit inconsistent on "cancelled" vs "canceled", we are consistently using `Cancellation` with two `L`s in our protos

temporal/api/workflow/v1/message.proto
```diff
@@ -229,3 +229,58 @@ message CallbackInfo {
...
+    int32 attempt = 7;
+
+    // The time when the last attempt completed.
+    google.protobuf.Timestamp last_attempt_complete_time = 8;
+    // The last attempt's failure, if any.
+    temporal.api.failure.v1.Failure last_attempt_failure = 9;
+    // The time when the next attempt is scheduled.
+    google.protobuf.Timestamp next_attempt_schedule_time = 10;
+
+    NexusOperationCancelationInfo cancelation_info = 11;
```
cretz, line 266: We don't seem to report this information for activities today, are we sure we have to put it on Nexus operations at this time?

temporal/api/workflow/v1/message.proto
```diff
@@ -229,3 +229,58 @@ message CallbackInfo {
...
+    // The number of attempts made to deliver the start operation request.
+    // This number represents a minimum bound since the attempt is incremented after the request completes.
+    int32 attempt = 7;
+
+    // The time when the last attempt completed.
+    google.protobuf.Timestamp last_attempt_complete_time = 8;
+    // The last attempt's failure, if any.
+    temporal.api.failure.v1.Failure last_attempt_failure = 9;
+    // The time when the next attempt is scheduled.
+    google.protobuf.Timestamp next_attempt_schedule_time = 10;
```
cretz, line 264: We don't seem to report these 3 values for activities today, are we sure we have to put it on Nexus operations at this time?

temporal/api/command/v1/message.proto
```diff
@@ -232,6 +232,37 @@ message ProtocolMessageCommandAttributes {
...
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 4;
+
+    // Header to attach to the Nexus request.
+    // Do not put sensitive information in this header as it is stored in workflow history and transmitted to external
+    // services as-is.
+    // This is useful propagating tracing information.
+    // Note that it is not using temporal.api.common.v1.Header because the Nexus protocol only supports string header
+    // values.
+    map<string, string> header = 5;
```
cretz, line 257: > Note that it is not using temporal.api.common.v1.Header because the Nexus protocol only supports string header values

We could only accept string payloads, but payloads are important because they imply encrypt-ability and plugin to every SDKs encryption tooling. Doing this as just strings means callers have to use separate encryption tooling which is a lot to ask inside of a workflow.

I wonder if we can say: "json/plain" strings are unwrapped and used as strings, all other "json/*" values are just bytes converted to strings, and all other values are just base64'd bytes.

While it may be worth thinking the SDK side can just accept strings and go through the data converter which will encrypt them, but then turn them back to strings here, several users have proxies that walk protos and encrypt all payloads.

temporal/api/nexus/v1/message.proto
cretz: How is this set by Nexus handlers and the Nexus HTTP protocol?

temporal/api/nexus/v1/message.proto
```diff
@@ -184,4 +183,10 @@ message OutgoingService {
 message OutgoingServiceSpec {
     // URL to invoke requests.
     string url = 1;
+    // Callback URL that is attached to every StartOperation request for this service.
+    // The URL should be accessable to the handler of the StartOperation requests.
```
cretz, line 187:
```suggestion
    // The URL should be accessible to the handler of the StartOperation requests.
```

temporal/api/nexus/v1/message.proto
```diff
@@ -184,4 +183,10 @@ message OutgoingService {
...
+    // The service handles callbacks in the /api/v1/namespaces/{namespace_name}/nexus/callback route. Note that if the
```
cretz, line 188: There is discussion of removing the `/api/v1` prefix from our HTTP API annotations (it is a bit redundant/unnecessary). This decision should happen soon, don't bake too many Nexus expectations on this prefix yet.

cretz (commented):
How far along is the implementation? Is it possible to not put these history-affecting protos into `master` until the implementation is basically complete server side?

temporal/api/command/v1/message.proto
bergundy: > I wonder if we can say: "json/plain" strings are unwrapped and used as strings, all other "json/*" values are just bytes converted to strings, and all other values are just base64'd bytes.

I'm open to this idea but it would require additional metadata to decode and would complicate the integration between Temporal Nexus and "plain" Nexus.

> several users have proxies that walk protos and encrypt all payloads.

I'm aware, I didn't expect users to put sensitive information in here.
AFAIR, our SDK headers don't go through codecs today and it doesn't seem like it's been an issue yet although I anticipate it may become an issue if they propagate baggage with their traces.

temporal/api/nexus/v1/message.proto
bergundy: In Go there's a `nexus.Failure` object with a `details` field that's `json.RawMessage`.

https://github.com/nexus-rpc/sdk-go/blob/506896b352dacee3b5d7d62010a87bc08b2b8fd1/nexus/api.go#L48-L55

Here's how it's defined in the HTTP spec: https://github.com/nexus-rpc/api/blob/main/SPEC.md#failure

temporal/api/workflow/v1/message.proto
bergundy: I think this is important to be able to understand how your operation is progressing.
Activities are run by _your_ workers, operations requests are made by the server, there's no other means of inspecting the status.

temporal/api/workflow/v1/message.proto
bergundy: Same as above.

temporal/api/nexus/v1/message.proto
cretz: :+1: I agree that `bytes` is better than structured value here. May want to document that it is likely JSON based on SDK usage.

temporal/api/workflow/v1/message.proto
cretz: Arguably it's important for activities too and I can give obvious reasons why, but we don't wait on whether I think it's important, we wait on whether users do.

temporal/api/nexus/v1/message.proto
cretz: :+1: Just a heads up so there are no implementation-side expectations inside server dev

bergundy:
> How far along is the implementation? Is it possible to not put these history-affecting protos into `master` until the implementation is basically complete server side?

Implementation is pretty much done but not yet production ready.
I have a feature branch in the server that I want to merge pending this PR.

temporal/api/command/v1/message.proto
bergundy: I'm not convinced it should just work by default, it all depends on how much context you want to share with the party you're calling, which may or may not be the same org and using the same encryption keys.

temporal/api/workflow/v1/message.proto
bergundy: Activity has last started and last failure.
Operations don't record attempt starts, only completions, so it's basically exposing the same amount of information but in different format.

temporal/api/enums/v1/common.proto
bergundy: I knew we were inconsistent but never noticed we we writing `Cancellation` with two `L`s.
I'm totally okay changing for "consistency".

cretz: :+1: Yeah we're consistent on cancellation, it's canceled we're inconsistent on

temporal/api/command/v1/message.proto
cretz: Header propagation does not work by default, users must write it. No keys are auto-forwarded.

cretz: After off-PR discussion, Temporal headers are not going to be Nexus headers at this time unfortunately. But we should consider that in the future.

temporal/api/command/v1/message.proto
```diff
@@ -232,6 +232,37 @@ message ProtocolMessageCommandAttributes {
...
+    // Schedule-to-close timeout for this operation.
+    // Indicates how long the caller is willing to wait for operation completion.
+    // Calls are retried internally by the server.
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 4;
+
+    // Header to attach to the Nexus request.
+    // Do not put sensitive information in this header as it is stored in workflow history and transmitted to external
+    // services as-is.
```
cretz, line 253:
```suggestion
    // Users are responsible for encrypting sensitive data in this header as it is stored in workflow history and transmitted to external
    // services as-is.
```

There is a use-case for putting sensitive data here, they just have to be careful

temporal/api/command/v1/message.proto
```diff
@@ -232,6 +232,37 @@ message ProtocolMessageCommandAttributes {
...
+    // This is useful for propagating tracing information.
+    // Note that it is not using temporal.api.common.v1.Header because the Nexus protocol only supports string header
+    // values.
```
cretz, line 256:
```suggestion
    // Note these headers are not the same as Temporal headers on internal activities and child workflows, these are transmitted to Nexus operations that may be external and are not traditional payloads.
```

Clarifying why you didn't use the model I think is different than clarifying behavior

temporal/api/history/v1/message.proto
```diff
@@ -794,6 +794,84 @@ message WorkflowExecutionUpdateAdmittedEventAttributes {
...
+    // internally when sending the StartOperation request. On the handler side, if it is also backed by Temporal, the
+    // content is transformed back to the original Payload stored in this event.
+    temporal.api.common.v1.Payload input = 3;
+    // Schedule-to-close timeout for this operation.
+    // Indicates how long the caller is willing to wait for operation completion.
+    // Calls are retried internally by the server.
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 4;
+    // Header to attach to the Nexus request.
```
cretz, line 813:
```suggestion
    // Header to attach to the Nexus request. Note these headers are not the same as Temporal headers on internal activities and child workflows, these are transmitted to Nexus operations that may be external and are not traditional payloads.
```

Might as well add the same note as in the command

cretz (approved):
LGTM. I do think it's unfortunate that Temporal headers can't be used.

PR #364, opened 2024-03-08
Add batch api for signal with start workflow

alexseedkou:
**What changed?**
Add batch operation API for signal with start workflow.

**Why?**

Add batch API for signal with start workflow as the request [here](https://github.com/temporalio/temporal/issues/1036)
For the batch API for signal workflow, it is addressed by this [PR](https://github.com/temporalio/api/pull/197) if my understanding is correct

**Breaking changes**

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -2543,6 +2543,9 @@
         "signalOperation": {
           "$ref": "#/definitions/v1BatchOperationSignal"
         },
+        "signalStartWorkflowOperation": {
+          "$ref": "#/definitions/v1BatchOperationSignalStartWorkflow"
+        },
         "cancellationOperation": {
           "$ref": "#/definitions/v1BatchOperationCancellation"
         },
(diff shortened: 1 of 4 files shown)


stephanos:
I think there's a misunderstanding here. The issue you've linked to talks about sending a batch (or a set) of signals to a Workflow. Signal-With-Start only allows sending a single signal.

The Batch API, though, is something else. It is querying _existing_ Workflows and executing an operation on them. Signal-With-Start doesn't really make sense here.

Note that we are actively developing a new API that will likely eventually support this feature. So we are not inclined to add another option at this time.

I'm going to close the PR for now.

alexseedkou:
> Thank you for your PR.
> 
> I think there's a misunderstanding here. The issue you've linked to talks about sending a batch (or a set) of signals to a Workflow, since Signal-With-Start only allows sending a single signal.
> 
> The Batch API, though, is something else. It is querying _existing_ Workflows and executing an operation on them. Signal-With-Start doesn't really make sense here.
> 
> Note that we are actively developing a new API that will likely eventually support sending multiple signals. So we are not inclined to add another option at this time.
> 
> I'm going to close the PR for now.

Thanks for your feedback.

Do we have any plan to address the request in the link with the coming change? If no, can I work on that?

PR #365, opened 2024-03-11
Add target_ prefix for CreateOrUpdateNexusIncomingServiceRequest fields

bergundy:
**What changed?**

See title.

**Why?**

Server considers requests with a namespace field as being scoped to that namespace which this request should be cluster scoped.

**Breaking changes**

Yes


dnr (approved):

bergundy:
> LGTM assuming open source only. Would like to set a general suggestion of namespace of all requests being field 1 for compat assistance for edge proxies, but since this is open source only it's not a big deal.
> 
> Can you confirm all workflowservice calls for Nexus are `string namespace = 1`?

I didn't add any more workflowservice RPCs, but the operatorservice outgoing registry RPCs all include a `string namespace = 1`:

https://github.com/temporalio/api/pull/363/files#diff-43cdd5e82c4820416a2f6b04aa800be6229b4790d74e71f9873e608a3382c442R188

PR #366, opened 2024-03-11
Add root workflow execution to WorkflowExecutionStartedEventAttributes

rodrigozhou:
**What changed?**
Add fields for root workflow execution.

The root workflow execution is defined as follows:
1. A workflow without parent workflow is its own root workflow.
2. A workflow that has a parent workflow has the same root workflow as its parent workflow.

Examples:
- Scenario 1: Workflow W1 starts child workflow W2, and W2 starts child workflow W3.
  - The root workflow of all three workflows is W1.
- Scenario 2: Workflow W1 starts child workflow W2, and W2 continued as new W3.
  - The root workflow of all three workflows is W1.
- Scenario 3: Workflow W1 continued as W2.
  - The root workflow of W1 is W1 and the root workflow of W2 is W2.

**Why?**

**Breaking changes**

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -107,6 +107,25 @@ message WorkflowExecutionStartedEventAttributes {
 
     // Completion callbacks attached when this workflow was started.
     repeated temporal.api.common.v1.Callback completion_callbacks = 30;
+
+    // Contains information about the root workflow execution.
+    // The root workflow execution is defined as follows:
+    // 1. A workflow without parent workflow is its own root workflow.
+    // 2. A workflow that has a parent workflow has the same root workflow as its parent workflow.
+    // Note: workflows continued as new or reseted may or may not have parents, check examples below.
+    //
+    // Examples:
+    //   Scenario 1: Workflow W1 starts child workflow W2, and W2 starts child workflow W3.
+    //     - The root workflow of all three workflows is W1.
+    //   Scenario 2: Workflow W1 starts child workflow W2, and W2 continued as new W3.
+    //     - The root workflow of all three workflows is W1.
+    //   Scenario 3: Workflow W1 continued as new W2.
+    //     - The root workflow of W1 is W1 and the root workflow of W2 is W2.
+    //   Scenario 4: Workflow W1 starts child workflow W2, and W2 is reseted, creating W3
+    //     - The root workflow of all three workflows is W1.
+    //   Scenario 5: Workflow W1 is reseted, creating W2.
+    //     - The root workflow of W1 is W1 and the root workflow of W2 is W2.
+    temporal.api.common.v1.WorkflowExecution root_workflow_execution = 31;
 }
 
 message WorkflowExecutionCompletedEventAttributes {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -7848,6 +7848,10 @@
         "executionDuration": {
           "type": "string",
           "description": "Workflow execution duration is defined as difference between close time and execution time.\nThis field is only populated if the workflow is closed."
+        },
+        "rootExecution": {
+          "$ref": "#/definitions/v1WorkflowExecution",
+          "description": "Contains information about the root workflow execution.\nThe root workflow execution is defined as follows:\n1. A workflow without parent workflow is its own root workflow.\n2. A workflow that has a parent workflow has the same root workflow as its parent workflow.\nNote: workflows continued as new or reseted may or may not have parents, check examples below.\n\nExamples:\n  Scenario 1: Workflow W1 starts child workflow W2, and W2 starts child workflow W3.\n    - The root workflow of all three workflows is W1.\n  Scenario 2: Workflow W1 starts child workflow W2, and W2 continued as new W3.\n    - The root workflow of all three workflows is W1.\n  Scenario 3: Workflow W1 continued as new W2.\n    - The root workflow of W1 is W1 and the root workflow of W2 is W2.\n  Scenario 4: Workflow W1 starts child workflow W2, and W2 is reseted, creating W3\n    - The root workflow of all three workflows is W1.\n  Scenario 5: Workflow W1 is reseted, creating W2.\n    - The root workflow of W1 is W1 and the root workflow of W2 is W2."
         }
       }
     },
@@ -7999,6 +8003,10 @@
             "$ref": "#/definitions/v1Callback"
           },
           "description": "Completion callbacks attached when this workflow was started."
+        },
+        "rootWorkflowExecution": {
+          "$ref": "#/definitions/v1WorkflowExecution",
+          "description": "Contains information about the root workflow execution.\nThe root workflow execution is defined as follows:\n1. A workflow without parent workflow is its own root workflow.\n2. A workflow that has a parent workflow has the same root workflow as its parent workflow.\nNote: workflows continued as new or reseted may or may not have parents, check examples below.\n\nExamples:\n  Scenario 1: Workflow W1 starts child workflow W2, and W2 starts child workflow W3.\n    - The root workflow of all three workflows is W1.\n  Scenario 2: Workflow W1 starts child workflow W2, and W2 continued as new W3.\n    - The root workflow of all three workflows is W1.\n  Scenario 3: Workflow W1 continued as new W2.\n    - The root workflow of W1 is W1 and the root workflow of W2 is W2.\n  Scenario 4: Workflow W1 starts child workflow W2, and W2 is reseted, creating W3\n    - The root workflow of all three workflows is W1.\n  Scenario 5: Workflow W1 is reseted, creating W2.\n    - The root workflow of W1 is W1 and the root workflow of W2 is W2."
         }
       },
       "title": "Always the first event in workflow history"
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -5766,6 +5766,27 @@ components:
           description: |-
             Workflow execution duration is defined as difference between close time and execution time.
              This field is only populated if the workflow is closed.
+        rootExecution:
+          allOf:
+            - $ref: '#/components/schemas/WorkflowExecution'
+          description: |-
+            Contains information about the root workflow execution.
+             The root workflow execution is defined as follows:
+             1. A workflow without parent workflow is its own root workflow.
+             2. A workflow that has a parent workflow has the same root workflow as its parent workflow.
+             Note: workflows continued as new or reseted may or may not have parents, check examples below.
+
+             Examples:
+               Scenario 1: Workflow W1 starts child workflow W2, and W2 starts child workflow W3.
+                 - The root workflow of all three workflows is W1.
+               Scenario 2: Workflow W1 starts child workflow W2, and W2 continued as new W3.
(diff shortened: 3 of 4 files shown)


dandavison:
Could you add some documentation to the code explaining what the fields are and how they should be used? (Also the PR itself doesn't have any explanation, but that could come from the code comments in a proto API PR).

temporal/api/history/v1/message.proto
```diff
@@ -107,6 +107,10 @@ message WorkflowExecutionStartedEventAttributes {
 
     // Completion callbacks attached when this workflow was started.
     repeated temporal.api.common.v1.Callback completion_callbacks = 30;
+
+    string root_workflow_namespace = 31;
+    string root_workflow_namespace_id = 32;
+    temporal.api.common.v1.WorkflowExecution root_workflow_execution = 33;
```
cretz, line 113:
```suggestion
    string root_namespace = 31;
    string root_execution_run_id = 32;
```

So I think this makes sense to me. We don't store the namespace today and we should, and we don't store the first workflow in the chain and we should. Questions:

* Do we have to have the namespace_id?
* Do we have to have the word `workflow` in the field names?
* Do we need the workflow ID in the execution? When would it ever be different from field 28 here?
* Can you provide user documentation/comments above these fields for what they are?

temporal/api/history/v1/message.proto
```diff
@@ -107,6 +107,8 @@ message WorkflowExecutionStartedEventAttributes {
 
     // Completion callbacks attached when this workflow was started.
     repeated temporal.api.common.v1.Callback completion_callbacks = 30;
+
+    temporal.api.common.v1.WorkflowExecution root_workflow_execution = 31;
```
cretz, line 111: Had some other questions on another comment, but stuff has changed since. Questions:

* Do we need the workflow ID in the execution? When would it ever be different from field 28 here?
* Do we need the run ID in the execution? When would it ever be different from field 16 here?
* Can you provide user documentation/comments above these fields for what they are?

temporal/api/history/v1/message.proto
rodrigozhou: I removed the namespace fields because it's always the same as the current workflow execution.

temporal/api/history/v1/message.proto
rodrigozhou: I added a comment explaining the field. Let me know if it's not clear enough.

temporal/api/history/v1/message.proto
```diff
@@ -107,6 +107,9 @@ message WorkflowExecutionStartedEventAttributes {
 
     // Completion callbacks attached when this workflow was started.
     repeated temporal.api.common.v1.Callback completion_callbacks = 30;
+
+    // Root workflow execution corresponds to the last node in the parent workflow execution chain.
+    temporal.api.common.v1.WorkflowExecution root_workflow_execution = 31;
```
dandavison, line 112: It's not clear to me what "last" means here. Is a "parent workflow execution chain" the tree structure that results from child workflow relationships? The PR should have an introduction that explains this (or should link to a PR or issue where it's explained).

temporal/api/history/v1/message.proto
cretz: Unresolving. You missed my other two questions.

temporal/api/history/v1/message.proto
cretz: You missed the third question here

temporal/api/history/v1/message.proto
rodrigozhou: Both workflow ID and run ID are needed as they can be different from the fields you mentioned. Root workflow execution is related to the child workflow execution, not continue as new.

temporal/api/history/v1/message.proto
rodrigozhou: Updated PR with detailed explanation of root workflow execution.
Let me know if it's clear now.

temporal/api/history/v1/message.proto
```diff
@@ -107,6 +107,20 @@ message WorkflowExecutionStartedEventAttributes {
...
+    // 2. A workflow that has a parent workflow has the same root workflow as its parent workflow.
+    //
+    // Examples:
+    //   Scenario 1: Workflow W1 starts child workflow W2, and W2 starts child workflow W3.
+    //     - The root workflow of all three workflows is W1.
+    //   Scenario 2: Workflow W1 starts child workflow W2, and W2 continued as new W3.
+    //     - The root workflow of all three workflows is W1.
+    //   Scenario 3: Workflow W1 continued as W2.
+    //     - The root workflow of W1 is W1 and the root workflow of W2 is W2.
+    temporal.api.common.v1.WorkflowExecution root_workflow_execution = 31;
```
dnr, line 123: I think some of this may have been discussed last week but just to include it on the PR:

How does this interact with reset? In general we don't have "current run id" on the WorkflowExecutionStartedEvent or available to workflow logic because it changes on reset, and we fork history so we can't rewrite event 1. It seems like this would have the same problem, i.e.:
- Workflow W1 runs 10 tasks.
- It gets reset to task 5, creating W1b.
- Question: based on the definition, the root workflow of W1b is W1b, since it has no parent. But its start event will contain W1, not W1b.
- After 5 more tasks, W1b starts child W2.
- Question: What is the root of W2, is it W1 or W1b?

yycptt: 👍 Reset is also the reason why current workflow's runID is not stored in the started event. 
It feels like we need some special representation when root workflow is itself, like nil. If we want a better customer experience, when this field is nil, server it populates it with the right runID before returning the result. (this will make server logic more complex though)

re. reset + start child, I think it should be W1b.

cretz, line 123: Traditionally people have manually walked the chain to get parents (and still have to even with this change because this only gives top-level while increasing history size). Can you help me understand the impetus for this addition? Just trying to get an idea of the use case.

temporal/api/history/v1/message.proto
```diff
@@ -107,6 +107,20 @@ message WorkflowExecutionStartedEventAttributes {
...
+    // 1. A workflow without parent workflow is its own root workflow.
+    // 2. A workflow that has a parent workflow has the same root workflow as its parent workflow.
+    //
+    // Examples:
+    //   Scenario 1: Workflow W1 starts child workflow W2, and W2 starts child workflow W3.
+    //     - The root workflow of all three workflows is W1.
+    //   Scenario 2: Workflow W1 starts child workflow W2, and W2 continued as new W3.
+    //     - The root workflow of all three workflows is W1.
+    //   Scenario 3: Workflow W1 continued as W2.
+    //     - The root workflow of W1 is W1 and the root workflow of W2 is W2.
```
dandavison, line 122: Should scenario 3 say "continued as new W2"? Why is root(W2) = W2 in scenario 3 yet root(W3) = W1 in scenario 1? Sorry, I have not read any design document or proposal regarding the new concepts being introduced here; I'm just reviewing as a naive reader of these comments.

rodrigozhou: In scenario 2, the parent of W3 is W2.
In scenario 3, both W1 and W2 have no parents.

dandavison: What is the difference between "continued as new" and "continued as"?

> W2 continued as new W3
> W1 continued as W2

temporal/api/history/v1/message.proto
cretz: Based on what I heard of the original use case, just adding this "root" execution to the main info is enough to satisfy the use case

cretz (approved):
Approving from API POV, but would like to see reset behavior on this field documented (i.e. that it doesn't change across reset if that's the behavior)

temporal/api/history/v1/message.proto
rodrigozhou: That's my typo, I'll fix it before merging.

dandavison: So

```
Scenario 2: W1 --child--> W2 --CAN--> W3
Scenario 3:               W1 --CAN--> W2
```

If `root(W3) = W1` in scenario 2, then I would expect `root(W2) = W1` in scenario 3.

In other words, what I'm inferring from the comments is that both `startChildWorkflow` and `CAN` create a parent-child relationship. (If that is correct then it would be helpful to say that explicitly.)

temporal/api/history/v1/message.proto
```diff
@@ -107,6 +107,20 @@ message WorkflowExecutionStartedEventAttributes {
 
     // Completion callbacks attached when this workflow was started.
     repeated temporal.api.common.v1.Callback completion_callbacks = 30;
+
+    // Contains information about the root workflow execution.
+    // The root workflow execution is defined as follows:
+    // 1. A workflow without parent workflow is its own root workflow.
+    // 2. A workflow that has a parent workflow has the same root workflow as its parent workflow.
```
dandavison, line 114: OK I think I understand. I'm thinking that adding this would help:
```suggestion
    // 2. A workflow that has a parent workflow has the same root workflow as its parent workflow.
    // 3. When a workflow continues-as-new, the parent is retained.
```

temporal/api/history/v1/message.proto
dandavison: Thanks, I think I understand now. I've suggested an addition above.

rodrigozhou: The better visualization is in tree structure:

Scenario 2:
```
   W1
  /
 /
W2 -CAN-> W3
```

Scenario 3:
```
W1 -CAN-> W2
```

In scenario 2, the parent of W3 is W1. Therefore, the root of W3 is W1.
In scenario 3, W2 inherit the parent of W1 which is nil. So, W2 doesn't have a parent, and therefore is its own root.

The definition of root workflow is tied to the definition of parent workflow.

dandavison (approved):
Thanks for explaining! I added an optional suggestion to make it explicit that if `A--CAN-->B` then `parent(B) = parent(A)`.

temporal/api/history/v1/message.proto
rodrigozhou: I'll add as a note. I don't think this would be part of the definition.
I think the definition given by rules 1 and 2 is complete.
Maybe we should clarify in the definition of parent workflow and how this works with continue as new.

temporal/api/history/v1/message.proto
rodrigozhou: Hm, that's an interesting case. I agree that if the top level workflow is reseted, then the new workflow should be its own root.
@yiminc Thoughts?

yiminc: sgtm

temporal/api/history/v1/message.proto
dandavison: Right, I agree that it's not clear that it's required in the definition.

temporal/api/history/v1/message.proto
yycptt: I think today we expose parent information to workflow logic (`workflow.GetInfo(ctx)`). Will we do the same thing for this root execution? 

If so, we'll have to populate those fields before returning the result to SDK. Can be done later as a separate feature though.

rodrigozhou:
The implementation in the Server is here: https://github.com/temporalio/temporal/pull/5520

PR #367, opened 2024-03-12
Add ExecuteMultiOperation

stephanos:
**What changed?**

Adding a new "ExecuteMultiOperation" API that will support sending a combined Start + Update.

The first version of the Server implementation can be found here: https://github.com/temporalio/temporal/pull/5577

Note that while the API has a working implementation, it might still change. It is therefore annotated with "NOTE: Experimental API".

**Why?**

New API for sending operations to the Server atomically.

**Breaking changes**


google/rpc/code.proto
stephanos, line 1: Copy-pasted from https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1428,3 +1429,27 @@ message RespondNexusTaskFailedRequest {
...
+    }
+}
+
+message ExecuteMultiOperationResponse {
+    repeated WorkflowOperationResult results = 1;
+}
+
+message WorkflowOperationResult {
+    oneof result {
+        google.rpc.Status error = 1;
```
cretz, line 1451: I think multi-operation failure should be a gRPC error.

I can't imagine a continue-on-error future (or even a non-transactional future) where this has value reported as a successful gRPC result with a failure embedded. I think for now we should consider multi-operation transactional (i.e. all fails or all succeeds) as that's where it has most value, and use gRPC errors to report failure. If we need to be more detailed about the error for the user, we can add error details which can include partial successes if we ever have that (I hope not, it's confusing and has limited value). Even if we did have the concept of partial completion at some point in the future, I think it'd be up to request options on whether error should be ignored, but it's important IMO to let gRPC failures be gRPC failures.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -107,6 +107,16 @@ service WorkflowService {
         };
     }
 
+    // ExecuteMultiOperation executes multiple operations.
```
cretz, line 110: Can we specify the transactional expectations of this call? Basically guarantee it all fails or all succeeds atomically/transactionally?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1428,3 +1429,27 @@ message RespondNexusTaskFailedRequest {
 
 message RespondNexusTaskFailedResponse {
 }
+
+message ExecuteMultiOperationRequest {
+    string namespace = 1;
+    repeated WorkflowOperation operations = 2;
```
cretz, line 1435: Can you clarify in comments here the exact constraints the initial implementation will have?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -107,6 +107,16 @@ service WorkflowService {
...
+    //
+    // NOTE: Experimental API.
+    rpc ExecuteMultiOperation (ExecuteMultiOperationRequest) returns (ExecuteMultiOperationResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/workflows/operations"
```
cretz, line 115:
```suggestion
            post: "/api/v1/namespaces/{namespace}/workflows/execute-operations"
```

I know it may seem a bit weird, but I think this should be `execute-operations`. This is not a referenceable resource you're creating, so I think it should be a verb for the call it's making. If we do ever have the concept of "operations" as a listable resource, that would make more sense. From https://github.com/temporalio/proposals/blob/master/api/http-api.md:

> * POST for everything else
>   * Noun as last URL term if for a create, verb if for an action

stephanos: Great point!

temporal/api/workflowservice/v1/service.proto
stephanos: Okay, I took a stab at it. I'm not quite sure yet how to express the subtleties in error behavior: when the Update fails to be admitted, there actually won't be a new Workflow (akin to a transaction). However, if the Update doesn't reach Accepted/Completed, only the Update operation will fail but not the Start. So it actually has _limited_ transactional guarantees.

temporal/api/workflowservice/v1/request_response.proto
stephanos: Okay, I added some clarification.

Sushisource: What does it mean if the message has multiple `StartWorkflow`s and they specify different things?

We should make some statement about what takes precedence (or just fail the call, which is probably preferable).

temporal/api/workflowservice/v1/request_response.proto
stephanos: I've alludes to it in the other comment; the issue here is that once the Update is admitted, the Workflow **needs** to start or the wait stage accepted/completed would never be reached. And once the Workflow has started, it cannot be undone (as of yet).

cretz: Update failure/rejection outcome is already on the successful update response. A gRPC failure would be failure to admit an update which should also fail to start the workflow (same as failure to admit a signal on signal with start).

temporal/api/workflowservice/v1/request_response.proto
stephanos: I've added
```
- The workflow ids must match across operations.
- The only valid set of operations is [StartWorkflow, UpdateWorkflow].
```
Is that sufficient?

temporal/api/workflowservice/v1/service.proto
cretz: > However, if the Update doesn't reach Accepted/Completed, only the Update operation will fail but not the Start. So it actually has limited transactional guarantees.

This depends on how you define fail. It won't fail from a gRPC perspective. I don't consider an update being rejected as an update "failing" from the POV of this API. If an update doesn't reach accepted/completed, the update response is capable of representing that. Still, that the start and update were accepted by the server and the update reached a state of durability, then the gRPC call succeeded, even if a single update had an application-level failure (not to be confused with server/gRPC failure).

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1:, though for me changing the last bullet to instead read:

```
- The only valid set of operations at this time is [StartWorkflow, UpdateWorkflow] in that order.
```

Just to clarify that this we're intentionally adding this limitation just at first, and that order matters.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1428,3 +1428,33 @@ message RespondNexusTaskFailedRequest {
...
+    // List of operations to execute within a single workflow.
+    // Preconditions:
+    // - The list of operations must not be empty.
+    // - The workflow ids must match across operations.
+    // - The only valid list of operations is [StartWorkflow, UpdateWorkflow] (in this order).
+    // - Update operation must be durable.
+    repeated WorkflowOperation operations = 2;
+}
+
+message WorkflowOperation {
```
cretz, line 1444: Sorry I didn't notice this earlier, would you consider moving this to a nested message of `ExecuteMultiOperationRequest` and same for result? This allows us to keep all top-level messages of this request/response file as requests/responses. Feel free to remove the `Workflow` prefix since it'll be plenty qualified.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1428,3 +1428,33 @@ message RespondNexusTaskFailedRequest {
...
+
+message ExecuteMultiOperationResponse {
+    repeated WorkflowOperationResult results = 1;
+}
+
+message WorkflowOperationResult {
+    oneof result {
+        StartWorkflowExecutionResponse start_workflow = 1;
+        UpdateWorkflowExecutionResponse update_workflow = 2;
+    }
```
cretz, line 1459: I don't have a strong opinion on the mixed use of "result" and "response" here, though may be worth using the word "response" everywhere.

temporal/api/workflowservice/v1/request_response.proto
stephanos: I had it that way at first, but I was worried that that would limited the reusability of `WorkflowOperation`. It's my understanding that they cannot be moved out after the fact and would always have to be used with the fully qualified path. Granted, there are no reusability plans right this moment, but I still thought this was the better tradeoff. I don't have strong feelings here, just wanted to bring this up.

temporal/api/workflowservice/v1/request_response.proto
stephanos: I had considered using "response" here at first but wasn't sure it was the right way to describe these nested values as response in this proto context implies "network call" and I didn't want to give the wrong impression to users that these are separate network call responses.

temporal/api/workflowservice/v1/request_response.proto
cretz: > I had it that way at first, but I was worried that that would limited the reusability of WorkflowOperation. 

If you're that worried, it can be placed in any other package. But this package's top-level messages are only requests and responses. I am not that worried though.

> It's my understanding that they cannot be moved out after the fact and would always have to be used with the fully qualified path.

It can. It's believe it is both wire compatible and JSON compatible to move a message. It is not code-gen compatible, but we don't guarantee that compatibility I don't believe.

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1: Up to you/others, I don't mind either way, just bringing up.

temporal/api/workflowservice/v1/request_response.proto
stephanos: > Update failure/rejection outcome is already on the successful update response. A gRPC failure would be failure to admit an update which should also fail to start the workflow (same as failure to admit a signal on signal with start).

You're right, I've come to see it the same way 👍 

> Also, I don't believe that multi-operation update should wait until update reaches a certain state, or that a worker even needs to run. I believe update "durably admitted" (which I heard we're adding support for so update-with-start works) is when it should return.

I hadn't considered this before. We are considering "durably admitted" now (approval pending) indeed. It's my understanding that being able to wait for accepted/completed allows the client to receive the response in a single gRPC call instead of having to initiate a poll.

cretz: > It's my understanding that being able to wait for accepted/completed allows the client to receive the response in a single gRPC call instead of having to initiate a poll.

This is my understanding too, I just think a supported wait-stage should be admitted with this call. One of the great aspects of Temporal is that when I submit something durably I know it will run, I don't have to wait for it to run. We may not be offering this for update, but since we're offering it for update-with-start, we should allow users to return without a worker running (i.e. basically immediately once made durable), same as signal-with-start. But I also support the ability to wait longer for update response for sure.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1428,3 +1428,33 @@ message RespondNexusTaskFailedRequest {
 
 message RespondNexusTaskFailedResponse {
 }
+
+message ExecuteMultiOperationRequest {
```
cretz, line 1432: Does this need a request ID for idempotency? What if I retry this call? What if I retry this call but workflow has already started and I had that workflow configured to fail-existing?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -107,6 +107,24 @@ service WorkflowService {
         };
     }
 
+    // ExecuteMultiOperation executes multiple operations within a single workflow.
+    //
+    // Operations are processed transactionally, meaning if *any* operation fails to be admitted, none are executed,
```
alexshtin, line 112:
```suggestion
    // Operations are admitted transactionally, meaning if *any* operation fails to be admitted, none are executed,
```
?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -107,6 +107,24 @@ service WorkflowService {
...
+    //
+    // Upon failure, the status code equals the status code of the *first* operation that failed to be admitted; and the
+    // list of details contains one `google.grpc.Status` for each requested operation. The failed
+    // operation(s) have the same error details as if it was executed separately. All other operations have the
+    // status code `Aborted` and `temporal.api.failure.v1.MultiOperationExecutionAborted` is added to the details field.
+    //
+    // NOTE: Experimental API.
+    rpc ExecuteMultiOperation (ExecuteMultiOperationRequest) returns (ExecuteMultiOperationResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/workflows/execute-operations"
```
alexshtin, line 123: Shouldn't it match gRPC name? I.e. `execute-multi-operation`?

alexshtin (approved):

temporal/api/workflowservice/v1/service.proto
stephanos: Good point.

cretz (approved):
Looks good. May want to clarify idempotency behavior (i.e. that users should set start workflow request ID since they may repeatedly invoke this w/ fail-if-existing), but don't need to necessarily.

(also, I disagree that multi operation should ever need to be speculative/lossy as that is very confusing for users, but this was already decided internally)

temporal/api/workflowservice/v1/service.proto
```diff
@@ -107,6 +107,24 @@ service WorkflowService {
         };
     }
 
+    // ExecuteMultiOperation executes multiple operations within a single workflow.
+    //
+    // Operations are admitted transactionally, meaning if *any* operation fails to be admitted, none are executed,
```
Quinn-With-Two-Ns, line 112: >Operations are admitted transactionally

Is this true? if we support non durable update then an update can become unadmitted so I don't see how it is transactional?

stephanos: Could you help me find a better description?

What I'm trying to say is that all operations are admitted together; or not. It doesn't mean that once it's admitted it is guaranteed to execute. Maybe "atomically" is clearer?

stephanos:
> May want to clarify idempotency behavior (i.e. that users should set start workflow request ID since they may repeatedly invoke this w/ fail-if-existing)

I'll think about this more ... I wonder if we should require a request ID 🤔 If we're not, I'll add a follow-up comment in a separate PR.

temporal/api/workflowservice/v1/service.proto
Quinn-With-Two-Ns: If it is not durable I don't think you can make any strong statement about transactional or atomic since it is not durable.

cretz:
> I wonder if we should require a request ID

We should not require it. It is a choice from the API caller whether they will ever re-invoke this, and request ID is only for that. Think about the user invoking this from the HTTP API. We really need to care about that developer's experience too (which is why lossy updates are rough on them).

temporal/api/workflowservice/v1/service.proto
```diff
@@ -107,6 +107,24 @@ service WorkflowService {
...
+    // and the request fails. Upon admission, the API returns only when *all* operations have a response.
+    //
+    // Upon failure, the status code equals the status code of the *first* operation that failed to be admitted; and the
```
Quinn-With-Two-Ns, line 115: If the update is lost by the server after admitted before `ExecuteMultiOperation` returns then what does the response from `ExecuteMultiOperationResponse ` look like?

stephanos: So if the Server cancels the Update (for internal reasons), Alex and I have discussed returning a `failurepb.Failure` with `NonRetryable=false`.

That way we can separate the error behavior very clearly into "pre-admitted" (gRPC error, nothing was executed) vs "post-admitted" (operation-specific error, operations were/are executed).

Note that this isn't implemented yet for Update, but will be soon.

Quinn-With-Two-Ns: How is this `failurepb.Failure` being returned to the user?

>That way we can separate the error behavior very clearly into "pre-admitted" (gRPC error, nothing was executed) vs "post-admitted" (operation-specific error, operations were/are executed).

Can we do that? If the server looses the update because the server crashed the server will have no record the update was admitted right? So how can the server separate the errors? I'm sure ya'll have thought about this a lot more then I have, I just am struggling to understand how it is possible.

stephanos: > How is this failurepb.Failure being returned to the user?

```
ExecuteMultiOperationResponse {
  responses: [
    ExecuteMultiOperationResponse { ... }, // Start
    ExecuteMultiOperationResponse {        // Update
       response: ExecuteMultiOperationResponse_Response {
         UpdateWorkflowExecutionResponse {
           Outcome: {
             value: Failure {
               Message: "Workflow Update was cancelled.",
               Source:  "Server",
               FailureInfo: Failure_ApplicationFailureInfo {
                 ApplicationFailureInfo: ApplicationFailureInfo: {
                   Type: "CanceledUpdate",
                   NonRetryable: false,
                 }
               }
             }
           }
           UpdateRef: ...
           Stage: ...
         }
       }
    } 
  ]
}
```

This rather verbose response seems to be the existing way we send a failure outcome regarding an Update. So we thought it would make sense to do the same thing here since the Server knowingly _cancelled_ the Update.

> Can we do that? If the server looses the update because the server crashed the server will have no record the update was admitted right? So how can the server separate the errors?

Right; that's a different kind of "lost" (I thought you were referring to the above one). In that case the client's request just times out.

Quinn-With-Two-Ns: >This rather verbose response seems to be the existing way we send a failure outcome regarding an Update

Yes, but failure can only come from the SDK right now. It isn't consistent with how we handle any other update error from server

>Right; that's a different kind of "lost" (I thought you were referring to the above one).

So I was referring to any type of "lost"

Quinn-With-Two-Ns (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1430,3 +1430,39 @@ message RespondNexusTaskFailedRequest {
...
+    string namespace = 1;
+
+    // List of operations to execute within a single workflow.
+    //
+    // Preconditions:
+    // - The list of operations must not be empty.
+    // - The workflow ids must match across operations.
+    // - The only valid set of operations at this time is [StartWorkflow, UpdateWorkflow], in this order.
```
Sushisource, line 1442:
```suggestion
    // - The only valid list of operations at this time is [StartWorkflow, UpdateWorkflow], in this order.
```
Since a set is unordered by definition

Sushisource (approved):

temporal/api/workflowservice/v1/service.proto
dandavison: One issue here is that we are starting to use the term "admission" for MultiOperation and for StartWorkflow, but we haven't defined what it means in those cases. But also, I think that it will help to expand the text slightly and be more explicit about what we are saying. Here's an attempt at re-drafting this section; feel free to take any bits you think help.

```proto
// ExecuteMultiOperation executes multiple operations within a single workflow.
//
// In general, a MultiOperation is atomic in the sense that when it returns, all
// constituent operations have started, or none have.
//
// However, note that
// - the only supported MultiOperation is [StartWorkflow, UpdateWorkflow],
// - an Update "starts" when it is admitted by the server,
// - an update is not durably persisted on admission (this happens later,
//   when it is accepted by a worker executing the workflow).
//
// Therefore, although a MultiOperation is atomic with respect to *starting* its
// operations, in the case of [StartWorkflow, UpdateWorkflow] it is possible for the
// MultiOperation to succeed and later reach a state in which the workflow is started and
// the server has no record of the Update request.
//
// In the future we may support additional MultiOperations, for example [StartWorkflow,
// SignalWorkflow] and [StartWorkflow, UpdateWorkflow(durableOnAdmission)]. In those
// cases, the MultiOperation will be atomic in the stronger sense that either all
// operations start durably, or none does.
//
// The API returns only when *all* operations have a response.
//
// Upon failure, the status code equals the status code of the *first* operation that
// failed, and the list of details contains one `google.grpc.Status` for each requested
// operation.
```

temporal/api/workflowservice/v1/service.proto
dandavison: > Upon failure, the status code equals the status code of the *first* operation that failed to be admitted

The update could be admitted, but the workflow could fail to start due to a failure to commit to the persistence layer. We haven't defined "admission" for StartWorkflow yet, but I suspect it will be better to drop the "to be admitted" part of this sentence, and restrict the "admission" concept to update only.

temporal/api/workflowservice/v1/service.proto
stephanos: I've replaced "admitted" with "started" and will merge now to get off of the feature branch. I have filed a follow-up ticket to finalize these docs based on what I learn from implementing the rest of the feature and resolving the last unknowns.

temporal/api/workflowservice/v1/service.proto
stephanos: I've made a small change to that effect now.

PR #368, opened 2024-03-13
Make it easier to see linter issues

MichaelSnowden:
**What changed?**
I changed the format of the api-linter to:
- Only show the rule URL and location
- Make the location a clickable link
- Group violations by rule

**Why?**
The output previously was not very easy to parse:

Before:

Screenshot 2024-03-13 at 12 55 27 PM
Screenshot 2024-03-13 at 12 55 39 PM

After:

Screenshot 2024-03-13 at 12 54 45 PM

**Breaking changes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,16 +20,6 @@ STAMPDIR := .stamp
 
 COLOR := "\e[1;36m%s\e[0m\n"
 
-# Only prints output if the exit code is non-zero
-define silent_exec
-    @output=$$($(1) 2>&1); \
-    status=$$?; \
-    if [ $$status -ne 0 ]; then \
-        echo "$$output"; \
-    fi; \
-    exit $$status
-endef
-
 PROTO_ROOT := .
 PROTO_FILES = $(shell find temporal -name "*.proto")
 PROTO_DIRS = $(sort $(dir $(PROTO_FILES)))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -104,7 +94,7 @@ buf-install:
 ##### Linters #####
 api-linter:
 	printf $(COLOR) "Run api-linter..."
-	$(call silent_exec, api-linter --set-exit-status $(PROTO_IMPORTS) --config $(PROTO_ROOT)/api-linter.yaml $(PROTO_FILES))
+	@api-linter --set-exit-status $(PROTO_IMPORTS) --config $(PROTO_ROOT)/api-linter.yaml --output-format json $(PROTO_FILES) | jq -r 'map(select(.problems != []) | . as $$file | .problems[] | {rule: .rule_doc_uri, location: "\($$file.file_path):\(.location.start_position.line_number)"}) | group_by(.rule) | .[] | .[0].rule + ":\n" + (map("\t" + .location) | join("\n"))'
```
stephanos, line 97: The `jq` requirement might be problematic. I don't know if we can expect everyone to have that installed?

stephanos: FWIW, since we expect `go` to be installed, there seems to be a Go port (https://github.com/itchyny/gojq) which could be installed like the other tools.

bergundy: Maybe instead of requiring `jq`, you can add a separate target `api-linter-pretty` and recommend running that for pretty output if this target fail.

cretz:
Wish we didn't have to have platform dependent build scripting and require extra dependencies like "jq". We have a perfectly fine language for executing processes, parsing JSON, etc.

MichaelSnowden:
Thanks for the suggestion, @stephanos! `gojq` seems to virtually be a drop-in replacement for `jq`. Also, the install already has a good home in `api-linter-install`. Hopefully this addresses your concerns about portability, @cretz and your concern about installing jq @bergundy.

cretz (approved):

stephanos (approved):

PR #371, opened 2024-03-18
User metadata on workflows/events for user interface usage

cretz:
**What changed?**

* Added `temporal.api.sdk.v1.UserMetadata` containing payloads for `summary` and `details` for general metadata use
* Added `temporal.api.command.v1.Command.user_metadata` for general use (initially only for timers and child workflow)
* Added `temporal.api.history.v1.HistoryEvent.user_metadata` for general use (initially only for timers)
* Added `temporal.api.workflowservice.v1.StartWorkflowExecutionRequest.user_metadata` for general use (initially for populating execution config)
* Added `temporal.api.workflowservice.v1.SignalWithStartWorkflowExecutionRequest.user_metadata` for same reason as start
* Added `temporal.api.workflow.v1.NewWorkflowExecutionInfo.user_metadata` for same reason as start
* Added `temporal.api.workflow.v1.WorkflowExecutionConfig.user_metadata` for general use (initially for fixed as-of-start summary/details)
* Added `temporal.api.sdk.v1.WorkflowMetadata.current_details` for the ability to have queryable long-form text representing current details of a workflow
* Removed `temporal.api.sdk.v1.WorkflowDefinition.definition` because a workflow definition's description isn't best placed on a query (but "current details" is good on the metadata itself)

**Why?**

* Users need a way to set fixed summary/details when starting a workflow for use by UIs
* Users need a way to have query-obtainable live "current details" for use by UIs
* Users need a way to set a summary when starting a timer for use by UIs
* All of this needs to be encrypted if the user wants
* We need to take into account future user-metadata use

**Breaking changes**

* Yes for workflow metadata which was only used in JSON form by one SDK (TypeScript) and not even invoked by UI yet


temporal/api/sdk/v1/user_metadata.proto
```diff
@@ -0,0 +1,48 @@
...
+message UserMetadata {
+  // Short-form text that provides a summary. This payload should be a "json/plain"-encoded payload
+  // that is a single JSON string. Formatting may not apply to this text when used in "title"
+  // situations. The payload data section is currently limited to 400 bytes.
+  temporal.api.common.v1.Payload summary = 1;
+
+  // Long-form text that provides details. This payload should be a "json/plain"-encoded payload
+  // that is a single JSON string. Formatting may apply to this text in common use. The payload data
+  // section is currently limited to 20000 bytes.
+  temporal.api.common.v1.Payload details = 2;
```
cretz, line 47: I considered "description", but it became clear that wasn't a general-purpose-enough term

cretz, line 47: Notes:

1. Considered "description" instead of "details", but description isn't a general-purpose-enough term
1. The 400 and 20000 bytes is an arbitrary ceiling based on expected use, happy to change
1. These are payloads so they can be encrypted
1. The use of JSON string as the type is a simple compatibility thing
1. At this time, we are not defining the long-form format (but I suspect some very limited form of markdown)

temporal/api/sdk/v1/workflow_metadata.proto
cretz, line 1: This file is safe for reworking in an incompatible way because it's just models for a well-known workflow query. See description to understand the changes.

temporal/api/command/v1/message.proto
```diff
@@ -236,6 +237,18 @@ message ProtocolMessageCommandAttributes {
...
+    // the interface that reads it.
+    //
+    // Current well-known uses:
+    //  * start_child_workflow_execution_command_attributes - populates
+    //    temporal.api.workflow.v1.WorkflowExecutionInfo.user_metadata where the summary and details
+    //    are used by user interfaces to show fixed as-of-start workflow summary and description.
+    //  * start_timer_command_attributes - populates temporal.api.history.v1.HistoryEvent for timer
+    //    started where the summary is used to identify the timer.
+    temporal.api.sdk.v1.UserMetadata user_metadata = 301;
+    // The command details. The type must match taht in `command_type`.
```
bergundy, line 251:
```suggestion
    // The command details. The type must match that in `command_type`.
```

temporal/api/sdk/v1/user_metadata.proto
```diff
@@ -0,0 +1,48 @@
...
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+
+import "temporal/api/common/v1/message.proto";
+
+// Information a user can set, often for use by user interfaces.
+message UserMetadata {
+  // Short-form text that provides a summary. This payload should be a "json/plain"-encoded payload
+  // that is a single JSON string. Formatting may not apply to this text when used in "title"
+  // situations. The payload data section is currently limited to 400 bytes.
```
bergundy, line 41: Maybe instead of "currently limited" mention that this is the default limit on the server.

Just a slight distinction that may be worth calling out. Self-hosted users may be able to up this value.

temporal/api/sdk/v1/workflow_metadata.proto
```diff
@@ -35,20 +35,19 @@ option csharp_namespace = "Temporalio.Api.Sdk.V1";
 message WorkflowMetadata {
   // Metadata provided at declaration or creation time.
   WorkflowDefinition definition = 1;
+  // Current long-form details of the workflow's state. This is used by user interfaces to show
+  // long-form text. This text may be formatted by the user interface.
+  string current_details = 2;
```
bergundy, line 40: This is always returned in query responses and goes through the user's DC, right?

temporal/api/sdk/v1/workflow_metadata.proto
```diff
@@ -35,20 +35,19 @@ option csharp_namespace = "Temporalio.Api.Sdk.V1";
...
 
 // (-- api-linter: core::0203::optional=disabled --)
 message WorkflowDefinition {
   // A name scoped by the task queue that maps to this workflow definition.
   // If missing, this workflow is a dynamic workflow.
   string type = 1;
-  // An optional workflow description provided by the application.
-  // By convention, external tools may interpret its first part,
-  // i.e., ending with a line break, as a summary of the description.
-  string description = 2;
```
bergundy, line 48: It this okay to break? Do we have any consumers of this as protos?

temporal/api/sdk/v1/user_metadata.proto
```diff
@@ -0,0 +1,48 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
```
bergundy, line 3: It's 2024...

bergundy (commented):
All of the fields are just plain text at the moment, right?
Would be nice to understand what the format is.

temporal/api/sdk/v1/workflow_metadata.proto
cretz: Right

temporal/api/sdk/v1/workflow_metadata.proto
cretz: Yes it's ok to break, no we don't have any consumers of it except TypeScript SDK internally (and not this field and we only use this as JSON not proto)

cretz:
>  Would be nice to understand what the format is.

Yes, this is documented on the field

> Would be nice to understand what the format is.

Mentioned [here](https://github.com/temporalio/api/pull/371#discussion_r1529177662), I am intentionally leaving that as out-of-scope for now pending some UI discussions. We should not block this progress on that.

temporal/api/sdk/v1/user_metadata.proto
cretz: Fixed

temporal/api/sdk/v1/user_metadata.proto
cretz: Fixed

temporal/api/command/v1/message.proto
cretz: Fixed

temporal/api/command/v1/message.proto
```diff
@@ -236,6 +237,18 @@ message ProtocolMessageCommandAttributes {
 
 message Command {
     temporal.api.enums.v1.CommandType command_type = 1;
+    // Metadata on the command. This is carried over to the event if there is an associated event.
+    // Most commands won't have this information, and how this information is used is dependent upon
+    // the interface that reads it.
```
Sushisource, line 242:
```suggestion
    // Metadata on the command. This is carried over to the history event if one is created as a result of the command.
    // Most commands won't have this information, and how this information is used is dependent upon
    // the interface that reads it.
```

Just to be slightly more clear

temporal/api/sdk/v1/user_metadata.proto
```diff
@@ -0,0 +1,48 @@
...
+// Information a user can set, often for use by user interfaces.
+message UserMetadata {
+  // Short-form text that provides a summary. This payload should be a "json/plain"-encoded payload
+  // that is a single JSON string. Formatting may not apply to this text when used in "title"
+  // situations. The payload data section is limited to 400 bytes by default.
+  temporal.api.common.v1.Payload summary = 1;
+
+  // Long-form text that provides details. This payload should be a "json/plain"-encoded payload
+  // that is a single JSON string. Formatting may apply to this text in common use. The payload data
+  // section is limited to 20000 bytes by default.
```
Sushisource, line 46: Probably worth it to mention that the `json/plain` suggestion is so that we can readily render it in UI / elsewhere.

temporal/api/workflow/v1/message.proto
```diff
@@ -151,6 +154,10 @@ message NewWorkflowExecutionInfo {
     temporal.api.common.v1.Memo memo = 11;
     temporal.api.common.v1.SearchAttributes search_attributes = 12;
     temporal.api.common.v1.Header header = 13;
+    // Metadata on the workflow if it is started. This is carried over to the WorkflowExecutionInfo
+    // for use by user interfaces to display the fixed as-of-start summary and description of the
+    // workflow.
```
Sushisource, line 159:
```suggestion
    // for use by user interfaces to display the fixed as-of-start summary and details of the
    // workflow.
```

Same for the other instances of this comment too. We're calling it `details` or `current_details` elsewhere

Sushisource (commented):
Looking good to me, just minor stuff.

bergundy (approved):

temporal/api/workflow/v1/message.proto
cretz: I used "description" specifically because for these uses "details" is used as a "description", but I can change to be more clear.

Sushisource: Ah, yeah that wasn't immediately obvious to me. I think using both words to connect this idea would be nice

Sushisource (approved):

temporal/api/sdk/v1/workflow_metadata.proto
dnr, line 40: Why a Payload in history and a string here?

temporal/api/sdk/v1/user_metadata.proto
```diff
@@ -0,0 +1,48 @@
...
+
+
+import "temporal/api/common/v1/message.proto";
+
+// Information a user can set, often for use by user interfaces.
+message UserMetadata {
+  // Short-form text that provides a summary. This payload should be a "json/plain"-encoded payload
+  // that is a single JSON string. Formatting may not apply to this text when used in "title"
+  // situations. The payload data section is limited to 400 bytes by default.
+  temporal.api.common.v1.Payload summary = 1;
```
dnr, line 42: I assume we considered this as a map from string to Payload.. why not do that? (For the record I like this better, just wondering what the reasoning was.)

temporal/api/command/v1/message.proto
```diff
@@ -236,6 +237,18 @@ message ProtocolMessageCommandAttributes {
...
+    //
+    // Current well-known uses:
```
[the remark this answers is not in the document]
cretz: I was planning on maintaining this list. We do not have a better public place to document this that I can think of other than on the API itself. The other alternative is to not document this at all.

temporal/api/sdk/v1/workflow_metadata.proto
cretz: Because this object is a query result and query results are payloads already (so already subject to encryption in transit) whereas in history if you want something encrypted at rest it needs to be a payload.

temporal/api/sdk/v1/user_metadata.proto
cretz: There were a few of reasons that was rejected (there's an internal document discussing this). The primary reasons were that 1) the stringly-typed contract was too loose, 2) the keys are not user-defined, and 3) it unnecessarily increased history size for fixed keys.

temporal/api/sdk/v1/user_metadata.proto
cretz: Updated

temporal/api/command/v1/message.proto
cretz: Updated

temporal/api/workflow/v1/message.proto
```diff
@@ -58,6 +59,8 @@ message WorkflowExecutionInfo {
     int64 history_size_bytes = 15;
     // If set, the most recent worker version stamp that appeared in a workflow task completion
     temporal.api.common.v1.WorkerVersionStamp most_recent_worker_version_stamp = 16;
+    // User metadata provided on start workflow.
+    temporal.api.sdk.v1.UserMetadata user_metadata = 17;
```
cretz, line 63: Is it acceptable to put all of this in visibility? I think it's only needed for describe. If not acceptable for visibility, where should it live? `WorkflowExecutionConfig`? On describe response itself?

buf.yaml
```diff
@@ -13,6 +13,8 @@ breaking:
     - google
     # Uncomment this after https://github.com/temporalio/api/pull/362 is merged
     - temporal/api/operatorservice/v1/request_response.proto
+    # Uncomment this after user metadata is merged
```
tdeebswihart, line 16: Is this comment necessary?

temporal/api/workflow/v1/message.proto
tdeebswihart: @rodrigozhou

buf.yaml
cretz: Yes, but it's bad wording. I should say "Remove this after user metadata is merged"

temporal/api/workflow/v1/message.proto
rodrigozhou: Describe API doesn't query visibility.

cretz: Right, but we're discussing `WorkflowExecutionInfo` which does come back from list calls too, correct? Should this move if we don't want it in visibility? If so, suggestions on where to?

buf.yaml
cretz: Fixed

gow (approved):

temporal/api/command/v1/message.proto
```diff
@@ -236,6 +237,18 @@ message ProtocolMessageCommandAttributes {
...
+    // created as a result of the command. Most commands won't have this information, and how this
+    // information is used is dependent upon the interface that reads it.
+    //
+    // Current well-known uses:
+    //  * start_child_workflow_execution_command_attributes - populates
+    //    temporal.api.workflow.v1.WorkflowExecutionInfo.user_metadata where the summary and details
+    //    are used by user interfaces to show fixed as-of-start workflow summary and details.
+    //  * start_timer_command_attributes - populates temporal.api.history.v1.HistoryEvent for timer
+    //    started where the summary is used to identify the timer.
+    temporal.api.sdk.v1.UserMetadata user_metadata = 301;
```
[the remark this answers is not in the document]
cretz: Same reason why the person who added `HistoryEvent.worker_may_ignore` as `300`, so that it would not interfere with our continually added, incrementally numbered attribute fields below. I chose 301 instead of 300 just to match history event's equivalent, but don't have to. The field number doesn't really matter much here, so it leaves room for attributes to keep adding.

yiminc (approved):

PR #372, opened 2024-03-18
Update Nexus Incoming Service CRUD to match Cloud API

bergundy:
**What changed?**

- Use IDs for incoming service endpoints
- Expose the URL prefix and don't tie it to service name
- Allow filtering list results by name
- Split Create and Update into separate APIs
- Added created_time and last_modified_time fields to both incoming and outgoing services
- Added `-o pipefail` to the Makefile to ensure commands with pipes fail

**Why?**

Be more consistent with the Cloud equivalent of these APIs:

https://github.com/temporalio/api-cloud/pull/16/files

**Breaking changes**

Changes are breaking but since there's no implementation yet it's still safe to break now.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -3982,20 +3982,20 @@
         }
       }
     },
+    "v1CreateNexusIncomingServiceResponse": {
+      "type": "object",
+      "properties": {
+        "service": {
+          "$ref": "#/definitions/v1IncomingService",
+          "description": "Data post acceptance. Can be used to issue additional updates to this record."
+        }
+      }
+    },
     "v1CreateNexusOutgoingServiceResponse": {
       "type": "object",
       "properties": {
         "service": {
           "$ref": "#/definitions/v1OutgoingService",
-          "description": "Data post acceptance. Can be used to issue additional updates to this record."
-        }
-      }
-    },
-    "v1CreateOrUpdateNexusIncomingServiceResponse": {
-      "type": "object",
-      "properties": {
-        "service": {
-          "$ref": "#/definitions/v1IncomingService",
           "description": "Data post acceptance. Can be used to issue additional updates to this record."
         }
       }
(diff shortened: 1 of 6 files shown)


temporal/api/operatorservice/v1/service.proto
```diff
@@ -78,23 +78,32 @@ service OperatorService {
...
-    // Set version to 0 to create a new service.
-    // Returns the updated service with the updated version, which can be used for subsequent updates.
-    rpc CreateOrUpdateNexusIncomingService(CreateOrUpdateNexusIncomingServiceRequest) returns (CreateOrUpdateNexusIncomingServiceResponse) {
+    // Create a Nexus service. This will fail if a service with the same name already exists in the namespace with a
+    // status of ALREADY_EXISTS.
+    // Returns the created service with its initial version. You may use this version for subsequent updates.
+    rpc CreateNexusIncomingService(CreateNexusIncomingServiceRequest) returns (CreateNexusIncomingServiceResponse) {
+    }
+
+    // Optimistically create or update a Nexus service based on provided version as obtained via the
```
MichaelSnowden, line 91: This still says "create or update"

MichaelSnowden (approved):

temporal/api/operatorservice/v1/service.proto
bergundy:
```suggestion
    // Optimistically update a Nexus service based on provided version as obtained via the
```

openapi/openapiv2.json
cretz, line 1: I wonder why `openapi/openapiv3.yaml` didn't require changes

cretz (commented):
Nothing obvious, but wondering if someone could debug the openapiv3 code gen thing and why this doesn't apply

openapi/openapiv2.json
bergundy: Not sure, I double checked and reran `make proto`.

cretz: Hrmm, @tdeebswihart - I think OpenAPI v2 generated docs are including requests/responses for APIs that don't have HTTP annotations. I am unsure if they have some exclusion/dead-code/tree-shaking thing that can prevent generating unreferenced types.

bergundy: I can't find that option: https://github.com/grpc-ecosystem/grpc-gateway/blob/4319001a1681546055079f148ad857b05753508a/protoc-gen-openapiv2/main.go#L18

We may have to manually prune.

Let's not let this block the PR though please.

cretz (approved):

openapi/openapiv2.json
bergundy: Looks like this option is false by default so it's probably not that:

```
	generateUnboundMethods         = flag.Bool("generate_unbound_methods", false, "generate swagger metadata even for RPC methods that have no HttpRule annotation")

```

tdeebswihart: Well that's unfortunate. We need to fix that if so

PR #374, opened 2024-03-19
Fix missing pipefail in api-linter

MichaelSnowden:
**What changed?**
I moved the `api-linter` code into its own script and made it actually propagate non-zero exit codes fro mthe api-linter before the output is formatted.

**Why?**
My previous PR, #368 , introduced an issue where failures in the api-linter would show up in the terminal output, but the exit code would be 0 because we didn't set `-o pipefail` in the shell. This fixes that.

Why not use something like `SHELL=/bin/bash -o pipefail`? Because, we don't want to depend on bash.

**Breaking changes**

### DIFF
--- a/scripts/lint-api.sh
+++ b/scripts/lint-api.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+# Usage: ./lint-api.sh [PROTO_IMPORTS] [PROTO_ROOT] [PROTO_FILES]
+
+PROTO_IMPORTS="$1"
+PROTO_ROOT="$2"
+shift 2
+
+api_linter_output=$(api-linter --set-exit-status "$PROTO_IMPORTS" --config "$PROTO_ROOT/api-linter.yaml" --output-format json "$@")
+api_linter_status=$?
+
+# Exit early if api-linter finds no issues
+if [ $api_linter_status -eq 0 ]; then
+    exit 0
+fi
+
+# Continue with the script if issues are found
+# shellcheck disable=SC2016
+jq_expr='
+    map(select(.problems != [])
+        | . as $file
+        | .problems[]
+        | {
+            rule: .rule_doc_uri,
+            location: "\($file.file_path):\(.location.start_position.line_number)"
+        }
+    )
+    | group_by(.rule)
+    | .[]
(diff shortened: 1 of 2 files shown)


bergundy:
> Why not use something like SHELL=/bin/bash -o pipefail? Because, we don't want to depend on bash.

You're depending on a unix-like shell in your script. I would expect bash to be available in those OSes.

MichaelSnowden:
I believe the portability of our Makefile remains unaffected by these changes. Rewriting it as a Go script wouldn't enhance the portability of the Makefile or its targets. Here's my observation:

1. Before my original PR, the environment requirements included Go, make, and a shell.
2. With your PR, the requirements are the same as above, with the addition of bash.
3. My PR maintains the original environment requirements.

scripts/lint-api.sh
```diff
@@ -0,0 +1,43 @@
...
+            location: "\($file.file_path):\(.location.start_position.line_number)"
+        }
+    )
+    | group_by(.rule)
+    | .[]
+    | .[0].rule + ":\n"
+      + (map("\t" + .location) | join("\n"))
+'
+
+gojq_output=$(echo "$api_linter_output" | gojq -r "$jq_expr")
```
dnr, line 34: just a thought: check if gojq is present, if so pipe it through, if not just echo it?

MichaelSnowden: It's part of `api-linter-install`, which is a dependency of the `install` target, so I think it should be safe to assume it's present, and spitting out an error if it's not is ok.

MichaelSnowden: To guarantee it, I just made `api-linter-install` a dependency of the `api-linter` target. I've verified that it's very fast if everything is already installed.

scripts/lint-api.sh
cretz, line 1: I would like to request we stop making more and more platform-specific logic/code. We have better programming languages than bash at our disposal. I am already lobbying against the Makefile, this is just adds even more. Our software should be able to be built and contributed to on all platforms we support (and it is the case for SDKs of all languages).

Also not a big fan of making a new top-level directory in this repository (though I can't think of a better approach here).

PR #378, opened 2024-03-22
Cloud API

cretz:
**What changed?**

* Copied cloud protos from https://github.com/temporalio/api-cloud
* Updated cloud protos to have license headers
* Updated cloud protos to have language-specific options
* Updated cloud service to have base HTTP URLs of `/api/v1/cloud`
* Updated OpenAPI definitions to include cloud service
* Disabled API lint checks globally
* Updated some protos to disable some API lint checks
* Added CODEOWNERS for saas for the cloud directory

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,5 @@
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
 *                       @temporalio/server @temporalio/sdk
-api/temporal/api/sdk/*  @temporalio/sdk
+/temporal/api/sdk/      @temporalio/sdk
+/temporal/api/cloud/    @temporalio/saas @temporalio/sdk
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -1,0 +1,258 @@
+// The MIT License
+//
+// Copyright (c) 2024 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
+
+package temporal.api.cloud.cloudservice.v1;
+
+option go_package = "go.temporal.io/api/cloud/cloudservice/v1;cloudservice";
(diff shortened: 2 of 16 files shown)


temporal/api/cloud/cloudservice/v1/request_response.proto
cretz, line 1: @anekkanti or others - can you backport all the cloud/ changes to https://github.com/temporalio/api-cloud if you continue to work in that repository so that it is easier to bring over other changes y'all will still be making there in the interim?

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -0,0 +1,148 @@
...
+
+syntax = "proto3";
+
+package temporal.api.cloud.cloudservice.v1;
+
+option go_package = "go.temporal.io/api/cloud/cloudservice/v1;cloudservice";
+option java_package = "io.temporal.api.cloud.cloudservice.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ServiceProto";
+option ruby_package = "Temporalio::Api::Cloud::CloudService::V1";
```
Quinn-With-Two-Ns, line 31: Do we not define a python package name on purpose?

temporal/api/cloud/cloudservice/v1/request_response.proto
Quinn-With-Two-Ns: How long will we need to backport changes for?

.github/CODEOWNERS
```diff
@@ -2,4 +2,5 @@
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
 *                       @temporalio/server @temporalio/sdk
-api/temporal/api/sdk/*  @temporalio/sdk
+/temporal/api/sdk/      @temporalio/sdk
+/temporal/api/cloud/    @temporalio/saas @temporalio/sdk
```
Quinn-With-Two-Ns, line 6: Does this mean SDK team will be required to approve any changes to the cloud API?

temporal/api/cloud/cloudservice/v1/service.proto
cretz: It hasn't traditionally been needed. The default chosen package is acceptable and Python SDK has some post processing anyways (as do many SDKs)

.github/CODEOWNERS
cretz: Yes, which I think is reasonable

temporal/api/cloud/cloudservice/v1/service.proto
Quinn-With-Two-Ns: Fine with me, just curious since I saw dotnet did set one

.github/CODEOWNERS
Quinn-With-Two-Ns: Basically implies we are the owner of this Repo :D

cretz: Yup, SDK basically is (though most of it is shared responsibility depending on which protos). Server has internal protos (as may cloud) that SDK does not have ownership on.

Quinn-With-Two-Ns (approved):

tdeebswihart (approved):

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -0,0 +1,148 @@
...
+option csharp_namespace = "Temporalio.Api.Cloud.CloudService.V1";
+
+import "temporal/api/cloud/cloudservice/v1/request_response.proto";
+import "google/api/annotations.proto";
+
+service CloudService {
+    // Gets all known users
+    rpc GetUsers(GetUsersRequest) returns (GetUsersResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/users",
```
cretz, line 41: I think we should change these to `/api/v1/cloud/...` prefixed paths so that it is easier for proxies/edge that may share cloud and data plane API to differentiate the two

cretz: Updated

mastermanu (approved):

.github/CODEOWNERS
```diff
@@ -2,4 +2,5 @@
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
 *                       @temporalio/server @temporalio/sdk
-api/temporal/api/sdk/*  @temporalio/sdk
```
alexshtin, line 5: This effectively never worked because there is no `api` directory.

alexshtin (approved):

cretz:
This approach has been rejected. A new PR will be made with an approach that fetches from the other repository.

cretz:
Actually, due to how this repo and the cloud API repo are laid out, embedding that within this is too difficult. We will just submodule this where code is generated while waiting for the day where the cloud API can be merged into here.

PR #379, opened 2024-03-22
Add `TaskQueueVersionGroup` message

carlydf:
Use new TaskQueueVersionGroup message in api repo to specify which versions are of interest in a way that can be reused by server protos.

### DIFF
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -54,6 +54,26 @@ message TaskQueueMetadata {
     google.protobuf.DoubleValue max_tasks_per_second = 1;
 }
 
+// Used to describe a group of versions the caller is interested in.
+message TaskQueueVersionGroup {
+    // Optional. If not provided, the result for the default Build ID will be returned. The default Build ID is the one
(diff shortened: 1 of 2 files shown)


cretz (approved):

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,6 +54,26 @@ message TaskQueueMetadata {
     google.protobuf.DoubleValue max_tasks_per_second = 1;
 }
 
+// Used to describe a group of versions the caller is interested in.
+message TaskQueueVersionGroup {
+    // Optional. If not provided, the result for the default Build ID will be returned. The default Build ID is the one
+    // mentioned in the first unconditional Assignment Rule. If there is no default Build ID, the result for the
+    // unversioned queue will be returned.
+    // (-- api-linter: core::0140::prepositions --)
+    oneof versions {
```
ShahabT, line 63: Now that we extract this message, I think it may make sense to make this not a oneof. I can imagine an operator may want to get the result for a fixed list of build IDs plus any active build IDs that may be missing from its radar. Or if it wants to get the results for some build IDs and unversioned together.

Also, nit, `TaskQueueVersionSelection` may be more appropriate.

PR #382, opened 2024-03-26
Adding ResourceScope to ResourceExhaustedFailure

prathyushpv:
**What changed?**
Adding a new field ResourceScope in ResourceExhaustedFailure.

**Why?**
We need to differentiate whether a request has been rate limited because of namespace limit or system limit. 
This will help us better understand why requests are rate limited and which limits to adjust.

**Breaking changes**
No

**Server PR**
https://github.com/temporalio/temporal/pull/5605


temporal/api/enums/v1/failed_cause.proto
```diff
@@ -127,3 +127,10 @@ enum ResourceExhaustedCause {
     // Caller exceeds action per second limit.
     RESOURCE_EXHAUSTED_CAUSE_APS_LIMIT = 6;
 }
+
+enum ResourceScope {
```
cretz, line 131: Can you document these a bit? A simple sentence above the namespace and system ones would help.

cretz, line 131:
```suggestion
enum ResourceExhaustedScope {
```

So as not to be confused with other potential uses of a "resource scope" in this shared `enums` proto package.

temporal/api/enums/v1/failed_cause.proto
```diff
@@ -127,3 +127,10 @@ enum ResourceExhaustedCause {
...
+    RESOURCE_SCOPE_UNSPECIFIED = 0;
+    RESOURCE_SCOPE_NAMESPACE = 1;
```
pdoerner, line 133: Not sure what has already been agreed upon, but there seems to be a lot of overlap between `ResourceExhaustedCause` and `ResourceExhaustedScope`. For example, `RESOURCE_EXHAUSTED_CAUSE_APS_LIMIT` will always map to `RESOURCE_SCOPE_NAMESPACE`, right? Maybe the same result could be achieved by adding a few more specific `ResourceExhaustedCauses` rather than a separate field?

tdeebswihart: We discussed this internally (I'll send you the proposal), but that would require us to break down our causes into the power set of Scopes * Resources. We'd discussed fully replacing the RESOURCE_EXHAUSTION_CAUSE with the pair of RESOURCE_EXHAUSTION_SCOPE and RESOURCE_EXHAUSTION_RESOURCE but don't we don't wish to deprecate the old field quite yet.

prathyushpv: You are right. ResourceExhaustedCause has resource and scope in some cases. But it doesn't have in other cases.

Our initial plan was to add two fields Resource, and Scope. Then eventually depricate ResourceExhaustedCause. This way we don't have to enumerate every resource and scope. It also make sense to add Service field to identify which service returned this error.

But service names and some of these resources should not be leaked through API. That is when we decided to just add Scope field for now. Then later decide about adding resource field. We want to stop using ResourceExhaustedCause eventually.

temporal/api/enums/v1/failed_cause.proto
prathyushpv: Added comment.

cretz (approved):

tdeebswihart (approved):

PR #383, opened 2024-03-26
Add call options for Nexus outgoing services

bergundy:
**What changed?**

In description.

**Why?**

Both Temporal Cloud and Self-hosted users will want to have a way to configure how the server makes outgoing service calls.

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -191,3 +191,42 @@ message Callback {
         Nexus nexus = 2;
     }
 }
+
+// Options for making TLS requests in various contexts.
+message TLSOptions {
+    // If not empty, this path is expected to contain a valid x509 certificate.
+    // Either both or none of tls_cert_path and tls_key_path should be set.
+    string tls_cert_path = 1;
+
+    // If not empty, this path is expected to contain a valid private certificate key.
+    // Either both or none of tls_cert_path and tls_key_path should be set.
+    string tls_key_path = 2;
+
+    // If not empty, this path is expected to contain a valid CA certificate.
+    string tls_ca_path = 3;
+
+    // Disable TLS host verification - generally not recommended for production use.
+    bool tls_disable_host_verification = 4;
+
+    // Overrides the target name (SNI) used for TLS host name checking. If this attribute is not specified, the name
+    // used for TLS host name checking will be the URL host. Adding this override should be done with care.
+    string tls_server_name = 5;
+}
+
+message EncryptionKeySpec {
+    // Encryption algorithm. Either "AES-GCM" or "XChaCha20-Poly1305" are supported (case insensitive).
+    // AES-GCM can be very fast on certain platforms due to hardware accelaration but requires more frequent key
+    // rotations as it uses a 96 bit nonce. It is recommended to rotate AES-GCM keys for every 2^32 nonces generated.
+    // XChaCha20-Poly1305 uses a 192 bit nonce and requires much less frequent rotation.
+    string algo = 1;
+    // Encryption key data.
+    // For AES, length must be either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.
+    // For xchacha20-poly1305, key data must be exactly 32 bytes.
+    // Mutually exclusive with path.
+    bytes data = 2;
+    // File system path to load encryption key data from.
+    // The content is expected to be stable and the path is not watched by the server.
+    // Mutually exclusive with data.
+    string path = 3;
+}
+
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -198,7 +198,8 @@ message TLSOptions {
     // Either both or none of tls_cert_path and tls_key_path should be set.
     string tls_cert_path = 1;
 
-    // If not empty, this path is expected to contain a valid private certificate key.
+    // If not empty, this path is expected to contain a valid private key
+    // that corresponds to the certificate in tls_cert_path.
     // Either both or none of tls_cert_path and tls_key_path should be set.
     string tls_key_path = 2;
 
--- a/temporal/api/nexus/v1/message.proto
+++ b/temporal/api/nexus/v1/message.proto
@@ -184,4 +184,35 @@ message OutgoingService {
 message OutgoingServiceSpec {
     // URL to invoke requests.
     string url = 1;
+
+    // Header to attach to all outgoing requests.
+    map<string, string> header = 2;
+
+    // ScheduleNexusOperation commands containing header keys that start with these prefixes will be rejected.
+    // Header key comparison is case-insensitive.
+    repeated string disallowed_user_header_prefixes = 3;
+
+    // If this service has an HTTPS URL these options may be provided.
+    // Trying to set these options with with an HTTP URL will result in an INVALID_ARGUMENT error.
+    temporal.api.common.v1.TLSOptions tls_options = 5;
+
+    // Callback URL that is attached to every StartOperation request for this service.
+    // The URL should be accessable to the handler of the StartOperation requests.
+    // See https://github.com/nexus-rpc/api/blob/main/SPEC.md#query-parameters for more information.
+    string public_callback_url = 6;
+
+    // Map of key ID to encryption key spec for the callback token used for asynchronous completion.
+    // The callback token contains information that may be considered sensitive such as such as workflow ID.
+    // Any of the provided keys may be used for decryption. Only callback_token_encryption_key_id is used for
+    // encryption. Callback tokens are passed as a `Nexus-Callback-Temporal-Callback-Token` header in StartOperation
+    // requests. See https://github.com/nexus-rpc/api/blob/main/SPEC.md#request-headers for more information.
+    // Mutually exclusive with disable_callback_token_encryption.
+    map<string, temporal.api.common.v1.EncryptionKeySpec> callback_token_encryption_keys = 7;
+    // The key ID used for callback token encryption. Must be set and present in the callback_token_encryption_keys map.
+    // Mutually exclusive with disable_callback_token_encryption.
+    string callback_token_encryption_key_id = 8;
+
+    // Opt-out of callback token encryption. Not recommended if this service targets an untrusted party.
+    // Mutually exclusive with callback_token_encryption_keys and callback_token_encryption_key_id.
+    bool disable_callback_token_encryption = 9;
 }
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -4209,6 +4209,24 @@
       ],
       "default": "ENCODING_TYPE_UNSPECIFIED"
     },
+    "v1EncryptionKeySpec": {
+      "type": "object",
+      "properties": {
+        "algo": {
+          "type": "string",
+          "description": "Encryption algorithm. Either \"AES-GCM\" or \"XChaCha20-Poly1305\" are supported (case insensitive).\nAES-GCM can be very fast on certain platforms due to hardware accelaration but requires more frequent key\nrotations as it uses a 96 bit nonce. It is recommended to rotate AES-GCM keys for every 2^32 nonces generated.\nXChaCha20-Poly1305 uses a 192 bit nonce and requires much less frequent rotation."
+        },
+        "data": {
+          "type": "string",
+          "format": "byte",
+          "description": "Encryption key data.\nFor AES, length must be either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.\nFor xchacha20-poly1305, key data must be exactly 32 bytes.\nMutually exclusive with path."
+        },
+        "path": {
+          "type": "string",
+          "description": "File system path to load encryption key data from.\nThe content is expected to be stable and the path is not watched by the server.\nMutually exclusive with data."
+        }
+      }
+    },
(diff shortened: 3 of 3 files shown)


temporal/api/nexus/v1/message.proto
```diff
@@ -184,4 +184,35 @@ message OutgoingService {
...
+    // See https://github.com/nexus-rpc/api/blob/main/SPEC.md#query-parameters for more information.
+    string public_callback_url = 6;
+
+    // Map of key ID to encryption key spec for the callback token used for asynchronous completion.
+    // The callback token contains information that may be considered sensitive such as such as workflow ID.
+    // Any of the provided keys may be used for decryption. Only callback_token_encryption_key_id is used for
+    // encryption. Callback tokens are passed as a `Nexus-Callback-Temporal-Callback-Token` header in StartOperation
+    // requests. See https://github.com/nexus-rpc/api/blob/main/SPEC.md#request-headers for more information.
+    // Mutually exclusive with disable_callback_token_encryption.
+    map<string, temporal.api.common.v1.EncryptionKeySpec> callback_token_encryption_keys = 7;
```
dnr, line 210: Sorry if this is described elsewhere, but why is this per-outgoing-service and not global for the cluster? A cluster is just encrypting/decrypting its own keys, right?

bergundy: At minimum this would be per namespace to allow migration but it seemed better to put this on every service in the namespace where users can minimize the blast radius of a compromised key and if they choose they can reuse keys between services.

temporal/api/common/v1/message.proto
```diff
@@ -191,3 +191,43 @@ message Callback {
...
+    string algo = 1;
+    // Encryption key data.
+    // For AES, length must be either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.
+    // For xchacha20-poly1305, key data must be exactly 32 bytes.
+    // Mutually exclusive with path.
+    bytes data = 2;
+    // File system path to load encryption key data from.
+    // The content is expected to be stable and the path is not watched by the server.
+    // Mutually exclusive with data.
+    string path = 3;
```
cretz, line 231: How often is this reloaded? I think we should just accept `data` and if we need some external tooling to update `data` from a file, ok.

temporal/api/common/v1/message.proto
```diff
@@ -191,3 +191,43 @@ message Callback {
...
+
+    // Disable TLS host verification - generally not recommended for production use.
+    bool tls_disable_host_verification = 4;
+
+    // Overrides the target name (SNI) used for TLS host name checking. If this attribute is not specified, the name
+    // used for TLS host name checking will be the URL host. Adding this override should be done with care.
+    string tls_server_name = 5;
+}
+
+message EncryptionKeySpec {
```
cretz, line 217: I think you should wait to assume this is common until it's in common use. This seems Nexus specific.

temporal/api/common/v1/message.proto
```diff
@@ -191,3 +191,43 @@ message Callback {
...
+    // Encryption algorithm. Either "AES-GCM" or "XChaCha20-Poly1305" are supported (case insensitive).
+    // AES-GCM can be very fast on certain platforms due to hardware accelaration but requires more frequent key
+    // rotations as it uses a 96 bit nonce. It is recommended to rotate AES-GCM keys for every 2^32 nonces generated.
+    // XChaCha20-Poly1305 uses a 192 bit nonce and requires much less frequent rotation.
+    string algo = 1;
```
cretz, line 222: I think you should make different messages for each high-level algorithm and do a oneof here (even if the only two today only take one `data` field)

temporal/api/common/v1/message.proto
```diff
@@ -191,3 +191,43 @@ message Callback {
         Nexus nexus = 2;
     }
 }
+
+// Options for making TLS requests in various contexts.
+message TLSOptions {
+    // If not empty, this path is expected to contain a valid x509 certificate.
+    // Either both or none of tls_cert_path and tls_key_path should be set.
+    string tls_cert_path = 1;
```
cretz, line 199:
```suggestion
    bytes client_cert = 1;
```

Improvements (to apply to other fields too):

* Do not reference the filesystem from external API (if we need tooling to get from file to API, so be it)
* Do not prefix with `tls`, we're already in TLS options
* Do not make it seem like this is a normal TLS cert, this is very specifically for client mTLS use and so make it clear it's for `client`

Here's what the message should look like IMO:

```proto
message TLSOptions {
    // Client certificate for mTLS, if present client_key must be present too. PEM formatted.
    bytes client_cert = 1;

    // Client key for mTLS, if present client_cert must be present too. PEM formatted. Not returned on get.
    bytes client_key = 2;

    // Set of CAs to validate server certificate against (note: can limit to one now if we want). PEM formatted.
    repeated bytes server_ca_certs = 3;

    string server_name_override = 4;

    bool disable_host_verification = 5;
}
```

I would also understand if you wanted to be more explicit with certificate and key-pair objects here.

temporal/api/common/v1/message.proto
```diff
@@ -191,3 +191,43 @@ message Callback {
...
+    // Encryption algorithm. Either "AES-GCM" or "XChaCha20-Poly1305" are supported (case insensitive).
+    // AES-GCM can be very fast on certain platforms due to hardware accelaration but requires more frequent key
+    // rotations as it uses a 96 bit nonce. It is recommended to rotate AES-GCM keys for every 2^32 nonces generated.
+    // XChaCha20-Poly1305 uses a 192 bit nonce and requires much less frequent rotation.
+    string algo = 1;
+    // Encryption key data.
+    // For AES, length must be either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.
+    // For xchacha20-poly1305, key data must be exactly 32 bytes.
+    // Mutually exclusive with path.
+    bytes data = 2;
```
cretz, line 227: Should not return this on get btw (and document that we don't)

temporal/api/nexus/v1/message.proto
```diff
@@ -184,4 +184,35 @@ message OutgoingService {
 message OutgoingServiceSpec {
     // URL to invoke requests.
     string url = 1;
+
+    // Header to attach to all outgoing requests.
+    map<string, string> header = 2;
+
+    // ScheduleNexusOperation commands containing header keys that start with these prefixes will be rejected.
+    // Header key comparison is case-insensitive.
+    repeated string disallowed_user_header_prefixes = 3;
```
cretz, line 193: Can you help me understand this use case?

temporal/api/common/v1/message.proto
tdeebswihart: Are we planning to store these encryption keys?

temporal/api/common/v1/message.proto
bergundy: > * Do not reference the filesystem from external API (if we need tooling to get from file to API, so be it)

I think there's a want to get the certs from K8s secrets as is common practice.
But I'd also support other URLs so we could retrieve from KMS as a extension (not built into the server).

> * Do not prefix with `tls`, we're already in TLS options

Oh yeah, duh, I meant to do that but forgot to when I move this into its own message.

> * Do not make it seem like this is a normal TLS cert, this is very specifically for client mTLS use and so make it clear it's for `client`

Thanks.

temporal/api/common/v1/message.proto
bergundy: I'm now thinking I'll put it on namespace config and not make it nexus specific to allow encrypting other types of tokens. There's been requests for that.

temporal/api/common/v1/message.proto
bergundy: Yeah, that makes sense, thanks for pointing it out.

temporal/api/common/v1/message.proto
bergundy: I'm not opposed to taking your suggestion, just wondering why you think the configs may diverge for other algorithms.

temporal/api/common/v1/message.proto
bergundy: Not reloaded:

> The content is expected to be stable and the path is not watched by the server.

temporal/api/nexus/v1/message.proto
bergundy: For a situation where you trust the server and not the workflow author to provide an identity for example.

temporal/api/nexus/v1/message.proto
bergundy: Thanks for pointing this out. I'm considering moving this to be scoped to a namespace and reusing this mechanism to sign arbitrary tokens.

temporal/api/common/v1/message.proto
cretz: > I think there's a want to get the certs from K8s secrets as is common practice.

Anything that can put into a file can surely put into the body of an API call. Often I've seen config map and environment variables (or helm replacement or whatever) do similar. We don't need our API referencing the filesystem IMO. At the very least we _know_ we should allow people to put their cert/key bytes in the API, so we should at least offer that and if a use-local-filesystem-paths-in-frontend-API need arises, we can add later.

temporal/api/common/v1/message.proto
cretz: I think we should remove this field then and only accept `data` and not care about where it's retrieved. I don't expect you to have an `env_var` or `url` or all the other ways that `data` could be obtained either.

temporal/api/nexus/v1/message.proto
cretz: Sorry, I admit not being familiar enough here. Can you give me a more concrete example of what someone might set here and what it helps prevent? I worry about two things here: 1) we're assuming a validation need too prematurely, and 2) we're providing a limited validation filter (why "prefix" why not "contains" or "match" or something else).

temporal/api/common/v1/message.proto
cretz: Because they often do and I've often seen crypto configurations expressed as option sets of their algorithm instead of attempting string name and hoping for a common set of configuration among them. It's easiest not to assume that every algorithm just needs "data bytes" and that's it (I'd have to go dig up some asymm that have more options than this). It also makes it easier for users to see what's being chosen and what's accepted when using strongly typed messages instead of string algo names. At the very least, caveats of each algorithm and what's expected in their "data" can be separately documented if properly separately modeled.

temporal/api/common/v1/message.proto
bergundy: I don't think we want to store secrets in the server's persistence layer.

temporal/api/nexus/v1/message.proto
bergundy: I'll remove this, I decided instead to hardcode the system added headers.
I may reject, ignore or append when a user tries to set those.

temporal/api/common/v1/message.proto
bergundy: I think we'll likely use a path or URL option to avoid storing keys in the server's persistence.
But some users may prefer storing these keys there. Especially when using the dev server.

temporal/api/common/v1/message.proto
bergundy: I'm actually thinking this could be a URL and only have built-in support for `file://` URLs only.
We can allow plugging in handlers for other URL types.

temporal/api/common/v1/message.proto
cretz: If your concern is storing this information, you should use keys the same way config does and just reference them here with some identifier. Our API should neither reference disk nor should the user even be expected to know which node of disk this may finally read off of in a cluster.

temporal/api/common/v1/message.proto
cretz: What about my key I put in the auth header if I am not using mTLS?

temporal/api/common/v1/message.proto
cretz: I think it's becoming clear that configuration is better for this kind of information than API

temporal/api/common/v1/message.proto
cretz: We shouldn't assume those requests will manifest. I think you're making something unnecessarily generic here.

bergundy:
Converted to draft to iron out the design.

temporal/api/common/v1/message.proto
bergundy: My goto for that was to also allow loading headers from a URL (or file).

temporal/api/common/v1/message.proto
bergundy: Not sure if we don't want to reference FS directly, our config also references certs on disk and I don't think referencing the config is better than referencing FS or any arbitrary URL.

Answered below for HTTP auth headers.
It's not included in this PR but I have given it thought.

temporal/api/common/v1/message.proto
bergundy: Not sure about that. I had considered it but considering how dynamic this config is and that it's specific to (in this case) an outgoing nexus service, I think we want to be able to define this in the API.

Even if a group of services uses a common encryption key, I still think we'll want to have the references from the service objects to that key than have the key in config specify which services it applies to.

bergundy:
This isn't relevant anymore, it may be revisited in the future.

PR #384, opened 2024-03-29
Add redirect_counter field to task started events and more

ShahabT:
**What changed?**
- Add `redirect_counter` field to task started events.
- Rename TaskQueueVersionGroup to TaskQueueVersionSelection and get rid of the `oneof`.

**Why?**
- `redirect_counter` is used by server to synchronize build ID redirects when rebuilding history. it is needed to make sure we reapply the redirects correctly for workflows with concurrent activities.
- Regarding TaskQueueVersionSelection, in some scenarios caller may want to specify more than one option. For example, the operator may want to get the result for a fixed list of build IDs plus any active build IDs that may be missing from its radar. Or if it wants to get the results for some build IDs and "unversioned" together.

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -196,7 +196,7 @@ message WorkflowTaskStartedEventAttributes {
     temporal.api.common.v1.WorkerVersionStamp worker_version = 6;
     // internally used by server to properly reapply build ID redirects to an execution
     // when rebuilding it from events.
-    int32 redirect_counter = 7;
+    int64 redirect_counter = 7;
 }
 
 message WorkflowTaskCompletedEventAttributes {
@@ -314,7 +314,7 @@ message ActivityTaskStartedEventAttributes {
     temporal.api.common.v1.WorkerVersionStamp worker_version = 6;
     // internally used by server to properly reapply build ID redirects to an execution
     // when rebuilding it from events.
-    int32 redirect_counter = 7;
+    int64 redirect_counter = 7;
 }
 
 message ActivityTaskCompletedEventAttributes {
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -194,6 +194,9 @@ message WorkflowTaskStartedEventAttributes {
     int64 history_size_bytes = 5;
     // Version info of the worker to whom this task was dispatched.
     temporal.api.common.v1.WorkerVersionStamp worker_version = 6;
+    // internally used by server to properly reapply build ID redirects to an execution
+    // when rebuilding it from events.
+    int32 redirect_counter = 7;
 }
 
 message WorkflowTaskCompletedEventAttributes {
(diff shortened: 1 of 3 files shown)


cretz (approved):
I admit not understanding the implementation details of `redirect_counter`, so I'm approving purely from an API POV. I would request a server team member outside of the versioning project to confirm implementation need. Also, I assume we'll confirm implementation before merging `versioning-2` branch into `main`.

ShahabT:
thanks @cretz. I'll get appropriate approvals from the server team before merging the API or the implementation into main.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -875,7 +875,7 @@ message DescribeTaskQueueRequest {
     // mentioned in the first unconditional Assignment Rule. If there is no default Build ID, the result for the
     // unversioned queue will be returned.
     // (-- api-linter: core::0140::prepositions --)
-    temporal.api.taskqueue.v1.TaskQueueVersionGroup versions = 6;
+    temporal.api.taskqueue.v1.TaskQueueVersionSelection versions = 6;
```
carlydf, line 878: could you also add these fields to `UpdateWorkerVersioningRulesResponse` in this file?

```
     repeated temporal.api.taskqueue.v1.TimestampedBuildIdAssignmentRule assignment_rules = 1;
    repeated temporal.api.taskqueue.v1.TimestampedCompatibleBuildIdRedirectRule compatible_redirect_rules = 2;
```

carlydf (approved):

temporal/api/workflowservice/v1/request_response.proto
carlydf: so that it matches `GetWorkerVersioningRulesResponse`

temporal/api/history/v1/message.proto
```diff
@@ -309,6 +312,9 @@ message ActivityTaskStartedEventAttributes {
     temporal.api.failure.v1.Failure last_failure = 5;
     // Version info of the worker to whom this task was dispatched.
     temporal.api.common.v1.WorkerVersionStamp worker_version = 6;
+    // internally used by server to properly reapply build ID redirects to an execution
```
carlydf, line 315: I think you should capitalize this for consistency, since it's the start of a sentence?

carlydf:
```suggestion
    // Used by server internally to properly reapply build ID redirects to an execution
```

temporal/api/history/v1/message.proto
```diff
@@ -194,6 +194,9 @@ message WorkflowTaskStartedEventAttributes {
     int64 history_size_bytes = 5;
     // Version info of the worker to whom this task was dispatched.
     temporal.api.common.v1.WorkerVersionStamp worker_version = 6;
+    // internally used by server to properly reapply build ID redirects to an execution
+    // when rebuilding it from events.
+    int64 redirect_counter = 7;
```
cretz, line 199: Now that I think about it, since this is specific to build ID things, I think we should qualify this field:

```suggestion
    int64 build_id_redirect_counter = 7;
```

PR #385, opened 2024-03-29
Add field to update search attributes of a schedule

rodrigozhou:
**What changed?**
Add field to update search attributes of a schedule

**Why?**
Add functionality to update search attributes of the scheduler workflow.

**Breaking changes**
No.

**Server PR**
https://github.com/temporalio/temporal/pull/5632

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1071,6 +1071,12 @@ message UpdateScheduleRequest {
     string identity = 5;
     // A unique identifier for this update request for idempotence. Typically UUIDv4.
     string request_id = 6;
+    // Schedule search attributes to be updated.
+    // Do not set this field if you do not want to update the search attributes.
+    // A non-null empty object will set the search attributes to an empty map.
+    // Note: you cannot only update the search attributes with `UpdateScheduleRequest`,
+    // you must also set the `schedule` field; otherwise, it will unset the schedule.
+    temporal.api.common.v1.SearchAttributes search_attributes = 7;
 }
 
 message UpdateScheduleResponse {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -3432,6 +3432,10 @@
         "requestId": {
           "type": "string",
           "description": "A unique identifier for this update request for idempotence. Typically UUIDv4."
+        },
+        "searchAttributes": {
+          "$ref": "#/definitions/v1SearchAttributes",
+          "description": "Schedule search attributes to be updated.\nDo not set this field if you do not want to update the search attributes.\nA non-null empty object will set the search attributes to an empty map.\nNote: you cannot only update the search attributes with `UpdateScheduleRequest`,\nyou must also set the `schedule` field; otherwise, it will unset the schedule."
         }
       }
     },
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -6604,6 +6604,15 @@ components:
         requestId:
           type: string
           description: A unique identifier for this update request for idempotence. Typically UUIDv4.
+        searchAttributes:
+          allOf:
+            - $ref: '#/components/schemas/SearchAttributes'
+          description: |-
+            Schedule search attributes to be updated.
+             Do not set this field if you do not want to update the search attributes.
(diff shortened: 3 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1039,6 +1039,10 @@ message UpdateScheduleRequest {
     string identity = 5;
     // A unique identifier for this update request for idempotence. Typically UUIDv4.
     string request_id = 6;
+    // Schedule search attributes to be updated.
+    // If the value is not nil, then it will either set or overwrite the value for the key.
+    // If the value is nil, then the search attribute will be removed from the schedule.
+    temporal.api.common.v1.SearchAttributes search_attributes = 7;
```
cretz, line 1045: I think we should do a complete replacement if this is non-nil. The SDKs are built to give the user the current schedule to use to build the update if they need, but lots of users instead use their own known state to update to and don't want it applied as a partial.

rodrigozhou: I thought about this, but the issue is we only have the UpsertSearchAttribute API...

dnr: The two ways to go are:
1. Interface consistency: this should do a complete replacement, and actually I'd say if this is nil, it should remove all search attributes. That matches how `schedule` works. The server would have to diff and unset ones that aren't given.
2. Server implementation simplicity: This should be an upsert, and if someone wants to remove one they they can explicitly set it to nil, the same way that other search attribute upsert works. The sdk can do the diff and set to nil if not specified.

I have a slight preference for 1, but I don't care that much.

cretz: I have a bit stronger preference for 1, but would remove the null-means-remove-all requirement due to backwards compatibility reasons with people using the update API today. They can submit a non-null empty message if they want to remove all. The user-facing behavior should not be affected by our internal limitations if we can help it.

rodrigozhou: So, if the user leave this field empty, should we unset all search attributes? This would be a breaking change. But then, how could the user unset all search attributes?

dnr: True, we could have a non-nil message with an empty map mean unset all. Or a separate boolean that says whether to pay attention to this message. But the empty message thing is cleaner.

cretz: > So, if the user leave this field empty, should we unset all search attributes? This would be a breaking change. But then, how could the user unset all search attributes?

No, `null` should not make any changes to this field for compatibility reasons.

rodrigozhou: It will also have a discrepancy in behavior compared to the `schedule` field above. When it's `nil`, the frontend handler is setting to an empty object, and overwriting the existing schedule.
But I think it's still better than half-replacement-half-upsert.

rodrigozhou: Btw, because `schedule` of this ^, we cannot only update the search attributes map. If the `schedule` is `nil`, then it will unset the schedule. cc: @dnr

cretz: > It will also have a discrepancy in behavior compared to the schedule field above. When it's nil, the frontend handler is setting to an empty object, and overwriting the existing schedule.

This makes perfect sense, there is no compatibility concern on `schedule` field.

> Btw, because schedule of this ^, we cannot only update the search attributes map. If the schedule is nil, then it will unset the schedule. cc: @dnr

The approach to updating a schedule is call get, then mutate, then call update. So if you wanted to only update search attribute map, you call get, then you mutate the search attribute map, then you call update.

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
dnr: Yeah, I think we all agree: the ideal semantics is replace-everything, including search attributes. Because of compatibility, we can't do that with this field though. So the compromise is that if it's set at all, we replace search attributes, otherwise ignore that one thing and replace the rest of the schedule. Clients should always set everything.

dnr (approved):

PR #386, opened 2024-04-02
Add execution duration to WorkflowExecutionInfo

rodrigozhou:
**What changed?**
Add execution duration to WorkflowExecutionInfo

**Why?**
We don't want to have the UI infer the value, so it will be consistent to whatever the server populates.

**Breaking changes**
No.

**Server PR**
https://github.com/temporalio/temporal/pull/5651

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -58,6 +58,9 @@ message WorkflowExecutionInfo {
     int64 history_size_bytes = 15;
     // If set, the most recent worker version stamp that appeared in a workflow task completion
     temporal.api.common.v1.WorkerVersionStamp most_recent_worker_version_stamp = 16;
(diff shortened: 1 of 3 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -58,6 +58,7 @@ message WorkflowExecutionInfo {
     int64 history_size_bytes = 15;
     // If set, the most recent worker version stamp that appeared in a workflow task completion
     temporal.api.common.v1.WorkerVersionStamp most_recent_worker_version_stamp = 16;
+    google.protobuf.Duration execution_duration = 17;
```
cretz, line 61: Can you clarify whether this is always non-null and whether it's guaranteed to be a subtraction between two fields above (and which two fields)?

cretz (approved):

yycptt (approved):

PR #387, opened 2024-04-02
Rename Update Requested to Update Admitted

stephanos:
**What changed?**

Renamed "Update Requested" to "Update Admitted".

**Why?**

After some discussion, we came to the conclusion that the name should align with Update's wait stage in order to simplify the mental model for the user. "Requested" would be yet another concept, but "Admitted" already exists and should be used instead.

**Breaking changes**

Yes, this is a breaking change. To the best of my knowledge, the renamed types have not been released anywhere yet. Certainly not in Server v1.23.

**Server PR**

_still working on that..._


Makefile
```diff
@@ -37,7 +37,7 @@ $(PROTO_OUT):
 	mkdir $(PROTO_OUT)
 
 ##### Compile proto files for go #####
-grpc: buf-lint api-linter buf-breaking clean go-grpc fix-path
+grpc: buf-lint api-linter clean go-grpc fix-path
```
stephanos, line 40: this will be re-enabled after merge

dandavison (approved):
LGTM!

cretz (approved):
In the future, for larger bits of work like this, I wonder if we should be working in a branch until implementation is complete (similar to what versioning is doing).

PR #389, opened 2024-04-04
Add a flag to the API payload to force a completion of activity

alexseedkou:
**What changed?**

This is the first PR for the change request in this [issue](https://github.com/temporalio/temporal/issues/987)

Add a `bool` flag in `RespondActivityTask*` payload to control the behavior when `CompleteActivityByID ` is called.

**Why?**
In async workflow, a bug or transit issue may cause an activity to fail. This will continuously fail a callback of some other process as the previous activity is backing off. With this change, we introduce a flag to let the server know we want to force the completion of the activity even if it is backing off in order to unblock the workflow.

Please refer to the [issue](https://github.com/temporalio/temporal/issues/987) for more discussions.

**Breaking changes**
We need to think about the backward compatibility with this change. It is safe for the change in this PR. But in the following change (which may include the update on the client SDK), we need to be careful about backward compatibility as we need to ask for the flag information from an external system.

**Server PR**

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -2298,6 +2298,10 @@
         "identity": {
           "type": "string",
           "title": "The identity of the worker/client"
+        },
+        "forceCompletion": {
+          "type": "boolean",
+          "title": "Toggle to force a completion of an activity\nWhen set as true, worker will complete an activity no matter if there is any pending activites"
         }
       }
     },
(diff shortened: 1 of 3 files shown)


bergundy (commented):
This LGTM but we're discussing internally whether this should just be the default server behavior and not require the flag.

Thanks!

alexseedkou:
> This LGTM but we're discussing internally whether this should just be the default server behavior and not require the flag.
> 
> Thanks!

Sounds good, please keep me posted and I am happy to continue working on this.

Thank you!

bergundy (commented):
After some internal discussion, we've decided to change the server behavior to accept activity completions even if the activity is currently backing off by default.

We may add an `attempt` field to these APIs if a user wants to target a specific attempt but let's get some feedback first.

PR #393, opened 2024-04-09
Versioning 2 API

ShahabT:
**What changed?**
- `UpdateWorkerVersioningRules` API
- `GetWorkerVersioningRules` API
- Enhanced `DescribeTaskQueue` API
- Added fields to the following proto messages to track versioning info
  - WorkflowExecutionInfo
  - PendingActivityInfo
  - WorkflowExecutionStartedEventAttributes
  - WorkflowTaskStartedEventAttributes
  - ActivityTaskStartedEventAttributes
- Wire-safe renaming of `use_compatible_version` to `inherit_build_id`
- Getting rid of bundle_id field in WorkerVersionStamp. (the field is not used in any SDK, nor in server)

**Why?**
The new API provides better versioning experience for users, and provides a better foundation for us to build more features on top of.

**Breaking changes**
Removed bundle_id field in WorkerVersionStamp (but it's confirmed not to be used anywhere, so no breaking change really)

**Server PR**
https://github.com/temporalio/temporal/pull/5691


temporal/api/history/v1/message.proto
cretz, line 1: It should be noted, this actually will break workflow replays for users using versioning today. Today when people download/save history, they use JSON format, and they keep this JSON format to run on replayers in SDKs. Changing field names is a backwards incompatible history format change from the perspective of replayers.

This may be ok, but it needs to be made _very_ clear. We also may want to offer a way in CLI or something to "fix" invalid history, up to y'all.

cretz (commented):
LGTM. Notes/questions::

1. Can you confirm that, besides the SDK replaying now-changed history JSON, there are no other issues you can think of in any SDK with the _runtime_ component of workflows? Meaning no workers or workflows will break because of this? (but the administrative side it may make sense to break)
2. Let's add a "WARNING: Versioning is not yet stable and the API and behavior may change incompatibly" above the unstable pieces. We don't technically have to and we didn't last time and we for sure will mark them this way in the SDK interfacing with these APIs, but I think it's also good here.

temporal/api/history/v1/message.proto
ShahabT: Thanks for bringing up the replay history from JSON issue. I did not think about that. Will make sure we mention that in the docs. Maybe some instructions for the user to replace "use_compatible_version" with "inherit_build_id" in the JSON string would suffice.

Regarding the first question, I cannot think of anything else that can break. We have functional tests for old and new versioning (and of course unversioned wfs). The "features" test pass in the versioning-2 branch which use old/current SDKs. As you noticed, the changes are mostly for the admin side, not the data plane. the fields added to data-plain are all optional and mostly for the use of server itself, and only for versioned workflows.

Added the warning to the RPC calls.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,6 +54,47 @@ message TaskQueueMetadata {
...
+    bool unversioned = 2;
+    // Include all active versions. A version is considered active if it has had new
+    // tasks or polls recently.
+    bool all_active = 3;
+}
+
+message TaskQueueVersionInfo {
+    // Empty means unversioned.
+    string build_id = 1;
+    repeated TaskQueueTypeInfo types_info = 3;
```
dnr, line 71: It's not a huge difference, but consider making this a `map<temporal.api.enums.v1.TaskQueueType, TaskQueueTypeInfo>` and removing `type` from that? (sorry, I didn't review this on the feature branch)

temporal/api/taskqueue/v1/message.proto
```diff
@@ -54,6 +54,47 @@ message TaskQueueMetadata {
...
+    temporal.api.enums.v1.TaskQueueType type = 1;
+    // Unversioned workers (with `useVersioning=false`) are reported in unversioned result even if they set a Build ID.
+    repeated PollerInfo pollers = 2;
+    // BacklogInfo backlog_info = 3;
+}
+
+// For workflow task queues, we only report the normal queue metrics, not sticky queues. This means the stats
+// reported here do not count all workflow tasks. However, because the tasks queued in sticky queues only remain
+// valid for a few seconds, the inaccuracy becomes less significant as the backlog age grows.
+message BacklogInfo {
```
dnr, line 85: should we comment this whole thing out for now, if we're not using it?

temporal/api/enums/v1/task_queue.proto
```diff
@@ -78,3 +79,30 @@ enum TaskReachability {
...
+    // Not applicable to activity-only task queues.
+    BUILD_ID_TASK_REACHABILITY_CLOSED_WORKFLOWS_ONLY = 2;
+    // Build ID is not used for new executions, nor it has been used by any existing execution
+    // within the retention period.
+    BUILD_ID_TASK_REACHABILITY_UNREACHABLE = 3;
+}
+
+enum DescribeTaskQueueMode {
+    // Unspecified means legacy behavior.
+    DESCRIBE_TASK_QUEUE_MODE_UNSPECIFIED = 0;
```
dnr, line 105: this means if you don't want to use ENHANCED mode, you have to use UNSPECIFIED, which is kind of weird. should there be an explicit BASIC and then have UNSPECIFIED default to BASIC?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -878,17 +878,47 @@ message DescribeWorkflowExecutionResponse {
...
 
 message DescribeTaskQueueResponse {
+    // Deprecated. Use `versions_info.types_info.pollers` with `ENHANCED` mode instead.
+    // Not set in `ENHANCED` mode.
     repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
+    // Deprecated. Not set in `ENHANCED` mode.
     temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2;
+
+    // Only set in `ENHANCED` mode.
+    repeated temporal.api.taskqueue.v1.TaskQueueVersionInfo versions_info = 3;
```
dnr, line 921: maybe this should be a `map<string, TaskQueueVersionInfo>` and remove `build_id` from that?

temporal/api/enums/v1/task_queue.proto
ShahabT: the idea is no one should need to ask for basic mode

dnr: not even if they want a cheaper call?

temporal/api/history/v1/message.proto
cretz: Regarding the first question, there is a problem with older UI/CLI (like only recently changed IIRC) where even unset fields are present in the JSON, but most of our replayers ignore unknown JSON fields. Also, sorry I forget, but can you tell whether any of the fields that were changed in the history events are set _implicitly_ in the SDK instead of by explicitly by the workflow author (e.g. as a result of worker setting only)?

I will circulate the history JSON change internally to gather thoughts there.

temporal/api/enums/v1/task_queue.proto
ShahabT: If by cheaper you mean something that only looks at local info for root partition and does not fan out to all partition, I'd still want to put that as an option for the enhanced mode so we return that result in the new structure. (of course if we really thing the cheap option would be useful enough).

temporal/api/taskqueue/v1/message.proto
ShahabT: removed this for now

ShahabT:
The CI fails for expected reasons due to field renames. Need to an admin to bypass the check.

cretz (approved):
Confirmed with SDK team that the field name changes in JSON should be ok

temporal/api/enums/v1/task_queue.proto
dnr: Hmm, as long as we can turn off the reachability queries, which I think is the real expensive part, that's okay. Fan-out is fine to have by default (it should always have been doing that).

But what about the "internal" stuff that's deprecated? ack level, rate limit, etc. We don't have InternalTaskQueueStatus yet so now there's no "supported" way to get that info?

temporal/api/taskqueue/v1/message.proto
dnr: Huh. It could be int32 with a comment, though? Now it's weird to have one level be a map and the next level not.

temporal/api/enums/v1/task_queue.proto
ShahabT: The plan is to add a tdbg command for the internal info before CLIs stop supporting legacy mode.

dnr: ok, I'm fine then (not a big deal anyway)

dnr (approved):

PR #395, opened 2024-04-12
Including `approximate_backlog_counter` in Request/Response protos

Shivs11:
### Describe what has changed in this PR
- Including `approximate_backlog_counter` as part of the Request/Response protos so that it can be checked via the `DescribeTaskQueue` api.

- Moreover, when `versioning-2` was merged with `master`, a [PR](https://github.com/temporalio/api/pull/393/commits/d98f69bbecbaa22f24af4887344ded9f0ec67fc4) was opened to remove changes which would not be used until now. This PR also includes those changes as it needs the `BacklogInfo` structure.

### Tell your future self why have you made these changes
Functional testing `approximate_backlog_counter`
Allowing the feature to be present for clients/workers call `DescribeTaskQueue`.

### Are there any breaking changes on binary or code level? 
No.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -920,6 +920,9 @@ message DescribeTaskQueueResponse {
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
     // Only set in `ENHANCED` mode.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3;
+
+    // Should this also be set only in the `ENHANCED` mode?
```
ShahabT, line 924: Yes

ShahabT (approved):
Lgtm other than the unfinished comment

temporal/api/workflowservice/v1/request_response.proto
Shivs11: Thanks, will alter.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -76,6 +76,21 @@ message TaskQueueTypeInfo {
     repeated PollerInfo pollers = 1;
 }
 
+// For workflow task queues, we only report the normal queue metrics, not sticky queues. This means the stats
+// reported here do not count all workflow tasks. However, because the tasks queued in sticky queues only remain
+// valid for a few seconds, the inaccuracy becomes less significant as the backlog age grows.
+message BacklogInfo {
+    // The approximate number of tasks backlogged in this task queue. May count expired tasks but eventually converges
+    // to the right value.
+    int64 approximate_backlog_count = 1;
```
Sushisource, line 85: We get a sticky queue backlog hint from workflow task polling responses. Would it be feasible to include that here as well? If it's not a meaningful extra cost, I'd say it's worth putting that in here too.

ShahabT: @Sushisource If I understand the ask correctly, wouldn't it be confusing to have two counts in backlog info? The plan is to use this more accurate value in the poll response as well eventually.

Sushisource: Why would that be confusing? One of them would be specifically for sticky (if applicable).

And yeah, great that the more accurate value will be in the poll response too - I just don't see a reason to have both counts available in both places if they're easy to access. In the poll responses is useful for worker's internals, in describe is useful for external scalers.

I'm not hugely invested in it - just might as wall have that info if it's readily accessible.

ShahabT: I was suggesting to use the more accurate count that we track for `approximate_backlog_count` to populate the hint in the poll response as well, not to put both values in poll response.
For DescribeTaskQueue response, the enhanced mode is not supposed to support sticky queue (at least so far) as the idea is to keep it at user level ([TaskQueueFamily](https://github.com/temporalio/temporal/blob/main/common/tqid/task_queue_id.go#L51)). It aggregates stats across normal partitions but not sticky partitions. (And yes, we call sticky queues [sticky partitions](https://github.com/temporalio/temporal/blob/main/common/tqid/task_queue_id.go#L104) now).

dnr: From the matching service pov, the existing backlog count "hint" is not just for sticky, it's sent with every poll response. It only accounts for that one partition (and one physical queue for versioning), so for non-sticky it's less meaningful. Well, it's a lower bound I guess.

If we do eventually change the hint in the poll response, we should do it carefully since different semantics there can actually cause problems in certain situations.

In the meantime... if the "hint" is used for worker internals, and the new "approximate" is used for scaling, then it makes sense to me that the hint is only on poll responses and the approximate is only on describe.

PR #398, opened 2024-04-19
Add MultiOperationExecutionFailure

stephanos:
**What changed?**

Added `MultiOperationExecutionFailure` and updated comments.

Usage can be seen in https://github.com/temporalio/api-go/pull/158

**Why?**

The current MultiOperation error response isn't easy to parse; by adding this explicit type it can be parsed unamigously.

**Breaking changes**

Changes the error behavior of the experimental MultiOperation API. (not in use yet)

**Server PR**

Not needed.


google/rpc/status.proto
stephanos, line 1: Copied from https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto

cretz, line 1: I am a little worried about the inclusion of this object with our SDKs and how they code gen. Up until now we have only included third party protos that are either part of standard protobuf (so are never gen'd, they come with the protobuf library of that language) or are only used for annotations (so can be ignored at runtime).

To now include a third party proto for use at runtime that _also_ ships with gRPC libraries of languages (some SDKs don't include gRPC dependencies) may prevent some challenges/conflicts (you could get duplicate type registration errors when they _do_ add a gRPC dependency for their own use).

If we want to go this route, I think we need to confirm for every SDK that it does not clash with that language's gRPC library use. An alternative would be to create your own message that mimics the gRPC status message.

cretz (commented):
Requesting we don't merge this until we understand the ramifications of including Google gRPC protos into SDKs separately from gRPC libraries.

google/rpc/status.proto
tdeebswihart: I'd prefer we create our own status message type that mimics the GRPC one; that will make integrating new languages easier as we don't know if any future language's GRPC library will have this conflict.

It also gives us full control of the object, so we could annotate it with additional details like which operation each failure is for (which the current approach doesn't do, as far as I can tell).

stephanos: 👍 

> it also gives us full control of the object, so we could annotate it with additional details like which operation each failure is for

My intent was/is to use the exact same `Status` message so it can be re-used (after a conversion step) by the existing SDK code. Not knowing the other SDKs, at least in Go it unlocks straight-forward code re-use that makes the integration of this error type much easier.

cretz: Using `google.rpc.Status` _inside_ an `Any` has the same problems. People have to have that proto to use it. Some of our SDKs don't have this proto by default and can't have their own version because it'd clash in the global type registry if a user did have gRPC w/ its form of the proto w/ the same qualified name (`Any` just contains the type URL as a qualified name of the message it usually looks up in a global type registry).

> Not knowing the other SDKs, at least in Go it unlocks straight-forward code re-use that makes the integration of this error type easier.

This is one of those (rare) cases where the different SDKs make a difference. The only reason it works in Go is because we depend on https://pkg.go.dev/google.golang.org/grpc which transitively depends on https://pkg.go.dev/google.golang.org/genproto/googleapis/rpc/status. Some other SDKs do not have a representation of the gRPC status proto because they do not depend on the gRPC library in lang (because Rust Core does their gRPC, not the lang themselves).

temporal/api/common/v1/message.proto
```diff
@@ -190,3 +191,29 @@ message Callback {
...
+    // A developer-facing error message, which should be in English. Any
+    // user-facing error message should be localized and sent in the
+    // [google.rpc.Status.details][google.rpc.Status.details] field, or localized
+    // by the client.
+    string message = 2;
+
+    // A list of messages that carry the error details.  There is a common set of
+    // message types for APIs to use.
+    repeated google.protobuf.Any details = 3;
+}
```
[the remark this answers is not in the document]
cretz: It's compatible to move a message later, might as well put this as a nested message in the only place that needs it for now IMO

google/rpc/status.proto
stephanos: Thank you for providing the SDK perspective, that's very helpful 👍

cretz (approved):
I wish we didn't have to have whole gRPC statuses, but I can see no better way

temporal/api/errordetails/v1/message.proto
```diff
@@ -112,3 +113,36 @@ message NewerBuildExistsFailure {
...
 }
+
+message MultiOperationExecutionFailure {
+    // One status for each requested operation from the failed MultiOperation. The failed
+    // operation(s) have the same error details as if it was executed separately. All other operations have the
+    // status code `Aborted` and `MultiOperationExecutionAborted` is added to the details field.
+    repeated RpcStatus statuses = 1;
+
+    // NOTE: `RpcStatus` is modelled after
+    // [`google.rpc.Status`](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto).
```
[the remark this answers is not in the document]
stephanos: That's helpful feedback 👍

tdeebswihart (approved):

temporal/api/errordetails/v1/message.proto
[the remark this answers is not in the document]
alexshtin: I don't like word `Rpc` here. It is "status of the step which is a single operation". `OperationStatus` looks better to me.

alexshtin (approved):

PR #401, opened 2024-04-24
Add HTTP annotations to Nexus registry endpoints

bergundy:
We're going to start working on UI soon for these registries.

### DIFF
--- a/temporal/api/operatorservice/v1/service.proto
+++ b/temporal/api/operatorservice/v1/service.proto
@@ -80,12 +80,19 @@ service OperatorService {
 
     // Get a registered incoming Nexus service by ID. The returned version can be used for optimistic updates.
     rpc GetNexusIncomingService(GetNexusIncomingServiceRequest) returns (GetNexusIncomingServiceResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/nexus/incoming-services/{id}"
+        };
     }
 
     // Create a Nexus service. This will fail if a service with the same name already exists in the namespace with a
     // status of ALREADY_EXISTS.
     // Returns the created service with its initial version. You may use this version for subsequent updates.
     rpc CreateNexusIncomingService(CreateNexusIncomingServiceRequest) returns (CreateNexusIncomingServiceResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/nexus/incoming-services"
+            body: "*"
+        };
     }
 
     // Optimistically update a Nexus service based on provided version as obtained via the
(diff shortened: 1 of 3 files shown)


cretz (commented):
I think it may have been more reasonable to have self-hosted use configuration for Nexus service management instead of making a full API/UI like we do for cloud where you can't set config. But meh, seems like the self-hosted-specific UI option has been chosen.

bergundy:
> I think it may have been more reasonable to have self-hosted use configuration for Nexus service management instead of making a full API/UI like we do for cloud where you can't set config. But meh, seems like the self-hosted-specific UI option has been chosen.

I'm surprised to hear you say this, you've been reviewing the additions to the operator service for a while and I haven't heard you raise this concern before. You've actually been advocating for a more consistent experience between self-hosted and cloud.

I definitely think that APIs are the right approach, especially considering that the dev server is so commonly used for local development.

pdoerner (approved):

temporal/api/operatorservice/v1/service.proto
```diff
@@ -94,46 +101,73 @@ service OperatorService {
...
+        };
     }
 
     // List all Nexus incoming services for the cluster, sorted by service ID in ascending order. Set page_token in the
     // request to the next_page_token field of the previous response to get the next page of results. An empty
     // next_page_token indicates that there are no more results. During pagination, a newly added service with an ID
     // lexicographically earlier than the previous page's last service name may be missed.
     rpc ListNexusIncomingServices(ListNexusIncomingServicesRequest) returns (ListNexusIncomingServicesResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/nexus/incoming-services"
```
rodrigozhou, line 123: Won't list have options? Wouldn't it better to be a POST?

temporal/api/operatorservice/v1/service.proto
```diff
@@ -94,46 +101,73 @@ service OperatorService {
...
+        };
     }
 
     // List all Nexus outgoing services for a namespace, sorted by service name in ascending order. Set page_token in
     // the request to the next_page_token field of the previous response to get the next page of results. An empty
     // next_page_token indicates that there are no more results. During pagination, a newly added service with a name
     // lexicographically earlier than the previous page's last service name may be missed.
     rpc ListNexusOutgoingServices(ListNexusOutgoingServicesRequest) returns (ListNexusOutgoingServicesResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/nexus/outgoing-services"
```
rodrigozhou, line 170: ditto

rodrigozhou (approved):

temporal/api/operatorservice/v1/service.proto
bergundy: Those would go in a query string.
POST shouldn't be used for read-only endpoints.
I also was being consistent with `ListWorkflows` where we take a query that's even harder to encode in a query string:

https://github.com/temporalio/api/blob/9670c6e8044342af5c71726939945e60413660c5/temporal/api/workflowservice/v1/service.proto#L388

PR #403, opened 2024-05-01
Nexus endpoints

bergundy:
**What changed?**

- Removed NexusOutgoingService definitions and endpoints
- Renamed NexusIncomingService to NexusEndpoint
- Added endpoint to workflow commands and events

**Why?**

We separated the concept of service and endpoint and decided to merge the incoming and outgoing registries.

**Breaking changes**

Very breaking but all of the usage is unreleased, behind a feature flag, and considered experiemental.

**Server PR**

https://github.com/temporalio/temporal/pull/5855


temporal/api/nexus/v1/message.proto
```diff
@@ -118,75 +117,55 @@ message Response {
...
-    // URL to invoke requests.
-    string url = 1;
-    // Callback URL that is attached to every StartOperation request for this service.
-    // The URL should be accessible to the handler of the StartOperation requests.
-    // The service handles callbacks in the /api/v1/namespaces/{namespace_name}/nexus/callback route. Note that if the
-    // namespace is renamed this URL will no longer be valid.
-    // See https://github.com/nexus-rpc/api/blob/main/SPEC.md#query-parameters for more information.
-    string public_callback_url = 2;
+// Target to route requests to.
+message EndpointTarget {
```
cretz, line 160: So this is combining outgoing and incoming endpoints? How would I provide, say, auth to outgoing or some other config to incoming? Or how would I say that this endpoint is not for outbound callers only for external callers?

Even if these use cases are not supported today, do you worry that by not retaining the separation between outbound endpoint configuration and inbound endpoint configuration, you are going to end up with a lot of "this only applies to outbound" and "this only applies to inbound" future options.

For user benefit, can we try to make this API have a similar shape as the cloud one except for cloud or self-hosted specifics?

bergundy: > So this is combining outgoing and incoming endpoints?

Yes

> How would I provide, say, auth to outgoing or some other config to incoming?

I was thinking we'd add another variant called "external" or something of that sort.

```proto
message External {
  string url = 1;
  TLSConfig tls_config = 2;
  map<string, string> header = 3;
}
```

Still need to figure out how to store secrets, but that's the idea.

> Or how would I say that this endpoint is not for outbound callers only for external callers?

Flag or enum.
There's value in saying this is an external endpoint that can be called externally, essentially making the frontend a proxy. It can be used when migrating namespaces between clusters when one of the clusters doesn't have access to the credentials (e.g. self-hosted to cloud or back).

> do you worry that by not retaining the separation between outbound endpoint configuration and inbound endpoint configuration, you are going to end up with a lot of "this only applies to outbound" and "this only applies to inbound" future options.

I think it's better this way, since these endpoints are specified in workflow code and we don't want ambiguity.
We can deal with the inbound and outbound specific options with a oneof.

> For user benefit, can we try to make this API have a similar shape as the cloud one except for cloud or self-hosted specifics?

Of course, Cloud will have to adjust.

cretz: > Of course, Cloud will have to adjust.

:+1: Will wait for the cloud one to compare

cretz (approved):
LGTM. A bit concerned of combining outbound and inbound endpoints, but it does work.

temporal/api/nexus/v1/message.proto
bergundy: Added `External`.

bergundy: No credentials yet, that's out of scope for now.

rodrigozhou (approved):

pdoerner (approved):

PR #406, opened 2024-05-20
Evolve documentation of update polling parameters

dandavison:


cretz (approved):

cretz (approved):

openapi/openapiv2.json
```diff
@@ -3457,7 +3457,7 @@
         },
         "waitPolicy": {
           "$ref": "#/definitions/v1WaitPolicy",
-          "description": "Describes when this request should return - basically whether the\nupdate is synchronous, asynchronous, or somewhere in between."
+          "description": "How long to wait before returning control to the caller."
```
yiminc, line 3460: How long implies that this specifies a duration?

dandavison: Yes, you're right. I've now changed it to
```proto
message UpdateWorkflowExecutionRequest {
    ...
    // When this request should return.
    temporal.api.update.v1.WaitPolicy wait_policy = 4;
    ...
}

message PollWorkflowExecutionUpdateRequest {
    ...
    // When this request should return. Omit to request a non-blocking poll.
    temporal.api.update.v1.WaitPolicy wait_policy = 4;
    ...
}
```

Sushisource (approved):
It would be cool if you could update the variants in the enum too to specify what they mean as a request (already exists) and what they mean as a return value (which I don't think they say, and would be nice to have).

PR #407, opened 2024-05-22
Replace GHA token in workflow to trigger api-go update

rodrigozhou:
**What changed?**
Replace GHA token in workflow to trigger api-go update

**Why?**
Current token is a PAT.

**Breaking changes**

**Server PR**

### DIFF
--- a/.github/workflows/trigger-api-go-update.yml
+++ b/.github/workflows/trigger-api-go-update.yml
@@ -16,14 +16,22 @@ jobs:
         shell: bash
 
     steps:
+      - name: Generate token
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.TEMPORAL_CICD_APP_ID }}
+          private-key: ${{ secrets.TEMPORAL_CICD_PRIVATE_KEY }}
+
       - name: Dispatch api-go Github Action
         env:
-          PAT: ${{ secrets.COMMANDER_DATA_TOKEN }}
-          PARENT_REPO: temporalio/api-go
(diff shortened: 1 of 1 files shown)


stephanos (approved):

.github/workflows/trigger-api-go-update.yml
```diff
@@ -16,14 +16,22 @@ jobs:
...
 
     steps:
+      - name: Generate token
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.TEMPORAL_CICD_APP_ID }}
+          private-key: ${{ secrets.TEMPORAL_CICD_PRIVATE_KEY }}
+
       - name: Dispatch api-go Github Action
```
tdeebswihart, line 26: Can we just use `uses: temporalio/api-go/blahblah@main` instead?

rodrigozhou: With `uses: temporalio/api-go/blahblah@main`, it would run the action in the `api` repo, not in `api-go`, which would probably mess up with `checkout` action for example.

cretz: @tdeebswihart - now I wonder if my suggestion at https://github.com/temporalio/cli/pull/575 was a bad one if you want to trigger a workflow on the other repo. May need to change that to do what's here too (sorry).

tdeebswihart: Well it's easy to revert @cretz :)

Quinn-With-Two-Ns (approved):

PR #411, opened 2024-05-28
Rename BacklogInfo to TaskQueueStats

ShahabT:
**What changed?**
Rename BacklogInfo to TaskQueueStats.

**Why?**
Some of the stats in this struct are not only about backlog and concerns the sync-matched tasks as well. e.g. task_add_rate.

**Breaking changes**

**Server PR**

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -1196,8 +1196,8 @@
             "collectionFormat": "multi"
           },
           {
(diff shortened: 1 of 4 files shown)


Shivs11 (approved):
LGTM. I agree with the renaming.

cretz (approved):

PR #414, opened 2024-05-29
Task Queue Statistics 

Shivs11:
**What changed?**
- aims to bring in the changes required for making task queue metrics visible via the `DescribeTaskQueue` API 

**Breaking changes**
None

**Server PR**
[#6022](https://github.com/temporalio/temporal/pull/6022)


ShahabT (approved):

temporal/api/taskqueue/v1/message.proto
```diff
@@ -84,6 +84,24 @@ message TaskQueueVersionInfo {
 message TaskQueueTypeInfo {
     // Unversioned workers (with `useVersioning=false`) are reported in unversioned result even if they set a Build ID.
     repeated PollerInfo pollers = 1;
+    TaskQueueStats stats = 3;
```
dnr, line 87: was tag number 2 used before? if so there should be a `reserved 2;` in there

dnr (approved):

cretz (approved):

temporal/api/taskqueue/v1/message.proto
ShahabT: no, this message was added with the versioning-2 branch. tag 2 has not been used.
```suggestion
    TaskQueueStats stats = 2;
```

temporal/api/taskqueue/v1/message.proto
```diff
@@ -84,6 +84,24 @@ message TaskQueueVersionInfo {
...
+}
+
+// For workflow task queues, we only report the normal queue stats, not sticky queues. This means the stats
+// reported here do not count all workflow tasks. However, because the tasks queued in sticky queues only remain
+// valid for a few seconds, the inaccuracy becomes less significant as the backlog age grows.
+message TaskQueueStats {
+    // The approximate number of tasks backlogged in this task queue. May count expired tasks but eventually converges
+    // to the right value.
```
Sushisource, line 95: More detail about how the approximation works would be helpful for SDKs

temporal/api/taskqueue/v1/message.proto
```diff
@@ -84,6 +84,24 @@ message TaskQueueVersionInfo {
...
+    int64 approximate_backlog_count = 1;
+    // Approximate age of the backlog based on the create timestamp of the task at the head of the queue.
```
Sushisource, line 97: Is this the approximate age of the _backlog estimate_ (count)? The approximate age of the _backlog_ would imply the time since there has been any task in the queue.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -84,6 +84,24 @@ message TaskQueueVersionInfo {
...
+    google.protobuf.Duration approximate_backlog_age = 2;
+    // Approximate tasks per second added to the task queue based on recent activity. This includes both backlogged and
+    // sync-matched tasks.
+    float tasks_add_rate = 3;
+    // Approximate tasks per second dispatched to workers based on recent activity. This includes both backlogged and
+    // sync-matched tasks.
```
Sushisource, line 103: A more concrete definition of "recent" would be useful too. Is it a fixed window? Configurable? If so which property?

temporal/api/taskqueue/v1/message.proto
Shivs11: Sure, the counter is incremented during task creation and is decremented as and when the Ack Level moves past a given task. 

Moreover, the reason why it's an approximation is due to the fact that there are scenarios where the counter could be an over or an under estimate, as explained [here](https://github.com/temporalio/temporal/blob/89fcc66d9bdf7aaa7012a1fa9c742fb40c2b1b49/service/matching/backlog_manager.go#L304)

temporal/api/taskqueue/v1/message.proto
Shivs11: No, this would be the age of the backlog based on the create timestamp of the task at the head of the queue.

temporal/api/taskqueue/v1/message.proto
Shivs11: Yes, the window is fixed to a maximum limit of 30 seconds.
It's not decided if it should be configurable or not yet, but I believe keeping 30 seconds for a window size seems reasonable. 

Thoughts?

temporal/api/taskqueue/v1/message.proto
Sushisource: I mean adding that to the docstring would be great :)

dnr: Actually I think we should keep this one a little vague. The implementation is totally different for SQL backends, for example, and we want to be able to change it in the future. What we should say here is the intention (this is the value you want to use for worker scaling) and the semantics (it's approximate, it may over-count, it may under-count too, more rarely, and should eventually converge).

temporal/api/taskqueue/v1/message.proto
dnr: Instead of "age of the backlog", how about "age of the oldest task in the backlog"

temporal/api/taskqueue/v1/message.proto
dnr: I'd also vote for keeping this vague. It'll probably be configurable, but isn't yet. For an api consumer, what they're doing with the value probably shouldn't depend on the exact window size. If they do care to that degree, they should keep their own metrics intstead.

temporal/api/taskqueue/v1/message.proto
Sushisource: Yep that works for me

temporal/api/taskqueue/v1/message.proto
Sushisource: I think an order of magnitude is at least reasonable to put in here. Beyond that, once it is configurable, referencing the property name is good - even if a consumer doesn't care it's just useful for our own reference.

Shivs11: I altered the comment to have a certain order of magnitude + keeping it as vague as possible - lmk if this is what we were going for

Sushisource (approved):
Thanks for the tweaks!

PR #415, opened 2024-05-30
Remove nexus and workflow_metadata from buf breaking ignore list

bergundy:


buf.yaml
```diff
@@ -13,8 +13,6 @@ breaking:
     - google
     # Remove this after user metadata is merged
     - temporal/api/sdk/v1/workflow_metadata.proto
-    # Remove after applying nexus spec metadata change
-    - temporal/api/nexus/v1/message.proto
```
cretz, line 17: Can remove the above one too if you make another commit here (if not no worries I'll do it soon).

cretz (approved):

buf.yaml
bergundy: Sure

bergundy: Done.

yycptt (approved):

PR #416, opened 2024-05-31
Replace GHA token in workflow to trigger api-go update

rodrigozhou:
**What changed?**
Replace GHA token in workflow to trigger api-go update (redo https://github.com/temporalio/api/pull/407 which was reverted in https://github.com/temporalio/api/pull/408).

Also added an input for `workflow_dispatch` event so we can actually manually trigger it.

**Why?**
Replace the token so it won't use my PAT.
Fix `workflow_dispatch` event since there are required fields in `update-proto.yml`.

**Breaking changes**

**Server PR**

### DIFF
--- a/.github/workflows/trigger-api-go-update.yml
+++ b/.github/workflows/trigger-api-go-update.yml
@@ -5,6 +5,11 @@ on:
     branches:
       - master
   workflow_dispatch:
+    inputs:
+      branch:
+        description: "Branch in api-go repo to trigger update protos (default: master)"
+        required: true
+        default: master
 
 jobs:
   notify:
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-api-go-update.yml
```diff
@@ -16,14 +21,51 @@ jobs:
...
+          owner: ${{ github.repository_owner }}
+          repositories: api-go # generate a token with permissions to trigger GHA in api-go repo
+
+      - name: Prepare inputs
+        id: prepare_inputs
+        run: |
+          case "${{ github.event_name }}" in
+            "push")
+              BRANCH=${{ github.event.ref }}
+              BRANCH=${BRANCH#refs/heads/}
```
cretz, line 39: Doesn't this appending duplicate work occurring below?

rodrigozhou: Oh, yes. Forgot to remove below.

cretz (approved):

stephanos (approved):

PR #418, opened 2024-06-05
Add new Enhanced Stack Trace proto

twin-drill:
**What changed?**
Added new Enhanced Stack Trace proto, as laid out in the design proposal:
https://github.com/temporalio/proposals/blob/master/all-sdk/enhanced-stack-trace-queries.md


temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
```diff
@@ -0,0 +1,71 @@
...
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "EnhancedStackTraceMetadataProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+// Internal structure used to create worker stack traces with references to code.
+message EnhancedStackTraceMetadata {
+    SDKInfoDefinition sdk = 1;
```
Sushisource, line 36: I'm debating whether or not we should just flatten this into `sdk_name` and `sdk_version` fields.

Anyone else have an opinion there?

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
```diff
@@ -0,0 +1,72 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
```
bergundy, line 3: nit: it's 2024.

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
```diff
@@ -0,0 +1,72 @@
...
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    uint32 column = 3;
+    string function_name = 4;
+    bool hide_default = 5;
+}
+
+// Collection of FileLocation messages from a single stack.
+message StackTraceDefinition {
+    repeated FileLocationDefinition locations = 1;
+}
```
bergundy, line 72: Drop the `Definition` from all of these messages, it's a given when you're defining schemas.

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
```diff
@@ -0,0 +1,66 @@
...
+message FileLocation {
+    string file_path = 1;
+    // (-- api-linter: core::0141::forbidden-types=disabled
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    uint32 line = 2;
+    // (-- api-linter: core::0141::forbidden-types=disabled
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    uint32 column = 3;
+    string function_name = 4;
+    bool hide_default = 5;
```
bergundy, line 60: What is this field? It's not in the spec / TS and not documented.

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
```diff
@@ -0,0 +1,66 @@
...
+
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "EnhancedStackTraceMetadataProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+// Internal structure used to create worker stack traces with references to code.
+message EnhancedStackTraceMetadata {
```
bergundy, line 35:
```suggestion
message EnhancedStackTrace {
```

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
bergundy, line 1: I would consider moving this into a sub package named `stack_trace` so all of the messages are in context.

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
```diff
@@ -0,0 +1,66 @@
...
+    string sdk_name = 1;
+    string sdk_version = 2;
```
bergundy, line 37: You've broken compatibility with the spec / TS. That's fine with me but please update both.

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
twin-drill: The `FileLocation` message contains the filename and other 'location' properties of the trace. I found it in the proposal (https://github.com/temporalio/proposals/blob/master/all-sdk/enhanced-stack-trace-queries.md#:~:text=export%20interface%20FileLocation%20{) and implemented the traces in python using it. I'll be documenting the protos better, so if you've got feedback on it let me know!

twin-drill: Ah, and the `hide_default` flag is something I discussed with Spencer -- if a user wishes to hide traces from certain file locations then the SDK when creating `FileLocations` can either not return the specific `FileLocation`s at all, or use the flag to communicate that to the UI.

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
cretz: I don't think this should be a separate package, but I do think each message name herein should be prefixed with `StackTrace`

temporal/api/sdk/v1/enhanced_stack_trace_metadata.proto
bergundy: I think the absence of sources or traces is a good enough indicator for the UI

temporal/api/sdk/v1/enhanced_stack_trace.proto
```diff
@@ -0,0 +1,78 @@
...
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "EnhancedStackTraceProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+// Internal structure used to create worker stack traces with references to code.
+message EnhancedStackTrace {
+    string sdk_name = 1;
+    string sdk_version = 2;
```
bergundy, line 37: I thought about this some more and I'd rather keep the format we have since I'd rather not have a different code path to support old TS versions.

temporal/api/sdk/v1/enhanced_stack_trace.proto
```diff
@@ -0,0 +1,78 @@
...
+
+// More specific location details of a file: its path, precise line and column numbers if applicable, and function name if available.
+// In essence, a pointer to a location in a file
+message FileLocation {
+    // Path to source file (absolute or relative).
+    // If the paths are relative, ensure that they are all relative to the same root.
+    string file_path = 1;
+    // (-- api-linter: core::0141::forbidden-types=disabled
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    // Optional; If possible, SDK should send this -- this is required for displaying the code location.
```
bergundy, line 61: Put the linter exceptions after the docstring please. It's better for readability.

```suggestion
    // Optional; If possible, SDK should send this -- this is required for displaying the code location.
    // (-- api-linter: core::0141::forbidden-types=disabled
    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
```

temporal/api/sdk/v1/enhanced_stack_trace.proto
```diff
@@ -0,0 +1,78 @@
...
+    uint32 line = 2;
+    // (-- api-linter: core::0141::forbidden-types=disabled
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    // Optional; if possible, SDK should send this.
+    uint32 column = 3;
+    // Function name this line belongs to, if applicable.
+    // Used for falling back to stack trace view.
+    string function_name = 4;
+    // Flag to communicate whether a location should be hidden by default in the stack view.
+    bool hide_default = 5;
```
bergundy, line 71: The SDK can simply omit this if it wants to hide the stack. Also the name `hide_default` isn't meaningful to me, there's no "default" to hide.

Sushisource: The SDK shouldn't omit it. The user might want to view something that makes sense to hide by default, but we don't want to eliminate the information entirely. Hence making it a display/UI concern rather than filtering it out send side.

It's just like the parts of a stack that are collapsed by default in IDE debugging.

temporal/api/sdk/v1/enhanced_stack_trace.proto
twin-drill: Resolved on local, will be in next commit.

temporal/api/sdk/v1/enhanced_stack_trace.proto
bergundy: I would just consider renaming this field to `hide_by_default` or `internal_code`.

cretz (approved):
LGTM, but should let other conversations/approvers get their say

temporal/api/sdk/v1/enhanced_stack_trace.proto
twin-drill: I like `internal_code`; the linter doesn't like the name `hide_by_default` which is what the field was initially.

twin-drill:
Quick note that I shared in a slack message:

If fields like `line` and `column` in `FileLocation` are optional, we may need some way of telling the UI or other interpreter of this message that those fields are actually omitted (as `0` is technically a valid line number, but is also the default value of the field when a value isn't provided).

temporal/api/sdk/v1/enhanced_stack_trace.proto
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// "Slice" of a file starting at line_offset -- a line offset and code fragment corresponding to the worker's stack.
+message StackTraceFileSlice {
+    // Only used possible to trim the file without breaking syntax highlighting.
+    // (-- api-linter: core::0141::forbidden-types=disabled
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    uint32 line_offset = 1;
+
+    // Slice of a file with `\n` (newline) line terminator.
```
bergundy, line 64: We might not want to commit to this being `\n` and just say it's an OS specific line terminator. (We can update the TS docstring too).

temporal/api/sdk/v1/enhanced_stack_trace.proto
```diff
@@ -0,0 +1,97 @@
...
+
+    // Optional; If possible, SDK should send this -- this is required for displaying the code location.
+    // (-- api-linter: core::0141::forbidden-types=disabled
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    uint32 line = 2;
+
+    // Optional; if possible, SDK should send this.
+    // (-- api-linter: core::0141::forbidden-types=disabled
+    //     aip.dev/not-precedent: These really shouldn't have negative values. --)
+    uint32 column = 3;
```
bergundy, line 83: We might need a way to determine whether these are zero because the value is omitted or if they're actually meant to be zero but you can start out with this for now.

bergundy (approved):

temporal/api/sdk/v1/enhanced_stack_trace.proto
twin-drill: Converted to `int32` as we discussed.

temporal/api/sdk/v1/enhanced_stack_trace.proto
twin-drill: Updated this comment to reflect that.

Sushisource (approved):

PR #419, opened 2024-06-06
Document reset_history_event_id

Quinn-With-Two-Ns:
Document what `reset_history_event_id` actually does.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -357,7 +357,9 @@ message RespondWorkflowTaskCompletedResponse {
     PollWorkflowTaskQueueResponse workflow_task = 1;
     // See `ScheduleActivityTaskCommandAttributes::request_start`
     repeated PollActivityTaskQueueResponse activity_tasks = 2;
-
+    // If non zero, indicates the server has discarded the workflow task that was being responed to.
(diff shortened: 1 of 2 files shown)


cretz (commented):
LGTM assuming server team confirms accuracy

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -357,7 +357,9 @@ message RespondWorkflowTaskCompletedResponse {
     PollWorkflowTaskQueueResponse workflow_task = 1;
     // See `ScheduleActivityTaskCommandAttributes::request_start`
     repeated PollActivityTaskQueueResponse activity_tasks = 2;
-
+    // If non zero, indicates the server has discarded the workflow task that was being responed to.
+    // Will be the event ID of the last workflow task started event in the history before the new workflow task.
+    // Server is only expected to discard a workflow task if it could not have modified the workflow state.
```
Sushisource, line 362:
```suggestion
    // If non zero, indicates the server has discarded the workflow task that was being responded to.
    // Will be the event ID of the last workflow task started event in the history before the new workflow task.
    // Server is only expected to discard a workflow task if it could not have modified the workflow state.
```

Sushisource (approved):

alexshtin (approved):

PR #420, opened 2024-06-11
Switch back to using description for Nexus Endpoints

bergundy:
**What changed?**

Use single `Payload description` field for Nexus Endpoints instead of re-using the UserMetadata message that we use for history events.

**Why?**

- `Description` applies better to Nexus Endpoints than `details`, we don't have a concept of `summary` yet.
- Avoid coupling.

**Breaking changes**

Yes, but not yet released.

**Server PR**

https://github.com/temporalio/temporal/pull/6121


temporal/api/nexus/v1/message.proto
```diff
@@ -162,7 +161,10 @@ message EndpointSpec {
     // Renaming an endpoint breaks all workflow callers that reference this endpoint, causing operations to fail.
     string name = 1;
 
-    temporal.api.sdk.v1.UserMetadata metadata = 2;
+    // Markdown description serialized as a single JSON string.
+    // If the Payload is encrypted, the UI and CLI may decrypt with the configured codec server endpoint.
+    // By default, the server enforces a limit of 20,000 bytes for this entire payload.
+    temporal.api.common.v1.Payload description = 2;
```
cretz, line 167: Not a fan of this Nexus-only description concept as opposed to general purpose user metadata

bergundy: I know but I'd rather stick to the original design and avoid coupling to user metadata that's more general purpose.

cretz: Ideally we wouldn't avoid general purpose thing if present. The whole point of "coupling" user metadata is so that we have a general view of user metadata (i.e. data server doesn't care about). If every project wanted to avoid coupling then it would not exist in its general state. I don't think that the name `details` vs `description` or the desire to avoid using a common user metadata approach is worth it, but can let others weigh in and approve.

cretz (approved):

pdoerner (approved):

PR #424, opened 2024-06-26
Add Nexus to GetSystemInfoResponse

pdoerner:
**What changed?**
Added a boolean to `workflowservice.GetSystemInfoResponse` to indicate whether the server supports Nexus.

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1008,6 +1008,9 @@ message GetSystemInfoResponse {
         // True if the server supports count group by execution status
         // (-- api-linter: core::0140::prepositions=disabled --)
         bool count_group_by_execution_status = 10;
+
+        // True if the server supports Nexus operations.
+        bool nexus = 11;
     }
 }
 


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1008,6 +1008,9 @@ message GetSystemInfoResponse {
         // True if the server supports count group by execution status
         // (-- api-linter: core::0140::prepositions=disabled --)
         bool count_group_by_execution_status = 10;
+
+        // True if the server supports Nexus operations.
```
bergundy, line 1012: nit: the other sentences don't end with a period (even though I think they should).

bergundy, line 1012:
```suggestion
        // True if the server supports Nexus operations.
        // This flag is dependent both on server version and for Nexus to be enabled via server configuration.
```

pdoerner: Some of them do, some of them don't. Didn't want to fix all of them in this PR.

bergundy: Yeah, definitely didn't ask you to change the others just to be consistent but I'm approving regardless. I don't think it's a blocker.

Quinn-With-Two-Ns (approved):

bergundy (approved):

PR #430, opened 2024-07-03
Add temporary additional bindings to previous HTTP API paths

cretz:
**What changed?**

Added temporary additional bindings to 1.34.0 HTTP paths. We were unaware when we changed these in 1.35.0 that they were in use by https://github.com/temporalio/ui-server. We are adding them now to allow ui-server to transition and then we will remove them.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -19,6 +19,2316 @@
     "application/json"
   ],
   "paths": {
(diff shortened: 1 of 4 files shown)


mjameswh (approved):
LGTM. I trust you that all the paths are correct.

bergundy (approved):

PR #432, opened 2024-07-09
Add Request ID to nexus completion events

bergundy:
**Why?**

We need this information to safely reapply completion events during namespace failover.
Without this information, we could potentially apply a completion for the wrong operation.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -8337,9 +8337,9 @@
           "$ref": "#/definitions/apifailurev1Failure",
           "description": "Cancellation details."
         },
-        "operationInitialVersionedTransition": {
-          "$ref": "#/definitions/v1VersionedTransition",
-          "description": "The execution's versioned transition at the time the operation was scheduled."
+        "requestId": {
+          "type": "string",
+          "description": "The request ID allocated at schedule time."
         }
       },
       "description": "Nexus operation completed as canceled. May or may not have been due to a cancellation request by the workflow."
@@ -8402,9 +8402,9 @@
           "$ref": "#/definitions/v1Payload",
           "description": "Serialized result of the Nexus operation. The response of the Nexus handler.\nDelivered either via a completion callback or as a response to a synchronous operation."
         },
-        "operationInitialVersionedTransition": {
-          "$ref": "#/definitions/v1VersionedTransition",
-          "description": "The execution's versioned transition at the time the operation was scheduled."
+        "requestId": {
+          "type": "string",
+          "description": "The request ID allocated at schedule time."
         }
       },
       "description": "Nexus operation completed successfully."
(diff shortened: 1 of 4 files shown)


bergundy:
Considering making this information internal and exposing this differently in the API.

cretz:
Can you give some background on how we solve this kind of failover problem with non-Nexus stuff today? Or rather, why this is uniquely a Nexus problem.

bergundy:
Today we only reapply `WorkflowExecutionSignaled`, `WorkflowExecutionUpdateAccepted`, and `WorkflowExecutionUpdateAdmitted`. Those events don't have a dependency on a scheduled event and can be blindly reapplied.

For nexus completions, the event is often generated from an asynchronous completion API and is essentially the equivalent of a signal. The difference between a signal and a nexus completion is that the completion corresponds to a specific operation. Scheduled event ID is not enough to identify the operation in reapply scenarios because there could be two operations scheduled on different clusters with the same event ID and different namespace failover versions.

cretz:
Is there a more general discussion about whether we want to reapply Nexus completion events? We don't seem to do so with activities or child workflows.

bergundy:
Changed to expose the request ID instead of leaking versioned transition.

bergundy:
> Is there a more general discussion to be had about whether we want to reapply Nexus completion events? We don't seem to do so with activities or child workflows. Should we consistently allow users to apply completion events or not instead of making Nexus completion events special?

We should discuss this in the context of workflow reset not for MRN conflict resolution - this fix is to prevent workflows from getting stuck.
For reset, if the started event is reapplied, the completion event should also be cherry-picked since that comes in asynchronously from an external actor. I don't think we should allow more control ATM, if we want to give more fine grained reset reapply control, it would have to be designed holistically to account for activities and child workflows, etc.

cretz:
> For reset, if the started event is reapplied, the completion event should also be cherry-picked since that comes in asynchronously from an external actor. 

What if the external actor is the external activity or child workflow task queue? I am just wondering why Nexus events are more special during reset than activity or child workflow. People seem to be able to live without activity/child reapply during reset.

cretz (commented):
Ah, this changed to just request ID. No problem from my POV.

cretz (approved):

bergundy:
Right now we don't support resetting workflows that have children because of similar complexity and acknowledge we need to fix that.
For activities, it's a bit different because the activity lifecycle is directly driven by the activity's workflow and the started and completed events are always committed together.

For nexus, we need to either take or discard both the started and completion events as a unit or the workflow may get stuck.

cretz:
IMO Nexus project being as good/bad as child workflows for reset is enough for MVP.

> For nexus, we need to either take or discard both the started and completion events as a unit or the workflow may get stuck.

I also wonder why scheduled event ID is not good enough here. You're going to have to rebuild that anyways when you're rebuilding the reset events. I don't know about reset logic, but I would have assumed you have the entire workflow history at your disposal and can therefore pick out the matching start/completed events based on information already present instead of adding new identifiers. Having said that, if this request ID is used for other things such as on the Nexus handler side, then it can be valuable information regardless.

yycptt (approved):

bergundy:
> IMO Nexus project being as good/bad as child workflows for reset is enough for MVP.

Currently reset with a pending operation may leave a workflow stuck, maybe it's worth blocking reset like we do for workflows with children in the meantime. I have a task to fix this later, more concerned with conflict resolution and MRN where we've seen workflows get stuck in edge cases where completion happens while there's a pending workflow task and namespace failover happens before those buffered events are flushed.

> I also wonder why scheduled event ID is not good enough here.

It's good enough for reset, not good enough for MRN for reasons I explained above.

PR #434, opened 2024-07-11
Expose scheduled event ID on PendingNexusOperationInfo

bergundy:
**Why?**

So the UI can correlate the info from describe with the history event.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -8725,6 +8725,11 @@
         },
         "cancellationInfo": {
           "$ref": "#/definitions/v1NexusOperationCancellationInfo"
+        },
+        "scheduledEventId": {
+          "type": "string",
+          "format": "int64",
+          "description": "The event ID of the NexusOperationScheduled event. Can be used to correlate an operation in the\nDescribeWorkflowExecution response with workflow history."
         }
       },
       "description": "PendingNexusOperationInfo contains the state of a pending Nexus operation."
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -6461,6 +6461,11 @@ components:
           format: date-time
         cancellationInfo:
           $ref: '#/components/schemas/NexusOperationCancellationInfo'
+        scheduledEventId:
+          type: string
+          description: |-
+            The event ID of the NexusOperationScheduled event. Can be used to correlate an operation in the
+             DescribeWorkflowExecution response with workflow history.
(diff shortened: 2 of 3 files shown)


cretz:
What is the use case? Does this use case not also apply to pending activity info?

bergundy:
Pending activity info can be correlated to the history event using `activity_id`. There's nothing to link the nexus operation today.

pdoerner (approved):

cretz:
Activity ID is (usually) generated by us, I wonder if the analogue here is request ID instead event ID?

bergundy:
> Activity ID is (usually) generated by us, I wonder if the equivalent here is the recently-added request ID instead of event ID?

Both can be used, the event ID is easier to work with though.

cretz:
My only concern is that we haven't traditionally used event identifiers outside of history. Maybe this is showing a need for an identifier unrelated to history for correlation reasons. When talking about a Nexus operation on a workflow, will we expect users to use event identifier as opposed to the other things (activities, child workflows, updates, etc) that have non-history-specific identifiers?

bergundy:
> My only concern is that we haven't traditionally used event identifiers outside of history. Maybe this is showing a need for an identifier unrelated to history for correlation reasons. When talking about a Nexus operation on a workflow, will we expect users to use event identifier as opposed to the other things (activities, child workflows, updates, etc) that have non-history-specific identifiers?
> 
> What will be put in OTel tracing spans for this operation end-to-end?

The thing with nexus operations is that callers don't get to set the ID, it's only available in async operations and set by the handler. Request ID is allocated by the server so doesn't have much meaning to the user.

The otel spans will be propagated via `nexus_header` on the command/event, we don't put IDs on otel spans AFAICT: https://github.com/temporalio/sdk-typescript/blob/a6849afe39cb51cafdc933e217c227206d4edacd/packages/interceptors-opentelemetry/src/workflow/index.ts#L80.

cretz:
I see, yeah we don't put outbound activity/child identifiers. I do think we should consider a way for users to identify Nexus operation invocations besides their scheduled event ID in history. But if others are ok with this being the identification identifier, no prob.

bergundy:
> I see, yeah we don't put outbound activity/child identifiers. I do think we should consider a way for users to identify Nexus operation invocations that isn't the scheduled event ID in history. But if others are ok with this being the identifier, no prob.

I don't think it hurts to add this field right now though. If we do come up with such an identifier, we could add that later too.
In the meantime users can get more information in the history event, such as request ID and inputs.

Sushisource (approved):

PR #437, opened 2024-07-19
GHA to create release

rodrigozhou:
**What changed?**
GHA to create, publish and delete releases.
- `create-release`: manually triggered, it validates the input, checks compatibility with sdk-go, and creates draft releases in `api` and `api-go`.
- `publish-release`: triggered when a release is published, it publishes the corresponding `api-go` release.
- `delete-release`: triggered when a release is deleted, it deletes the corresponding `api-go` release (only works for deleting draft releases).

**Why?**
Automation.

**Breaking changes**

**Server PR**

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -111,11 +111,6 @@ jobs:
           ref: ${{ needs.prepare-inputs.outputs.api_commit_sha }}
           token: ${{ steps.generate_token.outputs.token }}
 
-      - name: Set up Github credentials
-        run: |
-          git config --local user.name 'Temporal Data'
-          git config --local user.email 'commander-data@temporal.io'
-
       - name: Create release
         env:
           GH_TOKEN: ${{ steps.generate_token.outputs.token }}
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,4 +1,4 @@
-name: "Create a tag"
+name: "Create release"
 
 on:
   workflow_dispatch:
@@ -10,10 +10,6 @@ on:
       tag:
         description: "Tag for new version (v1.23.4)"
         required: true
-      create_release:
-        description: "Create release and set as latest"
-        type: boolean
-        default: true
       base_tag:
         description: "Base tag to generate commit list for release notes"
         required: false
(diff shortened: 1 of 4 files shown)


cretz:
Can you give some background on the savings here on starting a GH workflow vs publishing release in GH?

cretz (approved):
As mentioned at https://github.com/temporalio/api-go/pull/170#pullrequestreview-2488868880, I am a bit concerned about the maintenance of these two workflows for such an incredibly simple thing, but if y'all are resolving to maintain them and are now in full control of the release process of this and the API repo, ok.

cretz (approved):

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,152 @@
...
+        description: "Tag for new version (v1.23.4)"
+        required: true
+      create_release:
+        description: "Create release and set as latest"
+        type: boolean
+        default: true
+      base_tag:
+        description: "Base tag to generate commit list for release notes"
+        required: false
+      skip_sdk_check:
```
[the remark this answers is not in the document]
rodrigozhou: @cretz mentioned a use case here: https://github.com/temporalio/sdk-go/pull/1749#pullrequestreview-2490077991

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,140 @@
...
+      sdk_ref: latest
+      api_ref: ${{ needs.prepare-inputs.outputs.api_go_commit_sha }}
+
+  create-release:
+    name: "Create release"
+    needs: [prepare-inputs, check-compatibility-sdk-go]
+    if: |
+      !cancelled() &&
+      needs.prepare-inputs.result == 'success' &&
+      contains(fromJSON('["success", "skipped"]'), needs.check-compatibility-sdk-go.result)
```
bergundy, line 96: I would have imagined this would be the default, is it not so?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,140 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.prepare-inputs.outputs.api_commit_sha }}
+          token: ${{ steps.generate_token.outputs.token }}
+
+      - name: Set up Github credentials
+        run: |
+          git config --local user.name 'Temporal Data'
+          git config --local user.email 'commander-data@temporal.io'
```
bergundy, line 117: Do you even need this? You're not creating a commit, right?

.github/workflows/create-release.yml
rodrigozhou: Github actions doesn't work well when the dependency has a condition for running.
In this case, the `check-compatibility-sdk-go` job runs only if the flag is set. So, even if I add it to `needs`, the `create-release` job never runs. So, I need to manually check the conditions.

PR #438, opened 2024-07-24
Add internal callback variant

bergundy:
**What changed?**

Added an internal callback variant for callbacks that are delivered internally within the system.
This variant is not settable in the API and will be rejected by the service with an INVALID_ARGUMENT error.
The only reason that this is exposed is because callbacks are replicated across clusters via the
WorkflowExecutionStarted event, which is defined in the public API.

**Why?**

The first use case for this will be for the scheduler as a state machine rewrite. Later on we'll use this to allow arbitrary workflow to listen for "external" workflow completions.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -4653,6 +4653,17 @@
       "type": "object",
       "description": "Trigger for when the workflow is closed."
     },
+    "CallbackInternal": {
+      "type": "object",
+      "properties": {
+        "data": {
+          "type": "string",
+          "format": "byte",
+          "description": "Opaque internal data."
+        }
+      },
+      "description": "Callbacks to be delivered internally within the system.\nThis variant is not settable in the API and will be rejected by the service with an INVALID_ARGUMENT error.\nThe only reason that this is exposed is because callbacks are replicated across clusters via the\nWorkflowExecutionStarted event, which is defined in the public API."
+    },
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns (approved):

bergundy:
> I do wish the internal things didn't require _only_ public protos but instead wrapped them in internal protos that can have internal fields

We've been discussing this, I agree, long term we should have a better solution.

yycptt (approved):

openapi/openapiv2.json
```diff
@@ -4653,6 +4653,17 @@
...
+    "CallbackInternal": {
+      "type": "object",
+      "properties": {
+        "data": {
+          "type": "string",
+          "format": "byte",
+          "description": "Opaque internal data."
+        }
+      },
+      "description": "Callbacks to be delivered internally within the system.\nThis variant is not settable in the API and will be rejected by the service with an INVALID_ARGUMENT error.\nThe only reason that this is exposed is because callbacks are replicated across clusters via the\nWorkflowExecutionStarted event, which is defined in the public API."
```
yycptt, line 4665: once we do state-base migration we should be able to deprecate this field?

bergundy: Maybe, it depends on how we implement workflow import and rebuild. Currently we can only rebuild from history events as you obviously know.

PR #443, opened 2024-08-13
Add ShutdownStickyTaskQueue internal API

lina-temporal:
**What changed?**
- Adds the ShutdownStickyTaskQueue API. 

**Why?**
This API will allow workers to signal shutdown/restart, which Matching can use to unload the shutdown worker's sticky task queue. This will effectively remove the latency incurred from tasks being pushed to a sticky worker after that worker has already been shutdown. 

**Breaking changes**
- None. We aren't changing the existing `ResetStickyTaskQueue` API. This is a fully new API that will hit Matching, instead of History.

**Server PR**
- [Stub PR](https://github.com/temporalio/temporal/pull/6443)

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -869,12 +869,12 @@ message ResetStickyTaskQueueRequest {
 message ResetStickyTaskQueueResponse {
 }
 
-message ShutdownStickyTaskQueueRequest {
+message ShutdownWorkerRequest {
     string namespace = 1;
-    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
+    string sticky_task_queue = 2;
 }
 
-message ShutdownStickyTaskQueueResponse {
+message ShutdownWorkerResponse {
 }
 
 message QueryWorkflowRequest {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -869,6 +869,14 @@ message ResetStickyTaskQueueRequest {
 message ResetStickyTaskQueueResponse {
 }
 
+message ShutdownStickyTaskQueueRequest {
+    string namespace = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
+}
+
+message ShutdownStickyTaskQueueResponse {
+}
+
 message QueryWorkflowRequest {
     string namespace = 1;
     temporal.api.common.v1.WorkflowExecution execution = 2;
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -872,6 +872,8 @@ message ResetStickyTaskQueueResponse {
 message ShutdownWorkerRequest {
     string namespace = 1;
     string sticky_task_queue = 2;
+    string identity = 3;
+    string reason = 4;
 }
 
 message ShutdownWorkerResponse {
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -536,7 +536,10 @@ service WorkflowService {
     // in the normal task queue, eligible for any worker to pick up.
     //
     // ShutdownWorker should be called by workers while shutting down,
-    // after they've shut down their pollers.
+    // after they've shut down their pollers. If another sticky poll
+    // request is issued, the sticky task queue will be revived.
+    //
+    // As of Temporal Server v1.25.0, ShutdownWorker hasn't yet been implemented.
     //
     // (-- api-linter: core::0127::http-annotation=disabled
     //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -521,9 +521,26 @@ service WorkflowService {
     // 1. StickyTaskQueue
     // 2. StickyScheduleToStartTimeout
     //
+    // When possible, ShutdownStickyTaskQueue should be preferred over
+    // ResetStickyTaskQueue (particularly when a worker is shutting down or
+    // cycling).
+    //
     // (-- api-linter: core::0127::http-annotation=disabled
     //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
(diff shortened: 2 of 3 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -521,11 +521,28 @@ service WorkflowService {
...
+    // queue is no longer being polled by its worker. Following the completion of
+    // ShutdownStickyTaskQueue, newly-added workflow tasks will instead be placed
+    // in the normal task queue, eligible for any worker to pick up.
+    //
+    // ShutdownStickyTaskQueue should be called by workers while shutting down,
+    // after they've shut down their pollers.
+    //
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
+    rpc ShutdownStickyTaskQueue (ShutdownStickyTaskQueueRequest) returns (ShutdownStickyTaskQueueResponse) {
```
bergundy, line 543: I would call this `ShutdownWorker` instead and have the sticky task queue be a field on the request in case we want to add more worker shutdown related functionality in the future.

lina-temporal: Agreed, yeah! Renamed.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -521,11 +521,28 @@ service WorkflowService {
     // 1. StickyTaskQueue
     // 2. StickyScheduleToStartTimeout
     //
+    // When possible, ShutdownWorker should be preferred over
```
bergundy, line 524: Maybe mention that at the time of writing (server version 1.25.0) ShutdownWorker hasn't yet been implemented.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -521,11 +521,28 @@ service WorkflowService {
...
     rpc ResetStickyTaskQueue (ResetStickyTaskQueueRequest) returns (ResetStickyTaskQueueResponse) {
     }
 
+    // ShutdownWorker is used to indicate that the given sticky task
+    // queue is no longer being polled by its worker. Following the completion of
+    // ShutdownWorker, newly-added workflow tasks will instead be placed
+    // in the normal task queue, eligible for any worker to pick up.
+    //
+    // ShutdownWorker should be called by workers while shutting down,
+    // after they've shut down their pollers.
```
bergundy, line 539: Maybe document that this detail is important as a sticky poll request may revive the sticky task queue.

bergundy (approved):
Overall LGTM.

temporal/api/workflowservice/v1/service.proto
cretz: Disagree here in that a server is worker agnostic

temporal/api/workflowservice/v1/service.proto
```diff
@@ -521,11 +521,28 @@ service WorkflowService {
     // 1. StickyTaskQueue
     // 2. StickyScheduleToStartTimeout
     //
+    // When possible, ShutdownWorker should be preferred over
+    // ResetStickyTaskQueue (particularly when a worker is shutting down or
+    // cycling).
+    //
     // (-- api-linter: core::0127::http-annotation=disabled
     //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
     rpc ResetStickyTaskQueue (ResetStickyTaskQueueRequest) returns (ResetStickyTaskQueueResponse) {
```
cretz, line 530: Description doesn't contain justification for why this couldn't be a field on this existing RPC, bringing up internally...

temporal/api/workflowservice/v1/service.proto
Sushisource: I think it makes sense to be `ShutdownWorker`. It allows us to expand the scope of the RPC without the name ceasing to make sense. I can't really imagine any _other_ reason you'd want to call this beyond workers shutting down, and if we add more functionality then either the name stops making sense or you have to call multiple RPC.

temporal/api/workflowservice/v1/service.proto
Sushisource: +1 to that

Sushisource (approved):

temporal/api/workflowservice/v1/service.proto
lina-temporal: Agreed, will update

temporal/api/workflowservice/v1/service.proto
lina-temporal: Yup, will do

temporal/api/workflowservice/v1/service.proto
cretz: Discussed internally, they would rather a new API for shutting down workers

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -869,6 +869,14 @@ message ResetStickyTaskQueueRequest {
 message ResetStickyTaskQueueResponse {
 }
 
+message ShutdownWorkerRequest {
```
dnr, line 872: Are we going to add `identity` for removing from recent poller lists also? (Or at least marking with a hint that it's probably not polling anymore)

dnr: And just a random idea: If the worker will be displayed in recent poller lists (in ui/cli) with a message saying "shut down", then maybe there could be another string field `reason` that would get propagated to the ui for extra context?

lina-temporal: > Are we going to add identity for removing from recent poller lists also? (Or at least marking with a hint that it's probably not polling anymore)

I'm definitely onboard to add an `identity` field, though I hadn't added it as part of this PR since we won't be making use of it yet. Do you have a preference for requiring it as part of this request early?

> And just a random idea: If the worker will be displayed in recent poller lists (in ui/cli) with a message saying "shut down", then maybe there could be another string field reason that would get propagated to the ui for extra context?

I like the idea! From my understanding of how the recent poller lists work, we should be able to plumb that through by updating the `PollerInfo` protobuf and the `pollerHistory` cache, without needing to write to persistence; does that sound right to you?

dnr: No preference. It seems easy to do all at once but can also wait.

Yup, poller info is just in memory.

cretz (approved):

PR #445, opened 2024-08-19
Clarify workflow metadata interaction sort order

cretz:
**What changed?**

Clarified sort order for workflow interactions on workflow metadata


Sushisource (approved):
Maybe worth saying alphabetically just to be extra clear

bergundy (approved):

PR #447, opened 2024-08-20
Add task polling data to task responses

Sushisource:
**What changed?**
Add some additional information to polling responses about the task queue 

## Outcome of discussion w/ @ShahabT and @dnr 
Initially I would've liked `TaskQueueStats` to be returned on every poll response. This appears to be too much from both a performance and complexity perspective initially because regularly updating and distributing aggregated data across partitions would require more frequent internal RPCs, and would require them to go in a direction they don't currently (from sub-partition -> root rather than just the other way around).

The alternative is to have SDK acquire that data via `DescribeTaskQueue`. Calling that very frequently would also be problematic, since there is an RPS limit per-namespace.

The landed on middle ground, which is low complexity but ought to be reasonably efficient, is to include `TaskQueueStats` in the response whenever a poll hits the root partition, where it is readily available. If the SDK has not seen such a response in some reasonable TTL (~seconds), then it will call `DescribeTaskQueue` in order to have an updated view. In practice, it should be fairly unlikely for a worker with a reasonable number of pollers to somehow miss root, and low-traffic queues are also more likely to forward to the root to see if there is a task. Since the desired outcome SDK side is to only have a low number of pollers when there's low traffic, that should work out.

**Why?**
This will allow poller auto-tuning to have an idea of if it should be increasing the number of pollers because most matches are not sync (possibly in conjunction w/ a backlog hint)

If we are making sync matches, then we can reduce the number of pollers once the wait duration starts to exceed some threshold.

It can also inform the SDK what the minimum latency is when polling, since if we didn't make a sync match, then we were pulling something out of the backlog and delivered a task roughly as quickly as is possible.

The task queue stats info is useful to help to decide how quickly pollers should be ramped up, if at all.

**Breaking changes**
nope

**Server PR**
Will make one assuming we're OK with this add.

### DIFF
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -258,3 +258,18 @@ message TimestampedCompatibleBuildIdRedirectRule {
     CompatibleBuildIdRedirectRule rule = 1;
     google.protobuf.Timestamp create_time = 2;
 }
+
+// Data attached to successful poll responses (for any task type) which the SDK can use to optimize its polling
+// strategy.
+message TaskResponsePollingData {
+    // Set to true if the returned task was a result of a sync match.
+    bool was_sync_match = 1;
+    // Set to the amount of time a poll call waited for a task to be available. This value is allowed to be unset
+    // if the match was not sync. It is useful to help the SDK understand if it may reduce the number of pollers, since
+    // any substantial wait time here means the backlog is drained, and poll calls are idling waiting for tasks.
+    google.protobuf.Duration time_poller_waited = 2;
+    // See `TaskQueueStats`. This field is only set if the poll hit the root partition, where the data is readily
+    // available. SDKs should cache this data, and explicitly refresh it with a call to `DescribeTaskQueue` if some
+    // TTL has expired.
+    TaskQueueStats stats = 3;
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -315,6 +315,10 @@ message PollWorkflowTaskQueueResponse {
     map<string, temporal.api.query.v1.WorkflowQuery> queries = 14;
     // Protocol messages piggybacking on a WFT as a transport
     repeated temporal.api.protocol.v1.Message messages = 15;
+    // This field is set if the task was delivered as a result of a sync match, and it is set to the
+    // amount of time we waited to make such a match. If the task was delievered from the blacklog,
+    // then the field is unset.
+    google.protobuf.Duration sync_match_wait_duration = 16;
 }
 
 message RespondWorkflowTaskCompletedRequest {
@@ -447,6 +451,10 @@ message PollActivityTaskQueueResponse {
     // (or not) during activity scheduling. The service can override the provided one if some
     // values are not specified or exceed configured system limits.
     temporal.api.common.v1.RetryPolicy retry_policy = 17;
+    // This field is set if the task was delivered as a result of a sync match, and it is set to the
+    // amount of time we waited to make such a match. If the task was delievered from the blacklog,
+    // then the field is unset.
+    google.protobuf.Duration sync_match_wait_duration = 18;
 }
 
 message RecordActivityTaskHeartbeatRequest {
@@ -1588,6 +1596,10 @@ message PollNexusTaskQueueResponse {
     bytes task_token = 1;
     // Embedded request as translated from the incoming frontend request.
     temporal.api.nexus.v1.Request request = 2;
+    // This field is set if the task was delivered as a result of a sync match, and it is set to the
+    // amount of time we waited to make such a match. If the task was delievered from the blacklog,
+    // then the field is unset.
+    google.protobuf.Duration sync_match_wait_duration = 3;
 }
 
 message RespondNexusTaskCompletedRequest {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -315,10 +315,9 @@ message PollWorkflowTaskQueueResponse {
     map<string, temporal.api.query.v1.WorkflowQuery> queries = 14;
     // Protocol messages piggybacking on a WFT as a transport
     repeated temporal.api.protocol.v1.Message messages = 15;
-    // This field is set if the task was delivered as a result of a sync match, and it is set to the
-    // amount of time we waited to make such a match. If the task was delievered from the blacklog,
-    // then the field is unset.
-    google.protobuf.Duration sync_match_wait_duration = 16;
+    // See `TaskResponsePollingData`
+    // NOTE: Not yet implemented. Update this docstring when released with version.
+    temporal.api.taskqueue.v1.TaskResponsePollingData task_response_data = 16;
 }
 
 message RespondWorkflowTaskCompletedRequest {
@@ -451,10 +450,9 @@ message PollActivityTaskQueueResponse {
     // (or not) during activity scheduling. The service can override the provided one if some
     // values are not specified or exceed configured system limits.
     temporal.api.common.v1.RetryPolicy retry_policy = 17;
-    // This field is set if the task was delivered as a result of a sync match, and it is set to the
-    // amount of time we waited to make such a match. If the task was delievered from the blacklog,
-    // then the field is unset.
-    google.protobuf.Duration sync_match_wait_duration = 18;
+    // See `TaskResponsePollingData`
+    // NOTE: Not yet implemented. Update this docstring when released with version.
+    temporal.api.taskqueue.v1.TaskResponsePollingData task_response_data = 18;
 }
 
 message RecordActivityTaskHeartbeatRequest {
@@ -1596,10 +1594,9 @@ message PollNexusTaskQueueResponse {
     bytes task_token = 1;
     // Embedded request as translated from the incoming frontend request.
     temporal.api.nexus.v1.Request request = 2;
-    // This field is set if the task was delivered as a result of a sync match, and it is set to the
-    // amount of time we waited to make such a match. If the task was delievered from the blacklog,
-    // then the field is unset.
-    google.protobuf.Duration sync_match_wait_duration = 3;
+    // See `TaskResponsePollingData`
+    // NOTE: Not yet implemented. Update this docstring when released with version.
+    temporal.api.taskqueue.v1.TaskResponsePollingData task_response_data = 3;
 }
 
 message RespondNexusTaskCompletedRequest {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -8942,9 +8942,9 @@
           "$ref": "#/definitions/v1RetryPolicy",
           "description": "This is the retry policy the service uses which may be different from the one provided\n(or not) during activity scheduling. The service can override the provided one if some\nvalues are not specified or exceed configured system limits."
         },
-        "syncMatchWaitDuration": {
-          "type": "string",
-          "description": "This field is set if the task was delivered as a result of a sync match, and it is set to the\namount of time we waited to make such a match. If the task was delievered from the blacklog,\nthen the field is unset."
+        "taskResponseData": {
+          "$ref": "#/definitions/v1TaskResponsePollingData",
+          "description": "See `TaskResponsePollingData`\nNOTE: Not yet implemented. Update this docstring when released with version."
         }
       }
     },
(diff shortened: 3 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -313,6 +313,10 @@ message PollWorkflowTaskQueueResponse {
     map<string, temporal.api.query.v1.WorkflowQuery> queries = 14;
     // Protocol messages piggybacking on a WFT as a transport
     repeated temporal.api.protocol.v1.Message messages = 15;
+    // This field is set if the task was delivered as a result of a sync match, and it is set to the
+    // amount of time we waited to make such a match. If the task was delievered from the blacklog,
+    // then the field is unset.
+    google.protobuf.Duration sync_match_wait_duration = 16;
```
cretz, line 319: I wonder if we should break this off into `temporal.api.taskqueue.v1.TaskQueuePollInfo` or something. Any expectation of more information? I was under the impression that backlog hint was only useful in certain situations (something about being partition specific or something) so maybe the newer backlog info can be in this type too assuming it's very quick to access.

Sushisource: I'm happy with that idea yes. We may also add `TaskQueueStats` to this

dnr: I like the idea of encapsulating these in a message also, with clear semantics that the whole bunch are optional hints that SDKs can use to control polling and aren't good for much else.

dnr: Also as mentioned separately, I think separating wait time from a sync match flag makes more sense, even if a non-sync match would generally return immediately since it means there was a task in a backlog.

Sushisource: Upgraded w/ dedicated message & task queue stats per discussion

temporal/api/taskqueue/v1/message.proto
```diff
@@ -257,4 +257,19 @@ message TimestampedBuildIdAssignmentRule {
...
+    // Set to true if the returned task was a result of a sync match.
+    bool was_sync_match = 1;
+    // Set to the amount of time a poll call waited for a task to be available. This value is allowed to be unset
+    // if the match was not sync. It is useful to help the SDK understand if it may reduce the number of pollers, since
+    // any substantial wait time here means the backlog is drained, and poll calls are idling waiting for tasks.
+    google.protobuf.Duration time_poller_waited = 2;
+    // See `TaskQueueStats`. This field is only set if the poll hit the root partition, where the data is readily
+    // available. SDKs should cache this data, and explicitly refresh it with a call to `DescribeTaskQueue` if some
+    // TTL has expired.
+    TaskQueueStats stats = 3;
```
cretz, line 274: Can I get confirmation from server team that this information is available at no real extra cost at poll completion time?

cretz, line 274: Hrmm, I wonder a bit about this. I wasn't expecting SDKs to call describe task queue ever, only users. Is there information in here to drive decisions that we need?

cretz: Hrmm, can the poller scaling decisions be done without changing SDKs to add new describe task queue calls? Or is it a strong requirement at this point?

Sushisource: What's the concern here? Nothing is, strictly speaking, required - we could not do poller scaling at all, but the information is helpful in making those decisions. I already had some discussion with server folks about the load making this call would add and that's all fine.

cretz: The concern I think is adding the confusing results to the poll (we already got bit by this with backlog hint where it was confusing for everyone what it meant and when the data was accurate). Also there is the concern of adding logic burden on the workers to now start making additional, separate out of band calls beyond polling to support polling. I was hoping the poller scaling could work with the calls the worker already makes and if it needs more info, put the burden on the centralized part (i.e. server) to provide it as part of the calls already being made or acknowledge that we may need to work with limited information.

Sushisource: Yes, as mentioned I would've preferred that too but @ShahabT and @dnr explained that it was likely to cause more overall load than periodically calling describe, which is not all that difficult. Plus, this API does not _require_ things to work that way forever. This field could be always populated, later, if we are able to make that improvement.

cretz: Makes sense. So the remaining outstanding question is if this stats information is _that_ necessary for polling scaling decisions. Is it acceptable to have a more naive poller scaler that operates on the other two fields and not these stats? I wonder if we can implement that and come back and add this whole stats fetching mechanism to workers if it's the naive way is not good enough.

Sushisource: Well, we will see. That was my inclination as well but in talks with Max he was quite convinced it would be important to make use of the extra info. The good news is if I don't need to make the call nothing in this API is going to need to change, since sending this info on responses is very cheap. It may work out that the root partition is hit frequently enough that the periodic call is never necessary.

Quinn-With-Two-Ns: How is the worker expected to know the number of partitions on start up? To make sure we hit the root partition often we need to have at least number of partition number of pollers right? Do we need to make another call on start up

Sushisource: It's not - see the bit from the description for why I think that's unlikely to matter though:
> In practice, it should be fairly unlikely for a worker with a reasonable number of pollers to somehow miss root, and low-traffic queues are also more likely to forward to the root to see if there is a task. Since the desired outcome SDK side is to only have a low number of pollers when there's low traffic, that should work out.

And, if we connect and immediately see a backlog, well, then we're going to quickly add more pollers. 

There is some hesitation around exposing the number of partitions in the API from the server team, understandably, since it's an implementation detail that ideally we wouldn't need to know about externally. Even if we did know it, I think it's really desirable to be able to scale below that number too, since many many users are spending most of their time idle and minimizing the number of idling poll RPCs is a good thing.

It could make sense for an initial value on startup, though, and I actually almost had added that to the PR as `suggested_minimum_pollers` which could maybe instead be `suggested_initial_pollers`... but I think we can always add it later if it turns out that we aren't getting what we need on startup frequently enough.

Quinn-With-Two-Ns: Yeah I read the description, it wasn't clear how a worker knew what a "a reasonable number of pollers" is on startup because my understanding was the number is related to the number of partitions. I can understand the desire to not expose this to workers for cloud.

Quinn-With-Two-Ns: Yeah I read the description, it wasn't clear how a worker knew what a "a reasonable number of pollers" is on startup because my understanding was the number is related to the number of partitions. I can understand the desire to not expose this to workers for cloud.

Quinn-With-Two-Ns (approved):

temporal/api/taskqueue/v1/message.proto
cretz: > That was my inclination as well but in talks with Max he was quite convinced it would be important to make use of the extra info. 

Hrmm, I wonder if we should detail the poller scaling algorithm with or without this info present to understand its use (or maybe we already have).

> The good news is if I don't need to make the call nothing in this API is going to need to change, since sending this info on responses is very cheap

And not putting this info or even having the field in the API is cheaper/easier 😁 That's my only concern, is we're adding a field with a bit of confusing semantics and with new worker expectations, where the other option is not to add it.

Overall, if we know the poller scaling needs this new IMO-confusing field, ok. But if we don't know yet, would like to err on the side of not having the field.

cretz (approved):
Approving, but still a bit concerned w/ the new stats field and its necessity. Also concerned with whether workers need this new out-of-band RPC to drive poller scaling.

I think we should try simple poller scaling before adding these advanced machinations around task queue stats.

temporal/api/taskqueue/v1/message.proto
ShahabT: I agree with @cretz in that if we are not sure we need the stats we should not add it. My reason being even sending this info only for root partition polls is not free, server would still need to periodically call all the partitions and aggregate the stats and cache them in the root partitions.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -257,4 +257,19 @@ message TimestampedBuildIdAssignmentRule {
...
+
+// Data attached to successful poll responses (for any task type) which the SDK can use to optimize its polling
+// strategy.
+message TaskResponsePollingData {
+    // Set to true if the returned task was a result of a sync match.
+    bool was_sync_match = 1;
+    // Set to the amount of time a poll call waited for a task to be available. This value is allowed to be unset
+    // if the match was not sync. It is useful to help the SDK understand if it may reduce the number of pollers, since
+    // any substantial wait time here means the backlog is drained, and poll calls are idling waiting for tasks.
+    google.protobuf.Duration time_poller_waited = 2;
```
dnr, line 270: I'd call it `poll_wait_time` or even just `wait_time`

temporal/api/taskqueue/v1/message.proto
```diff
@@ -257,4 +257,19 @@ message TimestampedBuildIdAssignmentRule {
 message TimestampedCompatibleBuildIdRedirectRule {
     CompatibleBuildIdRedirectRule rule = 1;
     google.protobuf.Timestamp create_time = 2;
+}
+
+// Data attached to successful poll responses (for any task type) which the SDK can use to optimize its polling
+// strategy.
```
dnr, line 263: Maybe to head off potential confusion, add a comment like "Don't assume the accuracy of this data for any purpose other than poller scaling."

temporal/api/taskqueue/v1/message.proto
dnr: I don't think we need to talk about calling DescribeTaskQueue, or call it. If we're polling fairly frequently, then we'll hit the root and get the data. If we're not, then there's no tasks and no reason to scale pollers. Let's just say the SDK can make use of this field when it's provided, and do without it if it's missing.

dnr (approved):
As usual, please wait to merge until we have a server impl, since we usually discover things to tweak in the process of implementing

temporal/api/taskqueue/v1/message.proto
cretz: That workers won't be expected to call `DescribeTaskQueue` definitely alleviates one of my concerns. But it alludes to another - we don't know what we want workers to do yet it sounds like. Maybe we can even use this field's proto doc to define worker poller behavior. Or maybe, since it sounds like this is an optional field and poller scaling needs to work well without it, we can do poller scaling and come back and add this field and associated optimizations.

Sushisource: I've been writing some pseudocode for the algorithm that's nearly complete. I'm quite sure I want this data, and delivering it in the field is cheap. It's likely to be present often enough to be useful. I don't really see the cost of having it here as problematic at all.

More specifically, it's very useful in knowing when to scale down pollers. Which fits perfectly with this bit that is mentioned earlier but David writes clearly:

>If we're polling fairly frequently, then we'll hit the root and get the data. If we're not, then there's no tasks and no reason to scale pollers.

So we're going to get that data when we most need it.

cretz: :+1: So the docs make clear that it may not appear, but per what @dnr put a few messages up in this thread, can we remove the part about expecting SDKs to call describe task queue to get this info?

@ShahabT - can you expand on what you mean by aggregate-and-cache? How out of date might this value be compared to a regular describe call?

ShahabT: For the stats to reflect the whole task queue, and not only the root partition itself, we need to fan-out to all partitions and get their stats (via internal `DescribeTaskQueuePartition` API), aggregate all of them, and return the aggregated value. This is what Enhanced DescribeTaskQueue does.

Now, if we're to send such stats in the root partition's poll response, this fan-out and calling of `DescribeTaskQueuePartition` API should happen constantly (as opposed to now that it happens only when DescribeTaskQueue is called).

To manage the load, server would have to add caching in the root partition to limit the fan-out frequency (say to once per second).

It's not something that we cannot manage, but it is not free either.

Sushisource:
Leaving as an FYI here: I started the server work but it's likely I won't get back to it for a bit b/c of other Replay-related priorities, so this might have to sit for a bit.

Sushisource:
Superseded by https://github.com/temporalio/api/pull/553

PR #448, opened 2024-08-20
Add Links to HistoryEvent

rodrigozhou:
**What changed?**
Created `Link` struct, and added to `HistoryEvent` and Nexus StartOperation request/response

**Why?**
Support bi-directional links for Nexus calls.

**Breaking changes**

**Server PR**


temporal/api/history/v1/message.proto
```diff
@@ -842,6 +842,9 @@ message NexusOperationStartedEventAttributes {
 
     // The request ID allocated at schedule time.
     string request_id = 4;
+
+	// Links contains information about the operations started by the handler.
+    repeated temporal.api.common.v1.Link links = 5;
```
bergundy, line 847: You don't need this since HistoryEvent already contains links.

temporal/api/common/v1/message.proto
```diff
@@ -206,3 +206,28 @@ message Callback {
...
+        // Additional information about the workflow event.
+        // Eg: the caller workflow can send the history event ID that made the Nexus call.
+        oneof event {
+            int64 event_id = 100;
+            string signal_request_id = 101;
+            string update_id = 102;
+        }
+    }
+
+    reserved 1; // For a generic link to be added later
```
bergundy, line 229: Not necessary, I think we can start with just the WorkflowEvent variant.

temporal/api/common/v1/message.proto
```diff
@@ -206,3 +206,28 @@ message Callback {
...
+        string namespace = 1;
+        string workflow_id = 2;
+        string run_id = 3;
+
+        // Additional information about the workflow event.
+        // Eg: the caller workflow can send the history event ID that made the Nexus call.
+        oneof event {
+            int64 event_id = 100;
+            string signal_request_id = 101;
+            string update_id = 102;
```
bergundy, line 225: Let's add these later as we implement them.

temporal/api/common/v1/message.proto
```diff
@@ -206,3 +206,28 @@ message Callback {
...
+message Link {
+    message WorkflowEvent {
+        string namespace = 1;
+        string workflow_id = 2;
+        string run_id = 3;
+
+        // Additional information about the workflow event.
+        // Eg: the caller workflow can send the history event ID that made the Nexus call.
+        oneof event {
+            int64 event_id = 100;
```
bergundy, line 223: Let's call this `started_event_id` to express that this is a link to a workflow start event. Alternatively, we can have a variant that has `event_type + event_id`.

temporal/api/nexus/v1/message.proto
```diff
@@ -66,6 +66,8 @@ message StartOperationRequest {
     temporal.api.common.v1.Payload payload = 5;
     // Header that is expected to be attached to the callback request when the operation completes.
     map<string, string> callback_header = 6;
+    // Links to be associated to the operations.
+    repeated temporal.api.common.v1.Link links = 7;
```
bergundy, line 70: Hmm.. you can potentially use headers and let the nexus SDK build the links instead of having Temporal specific logic in your implementation.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -197,6 +197,8 @@ message StartWorkflowExecutionRequest {
     // for use by user interfaces to display the fixed as-of-start summary and details of the
     // workflow.
     temporal.api.sdk.v1.UserMetadata user_metadata = 23;
+    // Links to be associated with the workflow.
+    repeated temporal.api.common.v1.Link links = 24;
 }
 
 message StartWorkflowExecutionResponse {
```
bergundy, line 204: I would also return the event ID in the response to avoid making assumptions when generating the link in the response.

temporal/api/common/v1/message.proto
bergundy: Or you could put the event type outside of this `oneof`.

temporal/api/nexus/v1/message.proto
bergundy: Looking at how this is implemented in the Nexus SDK that's not possible because nexus headers can't have multiple values.

bergundy: In any case you should pass the links through as type and data and do the decoding in the SDK.

temporal/api/common/v1/message.proto
rodrigozhou: I thought this would be the Nexus scheduled event ID at this moment, and at this moment, we'd not return the event id of workflow started event. In any case, do we need the event type in the link?

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: The event ID is inside the link object.

temporal/api/nexus/v1/message.proto
rodrigozhou: I thought about this, but this would require duplicating the `nexus.Link` struct in protobuf because in Nexus it's not defined in protobuf.

temporal/api/common/v1/message.proto
cretz: @bergundy - Why does the link need to know the type of the event? Does a link to a signal request need to know the signal name or a link to an update ID need to know the workflow ID, etc? How do we decide what things that are also present on the thing linked we should bring here?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -197,6 +197,8 @@ message StartWorkflowExecutionRequest {
     // for use by user interfaces to display the fixed as-of-start summary and details of the
     // workflow.
     temporal.api.sdk.v1.UserMetadata user_metadata = 23;
+    // Links to be associated with the workflow.
+    repeated temporal.api.common.v1.Link links = 24;
```
cretz, line 201: Can you confirm this will end up on the started event? What about schedule workflow? What about child workflow? What about signal with start? We should try to make all ways of workflow starting have the same options.

We spent plenty of effort making sure user metadata was propagated in these same kinds of situations. Is everyone sure they don't want to use the user metadata propagation that already exists and just put this inside user metadata?

rodrigozhou: Yes, it's added here in the server PR: https://github.com/temporalio/temporal/pull/6420/files#diff-c0867c0ff803c196fa7bc72a15d01812b84ae90eb8b40e71de4ee11d807cfad3R165-R167

At this moment, we're only considering the case of starting new workflow.

Hm, do we want the links to propagate to child workflow, scheduled workflows, etc? I don't think it really make sense.

cretz: IMO we should be consistent in the options like these for all ways of starting a workflow. There are 4 main ones:

* Start workflow
* Signal with start workflow
* Schedule workflow
* Start child workflow

> I don't think it really make sense.

This is a general purpose link option, we shouldn't assume how it'll be used. Same type of thing with user metadata. In fact, arguably this is user metadata and should just be on that message, but that was rejected.

temporal/api/nexus/v1/message.proto
```diff
@@ -52,6 +52,12 @@ message UnsuccessfulOperationError {
     Failure failure = 2;
 }
 
+message Link {
+    // See https://github.com/nexus-rpc/api/blob/main/SPEC.md#links.
```
cretz, line 56: This URL does not work. Where can I read about this arbitrary byte array and string type?

cretz (approved):
I think the term event is a bit overloaded. It's a bit confusing that you can have workflow events that are not even history events, but meh. Not a big deal.

temporal/api/nexus/v1/message.proto
bergundy: It was discussed in an internal document, the public doc should be updated too.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I agree we should have the option to link in all of these methods. Not blocking this PR but I'd at least want an issue to add it later.

temporal/api/common/v1/message.proto
rodrigozhou: Discussed offline. Added another struct to represent a reference.

PR #450, opened 2024-08-27
Add UpdateActivityOptions API

ychebotarev:
**What changed?**
Add definitions for UpdateActivityOptions API

**Why?**
So user can change activities options while workflow is running.

**Breaking changes**

**Server PR**
https://github.com/temporalio/temporal/pull/6452


temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+option ruby_package = "Temporalio::Api::Activity::V1";
+option csharp_namespace = "Temporalio.Api.Activity.V1";
+
+import "temporal/api/common/v1/message.proto";
+import "temporal/api/taskqueue/v1/message.proto";
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/field_mask.proto";
+
+message Timeouts {
```
cretz, line 40: In some cases (e.g. Go) you can often use the package's qualifier to keep from having to qualify the type/message names, but in proto it's best to qualify it IMO

```suggestion
message ActivityTimeouts {
```

temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+  //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+  google.protobuf.Duration start_to_close = 3;
+  // Maximum permitted time between successful worker heartbeats.
+  google.protobuf.Duration heartbeat_timeout = 4;
+}
+
+message Options {
+  temporal.api.taskqueue.v1.TaskQueue task_queue = 1;
+
+  temporal.api.activity.v1.Timeouts timeouts = 2;
```
cretz, line 73: I think the fields in `Timeouts` should be inlined, I don't think there's a need for a separate message. And I think we should try to make this match the field names/structure of `ActivityTaskScheduledEventAttributes` as close as is reasonable. (also you don't have to qualify a reference to a message in the same package/file)

temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+}
+
+message OptionsUpdate {
+  string task_queue = 1;
+
+  temporal.api.activity.v1.Timeouts timeouts = 2;
+
+  temporal.api.common.v1.RetryPolicy retry_policy = 3;
+
+  google.protobuf.FieldMask update_mask = 4;
```
cretz, line 85: I think the field mask concept should be at the top-level of the request on the service side and not embedded into objects here. Note how at https://google.aip.dev/161 the field mask is at the top-level request even though it points to the resource within.

temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+
+message Options {
+  temporal.api.taskqueue.v1.TaskQueue task_queue = 1;
+
+  temporal.api.activity.v1.Timeouts timeouts = 2;
+
+  temporal.api.common.v1.RetryPolicy retry_policy = 3;
+}
+
+message OptionsUpdate {
```
cretz, line 78: If you're using field masks, why have separate messages for `Options` and `OptionsUpdate`?

temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+  // timeout is always retryable. Either this or `schedule_to_close_timeout` must be specified.
+  //
+  // (-- api-linter: core::0140::prepositions=disabled
+  //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+  google.protobuf.Duration start_to_close = 3;
+  // Maximum permitted time between successful worker heartbeats.
+  google.protobuf.Duration heartbeat_timeout = 4;
+}
+
+message Options {
```
cretz, line 70: These options should be visible on the `PendingActivityInfo` message IMO

temporal/api/enums/v1/activity.proto
```diff
@@ -0,0 +1,43 @@
...
+
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ActivityProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+
+enum ActivityState {
```
cretz, line 35: Where is this used?

temporal/api/common/v1/message.proto
```diff
@@ -87,6 +87,15 @@ message WorkflowType {
     string name = 1;
 }
 
+// Identifies a specific activity within a namespace. ActivityExecution is expected to be
+// globally unique for running workflows.
+// run_id in workflow_execution can be empty,
+// when run_id is empty the latest run of the workflow should be selected
```
dnr, line 93: Should we really allow run_id to be empty in this message? The same activity_id in two different runs might refer to two completely different activities. In other words, workflow_id->run_id->activity_id are narrowing down a selection of a single activity. It seems strange to leave out the middle of the three qualifiers.

temporal/api/activity/v1/message.proto
ychebotarev: it was like that before, but I change that after comments...
I checked, looks like in other proto definitions we repeat package qualifier.
Will do the same

temporal/api/activity/v1/message.proto
ychebotarev: yeah, will move it to the request

temporal/api/activity/v1/message.proto
ychebotarev: when getting options (which I will remove, it is not used yet) I'm returning TaskQueue, that has some other fields besides just name. Those fields can't be set (sticky queue name).

temporal/api/activity/v1/message.proto
ychebotarev: I will remove this structure.

temporal/api/enums/v1/activity.proto
ychebotarev: right, not used in UpdateActivityOptions, I will remove it

temporal/api/activity/v1/message.proto
ychebotarev: I prefer to have a single place where all timeouts are defined. Otherwise they will be copy/pasted, which is potential source of errors.  That is how it is in ActivityTaskScheduledEventAttributes which I can't reuse...And now we have 2 copies.

temporal/api/activity/v1/message.proto
ychebotarev: that said - just options is not a part of this PR, I will remove it

temporal/api/common/v1/message.proto
ychebotarev: it states in the comment then when run_id is empty the latest run of the workflow should be selected.
It is the same (as i understand) as in any other API.

dnr: Yes, but most APIs don't operate on a specific entity within a workflow run, they operate on the run as a whole, so the run_id is the narrowest qualifier, which makes sense to leave out

ychebotarev: I'm not sure i understand the argument. activity_id is unique per run. there is only one latest run. so workflow_id + acitvity_id is enough to target a particular activity

dnr: It's not ambiguous, as long as the latest run doesn't change between when you pick out the activity id from history or whatever, and when you make this api call. If it does... then you may get a different activity than you mean.

Of course this is true of any api that defaults run_id to latest, it just feels worse for this one because of the specificity. Compare to terminate or signal execution, say. Even if the latest run changes between when you decide to terminate/signal and when the server receives the api call, defaulting to latest is very likely to preserve the user's intent in that case.

This is definitely more of a theoretical concern. Defaulting to latest is probably more convenient than requiring it. As I said, it just feels weird here.

ychebotarev: Ok, I understand the problem ("latest' right now may be not the one user is expecting). But I don't feel this API being "more vulnerable" to it . It is expected to be called reactively, when workflow is running and stuck.

alexshtin: I agree with David here, although I also can't say for 100%. Losing requirement is backward compatibility change, while making them more strict is not. Start with require, and then make it optional if we get feedback? 

Another idea, if leave it optional, at least return `run_id` in response, for user to double check, that right activity execution was used.

ychebotarev: I understand "latest right now may be not the one user is expecting", but it is like that for every other API, and I prefer consistency and "easy to use" over very unlikely problems

temporal/api/activity/v1/message.proto
ychebotarev: ok talked to Alex. Apparently it is too late for the right thing, and I also want to move forward. I will remove it as a separate structure. Lets duplicate those fields.

alexshtin (approved):

temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,54 @@
...
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Activity::V1";
+option csharp_namespace = "Temporalio.Api.Activity.V1";
+
+import "temporal/api/common/v1/message.proto";
+
+import "google/protobuf/duration.proto";
+
+message ActivityOptionsUpdate {
+    string task_queue = 1;
```
cretz, line 39:
```suggestion
    temporal.api.taskqueue.v1.TaskQueue task_queue = 1;
```

I think this should match the history event, no real benefit to be inconsistent in this case from what I see

temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,54 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Activity::V1";
+option csharp_namespace = "Temporalio.Api.Activity.V1";
+
+import "temporal/api/common/v1/message.proto";
+
+import "google/protobuf/duration.proto";
+
+message ActivityOptionsUpdate {
```
cretz, line 38: This doesn't seem specific to an update. I think this makes more sense as just `ActivityOptions` and I think it there should be a field on `temporal.api.workflow.v1.PendingActivityInfo` as `temporal.api.activity.v1.ActivityOptions activity_options`.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -60,6 +61,7 @@ import "temporal/api/sdk/v1/user_metadata.proto";
 import "temporal/api/nexus/v1/message.proto";
 
 import "google/protobuf/duration.proto";
+import "google/protobuf/field_mask.proto";
```
cretz, line 64: For some generators we've found that we need this file also duplicated at `google/protobuf/` in this repo. But it seems CI passes so maybe not? May want to manually apply these changes to the `proto/api` submodule in https://github.com/temporalio/api-go and confirm the Makefile there succeeds.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,16 @@ message ExecuteMultiOperationResponse {
...
+    string namespace = 1;
+    temporal.api.common.v1.ActivityExecution activity_execution = 2;
+
+    temporal.api.activity.v1.ActivityOptionsUpdate activity_options_update = 3;
+    string identity = 4;
+
+    google.protobuf.FieldMask update_mask = 5;
+}
+
+message UpdateActivityOptionsResponse {
```
cretz, line 1671: Since this can be a partial update I wonder if we should return the entire options value in the result

temporal/api/workflowservice/v1/service.proto
```diff
@@ -835,4 +835,17 @@ service WorkflowService {
     //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
     rpc RespondNexusTaskFailed(RespondNexusTaskFailedRequest) returns (RespondNexusTaskFailedResponse) {
     }
+
+    rpc UpdateActivityOptions (UpdateActivityOptionsRequest) returns (UpdateActivityOptionsResponse)
```
cretz, line 839: I wonder if this should be `UpdateActivityOptionsById` to match the other RPC methods that do things by activity ID

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,16 @@ message ExecuteMultiOperationResponse {
         }
     }
 }
+
+message UpdateActivityOptionsRequest {
+    string namespace = 1;
+    temporal.api.common.v1.ActivityExecution activity_execution = 2;
```
cretz, line 1663: This is inconsistent with the other ways the activity-id-specific RPC calls accept activity ID information. They accept the 3 pieces of information inline instead of have a whole separate object. Are we sure we want to be inconsistent and make a whole new object just for this call?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -835,4 +835,17 @@ service WorkflowService {
...
+    {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/workflows/{activity_execution.workflow_execution.workflow_id}/activities/{activity_execution.activity_id}/update"
```
cretz, line 842: This is inconsistent with all other activity RPCs that are by ID. I think this should be `/namespaces/{namespace}/activities/update-options-by-id`. I don't think we want this call to somehow be special compared to other activity calls.

temporal/api/activity/v1/message.proto
ychebotarev: TaskQueue has parameters 3 fields. Only one - name - can be updated. Others like "kind" and "normalName" can't, to my beset knowledge. I think it will be weird to provide such a structure.

temporal/api/activity/v1/message.proto
ychebotarev: Ok I will change it to ActivityOptions. Changes to PendingActivityInfo are out of scope of this PR.

temporal/api/workflowservice/v1/service.proto
ychebotarev: Other RPC has "ByID" because there also another version, which is not called by clients, and that doesn't have "WithToken". I think it was wrong from the start (client names should be clean).

We are coming will be plenty of other APIs - reset, pause, resume, describe, may be more.
I think it is not optimal to have "PauseById" without "Pause". 
And so far we don't plan API that accept "task token".

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ... 3 plain field was like it started.,,

temporal/api/activity/v1/message.proto
cretz: Yes, we have chosen to accept this structure for activities knowing that we will only use one of its fields. I don't think here is where we should deviate from that choice.

temporal/api/activity/v1/message.proto
cretz: To confirm, it is not part of this project to let them know the options they are going to change? If possible I think it should be included in this project.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: yeah I can see how it may be useful.

temporal/api/activity/v1/message.proto
ychebotarev: talk over zoom. will change.

temporal/api/workflowservice/v1/service.proto
ychebotarev: talk over offline. It looks like client have an essential way to get "activity handle" by "task_token". Thus if we don't provide a way on the server side to do the same for new Activity API - it will be not be the best user experience. I still disagree with the naming, but guess that train has sailed.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,47 @@ message ExecuteMultiOperationResponse {
...
     }
 }
+
+message UpdateActivityOptionsRequest {
+    string namespace = 1;
+
+    // A unique identifier for this task
+    bytes task_token = 2;
+
+    // activity options. Partial updates are accepted and controlled by update_mask
```
cretz, line 1667:
```suggestion
    // Activity options. Partial updates are accepted and controlled by update_mask
```

Pedantic, but worth being consistent with capitalization of comments

temporal/api/workflowservice/v1/service.proto
```diff
@@ -835,4 +835,31 @@ service WorkflowService {
     //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
     rpc RespondNexusTaskFailed(RespondNexusTaskFailedRequest) returns (RespondNexusTaskFailedResponse) {
     }
+
+    rpc UpdateActivityOptions (UpdateActivityOptionsRequest) returns (UpdateActivityOptionsResponse)
+    {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/workflows/activities/update"
```
cretz, line 842:
```suggestion
            post: "/namespaces/{namespace}/activities/update-options"
```

Two things (here and applies to call below too):

* To match other APIs, we don't want the term `workflows` in here
* Clarifying this is for updating options only I think is important (who knows we may have a concept of activity update one day)

cretz (commented):
Looks great, only one minor needed thing

cretz (approved):
While I am approving, I do think we should consider not merging until the implementation is mostly done just to confirm the implementation doesn't affect the API (making backwards incompatible changes to protos, even unused ones, is a pain). If it is basically done, no problem.

temporal/api/workflowservice/v1/request_response.proto
Sushisource, line 1667:
```suggestion
    // Activity options. Partial updates are accepted and controlled by update_mask
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,47 @@ message ExecuteMultiOperationResponse {
...
+    temporal.api.activity.v1.ActivityOptions activity_options = 3;
+    // The identity of the client who initiated this request
+    string identity = 4;
+    // the way to control partial updates
```
Sushisource, line 1671:
```suggestion
    // Controls which fields from `activity_options` will be applied
```

I would use this same language in the other spots too

temporal/api/activity/v1/message.proto
```diff
@@ -0,0 +1,55 @@
...
+
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 2;
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_start_timeout = 3;
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration start_to_close_timeout = 4;
```
Sushisource, line 50: We should copy the docstrings for these from other places they're defined too - no reason to leave stuff blank.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,47 @@ message ExecuteMultiOperationResponse {
         }
     }
 }
+
+message UpdateActivityOptionsRequest {
+    string namespace = 1;
+
+    // A unique identifier for this task
+    bytes task_token = 2;
```
Sushisource, line 1665:
```suggestion
    // The task token attached to the `PollActivityTaskResponse` which identifies the activity whose options are being updated
    bytes task_token = 2;
```

Sushisource (approved):
Just some docstring improvements.

Personally I'm more into this being one rpc with a oneof separating the different ways to target the activity, but I get that this matches some existing stuff, and it's not a huge deal.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,47 @@ message ExecuteMultiOperationResponse {
...
+
+message UpdateActivityOptionsByIdRequest {
+    // Namespace of the workflow which scheduled this activity
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity
+    // if empty - latest workflow is used
+    string run_id = 3;
+    // Id of the activity we're updating
```
drewhoskins-temporal, line 1683: The most common question somebody will ask is probably, "how do I get one?"
The programmatic answer, I believe, is from `ActivityTaskScheduledEventAttributes` from history:
https://github.com/temporalio/api/blob/4a1ab98a84b94b59f82aab4b35ea505cff968804/temporal/api/history/v1/message.proto#L938

So something like:

"You can get one from the workflow's event history, either in the UI or programmatically from `ActivityTaskScheduledEventAttributes`"

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: The naming should be consistent with other existing activity related APIs. Otherwise, as I understand, it will be very inconvenient to write SDK code.
In all other such functions (RecordActivityTaskHeartbeatRequest, etc.) we have bytes task_token.

I don't what to add "attached to the `PollActivityTaskResponse`" because this is "how", not "what".  I'm also not quite sure that this is how SDK will get that token.

drewhoskins-temporal: IMO, it's great to answer "how" questions.  It's what users need to know, so tell them.  And there are many examples of us answering "how" in this API.
Yes, we need to maintain it if it ever changes, but it's a small price to pay for saving users time.

"A unique identifier for this task" -- there is no "this task" w.r.t. this API.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,47 @@ message ExecuteMultiOperationResponse {
         }
     }
 }
+
+message UpdateActivityOptionsRequest {
```
drewhoskins-temporal, line 1661: Help me understand where we intend people to use this (as opposed to "ByID" method)

temporal/api/workflowservice/v1/request_response.proto
cretz: You can also set it but I believe we don't make that well known because we discourage it in favor of using task tokens to refer to an activity. I do not have the historical context here to know why. From the option to set activity ID in Go:

```
 	// ActivityID - Business level activity ID, this is not needed for most of the cases if you have
	// to specify this then talk to temporal team.
```

Probably just because people were setting it unnecessarily.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I think Chad will give you a better answer, I can just explain how I understand this.
This is Server API, not SDK.
In SDK at least some SDK activity is implemented as an object, like this:
https://python.temporal.io/temporalio.client.Client.html#get_async_activity_handle

so the code looks like
```
activity_handle = get_async_activity_handle(token)
activity_handle.Pause()

```
or

```
activity_handle = get_async_activity_handle(namespace, workflow_id, activity_id)
activity_handle.Pause()

```
at the background SDK translates it into Pause or PauseById Server API call, depending on how it was initialized.

We already have that, so it will be really bad user experience if they will not be able to do this with new activity APIs, but still can do that with old ones:
[AsyncActivityHandle](https://python.temporal.io/temporalio.client.AsyncActivityHandle.html)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: We may add "here is how you get activity_id" in SDK definitions. And definitely will add it into the documentation.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev:
```
    // Id of the activity to confirm is cancelled
    string activity_id = 4;

```   
```
    // Id of the activity we're heartbeating
    string activity_id = 4;

```
```
    // Id of the activity to complete
    string activity_id = 4;

```
```
    // Id of the activity to fail
    string activity_id = 4;

```
All of other API  that use activity_id don't answer this question. From my point of view it is because this is Server proto API definitions, not SDK documentation.

temporal/api/workflowservice/v1/request_response.proto
drewhoskins-temporal: Thanks, this helps.
So for these operability APIs, which are called externally (for now), and in failure cases, do the task-token based APIs matter?  i.e. how would such a person retrieve the token in this snippet?

```
activity_handle = get_async_activity_handle(task_token_obtained_when_running_activity)
```

My basic question is, should we leave this version of the API out for now until we need it, and if we should include it, what's the use case?

cretz: > do the task-token based APIs matter? i.e. how would such a person retrieve the token in this snippet?

Yes, task token is the preferred way. The task token is available via `activity.info().task_token` inside the activity. See https://docs.temporal.io/develop/python/asynchronous-activity-completion. People usually pass this token to another service that will complete their activity.

> My basic question is, should we leave this version of the API out for now until we need it, and if we should include it, what's the use case?

To me it makes sense to have all actions to an activity via a client be able to use either the 3-ID form or the task token form. I would assume the same use cases apply to this activity action as the other 4, but maybe not. I do think consistency alone has value, but I'll leave it up to y'all on the project.

temporal/api/workflowservice/v1/request_response.proto
drewhoskins-temporal: One difference is that those APIs are called from inside activities, where it's evident where to get the activity ID.  
I don't think it's clear here.  So I'm not sure the consistency argument applies, but it's a minor detail, so up to you.

Most importantly: Last I checked, we're not planning to expose this API from inside an activity or workflow.  That is, activities can't edit their own properties, and workflows can't edit the properties of activities that are running inside them.  Let's make sure we're aligned on that, as it's a bigger question if we're not.

temporal/api/workflowservice/v1/request_response.proto
cretz: Like task token is obtained from the activity info, so is ID (but you can also set it even though we discourage it)

temporal/api/workflowservice/v1/request_response.proto
cretz: > Last I checked, we're not planning to expose this API from inside an activity or workflow

We don't expose any of the 4 activity client calls inside of activities or workflows, I don't think we would for this 5th one. I don't think there's a use case to invoke them from within an activity. But I wouldn't be surprised if one day (e.g. when pause comes about) that a use case may arise.

temporal/api/workflowservice/v1/request_response.proto
cretz: This ID is also available on pending activities when describing a workflow.

temporal/api/workflowservice/v1/request_response.proto
cretz: > Yes, task token is the preferred way.

Now that I think about it, this is the preferred way for those existing 4 activity things because you are doing things to the task not the activity as a whole. I suspect the preferred way to update options will be for the activity as a whole and therefore by ID (what if the task completes while you are calling). Therefore I can see not having a task token based approach, but I still think it may be valuable to call this `ById` to be consistent. Then again, I can see a task token option too that is still by activity as a whole and not the specific task even though the token is task specific.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: > All of other API that use activity_id don't answer this question. From my point of view it is because this is Server proto API definitions, not SDK documentation.

I don't think that's a great justification. Because it isn't useful in existing places doesn't mean we should continue to keep it so. The purpose of docstrings here is to explain the contract between server and API users. The primary API user is the SDK - but even ignoring that, the comment I suggest helps you understand, as an API user, where the task token you are supposed to supply comes from. This is important information. "A unique identifier for this task" doesn't tell you anything that isn't already in the type signature / field name.

Regarding the other discussion about inside or outside a task, the comment I suggest still applies.

temporal/api/workflowservice/v1/request_response.proto
drewhoskins-temporal: Chad, I ran this thread by Max based on your tagging him.  He said:

> The token based is not needed

AFAICT, an operator fixing a production issue would have no way of getting a token.  It's not like they would have proactively exfiltrated a token from their activity so they could call this method.

drewhoskins-temporal: To your question about `ByID`: no strong opinion.  My preference is to leave it out, since it seems to imply to look for another option.  But shippable either way.

temporal/api/workflowservice/v1/request_response.proto
drewhoskins-temporal: > This ID is also available on pending activities when describing a workflow.

That's good to know, and actually sounds like the more useful way to tell people to do this than what I said, since presumably you're always operating on pending activities.

temporal/api/workflowservice/v1/request_response.proto
cretz: > AFAICT, an operator fixing a production issue would have no way of getting a token. (Right?)

Right, they would have to be already obtaining it for other reasons. Task tokens really are for just tasks which may not make sense for activities actions as a whole like this one. I think we need to introduce another concept of activity handle in SDKs to support doing things to activities (as opposed to the async activity handles some SDKs have today which are only for async completion support).

temporal/api/workflowservice/v1/service.proto
```diff
@@ -225,7 +225,7 @@ service WorkflowService {
     rpc PollActivityTaskQueue (PollActivityTaskQueueRequest) returns (PollActivityTaskQueueResponse) {
     }
 
-    // RecordActivityTaskHeartbeat is optionally called by workers while they execute activities.
+        // RecordActivityTaskHeartbeat is optionally called by workers while they execute activities.
```
dnr, line 228: looks like extra whitespace snuck in?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,42 @@ message ExecuteMultiOperationResponse {
...
```
dnr, line 1661: I didn't follow the whole thread or the current consensus so sorry if this is reiterating things:

IMO this api should work by wf id/run id/activity id and not task_token. So I would remove this version and keep only the activity_id version.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,42 @@ message ExecuteMultiOperationResponse {
...
+    // Id of the activity we're updating
+    string activity_id = 4;
+
+    // Activity options. Partial updates are accepted and controlled by update_mask
+    temporal.api.activity.v1.ActivityOptions activity_options = 5;
+
+    // The identity of the client who initiated this request
+    string identity = 6;
+    // Controls which fields from `activity_options` will be applied
+    google.protobuf.FieldMask update_mask = 7;
```
dnr, line 1692: put the mask field immediately after the thing it applies to, so it's clear that it's using a mask

temporal/api/workflowservice/v1/request_response.proto
dnr: Reasoning: it's changing the options for the whole activity call, not just one task of it. Also, it's much easier to find the activity id when debugging/fixing operational issues than find a task token (actually I have no idea how you would do that if you didn't save it somewhere when the task came in)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1655,3 +1657,29 @@ message ExecuteMultiOperationResponse {
         }
     }
 }
+
+message UpdateActivityOptionsByIdRequest {
+    // Controls which fields from `activity_options` will be applied
+    google.protobuf.FieldMask update_mask = 1;
```
cretz, line 1663: Personally more of a fan of putting this just below the resource it references, but doesn't matter much

temporal/api/workflowservice/v1/service.proto
```diff
@@ -835,4 +835,21 @@ service WorkflowService {
     //     aip.dev/not-precedent: We do not expose worker API to HTTP. --)
     rpc RespondNexusTaskFailed(RespondNexusTaskFailedRequest) returns (RespondNexusTaskFailedResponse) {
     }
+
+    // UpdateActivityOptionsById is called by the client to update the options of an activity
+    // (-- api-linter: core::0136::prepositions=disabled
+    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
+    rpc UpdateActivityOptionsById (UpdateActivityOptionsByIdRequest) returns (UpdateActivityOptionsByIdResponse)
+    {
```
cretz, line 843:
```suggestion
    rpc UpdateActivityOptionsById (UpdateActivityOptionsByIdRequest) returns (UpdateActivityOptionsByIdResponse) {
```

cretz (approved):

dnr (approved):

PR #455, opened 2024-09-27
Add support to add links to more workflow APIs and link to a batch job

bergundy:
**Why?**

Better user experience to gain visibility into batch operations.

**Server PR**

Not implemented yet.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -639,6 +641,9 @@ message SignalWorkflowExecutionRequest {
     temporal.api.common.v1.Header header = 8;
     // Indicates that a new workflow task should not be generated when this signal is received.
     bool skip_generate_workflow_task = 9;
+
+    // Links to be associated with the WorkflowExecutionSignaled event.
+    repeated temporal.api.common.v1.Link links = 10;
```
Quinn-With-Two-Ns, line 646: What about workflow update ? Can we add link support there?

bergundy: Eventually yes, I guess I can just add that already.
We'll also want to support backlinks to updates and signals but those will come later as well, I'm specifically wanting this for batch operations because it's a no brainer to add.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -699,6 +704,10 @@ message SignalWithStartWorkflowExecutionRequest {
     // for use by user interfaces to display the fixed as-of-start summary and details of the
     // workflow.
     temporal.api.sdk.v1.UserMetadata user_metadata = 23;
+
+    // Links to be associated with the WorkflowExecutionSignaled event. Currently there is not a way to link to the
+    // WorkflowExecutionStarted event via this API.
+    repeated temporal.api.common.v1.Link signal_links = 24;
```
cretz, line 710: What about workflow links here? As we unify the SDK/CLI calls for starting with something (signal/update) and just starting, we need to try to make sure all the "with start" requests have at least the same thing as the workflow requests. I regret not catching this when start workflow links were added.

bergundy: We can have best effort workflow links here for the case where your request started a workflow but at that point are you really getting value over the signal link?

bergundy: Alternatively, we can duplicate the signal link to the workflow execution started event if a request starts a workflow.

cretz: Yes, that would work. I think it's understood on signal with start that if the workflow exists already some of the start options do not apply, especially these metadata ones.

pdoerner (approved):

temporal/api/workflowservice/v1/request_response.proto
cretz: > So why not have a single "links" field and apply it to the start event if the request started the workflow? Can you see a scenario where you want different links for start and signal?

So will the links only be for 1) start if started or discarded, or 2) start if started and signal if not started, or 3) start if started and signal always? If "1) start if started or discarded" that works for me since we can always come back and add signal links later. If either of the other two, I think it would be confusing.

bergundy: My proposal is have a single `links` field that gets populated on:
1. The WorkflowExecutionSignaled event - if the request succeeds
2. The WorkflowExecutionStarted event - if the request starts a new execution

cretz: I think that's confusing to users. Sometimes the link gets on a signal and sometimes not? I think it should consistently always set on signal or never. I believe Our other options are consistent regardless of whether started. Can you confirm the behavior of `header` and `user_metadata` (both of which are also metadata that technically could apply to both) in cases where a workflow is started vs not?

bergundy: > I think it should consistently always set on signal or never.

It is consistent with my proposal, just requires the request to be successful and the signal to actually be recorded. That's a given maybe so I could have just worded this differently.

> Can you confirm the behavior of header and user_metadata (both of which are also metadata that technically could apply to both) in cases where a workflow is started vs not?

Looks like we don't record the metadata on the signal or start events in this case (cc @gow to confirm): 

https://github.com/temporalio/temporal/blob/6346c70817ffcdb95f60c9edc0e6d39b9b772daf/service/history/api/signalwithstartworkflow/signal_with_start_workflow.go#L314-L320

https://github.com/temporalio/temporal/blob/428f71a0f96511b13c01c977cce9ee8a57f238e2/service/history/historybuilder/event_factory.go#L60

I see the header is applied to both events in the code (haven't confirmed with a test though).

So basically what I'm proposing is consistent with `header`.

gow: @bergundy `user_metadata` is added to workflow start event. It's added in [HistoryBuilder.AddWorkflowExecutionStartedEvent()](https://github.com/temporalio/temporal/blob/main/service/history/historybuilder/history_builder.go#L187)
Regarding signal events, I don't think the `user_metadata` is added there.

cretz: > I see the header is applied to both events in the code (haven't confirmed with a test though). [...] So basically what I'm proposing is consistent with header.

Ah, ok! I misinterpreted your original suggestion to link to _either_ signal or start, not _both_. I support both. So yes, I think this field name should be changed to `links` and we're good.

bergundy: @gow we should consider adding the metadata to the signal event as well. @cretz are we aligned on that?

cretz: Concerning user metadata, sure, though it has no use on signal today so this is not an urgent need or anything (it'll just duplicate workflow-level metadata which is probably fine for now).

cretz (approved):

PR #456, opened 2024-10-01
Add activiy APIs - pause, reset, resume, describe

ychebotarev:
Add definitions for DescribeActivity, PauseActivity, ResumeActivity and ResetActivity API

Part of Activity API work.

No.

**Server PR**
https://github.com/temporalio/temporal/pull/6584

### DIFF
--- a/temporal/api/enums/v1/activity.proto
+++ b/temporal/api/enums/v1/activity.proto
@@ -1,0 +1,41 @@
+// The MIT License
+//
+// Copyright (c) 2024 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
+
+package temporal.api.enums.v1;
+
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ActivityProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+enum ActivityState {
+  ACTIVITY_STATE_UNSPECIFIED = 0;
+  ACTIVITY_STATE_SCHEDULED = 1;
+  ACTIVITY_STATE_STARTED = 2;
+  ACTIVITY_STATE_PAUSED = 3;
+  ACTIVITY_STATE_WAITING_FOR_RETRY = 4;
+  ACTIVITY_STATE_FINISHED = 5;
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -31,6 +31,7 @@ option java_outer_classname = "RequestResponseProto";
 option ruby_package = "Temporalio::Api::WorkflowService::V1";
 option csharp_namespace = "Temporalio.Api.WorkflowService.V1";
 
+import "temporal/api/enums/v1/activity.proto";
 import "temporal/api/enums/v1/batch_operation.proto";
 import "temporal/api/enums/v1/common.proto";
 import "temporal/api/enums/v1/workflow.proto";
@@ -1694,4 +1695,93 @@ message UpdateActivityOptionsByIdRequest {
 message UpdateActivityOptionsByIdResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
+}
+
+message DescribeActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity
+    // if empty - latest workflow is used
+    string run_id = 3;
+    // Id of the activity we're updating
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request
+    string identity = 5;
+}
+
+message DescribeActivityByIdResponse {
+    // Activity options after an update
+    temporal.api.activity.v1.ActivityOptions activity_options = 1;
+
+    // Activity state
+    temporal.api.enums.v1.ActivityState activity_state = 2;
+}
+
+message PauseActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity
+    // if empty - latest workflow is used
+    string run_id = 3;
+    // Id of the activity we're updating
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request
+    string identity = 5;
+}
+
+message PauseActivityByIdResponse {
+
+}
+
+message ResumeActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity
+    // if empty - latest workflow is used
+    string run_id = 3;
+    // Id of the activity we're updating
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request
+    string identity = 5;
+
+    // Indicates that activity should be triggered immediately after resuming
+    bool trigger_immediately = 6;
+}
+
+message ResumeActivityByIdResponse {
+
+}
+
+message ResetActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity
+    // if empty - latest workflow is used
+    string run_id = 3;
+    // Id of the activity we're updating
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request
+    string identity = 5;
+
+    // Starts new activity even if the activity is already running
+    bool force_new_run = 6;
+
+    // If activity was paused, it will be kept paused after reset
+    bool keep_paused = 7;
+}
+
+message ResetActivityByIdResponse {
+
 }
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -856,7 +856,7 @@ service WorkflowService {
     rpc RespondNexusTaskFailed(RespondNexusTaskFailedRequest) returns (RespondNexusTaskFailedResponse) {
     }
 
-    // UpdateActivityOptionsById is called by the client to update the options of an activity
+    // UpdateActivityOptionsById is called by the client to update the options of an activity.
     // (-- api-linter: core::0136::prepositions=disabled
     //     aip.dev/not-precedent: "By" is used to indicate request type. --)
     rpc UpdateActivityOptionsById (UpdateActivityOptionsByIdRequest) returns (UpdateActivityOptionsByIdResponse)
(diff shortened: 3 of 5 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1695,3 +1696,92 @@ message UpdateActivityOptionsByIdResponse {
...
+    // The identity of the client who initiated this request
+    string identity = 5;
+}
+
+message DescribeActivityByIdResponse {
+    // Activity options after an update
+    temporal.api.activity.v1.ActivityOptions activity_options = 1;
+
+    // Activity state
+    temporal.api.enums.v1.ActivityState activity_state = 2;
```
cretz, line 1720: I wonder if there is value in `temporal.api.activity.v1.ActivityDescription` that encompasses these and maybe more in the future. Arguably this API and pending activity API have a lot of model overlap.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,63 @@ service WorkflowService {
         };
     }
 
-}
+    // DescribeActivityById is called by the client to get current activity options and state.
+    // (-- api-linter: core::0136::prepositions=disabled
+    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
+    rpc DescribeActivityById (DescribeActivityByIdRequest) returns (DescribeActivityByIdResponse)
+    {
```
cretz, line 878:
```suggestion
    rpc DescribeActivityById (DescribeActivityByIdRequest) returns (DescribeActivityByIdResponse) {
```

(and `UpdateActivityOptionsById` above where we didn't notice this)

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,63 @@ service WorkflowService {
...
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/activities/describe-by-id"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/activities/describe-by-id"
+                body: "*"
```
cretz, line 884:
```suggestion
            get: "/namespaces/{namespace}/activities/description-by-id"
            additional_bindings {
                get: "/api/v1/namespaces/{namespace}/activities/description-by-id"
```

Or something like that. Basically this should be a `GET` IMO.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,63 @@ service WorkflowService {
...
+            post: "/namespaces/{namespace}/activities/resume-by-id"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/activities/resume-by-id"
+                body: "*"
+            }
+        };
+    }
+
+    // ResetActivityById is called by the client to reset the activity to its "original" state.
```
cretz, line 919: Can you clarify what this does a bit? Is this only removing all retries? My concern is we have the verb "reset" for workflows which is not necessarily the same and may be a bit confusing. But I can't think of a better term.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,63 @@ service WorkflowService {
...
+            post: "/namespaces/{namespace}/activities/describe-by-id"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/activities/describe-by-id"
+                body: "*"
+            }
+        };
+    }
+
+    // PauseActivityById is called by the client to pause an activity.
```
cretz, line 889: Can you clarify what happens if it's already paused? Similarly, can you clarify on the resume call what happens if it's not paused? May also want to clarify what these do if the activity is already complete.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: This does has a lot of in common with PendingActivityInfo. And I wish I can just use PendingActivityInfo. But - PendingActivityInfo is more "internal". It has lots of fields that are at best "situational", etc. 
I though it will be cleaner if we have a separate structure.

temporal/api/workflowservice/v1/service.proto
ychebotarev: done. what about the others?

temporal/api/workflowservice/v1/service.proto
ychebotarev: I added more comments.

temporal/api/workflowservice/v1/service.proto
ychebotarev: add more comments

temporal/api/workflowservice/v1/request_response.proto
cretz: Ignoring the pending activity info part, I still wonder if a combination of these things (and potentially more in the future) deserves an encompassing `ActivityDescription` message.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,68 @@ service WorkflowService {
...
+            }
+        };
+    }
 
+    // ResetActivityById is called by the client to reset the activity to its "original" state.
+    // This is useful when an activity is stuck in a state that cannot be resolved by other means.
+    // With this call the number of attempts will be reset to 0 and the activity can be rescheduled.
+    // This call will fail if the activity is in a terminal state (finished).
+    // If activity is Pause state, it will be resumed if the corresponding flag is provided.
+    // If activity currently running - it will be terminated and rescheduled if the corresponding flag is provided.
```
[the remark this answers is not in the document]
ychebotarev: I updated the comments and add more details.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: Just like "DescribeWorkflowExecutionResponse", I think this structure is the one to have all potential fields, and wrapping those fields into ActivityDescription message doesn't add much. It is an extra level of indirection.

cretz (commented):
This looks good to me.

The only reason I am hesitant to approve is I would like to see if we can get implementation (mostly) done before merging just in case we learn things during implementation. But otherwise, effectively "approved" from my POV.

temporal/api/enums/v1/activity.proto
```diff
@@ -0,0 +1,41 @@
...
+option java_outer_classname = "ActivityProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+enum ActivityState {
+  ACTIVITY_STATE_UNSPECIFIED = 0;
+  ACTIVITY_STATE_SCHEDULED = 1;
+  ACTIVITY_STATE_STARTED = 2;
+  ACTIVITY_STATE_PAUSED = 3;
+  ACTIVITY_STATE_WAITING_FOR_RETRY = 4;
```
bergundy, line 39: Can you call this BACKING_OFF to match the terminology we use for Nexus?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,71 @@ service WorkflowService {
         };
     }
 
-}
+    // DescribeActivityById is called by the client to get current activity options and state.
```
bergundy, line 873: Does this only return pending activities? Otherwise, how does this handle when two activities with the same ID (yes, that's supported as long as they're not both pending).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1695,3 +1696,92 @@ message UpdateActivityOptionsByIdResponse {
...
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request
+    string identity = 5;
+
+    // Starts new activity even if the activity is already running
+    bool force_new_run = 6;
+
+    // If activity was paused, it will be kept paused after reset
+    bool keep_paused = 7;
```
bergundy, line 1782: Should we support reset and pause in the same request?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1695,3 +1696,92 @@ message UpdateActivityOptionsByIdResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
 }
+
+message DescribeActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity
```
bergundy, line 1701: Please put `.` at the end of sentences to be consistent with the rest of the docstrings in this file.

temporal/api/enums/v1/activity.proto
```diff
@@ -0,0 +1,41 @@
...
+package temporal.api.enums.v1;
+
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ActivityProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+enum ActivityState {
```
bergundy, line 34: Is this PendingActivityState or general ActivityState?

bergundy: Also, no need for a new file here IMHO, we already have most of the enums, including one for PendingActivityState here: https://github.com/temporalio/api/blob/b0c0ac92cf14bba89e2465fa551fb71fab44a2d8/temporal/api/enums/v1/workflow.proto#L104.

temporal/api/workflowservice/v1/service.proto
bergundy: I'm asking because DescribeWorkflowExecution only returns pending activities as those are stored in mutable state. If you want to return finished activities, you'll need to load history.

temporal/api/enums/v1/activity.proto
```diff
@@ -0,0 +1,41 @@
...
+  ACTIVITY_STATE_UNSPECIFIED = 0;
+  ACTIVITY_STATE_SCHEDULED = 1;
+  ACTIVITY_STATE_STARTED = 2;
+  ACTIVITY_STATE_PAUSED = 3;
```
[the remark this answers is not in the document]
bergundy: If you do think that PAUSED is a valid state, please update `PendingActivityState` to reflect this as well.

bergundy:
> This looks good to me.
> 
> The only reason I am hesitant to approve is I would like to see if we can get implementation (mostly) done before merging just in case we learn things during implementation. But otherwise, effectively "approved" from my POV.

+1 to this, please use a feature branch until the implementation solidifies and you're confident in the API.

temporal/api/workflowservice/v1/request_response.proto
cretz: > Should we support reset and pause in the same request?

I think they are different operations entirely. Whether we want to combine operations or not in general I don't have a strong opinion, but I don't think these two are special.

> To give us time to collect more opinions, can we leave it out for this PR

Can you clarify which part to leave out of the PR? Pausing? Reset? If both are still in, we need to define _some_ behavior for resetting a paused activity and a default. Whether that default is unpause or remain paused I have no opinion on.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1695,3 +1696,92 @@ message UpdateActivityOptionsByIdResponse {
...
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity
+    // if empty - latest workflow is used
+    string run_id = 3;
+    // Id of the activity we're updating
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request
```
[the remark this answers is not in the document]
cretz: This is present in almost all actionable RPCs and is a unique client identifier sent by the SDK. It allows you to uniquely identify a client that did something (we often store this on events and other places). This is a settable option in every SDK and usually defaults to a string with process ID and hostname.

temporal/api/enums/v1/activity.proto
ychebotarev: ok

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: This is after long discussions with Drew. Basically if user paused and then reset - in some cases  he may want to keep paused activities to stay paused. We right now discussing what should be "by default" - keep paused and specify a flag to resume, or resume and specify a flag to keep paused.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: Skipping UpdateActivityOptionsByIdRequest (which doesn't have dots for the same reason), scrolling above XXXNexus request/responses:

PollWorkflowExecutionUpdateResponse - no dots
PollWorkflowExecutionUpdateRequest - no dots
ListBatchOperationsResponse - no dots
ListBatchOperationsRequest - no dots
DescribeBatchOperationResponse - no dots
StopBatchOperationRequest - no dots

activity related:
PollActivityTaskQueueResponse - no dots
RecordActivityTaskHeartbeatRequest - no dots
RecordActivityTaskHeartbeatByIdRequest- no dots

from what I see not having `. ` at the end is more consistent with the rest of the docstrings in this file. 

I do also see some with dots, but seems they are minority.
Though if we go top-to-bottom - first few have dots...
Hm, also it looks like this is a subject of some heated discussions online. And Google proto style guidelines don't have recommendations.

temporal/api/enums/v1/activity.proto
ychebotarev: I will remove this. I try to keep it separated, but I think it will just create more confusion. I will (re)use PendingActivityState

temporal/api/enums/v1/activity.proto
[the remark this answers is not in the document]
ychebotarev: I will add PAUSED and BACKING_OFF to PendingActivityState

temporal/api/workflowservice/v1/service.proto
ychebotarev: only pending activities

temporal/api/workflowservice/v1/request_response.proto
bergundy: Let's have periods in our doc strings. I think the majority do.

temporal/api/workflowservice/v1/service.proto
bergundy: Put this in the docstring.

PR #457, opened 2024-10-07
add more states to pending activift info

ychebotarev:
**What changed?**

**Why?**

**Breaking changes**

**Server PR**


temporal/api/enums/v1/workflow.proto
```diff
@@ -106,6 +106,8 @@ enum PendingActivityState {
     PENDING_ACTIVITY_STATE_SCHEDULED = 1;
     PENDING_ACTIVITY_STATE_STARTED = 2;
     PENDING_ACTIVITY_STATE_CANCEL_REQUESTED = 3;
+    PENDING_ACTIVITY_STATE_BACKING_OFF = 4;
```
bergundy, line 109: As mentioned in our slack conversation. This is an artificial state and the state machine doesn't record this explicitly to save DB updates. I think it's nice to be able to surface this state to users though so I support adding it to the API.

temporal/api/enums/v1/workflow.proto
```diff
@@ -106,6 +106,8 @@ enum PendingActivityState {
...
+    PENDING_ACTIVITY_STATE_PAUSED = 5;
```
bergundy, line 110: As discussed in another PR. I wouldn't model this as a distinct state and instead have another dimension for expressing pause status to be able to represent scenarios such as: an attempt is currently started and the activity has been requested to pause.

ychebotarev: From user perspective I don't see a scenario when it matters from what state activity comes to the PAUSE state. 
Activity cannot be started and paused, it is just paused.

cretz: Agreed. These states are more where the activity is at in general (started, scheduled, etc) and not specific things about the already-started activity.

bergundy: I'm not 100% sure what you agree with @cretz.
FTR the scheduler state has a separate bool for paused: https://github.com/temporalio/api/blob/c038641453772f879638ceffb1cb2664e0ad1765/temporal/api/schedule/v1/message.proto#L281.

I would rather have this consistency.

bergundy (approved):
Discussed with @ychebotarev offline. I'm okay with the way he decided to represent these states and he's clarified that internally we don't lose any information.

temporal/api/enums/v1/workflow.proto
cretz: Sorry, I see my message posted at the same time as @ychebotarev.

I agree with the original concept that should not be in the enumerate for activity state. A paused or backing off activity is still "scheduled" or "started" based on how this enum is used. Not only is it a backwards incompatible change to suddenly start making backing-off activities be a different enum, but I agree paused/backing-off is a different dimension from whether the activity is scheduled/started.

ychebotarev: @cretz - there will be bool variable in PendingActivityInfo that indicates that activity is paused. PendingActivityInfo will also be part of the DescribeActivity response. This state is a  separate thing. It is more about "give it quick look and understand where are we".

cretz: "started" is not "running" and to make people's code that expects backing-off activities to be this new enum instead of what it always was is backwards incompatible. If we need a more detailed value somewhere because "started" needs to be split into multiple values, that makes sense.

PR #458, opened 2024-10-07
Add capabilities field to RespondWorkflowTaskCompletedResponse

Quinn-With-Two-Ns:
Add capabilities field to RespondWorkflowTaskCompletedResponse. Add `may_discard_speculative_workflowtask_with_command_events` that will allow the SDK to signal to the Server that it supports discarding speculative workflow tasks with command events. When this is set to true it will allow the server to discard a lot more speculative workflow tasks then it can today.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -372,7 +372,7 @@ message RespondWorkflowTaskCompletedResponse {
         // True if the SDK can handle speculative workflow task with command events.
         // If true, the server may chosse, at its discretion, to discard a speculative workflow task
         // even if that speculative task included command events the SDK had not previouly processed.
-        bool may_discard_speculative_workflowtask_with_command_events = 1;
+        bool may_discard_speculative_workflow_task_with_command_events = 1;
     }
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -371,7 +371,10 @@ message RespondWorkflowTaskCompletedResponse {
     message Capabilities {
         // True if the SDK can handle speculative workflow task with command events.
         // If true, the server may chosse, at its discretion, to discard a speculative workflow task
-        // even if that speculative task included command events the SDK had not previouly processed.
+        // even if that speculative task included command events the SDK had not previouly processed
+        //
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: "with" used to describe the workflow task. --)
         bool discard_speculative_workflow_task_with_command_events = 1;
     }
 }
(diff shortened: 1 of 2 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -365,6 +365,15 @@ message RespondWorkflowTaskCompletedResponse {
...
     // Server is only expected to discard a workflow task if it could not have modified the workflow state.
     int64 reset_history_event_id = 3;
+    // All capabilities the SDK supports.
+    Capabilities capabilities = 4;
+    // SDK capability details.
+    message Capabilities {
+        // True if the SDK can handle speculative workflow task with command events.
+        // If true, the server may chosse, at its discretion, to discard a speculative workflow task
+        // even if that speculative task included command events the SDK had not previouly processed.
+        bool may_discard_speculative_workflowtask_with_command_events = 1;
```
cretz, line 375:
```suggestion
        bool may_discard_speculative_workflow_task_with_command_events = 1;
```

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -365,6 +365,18 @@ message RespondWorkflowTaskCompletedResponse {
...
+    Capabilities capabilities = 4;
+    // SDK capability details.
+    message Capabilities {
+        // True if the SDK can handle speculative workflow task with command events.
+        // If true, the server may chosse, at its discretion, to discard a speculative workflow task
+        // even if that speculative task included command events the SDK had not previouly processed
+        //
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: "with" used to describe the workflow task. --)
+        bool discard_speculative_workflow_task_with_command_events = 1;
```
alexshtin, line 378:
```suggestion
        bool discard_speculative_workflow_task_with_events = 1;
```
`command_events` is not a common well known name. And in fact, all events falls in this category. Let's say we enable signal without WFT one day.

Sushisource: We use "command_events" in the SDKs all over the replace to refer to the events corresponding to commands. So, not all events are that - like signaled events.

Quinn-With-Two-Ns: Yeah `command_events ` is a common term in the SDK

Quinn-With-Two-Ns:
>Are there any SDKs with this capability today?

Yes, all latest SDK now do. but old versions didn't

temporal/api/workflowservice/v1/request_response.proto
Quinn-With-Two-Ns: that being said I am not opposed to removing it

bergundy:
> > Are there any SDKs with this capability today?
> 
> Yes, all latest SDK now do. but old versions didn't

I would maybe consider hard coding these SDK versions on the server instead of adding this mechanism since it's easy enough for us to track those but I'm okay adding it if people see this useful beyond the single use case.

cretz:
> Yes, all latest SDK now do. but old versions didn't

@Quinn-With-Two-Ns - what happens to older SDKs if the server were to just behave as if this new field was always true? Are all SDKs that don't work with this part of update pre-release instead of public preview? Just trying to help justify a new proto field just for pre-release update.

PR #459, opened 2024-10-08
Feature request - add more field for activity info

ychebotarev:
**What changed?**
Add few more fields to the PendingActivityInfo

**Why?**
Feature request. Users want to know what will be the next retry interval, or when exactly last failure happens.

No as far as I know.

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -143,8 +143,8 @@ message PendingActivityInfo {
     // If there will be no retry it will be null.
     google.protobuf.Duration current_retry_interval = 16;
 
-    // The time when the last activity attempt failed. If there was no failure then it will be null.
-    google.protobuf.Timestamp last_failure_time = 17;
+    // The time when the last activity attempt was completed. If activity has not been completed yet then it will be null.
+    google.protobuf.Timestamp last_attempt_complete_time = 17;
 
     // Next time when activity will be scheduled.
     // If activity is currently scheduled or started it will be null.
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -140,14 +140,14 @@ message PendingActivityInfo {
     // The time activity will wait until the next retry.
     // If activity is currently running it will be next retry interval if activity failed.
     // If activity is currently waiting it will be current retry interval.
-    // If there will be no retry it will be empty.
+    // If there will be no retry it will be null.
     google.protobuf.Duration current_retry_interval = 16;
 
-    // The time when activity failed last time. If there was no failure then it will be empty.
+    // The time when the last activity attempt failed. If there was no failure then it will be null.
     google.protobuf.Timestamp last_failure_time = 17;
 
     // Next time when activity will be scheduled.
-    // If activity is currently running it will be empty.
+    // If activity is currently scheduled or started it will be null.
     google.protobuf.Timestamp next_attempt_schedule_time = 18;
 }
 
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -136,6 +136,19 @@ message PendingActivityInfo {
     }
     // The version stamp of the worker to whom this activity was most recently dispatched
     temporal.api.common.v1.WorkerVersionStamp last_worker_version_stamp = 15;
+
+    // The time activity will wait until the next retry.
+    // If activity is currently running it will be next retry interval if activity failed.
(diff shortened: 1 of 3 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +136,19 @@ message PendingActivityInfo {
...
+    // If activity is currently waiting it will be current retry interval.
+    // If there will be no retry it will be empty.
+    google.protobuf.Duration current_retry_interval = 16;
+
+    // The time when activity failed last time. If there was no failure then it will be empty.
+    google.protobuf.Timestamp last_failure_time = 17;
+
+    // Next time when activity will be scheduled.
+    // If activity is currently running it will be empty.
+    google.protobuf.Timestamp next_attempt_schedule_time = 18;
```
cretz, line 151: I'm curious, in what situations is this not just `last_failure_time` + `current_retry_interval`?

cretz: Works for me. I am not the biggest fan of adding derived fields, but it sounds like y'all have decided that we do want to add them.

bergundy: This is just for convenience and consistency with what we did for Nexus.

temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +136,19 @@ message PendingActivityInfo {
     }
     // The version stamp of the worker to whom this activity was most recently dispatched
     temporal.api.common.v1.WorkerVersionStamp last_worker_version_stamp = 15;
+
+    // The time activity will wait until the next retry.
+    // If activity is currently running it will be next retry interval if activity failed.
+    // If activity is currently waiting it will be current retry interval.
+    // If there will be no retry it will be empty.
```
cretz, line 143:
```suggestion
    // If there will be no retry it will be null.
```

Same with the other two comments. There is a difference between empty and null for these values, so worth using the word `null` clearly.

cretz (approved):

temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +136,19 @@ message PendingActivityInfo {
...
+    google.protobuf.Duration current_retry_interval = 16;
+
+    // The time when activity failed last time. If there was no failure then it will be empty.
```
bergundy, line 146:
```suggestion
    // The time when the last activity attempt failed. If there was no failure then it will be empty.
```

temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +136,19 @@ message PendingActivityInfo {
...
+    google.protobuf.Timestamp last_failure_time = 17;
+
+    // Next time when activity will be scheduled.
+    // If activity is currently running it will be empty.
```
bergundy, line 150:
```suggestion
    // If activity is currently scheduled or started it will be empty.
```

temporal/api/workflow/v1/message.proto
bergundy: You can use `last_attempt_complete_time`. If we ever have a special status for polling, we can use this field to represent that status too.

temporal/api/workflow/v1/message.proto
ychebotarev: done

temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +136,19 @@ message PendingActivityInfo {
...
     temporal.api.common.v1.WorkerVersionStamp last_worker_version_stamp = 15;
+
+    // The time activity will wait until the next retry.
+    // If activity is currently running it will be next retry interval if activity failed.
+    // If activity is currently waiting it will be current retry interval.
+    // If there will be no retry it will be null.
+    google.protobuf.Duration current_retry_interval = 16;
+
+    // The time when the last activity attempt failed. If there was no failure then it will be null.
+    google.protobuf.Timestamp last_failure_time = 17;
```
bergundy, line 147: Looks like my comment here was lost:

> You can use last_attempt_complete_time. If we ever have a special status for polling, we can use this field to represent that status too.

bergundy (commented):
I'd wait until we confirm this with a server PR.

temporal/api/workflow/v1/message.proto
ychebotarev: done

temporal/api/workflow/v1/message.proto
ychebotarev: renamed

bergundy (approved):

PR #460, opened 2024-10-08
Fix Batch job ID in Link should have been a string

bergundy:
It's a breaking change but the API hasn't been in use in any server or SDK release yet.

I messed this up in #455


gow (approved):

temporal/api/common/v1/message.proto
```diff
@@ -234,7 +234,7 @@ message Link {
     // Batch jobs can be used to perform operations on a set of workflows (e.g. terminate, signal, cancel, etc).
     // This link can be put on workflow history events generated by actions taken by a batch job.
     message BatchJob {
-        int64 job_id = 1;
```
ychebotarev, line 237: talk offline. It is a breaking change, not it is not in use yet. So we good.

ychebotarev (approved):

PR #462, opened 2024-10-10
Add DescribeActivityById definition

ychebotarev:
Add definition for DescribeActivityById APi.

**Why?**
Part of Activity API work.

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -31,7 +31,10 @@ option java_outer_classname = "MessageProto";
 option ruby_package = "Temporalio::Api::Activity::V1";
 option csharp_namespace = "Temporalio.Api.Activity.V1";
 
+import "google/protobuf/timestamp.proto";
+
 import "temporal/api/common/v1/message.proto";
+import "temporal/api/enums/v1/activity_state.proto";
 import "temporal/api/taskqueue/v1/message.proto";
 
 import "google/protobuf/duration.proto";
@@ -64,4 +67,23 @@ message ActivityOptions {
     google.protobuf.Duration heartbeat_timeout = 5;
 
     temporal.api.common.v1.RetryPolicy retry_policy = 6;
+}
+
+// ActivityInfo contains information about an activity.
+// It is used to report activity state to the users.
+// It is expected to return information about pending activities and completed activities.
+message ActivityInfo {
+    // ActivityOptions contains the options that were used to operate the activity.
+    temporal.api.activity.v1.ActivityOptions options = 1;
+    // ActivityType is the type of the activity.
+    temporal.api.common.v1.ActivityType type = 2;
+    // ActivityState is the current state of the activity.
+    temporal.api.enums.v1.ActivityState state = 3;
+    // ScheduledTime is the time last the activity was scheduled. .
+    google.protobuf.Timestamp scheduled_time = 4;
+    // StartedTime is the time last the activity was started.
+    google.protobuf.Timestamp started_time = 5;
+    // CompletedTime is the time last the activity was completed.
+    // If activity is not yet completed, this field will be null.
+    google.protobuf.Timestamp completed_time = 6;
 }
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -69,21 +69,21 @@ message ActivityOptions {
     temporal.api.common.v1.RetryPolicy retry_policy = 6;
 }
 
-// ActivityInfo contains information about an activity.
-// It is used to report activity state to the users.
-// It is expected to return information about pending activities and completed activities.
+// Information about an activity in any state including pending and completed activities.
 message ActivityInfo {
-    // ActivityOptions contains the options that were used to operate the activity.
+    // Contains the options that were used to operate the activity.
     temporal.api.activity.v1.ActivityOptions options = 1;
-    // ActivityType is the type of the activity.
+    // Type of the activity.
     temporal.api.common.v1.ActivityType type = 2;
-    // ActivityState is the current state of the activity.
+    // Current state of the activity.
     temporal.api.enums.v1.ActivityState state = 3;
-    // ScheduledTime is the time last the activity was scheduled. .
-    google.protobuf.Timestamp scheduled_time = 4;
-    // StartedTime is the time last the activity was started.
-    google.protobuf.Timestamp started_time = 5;
-    // CompletedTime is the time last the activity was completed.
+    // The last time the activity was scheduled.
+    google.protobuf.Timestamp last_scheduled_time = 4;
+    // Last time the activity was started.
+    google.protobuf.Timestamp last_started_time = 5;
+    // Last time the activity was completed.
     // If activity is not yet completed, this field will be null.
-    google.protobuf.Timestamp completed_time = 6;
+    google.protobuf.Timestamp last_completed_time = 6;
+    // Indicates that activity receive cancel signal.
+    bool cancel_requested = 14;
 }
--- a/temporal/api/enums/v1/activity_state.proto
+++ b/temporal/api/enums/v1/activity_state.proto
@@ -1,0 +1,54 @@
+// The MIT License
+//
+// Copyright (c) 2024 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
+
+package temporal.api.enums.v1;
+
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ActivityStateProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+// More detailed state of an activity.
+enum ActivityState {
+    ACTIVITY_STATE_UNSPECIFIED = 0;
+    // Activity was scheduled but not yet started.
+    ACTIVITY_STATE_SCHEDULED = 1;
+    // Activity has started executing.
+    ACTIVITY_STATE_STARTED = 2;
+    // Activity was requested to be canceled.
+    ACTIVITY_STATE_CANCEL_REQUESTED  = 3;
+    // Activity failed and is backing off before retrying.
+    ACTIVITY_STATE_CANCEL_BACKING_OFF = 4;
+    // Activity was paused.
+    ACTIVITY_STATE_PAUSED = 5;
+    // Activity has completed successfully.
+    ACTIVITY_STATE_COMPLETED = 6;
+    // Activity has failed.
+    ACTIVITY_STATE_FAILED = 7;
+    // Activity has timed out.
+    ACTIVITY_STATE_TIMED_OUT = 8;
+}
+
--- a/temporal/api/enums/v1/activity_state.proto
+++ b/temporal/api/enums/v1/activity_state.proto
@@ -31,24 +31,24 @@ option java_outer_classname = "ActivityStateProto";
 option ruby_package = "Temporalio::Api::Enums::V1";
 option csharp_namespace = "Temporalio.Api.Enums.V1";
 
-// More detailed state of an activity.
+// Detailed state of an activity for both pending and completed activities.
 enum ActivityState {
     ACTIVITY_STATE_UNSPECIFIED = 0;
     // Activity was scheduled but not yet started.
     ACTIVITY_STATE_SCHEDULED = 1;
     // Activity has started executing.
-    ACTIVITY_STATE_STARTED = 2;
-    // Activity was requested to be canceled.
-    ACTIVITY_STATE_CANCEL_REQUESTED  = 3;
+    ACTIVITY_STATE_RUNNING = 2;
     // Activity failed and is backing off before retrying.
-    ACTIVITY_STATE_CANCEL_BACKING_OFF = 4;
+    ACTIVITY_STATE_BACKING_OFF = 3;
     // Activity was paused.
-    ACTIVITY_STATE_PAUSED = 5;
+    ACTIVITY_STATE_PAUSED = 4;
     // Activity has completed successfully.
-    ACTIVITY_STATE_COMPLETED = 6;
+    ACTIVITY_STATE_COMPLETED = 5;
     // Activity has failed.
-    ACTIVITY_STATE_FAILED = 7;
+    ACTIVITY_STATE_FAILED = 6;
     // Activity has timed out.
-    ACTIVITY_STATE_TIMED_OUT = 8;
+    ACTIVITY_STATE_TIMED_OUT = 7;
+    // Activity has been cancelled.
+    ACTIVITY_STATE_CANCELLED = 8;
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1715,14 +1715,14 @@ message DescribeActivityByIdRequest {
     // Run ID of the workflow which scheduled this activity.
     // Can be empty, in this case the latest workflow will be used.
     string run_id = 3;
-    // Id of the activity we're updating.
+    // Id of the activity we're describing.
     string activity_id = 4;
 
-    // The identity of the client that initiated this request,
+    // The identity of the client that initiated this request.
     string identity = 5;
 }
 
 message DescribeActivityByIdResponse {
-    // Activity options.
+    // Activity information.
     temporal.api.activity.v1.ActivityInfo activity_info = 1;
 }
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1706,3 +1706,26 @@ message UpdateActivityOptionsByIdResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
 }
+
+message DescribeActivityByIdRequest {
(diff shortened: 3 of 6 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,16 @@ service WorkflowService {
         };
     }
 
+    // DescribeActivityById is called by the client to get activity options and info
+    // for started or scheduled activities.
```
bergundy, line 875: Matches the terminology elsewhere:

```suggestion
    // for pending activities.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,26 @@ message UpdateActivityOptionsByIdResponse {
...
 }
+
+message DescribeActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity.
+    // Can be empty, in this case the latest workflow will be used.
+    string run_id = 3;
```
bergundy, line 1717: nit: can use `WorkflowExecution` here.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,26 @@ message UpdateActivityOptionsByIdResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
 }
+
+message DescribeActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity.
```
bergundy, line 1713:
```suggestion
    // ID of the workflow which scheduled this activity.
```

Here and everywhere else.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,26 @@ message UpdateActivityOptionsByIdResponse {
...
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity.
+    // Can be empty, in this case the latest workflow will be used.
+    string run_id = 3;
+    // Id of the activity we're updating.
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request,
```
bergundy, line 1721:
```suggestion
    // The identity of the client that initiated this request.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,26 @@ message UpdateActivityOptionsByIdResponse {
...
+    string identity = 5;
+}
+
+message DescribeActivityByIdResponse {
+    // Activity options after an update.
```
bergundy, line 1726: What do you mean by "after an update"? Doesn't seem to fit here.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: It was a longer then needed discussion, and we decided agains it.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: copy/paste. will remove

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,15 @@ service WorkflowService {
         };
     }
 
+    // DescribeActivityById is called by the client to get activity options and info for pending activities.
```
cretz, line 874: > and info for pending activities.

Can you clarify the behavior of this call if the activity is complete? Does it succeed? Is `pending_activity_info` set? If it does succeed and that info is not set, is the user expected to crawl history to find the activity events to get some of the details obtainable on pending info?

It is unfortunate that pending info is just for pending, but it has valuable stuff even for non-pending we may want to make available at the top level here or in a more general purpose "activity info" that is unrelated to its state.

ychebotarev: So far it was for pending activities only (those that are scheduled, started, or waiting between retries). And if activity is finished this call will fail.
I never saw.a request "what was a retry policy/timers" for an activity that is complete.
If we have this scenario - lets talk. I guess it is doable, we can use event history to populate such info.

cretz: Yeah, it might surprise a user calling this that at one moment the call works and another it stops working just because the activity finished in the meantime. I could just call describe workflow if I wanted only the pending activities. And arguably I should be able to get the activity options on those.

> I never saw.a request "what was a retry policy/timers" for an activity that is complete.

Describe workflow will provide some of this information for you regardless of whether workflow is complete. I think describe should always work if the item exists.

It's really unfortunate that the "pending activity info" structure has all sorts of good information but is named that. I am wondering if we should have a general purpose activity info that has everything from pending that applies to complete too. Otherwise, users might be expected to crawl history events once this RPC stops. I am struggling on whether pending activity info should still be visible for pending activities.

I do think the activity options should be in the pending activity info, not just here (so people doing describe workflow can see them too). So your options may be:

1) General data and pending data

```proto
message DescribeActivityByIdResponse {
    // Always present
    temporal.api.activity.v1.ActivityInfo info = 1;

    // Only if activity not complete
    temporal.api.workflow.v1.PendingActivityInfo pending_info = 2;
}
```

But this has a lot of duplicate data. Or:

2) Oneof

```proto
message DescribeActivityByIdResponse {
    oneof info {
        temporal.api.workflow.v1.CompletedActivityInfo completed = 1;
        temporal.api.workflow.v1.PendingActivityInfo pending = 2;
    }
}
```

But that's a bit hard on users to use. Or:

3) Leave pending info only for workflow describe

```proto
message DescribeActivityByIdResponse {
    temporal.api.activity.v1.ActivityInfo info = 1;
}
```

They all have unfortunate tradeoffs. May need to run this by developer experience.

ychebotarev: Discussed offline. I will add more info in a restricted way, but so we can add more if needed later

ychebotarev: @cretz I updated the PR, and add info we discussed.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,23 @@ message UpdateActivityOptionsByIdResponse {
...
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // Can be empty, in this case the latest workflow will be used.
+    string run_id = 3;
+    // Id of the activity we're updating.
+    string activity_id = 4;
+
+    // The identity of the client that initiated this request,
```
bergundy, line 1721:
```suggestion
    // The identity of the client that initiated this request.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,23 @@ message UpdateActivityOptionsByIdResponse {
...
+
+message DescribeActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // Can be empty, in this case the latest workflow will be used.
+    string run_id = 3;
+    // Id of the activity we're updating.
```
bergundy, line 1718:
```suggestion
    // ID of the activity we're updating.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,23 @@ message UpdateActivityOptionsByIdResponse {
...
+    string activity_id = 4;
+
+    // The identity of the client that initiated this request,
+    string identity = 5;
+}
+
+message DescribeActivityByIdResponse {
+    // Activity options.
```
bergundy, line 1726: These aren't options.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,15 @@ service WorkflowService {
         };
     }
 
+    // DescribeActivityById is called by the client to get activity and for pending and completed activities.
```
bergundy, line 874: This would require us to scan the entire workflow history to obtain the activity information. I think it's probably fine but wondering how we want to price this given that it's not a cheap operation anymore.

ychebotarev: I think it is not different from 'temporal workflow show' that shows the history. I'm also agains charging for "get info" operations.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: renamed

temporal/api/workflowservice/v1/service.proto
bergundy: After internal discussion and thinking more about it I tend to agree.
If this becomes a perf problem, we can add an index in mutable state to quickly locate the event ID and avoid a full history scan.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1706,3 +1706,23 @@ message UpdateActivityOptionsByIdResponse {
...
+
+message DescribeActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // Can be empty, in this case the latest workflow will be used.
+    string run_id = 3;
+    // Id of the activity we're describing.
```
bergundy, line 1718:
```suggestion
    // ID of the activity we're describing.
```

temporal/api/enums/v1/activity_state.proto
```diff
@@ -0,0 +1,54 @@
...
+    // Activity was requested to be canceled.
+    ACTIVITY_STATE_CANCEL_REQUESTED  = 3;
+    // Activity failed and is backing off before retrying.
+    ACTIVITY_STATE_CANCEL_BACKING_OFF = 4;
+    // Activity was paused.
+    ACTIVITY_STATE_PAUSED = 5;
+    // Activity has completed successfully.
+    ACTIVITY_STATE_COMPLETED = 6;
+    // Activity has failed.
+    ACTIVITY_STATE_FAILED = 7;
```
bergundy, line 50: There's also an `ACTIVITY_STATE_CANCELED` state corresponding to the ActivityTaskCanceled event.

temporal/api/activity/v1/message.proto
```diff
@@ -64,4 +67,23 @@ message ActivityOptions {
...
+    temporal.api.common.v1.ActivityType type = 2;
+    // ActivityState is the current state of the activity.
+    temporal.api.enums.v1.ActivityState state = 3;
+    // ScheduledTime is the time last the activity was scheduled. .
+    google.protobuf.Timestamp scheduled_time = 4;
+    // StartedTime is the time last the activity was started.
+    google.protobuf.Timestamp started_time = 5;
+    // CompletedTime is the time last the activity was completed.
+    // If activity is not yet completed, this field will be null.
+    google.protobuf.Timestamp completed_time = 6;
```
bergundy, line 88: No need to use Go style comments. You can remove the name of the field from the docstring.

temporal/api/activity/v1/message.proto
```diff
@@ -64,4 +67,23 @@ message ActivityOptions {
...
+// It is used to report activity state to the users.
+// It is expected to return information about pending activities and completed activities.
+message ActivityInfo {
+    // ActivityOptions contains the options that were used to operate the activity.
+    temporal.api.activity.v1.ActivityOptions options = 1;
+    // ActivityType is the type of the activity.
+    temporal.api.common.v1.ActivityType type = 2;
+    // ActivityState is the current state of the activity.
+    temporal.api.enums.v1.ActivityState state = 3;
+    // ScheduledTime is the time last the activity was scheduled. .
```
bergundy, line 82:
```suggestion
    // The time last the activity was scheduled.
```

bergundy: I would have a field for the current / next attempt schedule time and the initial schedule time (the time of the ActivityTaskScheduled event).

bergundy: And clarify in each of these fields what it's referring to.

temporal/api/activity/v1/message.proto
```diff
@@ -64,4 +67,23 @@ message ActivityOptions {
     google.protobuf.Duration heartbeat_timeout = 5;
 
     temporal.api.common.v1.RetryPolicy retry_policy = 6;
+}
+
+// ActivityInfo contains information about an activity.
+// It is used to report activity state to the users.
+// It is expected to return information about pending activities and completed activities.
```
bergundy, line 74:
```suggestion
// Information about an activity in any state including pending and completed activities.
```

temporal/api/enums/v1/activity_state.proto
```diff
@@ -0,0 +1,54 @@
...
+package temporal.api.enums.v1;
+
+option go_package = "go.temporal.io/api/enums/v1;enums";
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ActivityStateProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+// More detailed state of an activity.
```
bergundy, line 34:
```suggestion
// Detailed state of an activity for both pending and completed activities.
```

temporal/api/workflowservice/v1/service.proto
bergundy, line 874: Docstring can be improved here.

Please also mention that an activity ID isn’t unique for an entire workflow run only for pending activities and that the API will return the latest matching activity for the given ID. May also want to consider letting a user use the scheduled event ID in some cases because that is unique for the entire run.

temporal/api/enums/v1/activity_state.proto
```diff
@@ -0,0 +1,54 @@
...
+enum ActivityState {
+    ACTIVITY_STATE_UNSPECIFIED = 0;
+    // Activity was scheduled but not yet started.
+    ACTIVITY_STATE_SCHEDULED = 1;
+    // Activity has started executing.
+    ACTIVITY_STATE_STARTED = 2;
+    // Activity was requested to be canceled.
+    ACTIVITY_STATE_CANCEL_REQUESTED  = 3;
```
cretz, line 42: I am a little concerned about this as a state. I think this is better as a boolean on the `ActivityInfo`. While I understand it's a state and not an event (in the sense that you can't send a second cancel), I think there's two benefits to a bool on the info:

1. Because this is really `ACTIVITY_STATE_RUNNING_AND_CANCEL_REQUESTED` which is a bit confusing I can't just check for "running" to see if running
2. There is value in even knowing a cancel was requested regardless of whether swallowed (unsure we have this information for completed activities though)

I am curious about a behavior though. What if I have an activity that is requested to cancel but ignores/swallows it returns a general activity failure? Does it retry? Or is it marked failed? I will ask internally because it affects my opinion here.

cretz (commented):
LGTM overall, just a bit of concern on the new activity state enum

temporal/api/workflowservice/v1/service.proto
cretz: Thanks! Will close this thread.

temporal/api/enums/v1/activity_state.proto
ychebotarev: will add it

temporal/api/enums/v1/activity_state.proto
ychebotarev: I add it because it was there in pending activity state. I agree - I will remove the state and add a flag to activity info. It will be true only for pending activities that received cancel request.

temporal/api/enums/v1/activity_state.proto
```diff
@@ -0,0 +1,54 @@
...
+option java_package = "io.temporal.api.enums.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ActivityStateProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+// More detailed state of an activity.
+enum ActivityState {
+    ACTIVITY_STATE_UNSPECIFIED = 0;
+    // Activity was scheduled but not yet started.
```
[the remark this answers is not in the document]
ychebotarev: we have a separate flag for "backing off". state transition is "backing off"=>"scheduled"-"started"->"completed" or "backing off"

temporal/api/enums/v1/activity_state.proto
```diff
@@ -0,0 +1,54 @@
...
+    ACTIVITY_STATE_SCHEDULED = 1;
+    // Activity has started executing.
+    ACTIVITY_STATE_STARTED = 2;
```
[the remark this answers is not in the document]
ychebotarev: make sense to me

temporal/api/workflowservice/v1/service.proto
ychebotarev: Add more info. I will not add 'event id' as a routing parameter. It if comes to that users can as well read the history.

temporal/api/activity/v1/message.proto
ychebotarev: I rename the fields to have "last_" prefix.
I want to keep only the essential fields in this iteration.
If users need more it will be easy to add them later.

temporal/api/workflowservice/v1/service.proto
ychebotarev: I modify the docstring. At this point I don't want to use scheduled event ID as a routing parameter. For this they can get history

temporal/api/enums/v1/activity_state.proto
```diff
@@ -0,0 +1,54 @@
...
+    // Activity was requested to be canceled.
+    ACTIVITY_STATE_CANCEL_REQUESTED  = 3;
+    // Activity failed and is backing off before retrying.
+    ACTIVITY_STATE_CANCEL_BACKING_OFF = 4;
```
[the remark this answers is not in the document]
bergundy: Please don't resolve comments. I keep seeing my suggestions disappear on PRs and not always being addressed or commented on.

temporal/api/workflowservice/v1/service.proto
bergundy: Works for me, just call out the behavior in the docstring please and make sure you test it when you get to the implementation.

temporal/api/activity/v1/message.proto
ychebotarev: done

temporal/api/enums/v1/activity_state.proto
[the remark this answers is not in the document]
cretz: I am surprised it starts as backing off, but otherwise makes sense.

temporal/api/enums/v1/activity_state.proto
```diff
@@ -0,0 +1,54 @@
...
+    // Activity was paused.
+    ACTIVITY_STATE_PAUSED = 4;
+    // Activity has completed successfully.
+    ACTIVITY_STATE_COMPLETED = 5;
+    // Activity has failed.
+    ACTIVITY_STATE_FAILED = 6;
+    // Activity has timed out.
+    ACTIVITY_STATE_TIMED_OUT = 7;
+    // Activity has been cancelled.
+    ACTIVITY_STATE_CANCELLED = 8;
```
cretz, line 52:
```suggestion
    ACTIVITY_STATE_CANCELED = 8;
```

We use the one-L form in the history message name.

cretz (commented):
Only a single-character change suggested, otherwise LGTM. I do think we will want to do most of the implementation before merging.

temporal/api/enums/v1/activity_state.proto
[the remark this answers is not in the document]
ychebotarev: tolked offline. I'l be resolving only after I push new version

ychebotarev:
Closing this one. I will require history service to execute unbound operation (reading the full history). Our design choice is to not allow this to happen.

PR #464, opened 2024-10-16
Add running_workflows to ScheduleListInfo, for ListSchedules

lina-temporal:
**What changed?**
- A new field `RunningWorkflows` was added to `ScheduleListInfo`, matching the field in `ScheduleInfo`.

**Why?**
- Customers would like to be able to drill down into running workflow executions without having to follow a `ListSchedule` call with fan-outs to `DescribeSchedule`).

**Breaking changes**
- None here

**Server PR**
- Pending

### DIFF
--- a/temporal/api/schedule/v1/message.proto
+++ b/temporal/api/schedule/v1/message.proto
@@ -388,6 +388,12 @@ message ScheduleListInfo {
     // From info (maybe fewer entries):
     repeated ScheduleActionResult recent_actions = 5;
     repeated google.protobuf.Timestamp future_action_times = 6;
+
+    // From state:
+    // Running workflows returned here are eventually consistent, and their
+    // status may be out-of-date. For a strongly consistent view of a schedule's
+    // running workflows, use the `DescribeSchedule` API.
+    repeated temporal.api.common.v1.WorkflowExecution running_workflows = 7;
 }
 
 // ScheduleListEntry is returned by ListSchedules.
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -10081,6 +10081,14 @@
             "type": "string",
             "format": "date-time"
           }
+        },
+        "runningWorkflows": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "$ref": "#/definitions/v1WorkflowExecution"
+          },
+          "description": "From state:\nRunning workflows returned here are eventually consistent, and their\nstatus may be out-of-date. For a strongly consistent view of a schedule's\nrunning workflows, use the `DescribeSchedule` API."
         }
       },
       "description": "ScheduleListInfo is an abbreviated set of values from Schedule and ScheduleInfo\nthat's returned in ListSchedules."
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -7528,6 +7528,15 @@ components:
           items:
             type: string
             format: date-time
+        runningWorkflows:
+          type: array
+          items:
+            $ref: '#/components/schemas/WorkflowExecution'
+          description: |-
+            From state:
+             Running workflows returned here are eventually consistent, and their
+             status may be out-of-date. For a strongly consistent view of a schedule's
+             running workflows, use the `DescribeSchedule` API.
       description: |-
         ScheduleListInfo is an abbreviated set of values from Schedule and ScheduleInfo
          that's returned in ListSchedules.


Quinn-With-Two-Ns (approved):

temporal/api/schedule/v1/message.proto
```diff
@@ -388,6 +388,12 @@ message ScheduleListInfo {
     // From info (maybe fewer entries):
     repeated ScheduleActionResult recent_actions = 5;
     repeated google.protobuf.Timestamp future_action_times = 6;
+
+    // From state:
+    // Running workflows returned here are eventually consistent, and their
+    // status may be out-of-date. For a strongly consistent view of a schedule's
+    // running workflows, use the `DescribeSchedule` API.
+    repeated temporal.api.common.v1.WorkflowExecution running_workflows = 7;
```
dnr, line 396: 1. This goes in the memo so I think we should keep it as short as possible.
2. Part of the request was to include the close state of recent workflows, right? Is that going to be another change?

I'm thinking that for the close state, we can put it in ScheduleActionResult and fill it in when we know it.

For running workflows... since we don't track allowall anymore, there will be 0 or 1 running workflows, and if it's 1, it'll be the most recent action. So instead of repeating the wf id and run id from recent_actions here, we can just use a flag in the ScheduleActionResult for whether it's running or not? But actually the close state enum is just that: if it's not set yet, then it's still running, if it's set to completed/failed/timedout/cancelled/terminated, it's not running.

(If we do track running workflows from allowall, then that's not enough since there could be an old one still running. But we have no plans to do that right now so maybe we can wait until we do.)

lina-temporal: > Part of the request was to include the close state of recent workflows, right? Is that going to be another change?

Discussed during standup, I'll do this as part of this change.

> I'm thinking that for the close state, we can put it in ScheduleActionResult and fill it in when we know it.

Sounds good, will do!

> So instead of repeating the wf id and run id from recent_actions here, we can just use a flag in the ScheduleActionResult for whether it's running or not?

Sounds good in terms of state, I'll update. For the API, however, that would make the `ListSchedulesResult` distinct compared to `DescribeSchedule` (which has the separate `running_workflows` field), which doesn't seem great from a UX perspective; what do you think about populating a `running_workflows` field  in API responses based on the flag in our mutable state memo's `ScheduleActionResult`?

dnr: For the reverse question, the presence of the new field in Describe: If we just add the extra field to the action result and fill it in in the state, it'll automatically come out in Describe (query) as well as List, and I think that's good, it's slightly more information than is there now (and doesn't need any versioning in the workflow).

For your actual question, about a running_workflows list in List: I'm concerned that we won't always have enough information to fill it in. E.g. if we use the state machine impl and have it track a limited number of concurrent workflows that aren't the latest (from AllowAll), and supply them in Describe, that's great, but we might not be able to fit them in the memo (state machines don't help there at all), so we'd still be inconsistent. Except then also misleading.

Or we might put a bunch in the memo, but only as int timestamps (to save space) and then reconstruct the workflow ids. But we wouldn't have the run ids.

So I'm leaning towards no, just the enum in recents in list results.

cretz (approved):
I like this solution much better

bergundy (approved):

temporal/api/schedule/v1/message.proto
lina-temporal: @bergundy I like the structure suggestion. I think it's probably too late for the existing API, but maybe we could consider introducing a V2 API when we support different schedule action types?

bergundy: Works for me. I approved the PR already.

temporal/api/schedule/v1/message.proto
```diff
@@ -269,6 +270,10 @@ message ScheduleActionResult {
 
     // If action was start_workflow:
     temporal.api.common.v1.WorkflowExecution start_workflow_result = 11;
+
+    // If the action was start_workflow, this field will reflect an
+    // eventually-consistent view of the started workflow's status.
+    temporal.api.enums.v1.WorkflowExecutionStatus status = 12;
```
dnr, line 276: if we do with this extra field and not an embedded message, we should name it `workflow_status` or `started_workflow_status` so it's clear it's just for workflows, if there are more action types in the future.

lina-temporal: > That's fully backwards compatible at the proto level, though source code will have to change.

I'm happy to make the change now, as long as we're poking at this API.

lina-temporal: Ah, is the usual thing here to do a pair of buf-breaking toggle commits?

```
Run buf breaking changes check against master branch...
temporal/api/schedule/v1/message.proto:282:9:Field "11" with name "workflow" on message "ScheduleActionResult" changed option "json_name" from "startWorkflowResult" to "workflow".
temporal/api/schedule/v1/message.proto:282:9:Field "11" on message "ScheduleActionResult" moved from outside to inside a oneof.
temporal/api/schedule/v1/message.proto:282:9:Field "11" on message "ScheduleActionResult" changed type from "temporal.api.common.v1.WorkflowExecution" to "temporal.api.schedule.v1.ScheduleActionResult.WorkflowExecutionWithStatus".
temporal/api/schedule/v1/message.proto:282:37:Field "11" on message "ScheduleActionResult" changed name from "start_workflow_result" to "workflow".
``` 

I think I'll do this as part of a separate follow-up change. I'll rename the new field to `started_workflow_status` for the moment.

dnr: We should pay attention to the breaking change warnings when they indicate actual problems. Specifically: we shouldn't rename the field, since that'll cause problems for the web UI and cli json output.

As I understand it, moving a single field from outside to inside a oneof is explicitly allowed, so I'm not sure why it's complaining about that. Although I would want to confirm that it shows up the same in json output. If not, then we can just not do the oneof and only do the message-extension thing.

dnr: Wait, so we're not doing the message-extension thing to include the status inside start_workflow_result? That's fine, it is a little tricky. But it means we can't then move these result fields into a oneof (since there's two of them for start workflow)

lina-temporal: If we do want to clean this up with a `oneof`, I'd rather do that as part of another change, since this is tripping `buf-breaking`.

dnr: I mean if we do add the new field now, then we can't do the oneof thing later (unless we populate both, which is weird). So we kind of have to pick now

lina-temporal: I'm lost; it looks to me like we can't do the `oneof` change at all, if it renames fields and causes `buf-breaking` to trip. 

> temporal/api/schedule/v1/message.proto:282:9:Field "11" with name "workflow" on message "ScheduleActionResult" changed option "json_name" from "startWorkflowResult" to "workflow".

So, we want a `oneof` variant`WorkflowExecutionWithStatus` message with the field `startWorkflowResult`, or there's some trick we're going to use to rename the field to `result`? I get that this is compatible at proto level, I'm just not seeing how it's compatible at JSON level.

dnr: I meant if the oneof option itself is named `start_workflow_result` (instead of `workflow`) then maybe it would be json-compatible? But I'm not sure, I don't know the json rules as well as the proto wire rules. If that doesn't work without buf-breaking complaints then I think we should just go with the separate field and be done with it.

i.e.
```proto
message ScheduleActionResult {
    ...

    message WorkflowExecutionWithStatus {  // superset of common.WorkflowExecution
        string workflow_id = 1;
        string run_id = 2;
        temporal.api.enums.v1.WorkflowExecutionStatus status = 3;
    }

    oneof variant {
        // If action was start_workflow:
        WorkflowExecutionWithStatus start_workflow_result = 11;
    }
}
```

lina-temporal: `buf-breaking`'s still unhappy, even with the same field name, unfortunately;

```
temporal/api/schedule/v1/message.proto:279:9:Field "11" on message "ScheduleActionResult" moved from outside to inside a oneof.
temporal/api/schedule/v1/message.proto:279:9:Field "11" on message "ScheduleActionResult" changed type from "temporal.api.common.v1.WorkflowExecution" to "temporal.api.schedule.v1.ScheduleActionResult.WorkflowExecutionWithStatus".
```

dnr (approved):
Ok, this is good then

PR #465, opened 2024-10-21
Add PauseActivity API

ychebotarev:
**What changed?**
Add PauseActivity API

**Why?**
Part of Activity API work

**[Server PR](https://github.com/temporalio/temporal/pull/6685)**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1718,3 +1718,21 @@ message UpdateActivityOptionsByIdResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
 }
+
+message PauseActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
+    // Id of the activity we're updating.
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request.
+    string identity = 5;
+}
+
+message PauseActivityByIdResponse {
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1696,12 +1696,12 @@ message ExecuteMultiOperationResponse {
 message UpdateActivityOptionsByIdRequest {
     // Namespace of the workflow which scheduled this activity
     string namespace = 1;
-    // Id of the workflow which scheduled this activity
+    // ID of the workflow which scheduled this activity
     string workflow_id = 2;
-    // Run Id of the workflow which scheduled this activity
+    // Run ID of the workflow which scheduled this activity
     // if empty - latest workflow is used
     string run_id = 3;
-    // Id of the activity we're updating
+    // ID of the activity we're updating
     string activity_id = 4;
 
     // The identity of the client who initiated this request
@@ -1722,12 +1722,12 @@ message UpdateActivityOptionsByIdResponse {
 message PauseActivityByIdRequest {
     // Namespace of the workflow which scheduled this activity.
     string namespace = 1;
-    // Id of the workflow which scheduled this activity.
+    // ID of the workflow which scheduled this activity.
     string workflow_id = 2;
-    // Run Id of the workflow which scheduled this activity.
+    // Run ID of the workflow which scheduled this activity.
     // If empty - latest workflow is used.
     string run_id = 3;
-    // Id of the activity we're updating.
+    // ID of the activity we're updating.
     string activity_id = 4;
 
     // The identity of the client who initiated this request.
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -871,5 +871,19 @@ service WorkflowService {
         };
     }
 
+    // PauseActivityById is called by the client to pause the execution of an activity
+    // (-- api-linter: core::0136::prepositions=disabled
+    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
+    rpc PauseActivityById (PauseActivityByIdRequest) returns (PauseActivityByIdResponse)
+    {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/activities/pause-by-id"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/activities/pause-by-id"
+                body: "*"
+            }
+        };
+    }
 }
 
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -859,8 +859,7 @@ service WorkflowService {
     // UpdateActivityOptionsById is called by the client to update the options of an activity
     // (-- api-linter: core::0136::prepositions=disabled
     //     aip.dev/not-precedent: "By" is used to indicate request type. --)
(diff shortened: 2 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1718,3 +1718,21 @@ message UpdateActivityOptionsByIdResponse {
...
+message PauseActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // Id of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run Id of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
+    // Id of the activity we're updating.
+    string activity_id = 4;
```
bergundy, line 1731: Use ID in English and Go please.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,19 @@ service WorkflowService {
         };
     }
 
+    // PauseActivityById is called by the client to pause the execution of an activity
+    // (-- api-linter: core::0136::prepositions=disabled
+    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
+    rpc PauseActivityById (PauseActivityByIdRequest) returns (PauseActivityByIdResponse)
+    {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/activities/pause-by-id"
```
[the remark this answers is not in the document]
ychebotarev: it was intentionally put into `/namespaces/{namespace}/activities` after  some discussions.
This way:
* it indicates that is is different from workflow commands
* it is consistent with other activity-related rpcs that has the same "routing" and "by-id" signature

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,19 @@ service WorkflowService {
         };
     }
 
+    // PauseActivityById is called by the client to pause the execution of an activity
+    // (-- api-linter: core::0136::prepositions=disabled
+    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
+    rpc PauseActivityById (PauseActivityByIdRequest) returns (PauseActivityByIdResponse)
+    {
```
cretz, line 878:
```suggestion
    rpc PauseActivityById (PauseActivityByIdRequest) returns (PauseActivityByIdResponse) {
```

(same comment for RPC above made at https://github.com/temporalio/api/pull/450#discussion_r1742493792

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,19 @@ service WorkflowService {
         };
     }
 
+    // PauseActivityById is called by the client to pause the execution of an activity
+    // (-- api-linter: core::0136::prepositions=disabled
+    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
+    rpc PauseActivityById (PauseActivityByIdRequest) returns (PauseActivityByIdResponse)
```
cretz, line 877: Can you clarify what the behavior of pause is here? Specifically, what happens if already paused, and what happens to activities (what happens if running and what happens if backing off).

cretz, line 877: I am not sure this should be added independent of the ability to unpause/resume. I think we would need to see both to ensure they are reasonable reciprocals. And I also wonder if we want to see reset and others at this same time too so we can evaluate how they all work/look together.

Similarly, I wonder if these are really all separate RPC calls instead of one call to update the state of the activity.

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
cretz: 👍 This matches other `*Activity*ById` RPC calls in this service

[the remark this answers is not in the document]
bergundy: I wonder if we should change this. For example if we have a standalone activity one day, we would want different routes for that.

temporal/api/workflowservice/v1/service.proto
bergundy: I tend to agree that it would be better to see all of the Activity APIs as a whole.

ychebotarev: Pause definitions will not change, and it doesn't depend on unpause (and even less on reset). I want to start working on it.

ychebotarev: as for the idea to combine then - separated they are much cleaner for me, not sure what we are getting by combining them into a single command besides one more layer of indirection.

ychebotarev: add more comments

temporal/api/workflowservice/v1/service.proto
ychebotarev: done

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
cretz: The RPC calls are generally named, they aren't called `WorkflowActivity`, so if we came with non-workflow-related activity calls, we'd have completely different RPCs and they could have different routes than these. Regardless if we want to change this, not sure this PR is where it would happen, so may be deserving of a general discussion separately.

temporal/api/workflowservice/v1/service.proto
cretz: Reducing the API surface area does have value if they do a common, mutually exclusive thing. I think it'll be easier to tell how the API looks once all together. You can probably start work without this being in `master`. That's often what we suggest for many of these types of things where we often learn things about how we want the API during implementation of the API.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,29 @@ service WorkflowService {
...
+    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    //
+    // Pausing an activity means:
+    // - If the activity is currently waiting for a retry or is running and subsequently fails,
+    //   it will not be rescheduled until it is unpaused.
+    // - If the activity is already paused, calling this method will have no effect.
+    // - If the activity is running and will finish successfully, activity will be completed.
+    //
+    // For long-running activities:
+    // - Pausing will send a cancellation with "paused" as the reason.
```
cretz, line 883: `RecordActivityTaskHeartbeatResponse` doesn't support a reason for canceled, but do you mean this is somewhere in the event?

cretz, line 883: I don't disagree with this behavior, but I just want to confirm there is not a use case at this time to _not_ cancel the activity task.

cretz, line 883: Do we need to update pending activity info? Is there any way for a user to know an activity is paused? I am mostly curious if an activity is considered "paused" if it is still finishing up its task after pause is sent. I am also curious if we are going to change the definition of some of the fields there regarding next retry.

temporal/api/workflowservice/v1/service.proto
Sushisource: +1 - we have a habit of making API definitions that don't do a great job of combining the shared values for operations operating on the same entity. I think I mentioned this in a previous review about `UpdateActivityOptionsById`. Having a more generic `UpdateActivityStateById` that has the identical first 5 fields (namespace - identity) and then an `operation` `oneof` feels quite a bit more natural to me.

I know sometimes we have to avoid that for routing reasons (some proxy or whatever needs to be able to avoid inspecting the proto) which makes sense, but, don't think that applies here.

ychebotarev:
closing this per discussion. will create new one in feature branch

PR #466, opened 2024-10-21
UpdateWorkflowExecutionOptions API

carlydf:
- Add `UpdateWorkflowExecutionOptions ` individual and batch API
- Add `WorkflowExecutionOptionsUpdatedEventAttributes` to track requested update and resulting `WorkflowExecutionOptions`

To provide tools for users to get out of bad situations when using worker versioning

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -102,3 +102,13 @@ message BatchOperationReset {
   temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
 
 }
+
+// BatchOperationModifyWorkflowProperties sends ModifyWorkflowExecutionProperties requests to batch workflows.
+// Keep the parameters in sync with temporal.api.workflowservice.v1.ModifyWorkflowExecutionPropertiesRequest.
+message BatchOperationModifyWorkflowProperties {
+  // The identity of the worker/client.
+  string identity = 1;
+
+  // Options to specify in both single-workflow modify and batch modify
+  temporal.api.common.v1.ModifyWorkflowExecutionPropertiesOptions options = 2;
+}
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -120,9 +120,4 @@ message BatchOperationUpdateWorkflowExecutionOptions {
   // Controls which fields from `workflow_execution_options` will be applied.
   // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
   google.protobuf.FieldMask update_mask = 4;
-
-  // If true and if FieldMask mutates versioning behavior, the update will be conditional
-  // on a successful replay test of the workflow execution's history on the new worker it
-  // would need to replay on after the change. Not yet implemented.
-  bool force_replay_test_on_version_behavior_change = 5;
 }
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -103,9 +103,9 @@ message BatchOperationReset {
 
 }
 
-// BatchOperationModifyWorkflowProperties sends UpdateWorkflowExecutionOptions requests to batch workflows.
+// BatchOperationUpdateWorkflowExecutionOptions sends UpdateWorkflowExecutionOptions requests to batch workflows.
 // Keep the parameters in sync with temporal.api.workflowservice.v1.UpdateWorkflowExecutionOptionsRequest.
-message BatchOperationModifyWorkflowProperties {
+message BatchOperationUpdateWorkflowExecutionOptions {
   // The identity of the worker/client.
   string identity = 1;
 
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -103,12 +103,12 @@ message BatchOperationReset {
 
 }
 
-// BatchOperationModifyWorkflowProperties sends ModifyWorkflowExecutionProperties requests to batch workflows.
-// Keep the parameters in sync with temporal.api.workflowservice.v1.ModifyWorkflowExecutionPropertiesRequest.
+// BatchOperationModifyWorkflowProperties sends UpdateWorkflowExecutionOptions requests to batch workflows.
+// Keep the parameters in sync with temporal.api.workflowservice.v1.UpdateWorkflowExecutionOptionsRequest.
 message BatchOperationModifyWorkflowProperties {
   // The identity of the worker/client.
   string identity = 1;
 
-  // Options to specify in both single-workflow modify and batch modify
-  temporal.api.common.v1.ModifyWorkflowExecutionPropertiesOptions options = 2;
+  // Config to specify update-options behavior in both single-workflow and batch mode.
+  temporal.api.common.v1.UpdateWorkflowExecutionOptionsConfig config = 2;
 }
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -111,20 +111,18 @@ message BatchOperationUpdateWorkflowExecutionOptions {
   // The identity of the worker/client.
   string identity = 1;
 
-  string reason = 2;
-
   // Used to de-dupe update-options requests
-  string request_id = 3;
+  string request_id = 2;
 
   // Workflow Execution options. Partial updates are accepted and controlled by update_mask.
-  temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options = 4;
+  temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options = 3;
 
   // Controls which fields from `workflow_execution_options` will be applied.
   // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
-  google.protobuf.FieldMask update_mask = 5;
+  google.protobuf.FieldMask update_mask = 4;
 
   // If true and if FieldMask mutates versioning behavior, the update will be conditional
   // on a successful replay test of the workflow execution's history on the new worker it
   // would need to replay on after the change. Not yet implemented.
-  bool force_replay_test_on_version_behavior_change = 6;
+  bool force_replay_test_on_version_behavior_change = 5;
 }
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -36,6 +36,8 @@ import "google/protobuf/timestamp.proto";
 import "temporal/api/common/v1/message.proto";
 import "temporal/api/enums/v1/batch_operation.proto";
 import "temporal/api/enums/v1/reset.proto";
+import "temporal/api/workflow/v1/message.proto";
+import "google/protobuf/field_mask.proto";
 
 message BatchOperationInfo {
   // Batch job ID
@@ -109,6 +111,20 @@ message BatchOperationUpdateWorkflowExecutionOptions {
   // The identity of the worker/client.
   string identity = 1;
 
-  // Config to specify update-options behavior in both single-workflow and batch mode.
-  temporal.api.common.v1.UpdateWorkflowExecutionOptionsConfig config = 2;
+  string reason = 2;
+
+  // Used to de-dupe update-options requests
+  string request_id = 3;
+
+  // Workflow Execution options. Partial updates are accepted and controlled by update_mask.
+  temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options = 4;
+
+  // Controls which fields from `workflow_execution_options` will be applied.
+  // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
+  google.protobuf.FieldMask update_mask = 5;
+
+  // If true and if FieldMask mutates versioning behavior, the update will be conditional
+  // on a successful replay test of the workflow execution's history on the new worker it
+  // would need to replay on after the change. Not yet implemented.
+  bool force_replay_test_on_version_behavior_change = 6;
 }
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -256,9 +256,9 @@ message UpdateWorkflowExecutionOptionsConfig {
         // All future tasks will be scheduled on this build ID.
         google.protobuf.StringValue pinned_build_id_override = 1;
 
-        // If set to true, the workflow execution will become unpinned.
+        // If true, the workflow execution will become unpinned.
         // If the workflow is already unpinned, this is a no-op.
-        // If set to false, this is a no-op; use pinned_build_id_override to pin a workflow execution.
+        // If unset/false, this is a no-op; use pinned_build_id_override to pin a workflow execution.
         bool make_unpinned = 2;
 
         // If false, the update will be conditional on a successful
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -38,7 +38,6 @@ import "google/protobuf/wrappers.proto";
 import "temporal/api/enums/v1/common.proto";
 import "temporal/api/enums/v1/event_type.proto";
 import "temporal/api/enums/v1/reset.proto";
-import "temporal/api/enums/v1/workflow.proto";
 
 message DataBlob {
     temporal.api.enums.v1.EncodingType encoding_type = 1;
@@ -251,25 +250,25 @@ message Link {
     }
 }
 
-message ModifyWorkflowExecutionPropertiesOptions {
-    // If set, the workflow execution's current build ID will be set to this.
-    // If pinned, future tasks will be scheduled on this build ID.
-    // If unpinned, future tasks will be scheduled per the rollout config at
-    // the time of dispatch. Until the next task is dispatched, the unpinned
-    // workflow execution's current build ID is what determines it's reachability.
-    google.protobuf.StringValue build_id = 1;
+message UpdateWorkflowExecutionOptionsConfig {
+    message UpdateVersioningOptions {
+        // If set, the workflow execution will be pinned to this build ID.
+        // All future tasks will be scheduled on this build ID.
+        google.protobuf.StringValue pinned_build_id_override = 1;
 
-    // If set, the workflow execution's versioning behavior will be set to this.
-    temporal.api.enums.v1.VersioningBehavior versioning_behavior = 2;
+        // If set to true, the workflow execution will become unpinned.
+        // If the workflow is already unpinned, this is a no-op.
+        // If set to false, this is a no-op; use pinned_build_id_override to pin a workflow execution.
+        bool make_unpinned = 2;
 
-    // If false, the property modification will be conditional on a successful
-    // replay test of the workflow execution's history on the new worker it
-    // would need to replay on after the change.
-    //
-    // In the first release, skip_replay_test=false will return an unimplemented error.
-    // User must explicitly pass skip_replay_test=true to use the API.
-    bool skip_replay_test = 3;
+        // If false, the update will be conditional on a successful
+        // replay test of the workflow execution's history on the new worker it
+        // would need to replay on after the change.
+        //
+        // In the first release, skip_replay_test=false will return an unimplemented error.
+        // User must explicitly pass skip_replay_test=true to use the API.
+        bool skip_replay_test = 3;
+    }
 
-    // If true, prints a diff of which properties would be changed without changing them.
-    // bool diff = 4;// omit this flag in initial release
+    UpdateVersioningOptions versioning_options = 1;
 }
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -189,9 +189,6 @@ message ResetOptions {
 
     // Event types not to be reapplied
     repeated temporal.api.enums.v1.ResetReapplyExcludeType reset_reapply_exclude_types = 12;
-
-    // If specified, atomically modify workflow properties with reset
-    temporal.api.common.v1.ModifyWorkflowExecutionPropertiesOptions options = 8;
 }
 
 // Callback to attach to various events in the system, e.g. workflow run completion.
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -33,11 +33,11 @@ option csharp_namespace = "Temporalio.Api.Common.V1";
 
 import "google/protobuf/duration.proto";
 import "google/protobuf/empty.proto";
-import "google/protobuf/wrappers.proto";
 
 import "temporal/api/enums/v1/common.proto";
 import "temporal/api/enums/v1/event_type.proto";
 import "temporal/api/enums/v1/reset.proto";
+import "temporal/api/enums/v1/workflow.proto";
 
 message DataBlob {
     temporal.api.enums.v1.EncodingType encoding_type = 1;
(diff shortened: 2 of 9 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1710,3 +1713,22 @@ message UpdateActivityOptionsByIdResponse {
...
+    // (-- api-linter: core::0203::optional=disabled
+    //     aip.dev/not-precedent: false positive triggered by the word "optional" --)
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 2;
+    string reason = 3;
+
+    // Used to de-dupe modify requests
+    string request_id = 4;
+
+    // Options to specify in both single-workflow modify and batch modify
+    temporal.api.common.v1.ModifyWorkflowExecutionPropertiesOptions options = 5;
```
cretz, line 1730: Hrmm, I wonder if we should just inline the fields here instead of trying to share with batch. We don't share options across batch and non-batch with other things (not even sure why we made a common `ResetOptions` tbh). But it's not that big of a deal.

temporal/api/common/v1/message.proto
```diff
@@ -248,3 +250,26 @@ message Link {
...
+    // If set, the workflow execution's versioning behavior will be set to this.
+    temporal.api.enums.v1.VersioningBehavior versioning_behavior = 2;
+
+    // If false, the property modification will be conditional on a successful
+    // replay test of the workflow execution's history on the new worker it
+    // would need to replay on after the change.
+    //
+    // In the first release, skip_replay_test=false will return an unimplemented error.
+    // User must explicitly pass skip_replay_test=true to use the API.
+    bool skip_replay_test = 3;
```
cretz, line 271: Hrmm, this is interesting. Has there been SDK team discussion on how this might work? We have never had a server-triggered replay before that I know of (that isn't expected to do anything else). If we want a concept of server-invoked workflow replay, it may need to be a general concept since I can see it having value in other situations. Also, since this writes an event, this will implicitly wake up the workflow to do its next task anyways, so is this a replay then an event that may trigger another replay? How are we adding a new event to show the build ID is updated during this replay, but asking the workflow to process it as if it was a replay?

Also, are we sure we want replay testing to be the default? For timeout and memo that already exist on the event, not sure they need replay. If we expect this to be general purpose, wouldn't most externally modified properties not really require replay?

Also not sure how this would work with memo. What if I used a memo update to trigger a local activity? How would a future memo inclusion into this modification do a replay?

I am wondering if replay testing may be unique to build ID modification. I may have to understand the details more.

temporal/api/common/v1/message.proto
```diff
@@ -248,3 +250,26 @@ message Link {
...
+
+    // If true, prints a diff of which properties would be changed without changing them.
+    // bool diff = 4;// omit this flag in initial release
```
cretz, line 274:
```suggestion
```

Can just remove from the API then IMO

temporal/api/common/v1/message.proto
```diff
@@ -248,3 +250,26 @@ message Link {
         BatchJob batch_job = 2;
     }
 }
+
+message ModifyWorkflowExecutionPropertiesOptions {
```
cretz, line 254: Do we want to support updating memo now too?

temporal/api/history/v1/message.proto
```diff
@@ -727,6 +719,15 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
     int64 started_event_id = 5;
 }
 
+message WorkflowPropertiesModifiedEventAttributes {
```
cretz, line 722: If moving this to be in order with how its used, may need to move below the next message

temporal/api/history/v1/message.proto
```diff
@@ -727,6 +719,15 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
...
+    // The `WORKFLOW_TASK_COMPLETED` event which this command was reported with
+    int64 workflow_task_completed_event_id = 1;
+    // If set, update the workflow memo with the provided values. The values will be merged with
+    // the existing memo. If the user wants to delete values, a default/empty Payload should be
+    // used as the value for the key being deleted.
+    temporal.api.common.v1.Memo upserted_memo = 2;
+}
+
 message WorkflowPropertiesModifiedExternallyEventAttributes {
```
cretz, line 731: I'm not sure any SDK supports this event today. I think there may need to be a discussion on how best to support from SDK POV (i.e. whether workflow author can react to this event explicitly), and I think we need to make sure `worker_may_ignore` does not get set to `true` for this event.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1710,3 +1713,22 @@ message UpdateActivityOptionsByIdResponse {
...
+}
+
+message ModifyWorkflowExecutionPropertiesResponse {
```
cretz, line 1733: Ideally mutating properties would also return the now-mutated full set if possible

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +871,18 @@ service WorkflowService {
         };
     }
 
+    // ModifyWorkflowExecutionProperties will modify the properties of an existing workflow execution,
+    // starting with just build_id and versioning_behavior, but eventually other things.
+    rpc ModifyWorkflowExecutionProperties (ModifyWorkflowExecutionPropertiesRequest) returns (ModifyWorkflowExecutionPropertiesResponse) {
```
cretz, line 876: Part of me wonders if this is a more generic `UpdateWorkflowExecution`

temporal/api/common/v1/message.proto
mfateev: Should we group versioning changes to its own structure? 

I think it should be more like force-change-on-replay-failure flag. And another option is to "force-replay-on-version-behaviour-change".

temporal/api/workflowservice/v1/request_response.proto
dnr: I made ResetOptions because we totally should share options across batch and non-batch. We do way too much duplication of fields across different messages and it makes the API verbose and confusing and hard to maintain. There should be one message for "options for modifying workflow execution properties" whether it's applied to one execution or many (and same for reset options).

Yes, there may occasionally be some semantics that only work in the individual or batch case, but that doesn't mean should always duplicate everything.

temporal/api/workflowservice/v1/service.proto
dnr: This doesn't have anything to do with workflow update, as I understand it. That's operating at the application level

temporal/api/common/v1/message.proto
dnr: I like the idea of grouping versioning into one message, to make it very clear when we're updating versioning stuff and when we're not. The top level can be a oneof of different update options, e.g. versioning stuff, memo, priority/fairness metadata, etc.

temporal/api/batch/v1/message.proto
```diff
@@ -102,3 +102,13 @@ message BatchOperationReset {
   temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 2;
 
 }
+
+// BatchOperationModifyWorkflowProperties sends ModifyWorkflowExecutionProperties requests to batch workflows.
+// Keep the parameters in sync with temporal.api.workflowservice.v1.ModifyWorkflowExecutionPropertiesRequest.
+message BatchOperationModifyWorkflowProperties {
+  // The identity of the worker/client.
+  string identity = 1;
```
dnr, line 110: I think it was wrong to put identity in each of these batch operations.. identity should just be at the top level of the WorkflowService rpc. But it's probably not worth changing now.

ychebotarev (commented):
we just added UpdateActivityOptions.
Can we call this UpdateWorfklowExecutionOptions to be consistant?

temporal/api/workflowservice/v1/request_response.proto
cretz: > we totally should share options across batch and non-batch

But we don't, so it's a `common` message without "common" use.

temporal/api/history/v1/message.proto
carlydf: The worker doesn't need to do anything with this event, so I think that's alright.

temporal/api/workflowservice/v1/request_response.proto
carlydf: I'd like to share them so that I don't have to maintain duplicates of the same fields

temporal/api/common/v1/message.proto
carlydf: I agree that replay testing only makes sense for the versioning changes, so I put those three fields into their own subsection of parameters

carlydf: > I think it should be more like force-change-on-replay-failure flag. And another option is to "force-replay-on-version-behaviour-change".

The issue with `force-change-on-replay-failure` is that implies we _always_ run replay tests (which add load to workers), and just sometimes ignore the results. I don't think there is any value in having an option to run the test and ignore a known failure, since that will just put the workflow execution into NDE state as soon as its next task is dispatched.

The value I'm trying to offer with the configurable flag is being able to opt-out of replay testing in case the caller doesn't want to add load to their workers, or doesn't want to wait for the replay tests to finish on a large batch.

`force-replay-on-version-behavior-change` makes sense, it just flips the default behavior. My perspective is that having a safer move should be the default, so that callers are protected from accidentally moving a bunch of workflows from one build to an incompatible build and breaking all of them. But, there's also an argument that adding load to workers by default is unsafe. As for the naming suggestion of `-on-version-behavior-change`, I think that suffix is no longer necessary now that the `skip_replay_test` boolean is grouped with the versioning changes, which I think makes it clear that the replay test flag does not pertain to any of the other workflow options

carlydf: re: @cretz 's comment about server-invoked replay, I've discussed it a bit, but not in depth with the SDK team. For expediency, I was thinking to force a replay test by sending a query rather than coming up with a whole new concept. Although I also could see a more general forced-replay mechanism having value in other ways.

The WorkflowModifiedExternally event can continue to be ignored by workers as far as versioning is concerned.

If I understand correctly, the interaction with the SDK / workers of an `UpdateWorkflowExecutionOptions` call with skip_replay_test=false would be:
1. Server receives `UpdateOptions` API call
2. Server determines what Build ID the workflow needs to be able to replay on (either the current default if unpinned, or the `pinned_build_id_override` if pinned)
3. Server (matching_engine) uses `QueryWorkflow` to send a `WorkflowTask` to the worker that we want to replay on
4. Server waits for query result, which (I assume) will say whether there was an NDE error
5. If no NDE, Server proceeds with the Build ID change (updates mutable state and visibility)

temporal/api/history/v1/message.proto
carlydf: I think I'll just move it back to simplify the diff

temporal/api/workflowservice/v1/service.proto
carlydf: we discussed with Yuri and Drew in a slack thread and landed on `UpdateWorkflowExecutionOptions` to align with Yuri's `UpdateActivityOptions` and CRUD verbs in general

temporal/api/common/v1/message.proto
carlydf: I'm not sure what's involved in that. I'm trying to keep my work tightly scoped to versioning so our crew can meet the release deadlines we have before Replay 2025

temporal/api/workflowservice/v1/request_response.proto
carlydf: do you think I should return a set of options based on / equivalent to `StartWorkflowOptions`? or just listing options in like is fine. I noticed that `StartWorkflowOptions` is defined in the go-sdk but not in the api -- not sure if I should repeat it all in the api

temporal/api/common/v1/message.proto
```diff
@@ -248,3 +249,26 @@ message Link {
         BatchJob batch_job = 2;
     }
 }
+
+message UpdateWorkflowExecutionOptionsConfig {
```
dnr, line 253: Can we just name it `UpdateWorkflowExecutionOptionsConfig`? `OptionsConfig` sounds weird

temporal/api/workflowservice/v1/request_response.proto
dnr: Sorry, this was my fault. I intended to update the single-Reset request but plans changed and I didn't finish it. I still think we should do it eventually.

temporal/api/common/v1/message.proto
```diff
@@ -248,3 +249,26 @@ message Link {
...
+    message UpdateVersioningOptions {
+        // If set, the workflow execution will be pinned to this build ID.
+        // All future tasks will be scheduled on this build ID.
+        google.protobuf.StringValue pinned_build_id_override = 1;
+
+        // If true, the workflow execution will become unpinned.
+        // If the workflow is already unpinned, this is a no-op.
+        // If unset/false, this is a no-op; use pinned_build_id_override to pin a workflow execution.
+        bool make_unpinned = 2;
```
dnr, line 262: I assume setting both make_unpinned and pinned_build_id_override is an error?

temporal/api/common/v1/message.proto
cretz: > I was thinking to force a replay test by sending a query rather than coming up with a whole new concept

This may be ok, especially now that we have made it clear this option is only for versioning and not general purpose. We expect to charge an action for querying (even if we don't today on query failures).

> Server waits for query result, which (I assume) will say whether there was an NDE error

I think there may need to be a bit more detail. Are you going to send an intentionally not-found query and parse the error message from every language SDK? Are you going to fail on any non-not-found query error or are you going to fail on NDE only? Former may make more sense. String message expectations are brittle. We should update every SDK in every language with a more specific piece of data (e.g. error type or details) that you can be reliably check.

temporal/api/history/v1/message.proto
cretz: Are you planning on setting `worker_may_ignore` as `true` on the event? If you are, we need to update SDKS to explicitly ignore this event and this will fail older SDKs (what we want IMO). If you aren't, we need to discuss future uses of this event where ignoring may not be acceptable, e.g. memo update.

temporal/api/workflowservice/v1/request_response.proto
cretz: Unsure to be honest. Just thinking from a user POV usually if you update something, especially partial updates, you get to see the now-updated results. Maybe we don't need that here, unsure.

temporal/api/common/v1/message.proto
```diff
@@ -248,3 +249,26 @@ message Link {
...
+        // If false, the update will be conditional on a successful
+        // replay test of the workflow execution's history on the new worker it
+        // would need to replay on after the change.
+        //
+        // In the first release, skip_replay_test=false will return an unimplemented error.
+        // User must explicitly pass skip_replay_test=true to use the API.
+        bool skip_replay_test = 3;
+    }
+
+    UpdateVersioningOptions versioning_options = 1;
```
cretz, line 273:
```suggestion
    UpdateVersioningOptions versioning = 1;
```

(pedantic, don't have to change, just reads a bit better to me thinking about HTTP API POV, though even users having to bury this under a "config" key is a bit strange for a general update)

carlydf: yeah.. I don't love the Config user experience either.. I was just coming from the goal of not repeating identical code in batch update options vs single workflow mode. Perhaps in this situation the user-developer experience takes priority over the developer-of-the-api-developer experience, since people use the sdk much more often than people make changes to the API

cretz: I have no strong preference here, this is just a consequence of proto not having inheritance or includes. Tradeoffs both ways, your choice.

temporal/api/common/v1/message.proto
carlydf: I was planning to use the stack trace query that afaik all workflows support. How to handle different types of errors is something I definitely need to think through further, but all the replay-test stuff is due after December, so I'm just pre-planning here

temporal/api/common/v1/message.proto
carlydf: yes, it would be

carlydf: good call out -- I'm changing it to a oneof

temporal/api/common/v1/message.proto
cretz: Can we resolve to make sure this replay-test option doesn't reach `master` without a tested implementation? Meaning we can add to `versioning-3` branch here, implement, etc but if we can't implement cleanly/confidently, be willing to rethink before adding to stable/released branch.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1710,3 +1715,22 @@ message UpdateActivityOptionsByIdResponse {
...
 }
+
+message UpdateWorkflowExecutionOptionsRequest {
+    // The namespace name of the target Workflow.
+    string namespace = 1;
+    // The target Workflow Id and (optionally) a specific Run Id thereof.
+    // (-- api-linter: core::0203::optional=disabled
+    //     aip.dev/not-precedent: false positive triggered by the word "optional" --)
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 2;
+    string reason = 3;
```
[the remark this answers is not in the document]
carlydf: I could have put it in logs or something, but I don't have any other burning use cases for storing it somewhere, so I removed it

cretz (approved):
LGTM for the `versioning-3` branch. I would like a chance to re-review our implementation approach to `force_replay_test_on_version_behavior_change` before we merge to `main`.

temporal/api/history/v1/message.proto
```diff
@@ -741,6 +741,13 @@ message WorkflowPropertiesModifiedExternallyEventAttributes {
     // the existing memo. If the user wants to delete values, a default/empty Payload should be
     // used as the value for the key being deleted.
     temporal.api.common.v1.Memo upserted_memo = 5;
+
+    // If not null, the versioning behavior override on this workflow execution was either set or unset.
+    // If it was set, 'override_set' describes how. If unset, 'override_unset' is true.
+    oneof version_override_mutation {
```
ShahabT, line 747: `WorkflowPropertiesModifiedExternallyEventAttributes` is not yet used anywhere. Should we consider renaming it to something like `WorkflowOptionsUpdatedEventAttributes` to match the API name?

temporal/api/history/v1/message.proto
carlydf: I feel like update-object + fieldmask + result-object is a better way to capture what was changed in the event than the fields in `WorkflowPropertiesModifiedExternally`, and I want the event name to match the API name, so I made a new `WorkflowExecutionOptionsUpdated` event type. I'm thinking we can decide what to do with `WorkflowPropertiesModifiedExternally` event sometime in the future

carlydf: Since having the sdk ignore the event in the future for memo update is not acceptable, what would happen if I did not set `worker_may_ignore=true` on this event now?

For SDKs that don't know about this event type, does `worker_may_ignore=false` mean they crash if they see this event? Since using versioning-3 requires the newest sdk anyways, maybe that's fine.

For the first iteration of `UpdateWorkflowExecutionOptions`, where the sdk does not need to do anything, I could have `worker_may_ignore=false` but the code that "handles" the event in the SDK can be blank.

Later, when we add memo update and do need to react to the event, we could add logic in the sdk event handler to do what needs to be done. I think this makes sense to me but lmk what you think

temporal/api/common/v1/message.proto
carlydf: removed it since it won't be implemented yet!

temporal/api/workflowservice/v1/request_response.proto
carlydf: solved with fieldmask + returned `WorkflowExecutionOptions`

temporal/api/common/v1/message.proto
carlydf: outdated, resolving

temporal/api/common/v1/message.proto
carlydf: outdated, resolving

temporal/api/common/v1/message.proto
carlydf: outdated now that params are `WorkflowExecutionOptions` + FieldMask

temporal/api/history/v1/message.proto
carlydf: instead of renaming I made a new separate event. we can remove the unused one at a later time

temporal/api/common/v1/message.proto
```diff
@@ -156,6 +157,20 @@ message WorkerVersionCapabilities {
...
+// Used in both UpdateWorkflowExecutionOptions and StartWorkflowOptions to override
+// the versioning behavior of a specific workflow execution. If set, takes precedence
+// over the sdk-sent Versioning Behavior for the specific Workflow Execution it is set on.
+// To remove the override, call UpdateWorkflowExecutionOptions with a null VersioningBehaviorOverride,
+// and use the FieldMask to indicate that it should be mutated.
+message VersioningBehaviorOverride {
+    // Required.
+    temporal.api.enums.v1.VersioningBehavior behavior = 1;
+
+    // Required if behavior is 'pinned'. Must be null if behavior is 'unpinned'.
```
ShahabT, line 169: Nit: 
```suggestion
// Required if behavior is `PINNED`. Must be null if behavior is `AUTO_UPGRADE`.
```

temporal/api/history/v1/message.proto
```diff
@@ -746,6 +747,19 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
...
+    // Workflow Execution options that were sent in the UpdateWorkflowExecutionOptions request.
+    // Partial updates are accepted and controlled by update_mask.
+    temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options_update = 1;
+
+    // Update mask that was sent in the UpdateWorkflowExecutionOptions request.
+    // Indicates which fields from `workflow_execution_options_for_update` were applied.
+    google.protobuf.FieldMask update_mask = 2;
+
+    // Workflow Execution options after the update. Also returned as gRPC response.
+    temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options_result = 3;
```
ShahabT, line 760: Hmm... thinking if this is needed or just the updated fields are enough to put in the event :-?

carlydf: True we don't need this! The current options can be reconstructed by going through each update, so this is unecessary data. I should have thought of that.. I'll remove to not make History too big

temporal/api/history/v1/message.proto
antlai-temporal: we could also change the worker_may_ignore=false when we actually modify something that the sdk needs to act upon, like memo. This will more accurately represent whether the SDK needs to act...

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -358,6 +355,11 @@ message RespondWorkflowTaskCompletedRequest {
     temporal.api.sdk.v1.WorkflowTaskCompletedMetadata sdk_metadata = 12;
     // Local usage data collected for metering
     temporal.api.common.v1.MeteringMetadata metering_metadata = 13;
+    reserved 14;
+    reserved "capabilities";
+    // Must be set when versioning-3 is used (i.e. `worker_version_stamp.use_versioning` is `true`
+    // and `worker_version_stamp.deployment_name` is provided).
+    temporal.api.enums.v1.VersioningBehavior versioning_behavior = 15;
 }
```
antlai-temporal, line 363: We are changing the tag to 15 from 14. My sdk version is already using this field, hopefully just for testing, but I wonder whether we should try in general to not change tags meaning just in case...

PR #467, opened 2024-10-23
Fix RespondWorkflowTask capabilities

Quinn-With-Two-Ns:
Fix RespondWorkflowTask capabilities, I put it on the `Response` instead of the `Request` because I am dumb.

Note: This is technically a breaking change, but this was not released anywhere
* https://github.com/temporalio/temporal/blob/v1.26.2-122.1/go.mod -> https://github.com/temporalio/api-go/commit/c02089051a35
* https://github.com/temporalio/temporal/blob/v1.26.1/go.mod -> https://github.com/temporalio/api-go/commit/b13574e18f3c
* https://github.com/temporalio/temporal/blob/main/go.mod#L58 -> https://github.com/temporalio/api-go/commit/c02089051a35

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -12,6 +12,8 @@ breaking:
     - WIRE_JSON
   ignore:
     - google
+    # TODO(klassenq): remove me
+    - temporal/api/workflowservice/v1/request_response.proto
 lint:
   use:
     - DEFAULT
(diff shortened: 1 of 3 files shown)


alexshtin:
Two reviewers approved original #458 PR for the same reason.

alexshtin (approved):

cretz (approved):

PR #469, opened 2024-10-28
Pause/Unpause/Reset definitions

ychebotarev:
Add definitions for Pause/Unpause/Reset Activity API.

Working on activity API

Corresponding server PR: https://github.com/temporalio/temporal/pull/6736

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -149,6 +149,9 @@ message PendingActivityInfo {
     // Next time when activity will be scheduled.
     // If activity is currently scheduled or started it will be null.
     google.protobuf.Timestamp next_attempt_schedule_time = 18;
+
+    // Indicates if activity was paused.
+    bool is_paused = 19;
 }
 
 message PendingChildExecutionInfo {
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -150,7 +150,7 @@ message PendingActivityInfo {
     // If activity is currently scheduled or started it will be null.
     google.protobuf.Timestamp next_attempt_schedule_time = 18;
 
-    // Indicates if activity was paused.
+    // Indicates if activity is paused.
     bool paused = 19;
 }
 
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -151,7 +151,7 @@ message PendingActivityInfo {
     google.protobuf.Timestamp next_attempt_schedule_time = 18;
 
     // Indicates if activity was paused.
-    bool is_paused = 19;
+    bool paused = 19;
 }
 
 message PendingChildExecutionInfo {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1768,8 +1768,8 @@ message UnpauseActivityByIdRequest {
 
     // there are two options to resume the activity - with 'resume' or with 'reset'
     oneof operation{
-        ResumeAttributes resume_attributes = 6;
-        ResetAttributes reset_attributes = 7;
+        ResumeAttributes resume = 6;
+        ResetAttributes reset = 7;
     }
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -479,7 +479,7 @@ message RecordActivityTaskHeartbeatResponse {
     bool cancel_requested = 1;
 
     // Will be set to true if the activity has been put on pause.
-    bool activity_pause = 2;
+    bool activity_paused = 2;
 }
 
 message RecordActivityTaskHeartbeatByIdRequest {
@@ -1742,14 +1742,15 @@ message PauseActivityByIdResponse {
 
 message UnpauseActivityByIdRequest {
     message ResumeAttributes {
-        // Indicates that activity should be scheduled immediately.
+        // Indicates that if activity is waiting for retry, it will  be scheduled immediately.
         bool no_wait = 1;
     }
 
     message ResetAttributes {
         // Indicates that activity should be scheduled immediately.
+        // Note that this may run simultaneously with any existing executions of the activity.
         bool no_wait = 1;
-        // Indicates that activity should reset the heartbeats.
+        // If set, the Heartbeat Details will be cleared out to make the Activity start over from the beginning
         bool reset_heartbeat = 2;
     }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -477,6 +477,9 @@ message RecordActivityTaskHeartbeatResponse {
     // Will be set to true if the activity has been asked to cancel itself. The SDK should then
     // notify the activity of cancellation if it is still running.
     bool cancel_requested = 1;
+
+    // Will be set to true if the activity has been put on pause.
+    bool activity_pause = 2;
 }
 
 message RecordActivityTaskHeartbeatByIdRequest {
@@ -1696,12 +1699,12 @@ message ExecuteMultiOperationResponse {
 message UpdateActivityOptionsByIdRequest {
     // Namespace of the workflow which scheduled this activity
     string namespace = 1;
-    // Id of the workflow which scheduled this activity
+    // ID of the workflow which scheduled this activity
     string workflow_id = 2;
-    // Run Id of the workflow which scheduled this activity
+    // Run ID of the workflow which scheduled this activity
     // if empty - latest workflow is used
     string run_id = 3;
-    // Id of the activity we're updating
+    // ID of the activity we're updating
     string activity_id = 4;
 
     // The identity of the client who initiated this request
@@ -1718,3 +1721,86 @@ message UpdateActivityOptionsByIdResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
 }
+
+message PauseActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
+    // ID of the activity we're updating.
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request.
+    string identity = 5;
+}
+
+message PauseActivityByIdResponse {
+}
+
+message UnpauseActivityByIdRequest {
+    message ResumeAttributes {
+        // Indicates that activity should be scheduled immediately.
+        bool no_wait = 1;
+    }
+
+    message ResetAttributes {
+        // Indicates that activity should be scheduled immediately.
+        bool no_wait = 1;
+        // Indicates that activity should reset the heartbeats.
+        bool reset_heartbeat = 2;
+    }
+
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
+    // ID of the activity we're updating.
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request.
+    string identity = 5;
+
+    // there are two options to resume the activity - with 'resume' or with 'reset'
+    oneof operation{
+        ResumeAttributes resume_attributes = 6;
+        ResetAttributes reset_attributes = 7;
+    }
+}
+
+message UnpauseActivityByIdResponse {
+}
+
+message ResetActivityByIdRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
+    // ID of the activity we're updating.
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request.
+    string identity = 5;
+
+    // Indicates that activity should be scheduled immediately.
+    // If this flag doesn't set, and activity currently running - temporal will wait until activity is completed.
+    bool no_wait = 6;
+
+    // Indicates that activity should reset the heartbeats.
+    // This flag will be applied only to the new instance of the activity.
+    bool reset_heartbeat = 7;
+
+}
+
+message ResetActivityByIdResponse {
+}
+
+
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -478,7 +478,7 @@ message RecordActivityTaskHeartbeatResponse {
     // notify the activity of cancellation if it is still running.
     bool cancel_requested = 1;
 
-    // Will be set to true if the activity has been put on pause.
+    // Will be set to true if the activity is paused.
     bool activity_paused = 2;
 }
 
@@ -1748,12 +1748,12 @@ message PauseActivityByIdResponse {
 
 message UnpauseActivityByIdRequest {
     message ResumeOperation {
-        // Indicates that if activity is waiting for retry, it will  be scheduled immediately.
+        // Indicates that if the activity is waiting to retry, it will  be scheduled immediately.
         bool no_wait = 1;
     }
 
     message ResetOperation {
-        // Indicates that activity should be scheduled immediately.
+        // Indicates that the activity should be scheduled immediately.
         // Note that this may run simultaneously with any existing executions of the activity.
         bool no_wait = 1;
         // If set, the Heartbeat Details will be cleared out to make the Activity start over from the beginning
@@ -1807,7 +1807,7 @@ message ResetActivityByIdRequest {
     // If this flag doesn't set, and activity currently running - temporal will wait until activity is completed.
     bool no_wait = 7;
 
-    // Indicates that activity should reset the heartbeats.
+    // Indicates that activity should reset heartbeat details.
     // This flag will be applied only to the new instance of the activity.
     bool reset_heartbeat = 8;
 }
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1716,7 +1716,7 @@ message UpdateActivityOptionsByIdRequest {
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 7;
 
-    // Used to de-dupe reset requests
+    // Used to de-dupe requests
     string request_id = 8;
 }
 
@@ -1739,7 +1739,7 @@ message PauseActivityByIdRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe reset requests
+    // Used to de-dupe requests
     string request_id = 6;
 }
 
@@ -1773,7 +1773,7 @@ message UnpauseActivityByIdRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe reset requests
+    // Used to de-dupe requests
     string request_id = 6;
 
     // there are two options to resume the activity - with 'resume' or with 'reset'
@@ -1800,9 +1800,8 @@ message ResetActivityByIdRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe reset requests
+    // Used to de-dupe requests
     string request_id = 6;
-
 
     // Indicates that activity should be scheduled immediately.
     // If this flag doesn't set, and activity currently running - temporal will wait until activity is completed.
@@ -1811,7 +1810,6 @@ message ResetActivityByIdRequest {
     // Indicates that activity should reset the heartbeats.
     // This flag will be applied only to the new instance of the activity.
     bool reset_heartbeat = 8;
-
 }
 
 message ResetActivityByIdResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1739,7 +1739,7 @@ message PauseActivityByIdRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe requests
+    // Used to de-dupe requests.
     string request_id = 6;
 }
 
@@ -1747,12 +1747,12 @@ message PauseActivityByIdResponse {
 }
 
 message UnpauseActivityByIdRequest {
-    message ResumeAttributes {
+    message ResumeOperation {
         // Indicates that if activity is waiting for retry, it will  be scheduled immediately.
         bool no_wait = 1;
     }
 
-    message ResetAttributes {
+    message ResetOperation {
         // Indicates that activity should be scheduled immediately.
         // Note that this may run simultaneously with any existing executions of the activity.
         bool no_wait = 1;
@@ -1773,13 +1773,13 @@ message UnpauseActivityByIdRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe requests
+    // Used to de-dupe requests.
     string request_id = 6;
 
-    // there are two options to resume the activity - with 'resume' or with 'reset'
+    // There are two options to resume the activity - with 'resume' or with 'reset'.
     oneof operation{
-        ResumeAttributes resume = 7;
-        ResetAttributes reset = 8;
+        ResumeOperation resume = 7;
+        ResetOperation reset = 8;
     }
 }
 
@@ -1800,7 +1800,7 @@ message ResetActivityByIdRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe requests
+    // Used to de-dupe requests.
     string request_id = 6;
 
     // Indicates that activity should be scheduled immediately.
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1715,6 +1715,9 @@ message UpdateActivityOptionsByIdRequest {
 
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 7;
+
+    // Used to de-dupe reset requests
+    string request_id = 8;
 }
 
 message UpdateActivityOptionsByIdResponse {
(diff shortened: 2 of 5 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/activities/pause-by-id"
+                body: "*"
+            }
+        };
+    }
+
+    // UnpauseActivityById unpauses the execution of an activity specified by its ID.
+    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    // There are two 'modes' of unpausing an activity:
```
bergundy, line 902: Wondering about the semantics here. I get that some users will want to do both reset and unpause but what I find missing is a toggle to control whether to fast forward the time spent paused. Wondering if that has been discussed.

I think it'd be more intuitive to put unpause + reset in the reset API instead.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1718,3 +1721,86 @@ message UpdateActivityOptionsByIdResponse {
...
+    string activity_id = 4;
+
+    // The identity of the client who initiated this request.
+    string identity = 5;
+
+    // there are two options to resume the activity - with 'resume' or with 'reset'
+    oneof operation{
+        ResumeAttributes resume_attributes = 6;
+        ResetAttributes reset_attributes = 7;
+    }
```
bergundy, line 1773: As mentioned before, I don't think we want these two operations on this API but if we do decide to keep them, please rename to:

```suggestion
    oneof operation {
        ResumeAttributes resume = 6;
        ResetAttributes reset = 7;
    }
```

temporal/api/workflowservice/v1/service.proto
ychebotarev: >toggle to control whether to fast forward the time spent paused
it has been discussed. We decide the we don't 'skip' the time that activity spend paused. 

>I think it'd be more intuitive to put unpause + reset in the reset API instead.
This was also discussed. 'unpause' with 'reset' is for discoverability.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: will do

temporal/api/workflowservice/v1/service.proto
bergundy: > it has been discussed. We decide the we don't 'skip' the time that activity spend paused.

What's the rationale?

> This was also discussed. 'unpause' with 'reset' is for discoverability.

This doesn't have to affect the server API, SDKs and CLI can make this discoverable if they choose to.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+    // Pausing an activity means:
+    // - If the activity is currently waiting for a retry or is running and subsequently fails,
+    //   it will not be rescheduled until it is unpaused.
+    // - If the activity is already paused, calling this method will have no effect.
+    // - If the activity is running and will finish successfully, activity will be completed.
+    // - If the activity is running and will finish with failure:
+    //   * if there is no retry policy, or no retry left - activity will be completed.
+    //   * if there are more retries left - activity will be paused and will be retried later.
+    // For long-running activities:
+    // - Pausing will send a cancellation with "is_paused" set to 'true'.
```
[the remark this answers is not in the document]
ychebotarev: added more information. Keeping open.

temporal/api/workflowservice/v1/service.proto
ychebotarev: Rationale for not skipping time - a lot of complexity for no clear benefit. If users want to extend timers - they can use "UpdateOptions" for that.

second point will discuss offline.

bergundy: How much complexity is this though? If you centralize the access to the activity in the server codebase it should be easy to cover all of the paths.
Why do you think there's no clear benefit? Seems like we're forcing the user to deal with this by calling UpdateOptions.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1718,3 +1721,86 @@ message UpdateActivityOptionsByIdResponse {
...
+    string identity = 5;
+}
+
+message PauseActivityByIdResponse {
+}
+
+message UnpauseActivityByIdRequest {
+    message ResumeAttributes {
+        // Indicates that activity should be scheduled immediately.
+        bool no_wait = 1;
```
drewhoskins-temporal, line 1746: OOC, why is `no_wait` in both places rather than shared at the top level?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1718,3 +1721,86 @@ message UpdateActivityOptionsByIdResponse {
...
+    }
+
+    message ResetAttributes {
+        // Indicates that activity should be scheduled immediately.
+        bool no_wait = 1;
+        // Indicates that activity should reset the heartbeats.
```
drewhoskins-temporal, line 1752: This simply restates the field.
Consider: `If set, the Heartbeat Details will be cleared out to make the Activity start over from the beginning.`

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+            post: "/namespaces/{namespace}/activities/unpause-by-id"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/activities/unpause-by-id"
+                body: "*"
+            }
+        };
+    }
+
+    // ResetActivityById unpauses the execution of an activity specified by its ID.
```
drewhoskins-temporal, line 925: I thought we decided not to fail with a `NotFound` if the activity got completed to prevent failing on race conditions?

Minor stuff:

s/Resting/resetting/

> all timeout timers will be regenerated.

nit: This is an implementation-specific way of describing it.  Users would rather see "the activity will get its full timeout back"

> if 'no_wait' flag is provided, new instance of activity will be scheduled immediately.

I suggest:

if _the_ `no_wait` flag is provided, _a_ new instance of _the_ activity will be scheduled immediately.  _Note that this may run simultaneously with any existing executions of the activity._

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: they have same/similar name, but different meaning. For resume it means that if it is in retry it will not wait. But unpause will not start new instance it is already running

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ok.

temporal/api/workflowservice/v1/service.proto
ychebotarev: if this happens to be something users ask for - it can be done later by adding another flag

temporal/api/workflow/v1/message.proto
```diff
@@ -149,6 +149,9 @@ message PendingActivityInfo {
     // Next time when activity will be scheduled.
     // If activity is currently scheduled or started it will be null.
     google.protobuf.Timestamp next_attempt_schedule_time = 18;
+
+    // Indicates if activity was paused.
+    bool is_paused = 19;
```
cretz, line 154:
```suggestion
    bool paused = 19;
```

Pedantic, but probably don't need `is_` prefix (e.g. `ScheduleState.paused`)

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+    // PauseActivityById pauses the execution of an activity specified by its ID.
+    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    //
+    // Pausing an activity means:
+    // - If the activity is currently waiting for a retry or is running and subsequently fails,
+    //   it will not be rescheduled until it is unpaused.
+    // - If the activity is already paused, calling this method will have no effect.
+    // - If the activity is running and will finish successfully, activity will be completed.
+    // - If the activity is running and will finish with failure:
+    //   * if there is no retry policy, or no retry left - activity will be completed.
```
cretz, line 882:
```suggestion
    //   * if there is no retry left - activity will be completed.
```

From a user POV, no retry policy means indefinite retries

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+    //   * if there are more retries left - activity will be paused and will be retried later.
```
cretz, line 883:
```suggestion
    //   * if there are more retries left - activity will be paused
```

May not be retried later. Also, it seems like this is restating the second half of the first bullet above

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+    }
+
+    // ResetActivityById unpauses the execution of an activity specified by its ID.
+    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    // Resting an activity means:
+    // * number of attempts will be reset to 0.
+    // * all timeout timers will be regenerated.
+    // If activity currently running:
+    // *  if 'no_wait' flag is provided, new instance of activity will be scheduled immediately.
+    // *  if 'no_wait' flag is not provided, activity will be scheduled after current instance completes if needed.
```
cretz, line 932: Curious about something here. So there is technically a "pending reset" an activity could be stuck in while waiting for an attempt to complete? I suppose there's no way to back out of that?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+            }
+        };
+    }
+
+    // UnpauseActivityById unpauses the execution of an activity specified by its ID.
+    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    // There are two 'modes' of unpausing an activity:
+    // 'resume' - If the activity is paused, it will be resumed and scheduled for execution.
+    //    * If activity is currently running Unpause with 'resume' has no effect.
+    //    * if 'no_wait' flag is set and activity is waiting, the activity will be scheduled immediately.
```
cretz, line 905: These are a bit contradictory. I want no_wait but really you'll still wait for activity to complete?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -477,6 +477,9 @@ message RecordActivityTaskHeartbeatResponse {
     // Will be set to true if the activity has been asked to cancel itself. The SDK should then
     // notify the activity of cancellation if it is still running.
     bool cancel_requested = 1;
+
+    // Will be set to true if the activity has been put on pause.
+    bool activity_pause = 2;
```
cretz, line 482:
```suggestion
    bool activity_paused = 2;
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,80 @@ service WorkflowService {
...
+    // Pausing an activity means:
+    // - If the activity is currently waiting for a retry or is running and subsequently fails,
+    //   it will not be rescheduled until it is unpaused.
+    // - If the activity is already paused, calling this method will have no effect.
+    // - If the activity is running and will finish successfully, activity will be completed.
+    // - If the activity is running and will finish with failure:
+    //   * if there is no retry policy, or no retry left - activity will be completed.
+    //   * if there are more retries left - activity will be paused and will be retried later.
+    // For long-running activities:
+    // - activity in paused state will send a cancellation with "is_paused" set to 'true' in response to 'RecordActivityTaskHeartbeat'.
```
cretz, line 885:
```suggestion
    // - activity in paused state will send a cancellation with "activity_paused" set to 'true' in response to 'RecordActivityTaskHeartbeat'.
```

cretz (commented):
Some minor things, overall LGTM

temporal/api/workflow/v1/message.proto
ychebotarev: ok

temporal/api/workflowservice/v1/service.proto
ychebotarev: ok

temporal/api/workflowservice/v1/service.proto
ychebotarev: ok

temporal/api/workflowservice/v1/service.proto
ychebotarev: Yes, there will be internal "pending reset" state. Something like "reset_requested". Yes, there is no way for users to know that they are in this state. We can think about adding it to the external STATE, like CANCEL_REQUESTED.

temporal/api/workflowservice/v1/service.proto
ychebotarev: >These are a bit contradictory. I want no_wait but really you'll still wait for activity to complete?
We can't actually "pause" the activity that is running. The best we can do is to send signal via heartbeats for long running activities.
This flag is applicable only for activities that are in backoff state. I don't 100% like this flag name. Not loo late to propose better name.

>Also, there is a use case for triggering an activity to run another attempt immediately (not a pause/resume use case, but a ignore-backoff-run-next-attempt use case), are we handling that with these APIs?

What you are describing is "TriggerActivity". It was in the original proposal, but didn't make it to the final. I don't really remember why (just too many APIs?).

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ok

temporal/api/workflowservice/v1/service.proto
ychebotarev: ok

temporal/api/workflowservice/v1/service.proto
cretz: > We can't actually "pause" the activity that is running. The best we can do is to send signal via heartbeats for long running activities.

Not exactly. We can effectively "terminate". This is what happens on activity timeout. We send the cancel then ignore any response and move on. To me, this is what "no wait" means.

> What you are describing is "TriggerActivity". It was in the original proposal, but didn't make it to the final. I don't really remember why (just too many APIs?).

Hrmm, I would think the "my activity is in backoff, interrupt backoff and try again" is maybe the most commonly requested use case of all. People are just going to "pause" + "unpause + resume no wait" to get this behavior. @drewhoskins-temporal - are we intentionally not accounting for this use case?

temporal/api/workflowservice/v1/service.proto
ychebotarev: discussed offline. closing.

temporal/api/workflowservice/v1/service.proto
cretz: Ok, I just realized this is `reset` + `no_wait`. I am good here. (discoverability concerns may apply, but those can be addressed in the user-facing places)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1718,3 +1721,87 @@ message UpdateActivityOptionsByIdResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
 }
+
+message PauseActivityByIdRequest {
```
cretz, line 1725: Sorry, just realized, can we have a `request_id` on these calls for deduping? We do this on most other mutation calls. Granted for idempotent calls like pause it may not be needed, but unpause that has a resume-no-wait or reset, we need this idempotency key.

cretz (approved):
Everything LGTM, but just realized there was `request_id` missing

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: yes, it make sense, added.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1712,9 +1715,106 @@ message UpdateActivityOptionsByIdRequest {
 
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 7;
+
+    // Used to de-dupe reset requests
```
cretz, line 1719: This is not a reset request (same thing for other places in this commit where it says "reset requests" but it's not)

cretz (approved):
LGTM, minor note on proto comment. But don't want to merge on my approval alone, may need to make sure consensus is reached with those advocating for shared RPC call or any other point of contention first.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: fixed

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1712,9 +1715,104 @@ message UpdateActivityOptionsByIdRequest {
...
+
+    // Used to de-dupe requests
+    string request_id = 6;
+}
+
+message PauseActivityByIdResponse {
+}
+
+message UnpauseActivityByIdRequest {
+    message ResumeAttributes {
```
bergundy, line 1750:
```suggestion
    message ResumeOperation {
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1712,9 +1715,104 @@ message UpdateActivityOptionsByIdRequest {
...
+        // Indicates that if activity is waiting for retry, it will  be scheduled immediately.
+        bool no_wait = 1;
+    }
+
+    message ResetAttributes {
```
bergundy, line 1755:
```suggestion
    message ResetOperation {
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1712,9 +1715,104 @@ message UpdateActivityOptionsByIdRequest {
...
+    // The identity of the client who initiated this request.
+    string identity = 5;
+
+    // Used to de-dupe requests
+    string request_id = 6;
+
+    // there are two options to resume the activity - with 'resume' or with 'reset'
+    oneof operation{
+        ResumeAttributes resume = 7;
+        ResetAttributes reset = 8;
```
bergundy, line 1782:
```suggestion
        ResumeOperation resume = 7;
        ResetOperation reset = 8;
```

bergundy (approved):
Had a few small suggestions. The grammar could also be improved a bit in the comments and there's inconsistent use of punctuation but I won't block on that.

cretz:
@ychebotarev - is it possible to wait on merging this until your implementation is ready? Or if already done, all good.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -477,6 +477,9 @@ message RecordActivityTaskHeartbeatResponse {
     // Will be set to true if the activity has been asked to cancel itself. The SDK should then
     // notify the activity of cancellation if it is still running.
     bool cancel_requested = 1;
+
+    // Will be set to true if the activity has been put on pause.
```
Sushisource, line 481:
```suggestion
    // Will be set to true if the activity is paused.
```

temporal/api/workflow/v1/message.proto
```diff
@@ -149,6 +149,9 @@ message PendingActivityInfo {
     // Next time when activity will be scheduled.
     // If activity is currently scheduled or started it will be null.
     google.protobuf.Timestamp next_attempt_schedule_time = 18;
+
+    // Indicates if activity was paused.
```
Sushisource, line 153:
```suggestion
    // Indicates if activity is paused.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1712,9 +1715,104 @@ message UpdateActivityOptionsByIdRequest {
...
+    // Used to de-dupe requests
+    string request_id = 6;
+}
+
+message PauseActivityByIdResponse {
+}
+
+message UnpauseActivityByIdRequest {
+    message ResumeAttributes {
+        // Indicates that if activity is waiting for retry, it will  be scheduled immediately.
```
Sushisource, line 1751:
```suggestion
        // Indicates that if the activity is waiting to retry, it will be scheduled immediately.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1712,9 +1715,104 @@ message UpdateActivityOptionsByIdRequest {
...
+        bool no_wait = 1;
+    }
+
+    message ResetAttributes {
+        // Indicates that activity should be scheduled immediately.
+        // Note that this may run simultaneously with any existing executions of the activity.
```
Sushisource, line 1757:
```suggestion
        // Indicates that the activity should be scheduled immediately.
        // Note that this may run simultaneously with any existing executions of the activity.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1712,9 +1715,104 @@ message UpdateActivityOptionsByIdRequest {
...
+    string identity = 5;
+
+    // Used to de-dupe requests
+    string request_id = 6;
+
+    // Indicates that activity should be scheduled immediately.
+    // If this flag doesn't set, and activity currently running - temporal will wait until activity is completed.
+    bool no_wait = 7;
+
+    // Indicates that activity should reset the heartbeats.
```
Sushisource, line 1810:
```suggestion
    // Indicates that activity should reset heartbeat details.
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,79 @@ service WorkflowService {
...
+    // - If the activity is currently waiting for a retry or is running and subsequently fails,
+    //   it will not be rescheduled until it is unpaused.
+    // - If the activity is already paused, calling this method will have no effect.
+    // - If the activity is running and will finish successfully, activity will be completed.
+    // - If the activity is running and will finish with failure:
+    //   * if there is no retry left - activity will be completed.
+    //   * if there are more retries left - activity will be paused.
+    // For long-running activities:
+    // - activity in paused state will send a cancellation with "activity_paused" set to 'true' in response to 'RecordActivityTaskHeartbeat'.
+    // - The activity should respond to the cancellation accordingly.
```
Sushisource, line 886:
```suggestion
    // - If the activity is running and finishes successfully, the activity will be completed.
    // - If the activity is running and finishes with failure:
    //   * if there is no retry left - the activity will be completed.
    //   * if there are more retries left - the activity will be paused.
    // For long-running activities:
    // - activities in paused state will send a cancellation with "activity_paused" set to 'true' in response to 'RecordActivityTaskHeartbeat'.
    // - The activity should respond to the cancellation accordingly.
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,79 @@ service WorkflowService {
...
+    // There are two 'modes' of unpausing an activity:
+    // 'resume' - If the activity is paused, it will be resumed and scheduled for execution.
+    //    * If activity is currently running Unpause with 'resume' has no effect.
+    //    * if 'no_wait' flag is set and activity is waiting, the activity will be scheduled immediately.
+    // 'reset' - If the activity is paused, it will be reset to its initial state and (depending on parameters) scheduled for execution.
+    //    * If activity is currently running Unpause with 'reset' will reset the number of attempts.
+    //    * if 'no_wait' flag is set, the activity will be scheduled immediately.
+    //    * if 'reset_heartbeats' flag is set, the activity heartbeat timer and heartbeats will be reset.
+    // If activity is in waiting for retry and past it retry timeout, it will be scheduled immediately.
+    // Once activity is unpaused, all timeout timers will be regenerated.
```
Sushisource, line 911:
```suggestion
    //    * If the activity is currently running, Unpause with 'resume' has no effect.
    //    * if 'no_wait' flag is set and the activity is waiting, the activity will be scheduled immediately.
    // 'reset' - If the activity is paused, it will be reset to its initial state and (depending on parameters) scheduled for execution.
    //    * If the activity is currently running, Unpause with 'reset' will reset the number of attempts.
    //    * if 'no_wait' flag is set, the activity will be scheduled immediately.
    //    * if 'reset_heartbeats' flag is set, the activity heartbeat timer and heartbeats will be reset.
    // If the activity is in waiting for retry and past it retry timeout, it will be scheduled immediately.
    // Once the activity is unpaused, all timeout timers will be regenerated.
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -871,5 +870,79 @@ service WorkflowService {
...
+    }
+
+    // ResetActivityById unpauses the execution of an activity specified by its ID.
+    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    // Resetting an activity means:
+    // * number of attempts will be reset to 0.
+    // * activity timeouts will be resetted.
+    // If activity currently running:
+    // *  if 'no_wait' flag is provided, new instance of activity will be scheduled immediately.
+    // *  if 'no_wait' flag is not provided, activity will be scheduled after current instance completes if needed.
```
Sushisource, line 932:
```suggestion
    // If the activity currently running:
    // *  if 'no_wait' flag is provided, a new instance of the activity will be scheduled immediately.
    // *  if 'no_wait' flag is not provided, a new instance of the activity will be scheduled after the current instance completes, if needed.
```

Sushisource (approved):
Works for me, nothing blocking. Just some grammar help suggestions, mostly being consistent about treating specific instances of activities as definitive with "the".

I still think I'd prefer if all the "targeting" parameters that are duped per-rpc were wrapped up in a common message like `ActivityReference` or whatever name, but it's not a blocking concern.

temporal/api/workflowservice/v1/service.proto
ychebotarev: done

temporal/api/workflowservice/v1/service.proto
ychebotarev: done

temporal/api/workflowservice/v1/service.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflow/v1/message.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

PR #471, opened 2024-11-01
Add paused to retry state

ychebotarev:
Add RETRY_STATE_PAUSED.

Part of activity API. If activity is paused, retry shouldn't happen, but we also shouldn't return an error. Thus we need another state.

This is what will be returned by RetryActivity in mutable state if activity is paused. I don't think it is visible outside of server code.


cretz (commented):
Makes sense, I'm guessing we missed this one on #468. Do we want to wait to collect other things that we may add as part of this project for first implementation?

ychebotarev:
> Makes sense, I'm guessing we missed this one on #468. Do we want to wait to collect other things that we may add as part of this project for first implementation?

This is not a breaking change.

cretz:
Sorry, I mistyped, I mean #469. Yes, in this case it was good that what was missed is an addition and not a reworking.

ychebotarev:
ok lets not do that.

PR #472, opened 2024-11-05
Add versioning fields to WorkflowExecutionInfo

ShahabT:
**What changed?**
Add `versioning_info` to `WorkflowExecutionInfo`. For now `versioning_info` includes:
- `versioning_behavior`
- `deployment`
- `versioning_behavior_override`
- `deployment_override`
- `redirecting_deployment`

**Why?**
This info is important to users dealing with versioned workflows.

**Breaking changes**
Deprecated V2 fields.

**Server PR**

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -156,6 +156,12 @@ message WorkerVersionCapabilities {
     // Later, may include info like "I can process WASM and/or JS bundles"
 }
 
+// Identifies a versioned worker deployment.
+message WorkerDeployment {
+    string deployment_name = 1;
+    string build_id = 2;
+}
+
 // Describes where and how to reset a workflow, used for batch reset currently
 // and may be used for single-workflow reset later.
 message ResetOptions {
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -222,12 +222,13 @@ message WorkflowTaskStartedEventAttributes {
     // when rebuilding it from events.
     // Deprecated. This field should be cleaned up when versioning-2 API is removed. [cleanup-experimental-wv]
     int64 build_id_redirect_counter = 7;
-    // The worker deployment to which the task was dispatched. Empty means the task
-    // was dispatched to an unversioned worker.
-    temporal.api.common.v1.WorkerDeployment deployment = 8;
+    // When present, it means this task is dispatched to a new deployment and caused
+    // the workflow execution to be redirected to that deployment from this point on.
+    temporal.api.common.v1.WorkerDeployment redirect_deployment = 8;
     // Used by server internally to properly reapply deployment redirects to an
-    // execution when rebuilding it from history events.
-    int64 deployment_change_counter = 9;
+    // execution when rebuilding it from history events. This is non-zero if and
+    // only if `redirect_deployment` is present.
+    int64 deployment_redirect_counter = 9;
 }
 
 message WorkflowTaskCompletedEventAttributes {
@@ -352,12 +353,13 @@ message ActivityTaskStartedEventAttributes {
     // when rebuilding it from events.
     // Deprecated. This field should be cleaned up when versioning-2 API is removed. [cleanup-experimental-wv]
     int64 build_id_redirect_counter = 7;
-    // The worker deployment to which the task was dispatched. Empty means the task
-    // was dispatched to an unversioned worker.
-    temporal.api.common.v1.WorkerDeployment deployment = 8;
+    // When present, it means this task is dispatched to a new deployment and caused
+    // the workflow execution to be redirected to that deployment from this point on.
+    temporal.api.common.v1.WorkerDeployment redirect_deployment = 8;
     // Used by server internally to properly reapply deployment redirects to an
-    // execution when rebuilding it from history events.
-    int64 deployment_change_counter = 9;
+    // execution when rebuilding it from history events. This is non-zero if and
+    // only if `redirect_deployment` is present.
+    int64 deployment_redirect_counter = 9;
 }
 
 message ActivityTaskCompletedEventAttributes {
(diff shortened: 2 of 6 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -215,10 +216,18 @@ message WorkflowTaskStartedEventAttributes {
...
     // Used by server internally to properly reapply build ID redirects to an execution
     // when rebuilding it from events.
+    // Deprecated. This field should be cleaned up when versioning-2 API is removed. [cleanup-experimental-wv]
     int64 build_id_redirect_counter = 7;
+    // The worker deployment to which the task was dispatched. Empty means the task
+    // was dispatched to an unversioned worker.
+    temporal.api.common.v1.WorkerDeployment deployment = 8;
+    // Used by server internally to properly reapply deployment redirects to an
+    // execution when rebuilding it from history events.
+    int64 deployment_change_counter = 9;
```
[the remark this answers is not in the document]
ShahabT: This will not be set on every task. it will be zero when no redirect has happen on this task which should be 99% of the times. It's only set when this task is being sent to a new build ID.

temporal/api/history/v1/message.proto
```diff
@@ -337,10 +346,18 @@ message ActivityTaskStartedEventAttributes {
...
     temporal.api.failure.v1.Failure last_failure = 5;
     // Version info of the worker to whom this task was dispatched.
+    // Deprecated. This field should be cleaned up when versioning-2 API is removed. [cleanup-experimental-wv]
     temporal.api.common.v1.WorkerVersionStamp worker_version = 6;
     // Used by server internally to properly reapply build ID redirects to an execution
     // when rebuilding it from events.
+    // Deprecated. This field should be cleaned up when versioning-2 API is removed. [cleanup-experimental-wv]
     int64 build_id_redirect_counter = 7;
+    // The worker deployment to which the task was dispatched. Empty means the task
+    // was dispatched to an unversioned worker.
```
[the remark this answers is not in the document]
ShahabT: Good point. I changed the field. Now it's only populated when the deployment is different from that of the wf.

temporal/api/history/v1/message.proto
```diff
@@ -215,10 +216,19 @@ message WorkflowTaskStartedEventAttributes {
...
     // when rebuilding it from events.
+    // Deprecated. This field should be cleaned up when versioning-2 API is removed. [cleanup-experimental-wv]
     int64 build_id_redirect_counter = 7;
+    // When present, it means this task is dispatched to a new deployment and caused
+    // the workflow execution to be redirected to that deployment from this point on.
+    temporal.api.common.v1.WorkerDeployment redirect_deployment = 8;
+    // Used by server internally to properly reapply deployment redirects to an
+    // execution when rebuilding it from history events. This is non-zero if and
+    // only if `redirect_deployment` is present.
+    int64 deployment_redirect_counter = 9;
```
Shivs11, line 231: From the comments, it seems like `deployment_redirect_counter` is present only when we have `redirect_deployment` set. I wonder, from the lens of good API design, if it's worth making a new message encapsulating both these fields? I am no API expert by any means, but I recall discussions within Max and the team that it seems annoying to have multiple fields with individual caveats such as this one. (Unless I am missing the point and these two are actually independent of each other)

ShahabT: I considered that as well. The reason I eventually kept them separate was that `deployment_redirect_counter`  is internal and does not have any use for users and and I didn't want to add an additional nest just for one useful field.

ShahabT:
Merging this now because I need it to work on server changes. If more changes are asked, I'll make them in a future PR.

temporal/api/history/v1/message.proto
Shivs11: Gotcha - This isn't a blocker but I just wanted a confirmation on a quick thing:
- The API being internal/external should not impact the API design right? We should still strive for a cleaner API design (if possible) even if the API is internal
- I don't think it's just one useful field given that we will also be using `redirect_deployment` every-time we have `deployment_redirect_counter` and vice-versa. My thinking was that if we can group two fields together, we should.

Please don't consider this as me requiring this change given our time constraints. I just want to improve my understanding on API design and would love confirmation on these hunches.

temporal/api/common/v1/message.proto
```diff
@@ -156,6 +156,12 @@ message WorkerVersionCapabilities {
     // Later, may include info like "I can process WASM and/or JS bundles"
 }
 
+// Identifies a versioned worker deployment.
+message WorkerDeployment {
+    string deployment_name = 1;
+    string build_id = 2;
+}
```
antlai-temporal, line 163: docstrings?

temporal/api/history/v1/message.proto
antlai-temporal, line 231: I wonder if you need a top level struct for redirect with two fields, one the counter, the other just WorkerDeployment. WorkerDeployment will be used somewhere else, we don't want to pollute it with the internal counter

temporal/api/history/v1/message.proto
```diff
@@ -337,10 +347,19 @@ message ActivityTaskStartedEventAttributes {
...
```
antlai-temporal, line 362: ditto for a top level redirect struct

temporal/api/workflow/v1/message.proto
```diff
@@ -91,25 +92,43 @@ message WorkflowExecutionInfo {
...
+        // Manual override for versioning behavior. Takes precedence over
+        // `versioning_behavior`.
+        temporal.api.enums.v1.VersioningBehavior versioning_behavior_override = 3;
+        // Manual override for deployment. Takes precedence over `deployment`.
+        // Must be set when if and only if `versioning_behavior_override` is PINNED.
+        temporal.api.common.v1.WorkerDeployment deployment_override = 4;
+        // Indicates the workflow is being redirected to a different deployment.
+        // Workflow redirect finishes at the next WorkflowTaskCompleted event.
+        // Tasks that start while workflow has an ongoing redirect, will be
+        // dispatched to `redirecting_deployment`.
```
antlai-temporal, line 129: not clear whether override has preference over redirect in comment

PR #473, opened 2024-11-05
Clarify run ID may not be current in ResetWorkflowExecutionRequest

bergundy:


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -733,6 +733,9 @@ message SignalWithStartWorkflowExecutionResponse {
 
 message ResetWorkflowExecutionRequest {
     string namespace = 1;
+    // The workflow ID and optional run ID to reset to.
+    // The current run will be terminated upon request even if it has a different run ID than the one provided in this
```
yycptt, line 737:
```suggestion
    // The current run will be terminated upon request if it has a different run ID than the one provided in this
```

yycptt (approved):

temporal/api/workflowservice/v1/request_response.proto
bergundy: I think this wording is important because in both cases the current run will be terminated.

yycptt: ah I see what you are saying. makes sense.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -733,6 +733,9 @@ message SignalWithStartWorkflowExecutionResponse {
 
 message ResetWorkflowExecutionRequest {
     string namespace = 1;
+    // The workflow ID and optional run ID to reset to.
```
dandavison, line 736:
```suggestion
    // The workflow ID and optional run ID to reset.
```

temporal/api/workflowservice/v1/request_response.proto
dandavison, line 737: nit FFTI:

```suggestion
    // The current run will be terminated even if it has a different run ID than the one provided in this
```

dandavison: Are we saying the following:

> The provided `workflow_task_finish_event_id` will always be interpreted as addressing an event in the current workflow run, regardless of the presence or value of the run ID in the request?

dandavison: Presumably multiple runs might be terminated, if the request identifies a run far back in the chain?

Maybe some wording like the following?

> The run identified in the request, and any subsequent runs, will be terminated by the reset, and a new run will be created that becomes the current run. If the request does not identify a run, then it is interpreted as referring to the current run.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Hmm.. I think it's actually correct, because you can reset back to a previous run.

dandavison: OK, I had been thinking of it as I "reset _to_ an event ID", but no super strong opinion here, I think it's a little hard to describe in few words.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -733,6 +733,9 @@ message SignalWithStartWorkflowExecutionResponse {
...
+    // The current run will be terminated even if it has a different run ID than the one provided in this
+    // request.
```
dandavison, line 738: Presumably multiple runs might be terminated, if the request identifies a run far back in the chain?

Maybe some wording like the following?

> The run identified by the request, and any subsequent runs, will be terminated by the reset, and a new run will be created that becomes the current run. If the request does not identify a run, then it is interpreted as identifying the current run.

bergundy: > Presumably multiple runs might be terminated, if the request identifies a run far back in the chain?

No, there can only be one current run. Only the current one will be terminated.

bergundy: Let me know if that makes sense or you still want clarification.

dandavison: Haha yes, what I said made zero sense.

It's not a big problem but I'm still finding it slightly unclear. Here's another attempt:

```
// The workflow to reset. If this contains a run ID then the workflow will be reset back to the
// provided event ID in that run. Otherwise it will be reset to the provided event ID in the
// current run. In all cases the current run will be terminated and a new run started.
```

cretz (approved):

PR #475, opened 2024-11-08
 remove extra WorkflowExecutionOptions from WorkflowExecutionOptionsUpdatedEvent and fix comment

carlydf:
remove extra WorkflowExecutionOptions from WorkflowExecutionOptionsUpdatedEvent and fix comment

To reduce history size and use auto_upgrade instead of unpinned terminology

**Breaking changes**

**Server PR**


Shivs11 (approved):
🚀

cretz:
This change is a great example of why we did the right thing doing all of this API work in a branch instead of master.

PR #476, opened 2024-11-14
Remove skip_generate_workflow_task from signal APIs

alexshtin:
**What changed?**
Remove `skip_generate_workflow_task` flag from signal APIs and events.

**Why?**
See https://github.com/temporalio/temporal/pull/6810 for details.

**Breaking changes**
Yes, but shouldn't be used.

**Server PR**
https://github.com/temporalio/temporal/pull/6810 will be merged before this PR.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -472,7 +472,8 @@ message WorkflowExecutionSignaledEventAttributes {
     // Headers that were passed by the sender of the signal and copied by temporal 
     // server into the workflow task.
     temporal.api.common.v1.Header header = 4;
-    reserved 5;
+    // This field is deprecated and never respected. It should always be set to false.
+    bool skip_generate_workflow_task = 5 [deprecated = true];
     // When signal origin is a workflow execution, this field is set.
     temporal.api.common.v1.WorkflowExecution external_workflow_execution = 6;
 }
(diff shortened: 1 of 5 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -472,8 +472,7 @@ message WorkflowExecutionSignaledEventAttributes {
     // Headers that were passed by the sender of the signal and copied by temporal 
     // server into the workflow task.
     temporal.api.common.v1.Header header = 4;
-    // Indicates the signal did not generate a new workflow task when received.
-    bool skip_generate_workflow_task = 5;
+    reserved 5;
```
cretz, line 475: I do not believe this is a safe/compatible change. We have an obligation to have JSON/field-name compatibility, not just proto binary compatibility, especially for history but also for RPC requests since we have an HTTP API. Can we just document that it's deprecated/ignored?

pdoerner (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -654,8 +654,7 @@ message SignalWorkflowExecutionRequest {
     // Headers that are passed with the signal to the processing workflow.
     // These can include things like auth or tracing tokens.
     temporal.api.common.v1.Header header = 8;
-    // Indicates that a new workflow task should not be generated when this signal is received.
-    bool skip_generate_workflow_task = 9;
+    reserved 9;
```
cretz, line 657: I think we need to do the same thing here as we did with the history, we need to deprecate/document but not remove the field. These things are part of the HTTP API and need to be JSON compatible.

cretz (approved):
Approving because this is _probably_ safe, but overall we should be deprecating fields, not removing them. And we shouldn't have inconsistent deprecation approaches based on how we _think_ users may be using our public, stable, released API models.

PR #477, opened 2024-11-14
Add VersioningBehaviorOverride to StartWorkflowExecutionRequest

carlydf:
Add VersioningBehaviorOverride to StartWorkflowExecutionRequest

So that users can implement their own deployment checks by sending a test workflow to a specific deployment, before ramping traffic on that deployment.

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -713,9 +713,11 @@ message SignalWithStartWorkflowExecutionRequest {
     // for use by user interfaces to display the fixed as-of-start summary and details of the
     // workflow.
     temporal.api.sdk.v1.UserMetadata user_metadata = 23;
-
     // Links to be associated with the WorkflowExecutionStarted and WorkflowExecutionSignaled events.
     repeated temporal.api.common.v1.Link links = 24;
+    // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
+    // To unset the override after the workflow is running, use UpdateWorkflowExecutionOptions.
+    temporal.api.common.v1.VersioningBehaviorOverride versioning_behavior_override = 25;
 }
 
 message SignalWithStartWorkflowExecutionResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -39,12 +39,14 @@ import "temporal/api/enums/v1/failed_cause.proto";
 import "temporal/api/enums/v1/query.proto";
 import "temporal/api/enums/v1/reset.proto";
 import "temporal/api/enums/v1/task_queue.proto";
+import "temporal/api/enums/v1/deployment.proto";
 import "temporal/api/enums/v1/update.proto";
 import "temporal/api/activity/v1/message.proto";
 import "temporal/api/common/v1/message.proto";
 import "temporal/api/history/v1/message.proto";
 import "temporal/api/workflow/v1/message.proto";
 import "temporal/api/command/v1/message.proto";
+import "temporal/api/deployment/v1/message.proto";
 import "temporal/api/failure/v1/message.proto";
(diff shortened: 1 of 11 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -201,6 +201,9 @@ message StartWorkflowExecutionRequest {
     temporal.api.sdk.v1.UserMetadata user_metadata = 23;
     // Links to be associated with the workflow.
     repeated temporal.api.common.v1.Link links = 24;
+    // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
+    // To unset the override after the workflow is running, use UpdateWorkflowExecutionOptions.
+    temporal.api.common.v1.VersioningBehaviorOverride versioning_behavior_override = 25;
```
cretz, line 206: Does this need to apply to `SignalWithStartWorkflowExecutionRequest` for signal with start and `NewWorkflowExecutionInfo` for schedules?

carlydf: This is primarily to enable customers to write canary workflows and start/run them on a specific worker deployment that they want to test. I could see people wanting to do that with schedules, not sure if people would want to do that with `SignalWithStart`, but there's no harm in giving them the option

carlydf: I just added it to both of those

carlydf: I think it will also need to be added to `internal.StartWorkflowOptions` in `sdk-go`

antlai-temporal:
One thing about error handling. The override specifies a deployment and a build id. We also pass the task queue name when starting a workflow. Is there any check in the server that the task queue belongs to the deployment? Maybe the answer is no check and the workflow times out, but a comment on this would be useful...

antlai-temporal (approved):

temporal/api/workflowservice/v1/request_response.proto
cretz: From an SDK POV, probably to every start workflow options in every language (Go is not special)

cretz (approved):

Shivs11:
> One thing about error handling. The override specifies a deployment and a build id. We also pass the task queue name when starting a workflow. Is there any check in the server that the task queue belongs to the deployment? Maybe the answer is no check and the workflow times out, but a comment on this would be useful...

@antlai-temporal - asking just to ensure we are on the same page - you specifically mean using the override during an update to get out of bad situations right? I think care should be taken on the server side and we will have a check to verify that the task-queue is in the right deployment before we do move it out and place it into a different one

antlai-temporal:
> > One thing about error handling. The override specifies a deployment and a build id. We also pass the task queue name when starting a workflow. Is there any check in the server that the task queue belongs to the deployment? Maybe the answer is no check and the workflow times out, but a comment on this would be useful...
> 
> @antlai-temporal - asking just to ensure we are on the same page - you specifically mean using the override during an update to get out of bad situations right? I think care should be taken on the server side and we will have a check to verify that the task-queue is in the right deployment before we do move it out and place it into a different one

No, this API is to start a new workflow on  a particular deployment. When you start a workflow you also need to specify the task queue. The question is if we make a mistake the deployment may not include that queue. Since task queue membership in a deployment is eventually consistent (triggered by worker polling), we may not want to throw an error, and wait until the workflow timeout...

ShahabT (approved):

ShahabT:
> we may not want to throw an error, and wait until the workflow timeout

Correct. This is possible for pinned overrides. If they pinned to the wrong deployment its like starting the wf on a non-existing task queue. We assume user knows what they are doing. In UI we'll show a watning that "there is no poller for this TQ + deployment".

ShahabT:
@carlydf I think you should add the override to `WorkflowExecutionStartedEventAttributes` as well.

PR #478, opened 2024-11-16
Add `WORKFLOW_EXECUTION_OPTIONS_UPDATED` event type and remove FieldMask from `WorkflowExecutionOptionsUpdated` Event

carlydf:
- Add workflow options updated event type
- Remove field mask from WorkflowExecutionOptionsUpdatedEvent and only store resulting options. When memo is added to `WorkflowExecutionOptions`, we will only store it in this event when it was upserted, so that we don't increase history size by storing memo on every OptionsUpdate. This works because Memo is never nil, only empty or populated, so a nil Memo in the event can mean "no change".
- Change comments on `ActivityPropertiesModifiedExternallyEventAttributes` to indicate it was never used and is not intended to be used in the future.

To implement the `UpdateWorkflowExecutionOptions` API

**Breaking changes**

**Server PR**


temporal/api/history/v1/message.proto
```diff
@@ -760,10 +760,10 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
...
     // Workflow Execution options that were sent in the UpdateWorkflowExecutionOptions request.
     // Partial updates are accepted and controlled by update_mask.
-    temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options_update = 1;
+    temporal.api.workflow.v1.WorkflowExecutionOptions options = 1;
 
     // Update mask that was sent in the UpdateWorkflowExecutionOptions request.
-    // Indicates which fields from `workflow_execution_options_for_update` were applied.
+    // Indicates which fields from `options` were applied.
     google.protobuf.FieldMask update_mask = 2;
 }
```
cretz, line 768: So I didn't notice or I guess didn't say anything when this first appeared, but processing a field mask may be a bit rough to ask every SDK to do. Is it too much history to provide the full set of options as they now sit? Or maybe we can just document this field mask here that 1) it will always be set with the updated fields (i.e. even if the user did not set it at RPC time because https://google.aip.dev/134#request-message does not require it), and 2) it will always be a fixed, non-nested field (i.e. never have dots or wildcards or any of that.

cretz (commented):
LGTM, but I added a note where I think we need to at least clarify some simple rules for SDK on the field mask if we're going to start using field masks in history (sorry I did not see this earlier).

temporal/api/history/v1/message.proto
carlydf: hm yeah actually I was considering storing just the resulting options instead of the mask + request options.. I think the biggest thing would maybe be memo when/if we add memo to these? Do you know how big memo can be

carlydf: it actually would be slightly less repeated work in the server if I store the merged options only and no field..

ShahabT: I don't have a strong preference. For practical reasons, I'd suggest to get rid of mask for now and punt the mask decision to a later time as and if more fields are added. This is so we can move forward with the versioning work in the very tight schedule that we have. (Right now, there is only one field in WorkflowExecutionOptions and this event has to change that only field.)

ShahabT: I'd say we should support it. We can punt it to public preview though. The API can return "unimplimented" error in those cases meanwhile.

cretz: > Do you guys think it's ok if I want to say that asterisk and empty field mask is not supported

Can we include dotted properties at least for now (we can change the comment later if we decide to support more)?

> I'd say we should support it

I think there is confusion between whether it's supported in the RPC call and whether it's supported in history. I propose support whatever advanced form you want in the RPC call in the server, but turn it into a fixed, more limited form in history. The server should be able to know which fields the mask references and just enumerate those as fixed strings inside of here instead of just blindly passing over whatever wildcards the user passed to the server.

ShahabT: +1 for both of @cretz's comments.

carlydf: I've thought through both sides of this (wanting to support the recommendations vs wanting to reduce complexity and offer the API that guides our users to do the right thing), and actually I think we should not support asterisk = "PUT all fields".

Lazy people will just put "*" in their code right now, since there is only one field in the options, and then later when we add more fields, their old code will suddenly null things such as timeout or memo that it didn't before. I think that would be bad, and I could see users actually doing that and running into issues.

cretz: FWIW, asterisk is probably easy to support RPC side, it's just the normalization-for-history step that may take some effort. But I'm also ok limiting RPC side. Can you do me a favor and make clear what the absence of the update mask _in history_ means? (I know it means "all fields" in RPC call, just wanting to know whether you want the same behavior for this field in history event)

carlydf: @antlai-temporal and I are proposing this for the Go SDK:
```
["versioning_behavior_override"] -> invalid
["foo"]                                              -> invalid
["*"]                                                  -> invalid
[]                                                       -> valid, means "PUT nothing, GET the full current options"
["VersioningBehaviorOverride"]   -> valid
["VersioningBehaviorOverride.Behavior"] -> invalid
["VersioningBehaviorOverride.anything"] -> invalid
```

The RPC will differ slightly in that:
```
["versioning_behavior_override"] -> valid
["VersioningBehaviorOverride"]   -> invalid
```

My reason for not supporting `*` is above, also I think there is not much benefit, since it would be rare that a user wanted to update memo AND priority AND versioning behavior AND timeouts all at once.

My reason for not supporting dot notation / partial updates, is that it becomes fraught when the struct being partially-updated can have field-combinations that are invalid, and maybe two callers are updating it and don't realize that one person just changed it. The structs are simple enough that I don't see a convincing use case for supporting partial struct updates and adding a lot of complexity.

cretz: > I think an empty field to update everything is really dangerous.

We're on a thread about the history field here but I suppose we're also talking about the RPC.

I disagree. People need to be able to put the whole object by default and patch non-default. I believe this is what's happening with activity options I think today and what https://google.aip.dev/134#request-message says to follow (where the concept of using field masks basically came from). If you want the SDK API to disallow this, that's fine, but people expect to be able to opt-in to partial updates, not out. The only thing that could maybe make sense is failing on empty field mask. But turning an update into a getter with an empty field mask does not make sense IMO and may end up making us inconsistent with ourselves and others expected use of field masks in gRPC.

carlydf: Oh, I think we are interpreting this differently:

> The field must be optional, and the service must treat an omitted field mask as an implied field mask equivalent to all fields that are populated (have a non-empty value).

My interpretation of that rule is that an "omitted field mask" is different from an "empty field mask". Where "omitted field mask" means `update_mask == nil` and "empty field mask" means `update_mask.GetPaths() = []string{}`

carlydf: Apologies for jumping between the history event question and the RPC question, I'm just in the middle of implementing the RPC and want to decide what to accept there.

temporal/api/history/v1/message.proto
```diff
@@ -758,28 +757,21 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
...
-
-    // Update mask that was sent in the UpdateWorkflowExecutionOptions request.
-    // Indicates which fields from `workflow_execution_options_for_update` were applied.
-    google.protobuf.FieldMask update_mask = 2;
+    // Versioning override in the mutable state after event has been applied.
+    temporal.api.common.v1.VersioningBehaviorOverride versioning_override = 1;
 }
 
+// Not used anywhere. Use case is replaced by WorkflowExecutionOptionsUpdatedEventAttributes
 message WorkflowPropertiesModifiedExternallyEventAttributes {
```
cretz, line 765: I see we decided to not reuse this event and add your field to the bottom. Works for me (and I agree we can't really delete this), though would have also been ok just adding your field. Unsure if others have opinions here or not.

cretz (approved):

temporal/api/history/v1/message.proto
```diff
@@ -758,28 +757,21 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
...
 message WorkflowExecutionOptionsUpdatedEventAttributes {
-    // Workflow Execution options that were sent in the UpdateWorkflowExecutionOptions request.
-    // Partial updates are accepted and controlled by update_mask.
-    temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options_update = 1;
-
-    // Update mask that was sent in the UpdateWorkflowExecutionOptions request.
-    // Indicates which fields from `workflow_execution_options_for_update` were applied.
-    google.protobuf.FieldMask update_mask = 2;
+    // Versioning override in the mutable state after event has been applied.
+    temporal.api.common.v1.VersioningBehaviorOverride versioning_override = 1;
```
antlai-temporal, line 761: This is moving to common.v1.VersioningOverride as discussed. Do you want to include that  change in this PR?

antlai-temporal (approved):

temporal/api/history/v1/message.proto
carlydf: Shahab has an open PR with that change, so I'm going to let it happen there to reduce churn / merge issues

PR #479, opened 2024-11-17
Add Deployment Pre-Release APIs

ShahabT:
**What changed?**
Added the following APIs:
- DescribeDeployment
- ListDeployments
- GetCurrentDeployment
- SetCurrentDeployment (low-level placeholder for Rollout API)
- GetDeploymentReachability

**Why?**
Needed for Worker Versioning V3.

**Breaking changes**

**Server PR**


temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Holds information about a worker deployment.
+message DeploymentInfo {
+    Deployment deployment = 1;
+    enums.v1.DeploymentStatus status = 2;
+    google.protobuf.Timestamp status_change_time = 3;
+    // A user-defined set of fields. Can be updated via other write operations to the deployment,
+    // such as SetCurrentDeployment.
+    common.v1.Memo memo = 4;
```
cretz, line 54: Memo has a very specific workflow-visibility meaning. We have tried to avoid arbitrary user-defined fields lately (e.g. user metadata project didn't allow them), but if you must have it, can you just have a map instead of reusing the "memo" term here?

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+    repeated TaskQueueInfo task_queues_info = 5;
```
cretz, line 55:
```suggestion
    repeated TaskQueueInfo task_queue_infos = 5;
```

Pedantic, but this plural reads a bit better IMO, as it is not a single info about multiple task queues but rather multiple task queue info objects

temporal/api/history/v1/message.proto
```diff
@@ -255,7 +256,7 @@ message WorkflowTaskCompletedEventAttributes {
 message CompletedDeploymentRedirectInfo {
     // The target deployment of the redirect. Null means a previously versioned workflow just
     // completed a redirect to unversioned workers.
-    temporal.api.common.v1.WorkerDeployment deployment = 1;
+    deployment.v1.Deployment deployment = 1;
```
cretz, line 259: Everywhere else I think we fully qualify the package even if we technically can use a relative name, we might as well remain consistent (it is harmless to do so)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1737,3 +1739,69 @@ message UpdateWorkflowExecutionOptionsResponse {
...
+message DescribeDeploymentResponse {
+    deployment.v1.DeploymentInfo deployment_info = 1;
+}
+
+message ListDeploymentsRequest {
+    int32 page_size = 1;
+    bytes next_page_token = 2;
+    string namespace = 3;
+    // Optional. Use to filter based on deployment name.
+    string deployment_name = 4;
```
cretz, line 1757: Can you clarify whether this is exact-match string or some form of partial match?

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+// deployment name + build ID. Both values are required.
+message Deployment {
+    string deployment_name = 1;
+    string build_id = 2;
+}
+
+// Holds information about a worker deployment.
+message DeploymentInfo {
+    Deployment deployment = 1;
+    enums.v1.DeploymentStatus status = 2;
```
cretz, line 50: You are a bit inconsistent with your prefixes. For instance, in some objects you may have `deployment_name` on a deployment or and `task_queue_type` on a task queue info, but in others you may have just `status` or `memo` on a deployment info instead of `deployment_status` and `deployment_memo`. No opinion on which approach is best, just was really noticeable reading the API.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1737,3 +1739,69 @@ message UpdateWorkflowExecutionOptionsResponse {
...
+    repeated DeploymentInfo deployments = 2;
+
+    message DeploymentInfo {
+        deployment.v1.Deployment deployment = 1;
+        enums.v1.DeploymentStatus status = 2;
+        google.protobuf.Timestamp status_change_time = 3;
+    }
+}
+
+message SetCurrentDeploymentRequest {
```
cretz, line 1771: Is this call idempotent? If there are any concerns, you may want a `request_id` on this since our client may call this multiple times on certain gRPC codes.

temporal/api/workflowservice/v1/request_response.proto
Shivs11: I believe this will be an exact-match string. If absent, the response would just be all deployments present in a single namespace

Shivs11 (approved):

temporal/api/deployment/v1/message.proto
ShahabT: Good feedback. I'll make them consistent.

temporal/api/deployment/v1/message.proto
ShahabT: Makes sense, will change it to a `metadata` map.

temporal/api/deployment/v1/message.proto
ShahabT: I was not sure which is the right way, mostly because I was not sure about grammatically correctness of `infos`. But now after your comment (as a native speaker), I think I got my answer and |'ll change all `xs_info` to `x_infos` in this PR.

temporal/api/history/v1/message.proto
ShahabT: The verboseness bothered me a little, but I'll change it to remain consistent.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: I believe it is idempotent. Will leave it to @dnr and @Shivs11  to add request_id if they realize it is needed during the implementation.

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+
+import "google/protobuf/timestamp.proto";
+
+import "temporal/api/enums/v1/deployment.proto";
+import "temporal/api/enums/v1/task_queue.proto";
+
+// Identifies a worker deployment. A worker deployment is identified by the combination of
+// series name + build ID. Both values are required.
+message Deployment {
+    string series_name = 1;
```
Shivs11, line 42: just a thought - is it confusing for customers to have to see some places have `deployment_series_name` and `series_name` when they mean the same thing? I understand how this current attribute resolves to Deployment.series_name but just a thought I had right now...

Shivs11 (approved):

temporal/api/deployment/v1/message.proto
carlydf: I vote `Deployment.series_name` but it's a gentle vote :)

ShahabT: The only places that `deployment_series_name` is used is in the WorkerVersionStamp and WorkerVersionCapabilities structs and those are not something that user will interact with. I have another PR planned to get rid of these structs in the events and every other user-facing place. We should just talk in terms of `Deployment` and `VersioningBehavior`.

carlydf: I could see it either way / I don't care either way

temporal/api/enums/v1/deployment.proto
```diff
@@ -0,0 +1,66 @@
...
+// Status of a deployment within its deployment series. Any given deployment series can at most have
+// one `RAMPING` deployment, and one `CURRENT` deployment.
+//
+// (-- api-linter: core::0216::synonyms=disabled
+//     aip.dev/not-precedent: This enum specifies the status of the deployment among all deployments
+//     in the same series, therefore `DeploymentStatus` is a better name that `DeploymentState`. --)
+enum DeploymentStatus {
+    DEPLOYMENT_STATUS_UNSPECIFIED = 0;
+    // The deployment does not have a special status.
+    DEPLOYMENT_STATUS_NO_STATUS = 1;
```
antlai-temporal, line 43: Why do we need this? is UNSPECIFIED not enough? Clarify in the comments...

temporal/api/enums/v1/deployment.proto
```diff
@@ -0,0 +1,66 @@
...
+    DEPLOYMENT_REACHABILITY_OPEN_WORKFLOWS = 1;
+    // The deployment is not reachable by open workflows but might be still needed by closed
+    // workflows who can only run Queries on this deployment. The deployment can be decommissioned
+    // safely if user does not query closed workflows.
+    DEPLOYMENT_REACHABILITY_CLOSED_WORKFLOWS_ONLY = 2;
+    // The deployment is not reachable by any workflow because all the workflows who needed this
+    // deployment went out of retention period. The deployment can be decommissioned
+    // safely.
+    DEPLOYMENT_REACHABILITY_UNREACHABLE = 3;
+}
```
carlydf, line 66: We should say something about PINNED / AUTO_UPDRADE here I think..
I want there to be a reachability status that says, even if zero workflows have been started, if this deployment is RAMPING or CURRENT, it is reachable by NEW UNPINNED workflows.
```suggestion
// Specify the reachability level for a deployment so users can decide if it is time to
// decommission the deployment. 
enum DeploymentReachability {
    // Reachability level is not specified.
    DEPLOYMENT_REACHABILITY_UNSPECIFIED = 0;
    // The deployment is reachable by PINNED open workflows that can only run on this deployment, and
    // by new AUTO_UPGRADE workflows as long as the deployment's status is CURRENT or RAMPING.
    // The deployment cannot be decommissioned safely.
    DEPLOYMENT_REACHABILITY_NEW_AND_OPEN_WORKFLOWS = 1;
    // The deployment is reachable by PINNED open workflows that can only run on this deployment.
    // The deployment cannot be decommissioned safely.
    DEPLOYMENT_REACHABILITY_OPEN_WORKFLOWS = 2;
    // The deployment is not reachable by PINNED open workflows but might be still needed by PINNED 
    // closed workflows who can only run Queries on this deployment. The deployment can be decommissioned
    // safely if user does not query PINNED closed workflows.
    DEPLOYMENT_REACHABILITY_CLOSED_WORKFLOWS_ONLY = 3;
    // The deployment is not reachable by any workflow because all the PINNED workflows who needed
    // this deployment went out of retention period. The deployment can be decommissioned safely.
    DEPLOYMENT_REACHABILITY_UNREACHABLE = 4;
}
```

temporal/api/workflow/v1/message.proto
```diff
@@ -107,29 +108,27 @@ message WorkflowExecutionInfo {
...
-        temporal.api.common.v1.WorkerDeployment deployment = 2;
-        // Manual override for execution's versioning behavior. Takes precedence over `behavior`.
-        temporal.api.enums.v1.VersioningBehavior behavior_override = 3;
-        // Used to manually pin the execution to a deployment. Must be set when if and only if
-        // `behavior_override` is PINNED. Takes precedence over `deployment`.
-        temporal.api.common.v1.WorkerDeployment deployment_override = 4;
+        temporal.api.deployment.v1.Deployment deployment = 2;
+        // Manual override for execution's versioning behavior. Takes precedence over `behavior`
+        // and `deployment`.
+        VersioningBehaviorOverride manual_override = 3;
```
carlydf, line 119: Thanks, I like this :)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1737,3 +1739,69 @@ message UpdateWorkflowExecutionOptionsResponse {
...
+
+message ListDeploymentsResponse {
+    bytes next_page_token = 1;
+    repeated DeploymentInfo deployments = 2;
+
+    message DeploymentInfo {
+        deployment.v1.Deployment deployment = 1;
+        enums.v1.DeploymentStatus status = 2;
+        google.protobuf.Timestamp status_change_time = 3;
+    }
```
carlydf, line 1768: why redefine `DeploymentInfo` without metadata and task_queue_infos here?

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+
+// Holds information about a worker deployment.
+message DeploymentInfo {
+    Deployment deployment = 1;
+    temporal.api.enums.v1.DeploymentStatus status = 2;
+    google.protobuf.Timestamp status_change_time = 3;
+    // A user-defined set of key-values. Can be updated as part of write operations to the
+    // deployment, such as `SetCurrentDeployment`.
+    map<string, bytes> metadata = 4;
+    repeated TaskQueueInfo task_queue_infos = 5;
```
antlai-temporal, line 54: Ditto about the metadata values being strings...

temporal/api/workflowservice/v1/request_response.proto
carlydf: so the list is smaller? is the idea that if the caller wants more detailed info they would subsequently call `DescribeDeployment`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1737,3 +1739,69 @@ message UpdateWorkflowExecutionOptionsResponse {
...
+    string identity = 3;
+    // Optional. Use to add or remove user-defined metadata entries. Metadata entries are exposed
+    // when describing a deployment. It is a good place for information such as operator name,
+    // links to internal deployment pipelines, etc.
+    deployment.v1.UpdateDeploymentMetadata update_metadata = 4;
+}
+
+message SetCurrentDeploymentResponse {
+    deployment.v1.DeploymentInfo current_deployment = 1;
+}
```
antlai-temporal, line 1784: Do we want to return the previous deployment just overiden instead? It may be more useful that return what we just passed, and we also have GetCurrent...

temporal/api/workflowservice/v1/service.proto
```diff
@@ -772,6 +772,62 @@ service WorkflowService {
...
+            get: "/namespaces/{namespace}/deployments"
+            additional_bindings {
+                get: "/api/v1/namespaces/{namespace}/deployments"
+            }
+        };
+    }
+
+    // Returns the reachability level of a worker deployment to help users decide when it is time
+    // to decommission a deployment. Reachability level is calculated based on the deployment's
+    // `status` and existing workflows that depend on the given deployment for their execution.
```
antlai-temporal, line 798: I'm assuming that we don't return reachability in `DescribeDeployment` because it is expensive. Maybe a comment about costs here could clarify...

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Also we don't want to make the list API expensive. For example, TQ info will not be immediately accessible from Visibility.

temporal/api/workflow/v1/message.proto
```diff
@@ -364,5 +363,18 @@ message NexusOperationCancellationInfo {
...
-}
\ No newline at end of file
+    VersioningBehaviorOverride versioning_behavior_override = 1;
+}
+
+// Used to manually override the versioning behavior of a specific workflow execution. If set, takes
+// precedence over the SDK-sent Versioning Behavior.
+// To remove the override, call `UpdateWorkflowExecutionOptions` with a null
+// `VersioningBehaviorOverride`, and use the `update_mask` to indicate that it should be mutated.
+message VersioningBehaviorOverride {
```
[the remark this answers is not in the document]
ShahabT: or just `VersioningOverride`?

antlai-temporal (approved):

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: Can you call it something different to avoid confusion, like DeploymentSummary or DeploymentBriefs or something like that...

temporal/api/deployment/v1/message.proto
Shivs11: marking the conversation as "resolved" because I agree with what was said

carlydf (approved):

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+
+    message TaskQueueInfo {
+        string name = 1;
+        temporal.api.enums.v1.TaskQueueType type = 2;
+        // When server saw the first poller for this task queue in this deployment.
+        google.protobuf.Timestamp first_poller_time = 3;
```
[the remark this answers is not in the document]
ShahabT: The message is already nested in `DeploymentInfo` so I thought it's less verbose like this. But please lmk if you think it should still be renamed to `DeploymentTaskQueueInfo`.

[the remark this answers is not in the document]
ShahabT: 👍 That's possible and I agree we should have it. Will create a task for it (can't do right away because of more technical work).

temporal/api/workflow/v1/message.proto
```diff
@@ -107,29 +108,27 @@ message WorkflowExecutionInfo {
...
         // The worker deployment to which this workflow is assigned currently.
         // Must be present if and only if `behavior` is set.
-        temporal.api.common.v1.WorkerDeployment deployment = 2;
-        // Manual override for execution's versioning behavior. Takes precedence over `behavior`.
-        temporal.api.enums.v1.VersioningBehavior behavior_override = 3;
-        // Used to manually pin the execution to a deployment. Must be set when if and only if
-        // `behavior_override` is PINNED. Takes precedence over `deployment`.
-        temporal.api.common.v1.WorkerDeployment deployment_override = 4;
+        temporal.api.deployment.v1.Deployment deployment = 2;
+        // Manual override for execution's versioning behavior. Takes precedence over `behavior`
```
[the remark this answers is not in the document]
carlydf: WorkflowExecutionOverride maybe? I feel like "WorkflowSpecific" is a bit ambiguous because "workflow" could mean Workflow Execution or Workflow Type/Workflow Definition

temporal/api/enums/v1/deployment.proto
```diff
@@ -0,0 +1,66 @@
...
+//     aip.dev/not-precedent: This enum specifies the status of the deployment among all deployments
+//     in the same series, therefore `DeploymentStatus` is a better name that `DeploymentState`. --)
+enum DeploymentStatus {
+    DEPLOYMENT_STATUS_UNSPECIFIED = 0;
+    // The deployment does not have a special status.
+    DEPLOYMENT_STATUS_NO_STATUS = 1;
+    // The deployment is the ramping deployment for the deployment series.
+    DEPLOYMENT_STATUS_RAMPING = 2;
+    // The deployment is the current deployment for the deployment series.
+    DEPLOYMENT_STATUS_CURRENT = 3;
```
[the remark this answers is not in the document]
drewhoskins-temporal: We've thought about some of these cases:
- for task queues, at least on our current roadmap, we'll insist that users create a different deployment series if they want to rollout task queues separately.  (Note that we will support the ability to move a task queue from one deployment series to another)  We don't see this as being a big burden for such an advanced use case.
- for workflow types, we've decided to lean on the UpdateWorkflowOptions for these power-user scenarios for now.  But you could imagine adding more fine-grained control to the `rollout` API.  (i.e. instead of a single % ramp, we could provide various ramps for different types.  Shahab mentioned there could be multi-cursor problems to think about with such, though).  In that world, we could keep the high-level status of `DEPLOYMENT_STATUS_RAMPING` and create a more fine-grained API for power users.  So, it seems premature to think too deeply now.

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
ShahabT: Got rid of the "manual" in the field name and comment. Mentioned "execution-specific" in the comment. Renamed the message to `VersioningOverride` as discussed before. Also renamed the field name to be consistent to the message name.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1737,3 +1739,69 @@ message UpdateWorkflowExecutionOptionsResponse {
...
+    repeated DeploymentInfo deployments = 2;
+
+    message DeploymentInfo {
+        deployment.v1.Deployment deployment = 1;
+        enums.v1.DeploymentStatus status = 2;
+        google.protobuf.Timestamp status_change_time = 3;
+    }
+}
+
+message SetCurrentDeploymentRequest {
```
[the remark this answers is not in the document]
ShahabT: I annotated this and other APIs as "Experimental" in `service.proto` so it's clear that they all can change or be removed in the next release.

As @drewhoskins-temporal said, this one in particular is merely a low-level placeholder until the rollout API arrives before public preview.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Good point. Added, although the fate of this API is unknown.

temporal/api/workflowservice/v1/service.proto
ShahabT: Added a comment about expensiveness of the call and the possibility of returning cached values.

temporal/api/enums/v1/deployment.proto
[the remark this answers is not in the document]
ShahabT: Adding to @drewhoskins-temporal comments, basically we envision eventually adding task queue and workflow type granularity to **Rollout** so user can ramp them one-by-one. However, setting different **current** deployment for different workflow types or task queues is not currently on the road map and users should create new deployment series for such cases.

temporal/api/deployment/v1/message.proto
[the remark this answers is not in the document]
mfateev: Message names are top level. So someone looking at the name will not see where it is used.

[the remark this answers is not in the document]
ShahabT: @mfateev This message name is nested, not top level. Are suggesting we should not have nested message definitions and only top level?

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+message Deployment {
+    string series_name = 1;
+    string build_id = 2;
+}
+
+// Holds information about a worker deployment.
+message DeploymentInfo {
+    Deployment deployment = 1;
+    temporal.api.enums.v1.DeploymentStatus status = 2;
+    google.protobuf.Timestamp last_became_current_time = 3;
```
drewhoskins-temporal, line 50: This is confusingly worded and perhaps the concept isn't quite right.  
Would it be 0 if status is not equal to current?
And why do users care about that specific status transition, but not the others?

It's unclear to me where timestamps should go--on the rollout, on the deployment, or both?
For example, we might link to the associated rollouts, once we have that.

Do we need this for pre-release?  Maybe cut for now, or just stick with the normal "created time" and we can add stuff later.

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+// series name + build ID. Both values are required.
+message Deployment {
+    string series_name = 1;
+    string build_id = 2;
+}
+
+// Holds information about a worker deployment.
+message DeploymentInfo {
+    Deployment deployment = 1;
+    temporal.api.enums.v1.DeploymentStatus status = 2;
```
drewhoskins-temporal, line 49: This seems like a denormalization of the rollout status. Which might be fine, but it's also blurring the lines between a rollout and a deployment in a way that's a bit confusing.  (why isn't it `RolloutStatus`) ?
And it's making deployments stateful when perhaps rollouts are the stateful thing.
I'm okay thinking of this as a placeholder for pre-release, but let's flag this to revisit.

temporal/api/deployment/v1/message.proto
ShahabT: Initially it was more generic and was called `status_change_time`. But I realized for internal use while there is no rollout yet, we also need timestamp of the last time that a deployment becomes current. But you are right, that info should not go to the public API. Will revert this to status_change_time.

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+    google.protobuf.Timestamp last_became_current_time = 3;
+    // A user-defined set of key-values. Can be updated as part of write operations to the
+    // deployment, such as `SetCurrentDeployment`.
+    map<string, string> metadata = 4;
```
cretz, line 53: Should consider making the values payloads if users should be allowed to encrypt this information

cretz (approved):
Only one additional minor suggestion, otherwise LGTM (knowing it's going into non `master` and may be adjusted as implementation dictates)

temporal/api/enums/v1/deployment.proto
```diff
@@ -0,0 +1,66 @@
...
+option java_outer_classname = "DeploymentProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+// Status of a deployment within its deployment series. Any given deployment series can at most have
+// one `RAMPING` deployment, and one `CURRENT` deployment.
+//
+// (-- api-linter: core::0216::synonyms=disabled
+//     aip.dev/not-precedent: This enum specifies the status of the deployment among all deployments
+//     in the same series, therefore `DeploymentStatus` is a better name that `DeploymentState`. --)
```
drewhoskins-temporal, line 39: I don't understand the reasoning here.

temporal/api/deployment/v1/message.proto
drewhoskins-temporal: Thinking about this more, until we design the state machine, maybe best is just to have an `is_current` boolean that we plan to deprecate.  It might be more challenging to evolve if we put in this enum, which might be wrong in some way.

ShahabT: I suspected that in the UI, for example, we would want to label statuses such as RAMPING and CURRENT so the Describe and List APIs might want to return these info in some format. It does not necessarily mean that the Deployment itself is stateful, as the status could be logically populated based on the Rollout records.

For now, I can get rid of the status and the change timestamp if you prefer. That means for now the only way for the user to get currentness of a deployment is to compare it with the result of GetCurrentDeployment.

temporal/api/workflowservice/v1/request_response.proto
drewhoskins-temporal: This also seems odd to me.  Why do we care about making this API expensive?

ShahabT: @drewhoskins-temporal are you suggesting the list should return complete info of all deployments?

ShahabT: I can change the name to `DeploymentListEntry` or `DeploymentSummary`. Same pattern is used for ListSchedules where we return a abbreviated set of info in the list response: https://github.com/temporalio/api/blob/a3ed90f87395531e19467f8680d3bcbd34c7ea06/temporal/api/schedule/v1/message.proto#L372

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1737,3 +1739,70 @@ message UpdateWorkflowExecutionOptionsResponse {
...
+message DescribeDeploymentRequest {
+    string namespace = 1;
+    deployment.v1.Deployment deployment = 2;
+}
+
+message DescribeDeploymentResponse {
+    deployment.v1.DeploymentInfo deployment_info = 1;
+}
+
+message ListDeploymentsRequest {
```
Shivs11, line 1752: nit: For consistency purposes with other similar protos, we should maybe have this structure as:

`message ListDeploymentsRequest {
    string namespace = 1;
    int32 max_page_size = 2;
    bytes next_page_token = 3;
   
    // Optional. Use to filter based on exact series name match.
    string series_name = 4;
}
`

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+        string name = 1;
+        temporal.api.enums.v1.TaskQueueType type = 2;
+        // When server saw the first poller for this task queue in this deployment.
+        google.protobuf.Timestamp first_poller_time = 3;
+    }
+}
+
+// Used as part of Deployment write APIs to update metadata attached to a deployment.
+message UpdateDeploymentMetadata {
+    map<string, bytes> upsert_entries = 1;
```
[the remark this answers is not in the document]
ShahabT: Ended up making them Payload similar to the memo in workflows and schedules.

temporal/api/deployment/v1/message.proto
ShahabT: Makes sense. changed it to Payload, now it is consistent with workflow and schedules memo.

temporal/api/enums/v1/deployment.proto
ShahabT: getting rid of this enum for now. just putting an `is_current` bool in DeploymentInfo until we finish the rollout API design.

temporal/api/deployment/v1/message.proto
```diff
@@ -0,0 +1,69 @@
...
+import "temporal/api/enums/v1/task_queue.proto";
+
+// Identifies a worker deployment. A worker deployment is identified by the combination of
+// series name + build ID. Both values are required.
+message Deployment {
+    string series_name = 1;
+    string build_id = 2;
+}
+
+// Holds information about a worker deployment.
```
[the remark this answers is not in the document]
ShahabT: Added more information for both Deployment and DeploymentInfo.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Changed it to DeploymentListInfo similar to schedules.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1737,3 +1739,70 @@ message UpdateWorkflowExecutionOptionsResponse {
...
+    int32 page_size = 1;
+    bytes next_page_token = 2;
+    string namespace = 3;
+    // Optional. Use to filter based on exact series name match.
+    string series_name = 4;
+}
+
+message ListDeploymentsResponse {
```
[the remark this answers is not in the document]
ShahabT: This is going to be **roughly** sorted based on deployment creation time. Although if a deployment has a very high number of TQs (or very high number of events happening to it) it will CaN and the sort will be mixed up. In reality we are relying on Visibility sort which is based on WF start time.

I prefer to keep this unspecified until we have a guaranteed order in the future.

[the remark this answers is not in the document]
drewhoskins-temporal: Something's not right here.  I don't see a use case for listing all the deployments, intermixed, regardless of deployment name.  The two scenarios I can think of are users asking:
a) what are the deployment series?  (or perhaps the current deployment for each series)
b) what's the history of a given deployment series?

If users do want them all in one endpoint, it would be a list of lists.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: changed the order.

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
ShahabT: I believe the shuffle is sufficiently unlikely and we can fix it after pre-release. It also won't be a total shuffle, one/few over active deployment will be at the top, the fact that they are over active might mean they are also new builds.
Regarding list of lists, I agree such a UI will make sense and likely preferable. This is something I actually tried to bring up and get attention to but was not successful. Hopefully, it will be discussed during UI work and I'll be OK changing the list APIs then.

PR #483, opened 2024-11-20
Improve V3 info in events and ExecutionInfo

ShahabT:
**What changed?**
Touchups to versioning V3 fields in ExecutionInfo and History Events.
Deprecated `WorkerVersioningStamp` which is replaced completely with `Deployment` and `VersioningBehavior` in V3.

**Why?**

**Breaking changes**

**Server PR**


temporal/api/history/v1/message.proto
```diff
@@ -246,29 +246,14 @@ message WorkflowTaskCompletedEventAttributes {
     // Local usage data sent during workflow task completion and recorded here for posterity
     temporal.api.common.v1.MeteringMetadata metering_metadata = 13;
 
-    // Presence of this value means a deployment redirect completed in this workflow task.
-    // completion event.
-    CompletedDeploymentRedirectInfo completed_redirect = 8;
```
carlydf, line 251: why are you removing this?

temporal/api/history/v1/message.proto
```diff
@@ -305,8 +288,6 @@ message WorkflowTaskFailedEventAttributes {
     // populated to preserve compatibility).
     // Deprecated. Use the info inside the corresponding WorkflowTaskStartedEvent
     temporal.api.common.v1.WorkerVersionStamp worker_version = 10;
-    // Presence of this value means a deployment redirect failed with this failed workflow task.
-    FailedDeploymentRedirectInfo failed_redirect = 11;
```
carlydf, line 309: for this one and the one above, why are you removing the failed_redirect field?

temporal/api/history/v1/message.proto
```diff
@@ -759,7 +740,7 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
 
 message WorkflowExecutionOptionsUpdatedEventAttributes {
     // Versioning override in the mutable state after event has been applied.
-    temporal.api.common.v1.VersioningBehaviorOverride versioning_override = 1;
+    temporal.api.workflow.v1.VersioningOverride versioning_override = 1;
```
carlydf, line 743: ah I thought I did this -- thank you

temporal/api/workflow/v1/message.proto
```diff
@@ -191,7 +197,7 @@ message PendingActivityInfo {
 
     // The deployment this activity was dispatched to most recently. Present only if the activity
     // was dispatched to a versioned worker.
-    temporal.api.deployment.v1.Deployment last_started_deployment = 19;
+    temporal.api.deployment.v1.Deployment last_deployment = 19;
```
carlydf, line 200: maybe `last_used_deployment` would be more clear? this is not a strong opinion though

temporal/api/history/v1/message.proto
ShahabT: because the deployment and behavior must go outside of the redirect. Remember we want to update wf's deployment and behavior even if a transition is not happening (in case of override).

The only thing remaining would be an override, but I figured for safe override, we'll add an WorkflowExecutionOptionsUpdatedEvent immediately after the wftc event anyways.

temporal/api/history/v1/message.proto
ShahabT: no need because we are saying a WorkflowTaskFailedEvent automatically fails the ongoing redirect. It also is not informative to the users until we have safe redirects. (We can add it then with more complete context)

temporal/api/history/v1/message.proto
carlydf: ohh I see. this is replaced by deployment and behavior fields. got it

carlydf (approved):

PR #484, opened 2024-11-20
Deprecate WorkerVersionStamp and replace it with Deployment

ShahabT:
**What changed?**
Deprecate WorkerVersionStamp and replace it with Deployment and Versioning Behavior in all APIs.

**Why?**
WorkerVersionStamp does not fit well with the versioning 3 concepts and terminology.

**Breaking changes**
SDKs that support versioning 3 need to populate Deployment in the task completion/failure responses. Values populated in WorkerVersionStamp will still be honored if Deployment is not set.

**Server PR**


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -362,9 +363,12 @@ message RespondWorkflowTaskCompletedRequest {
     temporal.api.common.v1.MeteringMetadata metering_metadata = 13;
     reserved 14;
     reserved "capabilities";
-    // Must be set when versioning-3 is used (i.e. `worker_version_stamp.use_versioning` is `true`
-    // and `worker_version_stamp.deployment_series_name` is provided).
-    temporal.api.enums.v1.VersioningBehavior versioning_behavior = 15;
```
ShahabT, line 367: @antlai-temporal looking at the comment for versioning_behavior, I realized the stamp is not serving us well is v3 and is basically a twisted way of representing `Deployment` and versioned/unversioned boolean. I think Deployment and VersioningBehavior is perfect for these responsibilities, hence I'm deprecating the stamp. Please consider setting the deployment field instead of the stamp when v3 is used (WorkerDeploymentOptions is set).

carlydf (approved):

ShahabT:
@antlai-temporal merging this to unblock server. Will address change requests in a separate PR, if any.

temporal/api/common/v1/message.proto
```diff
@@ -134,6 +134,11 @@ message WorkerVersionStamp {
     // marker for workflow reset points and the BuildIDs search attribute.
     bool use_versioning = 3;
 
+    // Must be sent if user has set a deployment series name (versioning-3).
+    // Deprecated. SDK should return `Deployment` in the requests to server instead of populating
+    // this field.
+    string deployment_series_name = 4;
```
Sushisource, line 140: Wait why is this both newly added and already deprecated?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -351,6 +351,7 @@ message RespondWorkflowTaskCompletedRequest {
     // Version info of the worker who processed this task. This message's `build_id` field should
     // always be set by SDKs. Workers opting into versioning will also set the `use_versioning`
     // field to true. See message docstrings for more.
+    // Deprecated. Use `deployment` and `versioning_behavior` instead.
     temporal.api.common.v1.WorkerVersionStamp worker_version_stamp = 10;
```
Sushisource, line 355: I'm guessing we don't do this consistently everywhere, but, I think deprecation warnings really should go first at the top of the comment.

Also we probably want to use the official `[deprecated = true]` tag: https://protobuf.dev/programming-guides/proto3/

ShahabT: Good to know. I'll add the official deprecation tag in my next PR.

temporal/api/common/v1/message.proto
antlai-temporal: I think if we don't need it and it is just in my branch we could just delete it...

PR #490, opened 2024-11-22
Fix trigger-api-go-update GHA to escape backticks in commit message

rodrigozhou:
**What changed?**
Fix trigger-api-go-update GHA to escape backticks in commit message

**Why?**
Backticks in commit message are being executed as bash commands.

**Breaking changes**

**Server PR**

### DIFF
--- a/.github/workflows/trigger-api-go-update.yml
+++ b/.github/workflows/trigger-api-go-update.yml
@@ -37,6 +37,7 @@ jobs:
           EVENT_PUSH_COMMIT_AUTHOR: ${{ github.event.head_commit.author.name }}
           EVENT_PUSH_COMMIT_AUTHOR_EMAIL: ${{ github.event.head_commit.author.email }}
           EVENT_PUSH_COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
+          EVENT_WF_DISPATCH_BRANCH: ${{ github.event.inputs.branch }}
         run: |
           case "${{ github.event_name }}" in
             "push")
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-api-go-update.yml
```diff
@@ -39,7 +39,10 @@ jobs:
               BRANCH=${BRANCH#refs/heads/}
               COMMIT_AUTHOR=${{ toJSON(github.event.head_commit.author.name) }}
               COMMIT_AUTHOR_EMAIL=${{ toJSON(github.event.head_commit.author.email) }}
-              COMMIT_MESSAGE=${{ toJSON(github.event.head_commit.message) }}
+              COMMIT_MESSAGE=$(cat << 'EOF' | xargs
```
dnr, line 42: the correct way to do this is to use an `env:` section and use env var substitution in the script. there's no safe way to use `${{}}` in a script if it comes from external data

rodrigozhou: Updated to capture in `env:` section.

.github/workflows/trigger-api-go-update.yml
```diff
@@ -32,14 +32,18 @@ jobs:
...
-              COMMIT_AUTHOR=${{ toJSON(github.event.head_commit.author.name) }}
-              COMMIT_AUTHOR_EMAIL=${{ toJSON(github.event.head_commit.author.email) }}
-              COMMIT_MESSAGE=${{ toJSON(github.event.head_commit.message) }}
+              BRANCH="${EVENT_PUSH_BRANCH#refs/heads/}"
+              COMMIT_AUTHOR="${EVENT_PUSH_COMMIT_AUTHOR}"
+              COMMIT_AUTHOR_EMAIL="${EVENT_PUSH_COMMIT_AUTHOR_EMAIL}"
+              COMMIT_MESSAGE="${EVENT_PUSH_COMMIT_MESSAGE}"
               ;;
 
             "workflow_dispatch")
```
dnr, line 49: can we change the BRANCH on line 50 too? just for consistency.. I know it's less of a risk since only people in the org can do the manual dispatch

dnr (approved):

.github/workflows/trigger-api-go-update.yml
rodrigozhou: Done!

PR #491, opened 2024-11-22
Improve versioning info comments

ShahabT:
**What changed?**

**Why?**

**Breaking changes**

**Server PR**


temporal/api/workflow/v1/message.proto
```diff
@@ -103,31 +103,51 @@ message WorkflowExecutionInfo {
...
+    // queue's current deployment is different from the workflow's deployment. Also, if
+    // the AUTO_UPGRADE workflow is stuck due to backlogged activity or workflow tasks, those tasks
+    // will be redirected to the task queue's current deployment and as soon as a poller of that
+    // deployment is available to receive the task, the workflow will automatically start a
+    // transition to that deployment and continue execution there.
+    // A deployment transition can only exist when there is a pending or started workflow task.
+    // Once the pending workflow task completes on the transition target deployment, the transition
+    // completes and the workflow's `deployment` and possible `behavior` fields are updated.
+    // Pending activities will not start new attempts during a transition. Once the transition is
+    // completed, pending activities will start their next attempt on the new deployment.
```
carlydf, line 150:
```suggestion
    // Not applicable to workflows with PINNED behavior.
    // When a workflow with AUTO_UPGRADE behavior creates a new workflow task, it will automatically
    // start a transition to the task queue's current deployment if the task queue's current deployment is 
    // different from the workflow's deployment and if there is no override.
    // If the AUTO_UPGRADE workflow is stuck due to backlogged activity or workflow tasks, those tasks
    // will be redirected to the task queue's current deployment. As soon as a poller of that deployment
    // is available to receive the task, the workflow will automatically start a transition to that deployment
    // and continue execution there.
    // A deployment transition can only exist while there is a pending or started workflow task.
    // Once the pending workflow task completes on the transition's target deployment, the transition
    // completes and the workflow's `deployment` and `behavior` fields are updated per the SDK's 
    // completion request.
    // Pending activities will not start new attempts during a transition. Once the transition is
    // completed, pending activities will start their next attempt on the new deployment.
```

carlydf (approved):
I fixed some run-on sentences, hopefully made it more understandable for users. Approving so you can merge when you need to, but I hope you take my suggestions

PR #497, opened 2024-11-26
Fix api-go workflow dispatch

dnr:
**What changed?**
Collapse two actions into one to avoid problems with multi-line strings and GITHUB_OUTPUT

**Why?**
Fix breakage

### DIFF
--- a/.github/workflows/trigger-api-go-update.yml
+++ b/.github/workflows/trigger-api-go-update.yml
@@ -30,9 +30,9 @@ jobs:
           owner: ${{ github.repository_owner }}
           repositories: api-go # generate a token with permissions to trigger GHA in api-go repo
 
-      - name: Prepare inputs
-        id: prepare_inputs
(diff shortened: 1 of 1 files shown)


rodrigozhou:
> Use only the first line of the commit message to pass to the api-go workflow

I don't see where this is done.

dnr:
> > Use only the first line of the commit message to pass to the api-go workflow
> 
> I don't see where this is done.

Sorry, that was the initial version. I think it can handle multiple lines as long as you don't go through GITHUB_OUTPUT.

I think it was working before because of the toJSON that got removed, but I think that was probably not quite correct either since GITHUB_OUTPUT isn't json syntax

rodrigozhou (approved):

rodrigozhou:
I wonder if it's actually good to use only the first line though.

ShahabT (approved):

PR #498, opened 2024-11-26
[Reset] Define child handling policies

gow:
Added reset policies to handle child workflows.

**Why?**
Need these to allow resetting of workflows with pending children.

**Breaking changes**
No

**Server PR**
N/A

### DIFF
--- a/temporal/api/enums/v1/workflow.proto
+++ b/temporal/api/enums/v1/workflow.proto
@@ -77,6 +77,13 @@ enum ParentClosePolicy {
     PARENT_CLOSE_POLICY_REQUEST_CANCEL = 3;
 }
 
+// ResetChildPolicy specifies the policy to apply to child workflows when a parent is reset.
+enum ResetChildPolicy {
+    RESET_TYPE_POLICY_UNSPECIFIED = 0;
+    // Reconnects to the currently running child.
+    RESET_TYPE_POLICY_RECONNECT = 1;
+}
+
 enum ContinueAsNewInitiator {
     CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED = 0;
     // The workflow itself requested to continue as new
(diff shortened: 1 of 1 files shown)


temporal/api/enums/v1/workflow.proto
```diff
@@ -77,6 +77,13 @@ enum ParentClosePolicy {
     PARENT_CLOSE_POLICY_REQUEST_CANCEL = 3;
 }
 
+// ResetChildPolicy specifies the policy to apply to child workflows when a parent is reset.
+enum ResetChildPolicy {
+    RESET_CHILD_POLICY_UNSPECIFIED = 0;
+    // Reconnects to the currently running child.
+    RESET_CHILD_POLICY_RECONNECT = 1;
```
yycptt, line 84: I thought we initially only support terminateAndRerun for children initiated after the reset point?

gow: Yes. We don't need it immediately. But I need a way to know if the current run is a result of a reset. So wanted to add this.
But on second thought I think we should add an explicit flag in the new run to mark it as a reset-run. Doing that in this PR - https://github.com/temporalio/temporal/pull/6916

temporal/api/enums/v1/workflow.proto
```diff
@@ -77,6 +77,13 @@ enum ParentClosePolicy {
     PARENT_CLOSE_POLICY_REQUEST_CANCEL = 3;
 }
 
+// ResetChildPolicy specifies the policy to apply to child workflows when a parent is reset.
+enum ResetChildPolicy {
```
cretz, line 81: Where is this enum referenced/used in the API?

gow: I don't have the corresponding change yet. The intent is to use it in `ResetWorkflowExecutionRequest` .https://github.com/temporalio/api/blob/bae790a6cfe59936cb26e613af107780deea500e/temporal/api/workflowservice/v1/request_response.proto#L763

For now I'm holding off on this. Will update this PR when I get to it.

gow:
Not required.

PR #501, opened 2024-11-28
Replace deployment field in SetCurrentDeploymentRequest

ShahabT:

**What changed?**
Replaced the `deployment` field in `SetCurrentDeploymentRequest` with two top-level fields for `series_name` and `build_id`.

**Why?**
This structure is more consistent with `GetCurrentDeploymentRequest` and also allows for supporting "unset" operation by passing empty build ID. (unset part is not documented yet, and will be once we implement it).

**Breaking changes**
The change is breaking but this proto is added to the branch yesterday and it is not used in any release. Moreover, `SetCurrentDeploymentRequest` itself is a pre-release API.

**Server PR**
Server PR will be added by @dnr shortly.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1893,13 +1893,16 @@ message ListDeploymentsResponse {
 
 message SetCurrentDeploymentRequest {
     string namespace = 1;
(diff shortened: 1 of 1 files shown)


ShahabT:
Discussed with @antlai-temporal. Will make this change later if needed. The SetCurrentDeployment API may change significantly anyways before public preview. This can go in with the other changes.

PR #502, opened 2024-12-03
Reserve RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST

hai719:

**What changed?**
Reserve RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST option.

**Why?**
Cancel request should be reapplied only in conflict resolution case. There is no option to exclude it for reset.

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/enums/v1/reset.proto
+++ b/temporal/api/enums/v1/reset.proto
@@ -40,8 +40,8 @@ enum ResetReapplyExcludeType {
     RESET_REAPPLY_EXCLUDE_TYPE_UPDATE = 2;
     // Exclude nexus events when reapplying events beyond the reset point.
     RESET_REAPPLY_EXCLUDE_TYPE_NEXUS = 3;
-    reserved 4;
-    reserved "RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST";
+    // Exclude cancel request events when reapplying events beyond the reset point.
+    RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST = 4 [deprecated=true];
 }
 
 // Event types to include when reapplying events. Deprecated: applications
--- a/temporal/api/enums/v1/reset.proto
+++ b/temporal/api/enums/v1/reset.proto
@@ -40,8 +40,8 @@ enum ResetReapplyExcludeType {
     RESET_REAPPLY_EXCLUDE_TYPE_UPDATE = 2;
     // Exclude nexus events when reapplying events beyond the reset point.
     RESET_REAPPLY_EXCLUDE_TYPE_NEXUS = 3;
-    // Exclude cancel request events when reapplying events beyond the reset point.
-    RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST = 4 [deprecated=true];
+    reserved 4;
+    reserved "RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST";
 }
 
 // Event types to include when reapplying events. Deprecated: applications
--- a/temporal/api/enums/v1/reset.proto
+++ b/temporal/api/enums/v1/reset.proto
@@ -40,8 +40,8 @@ enum ResetReapplyExcludeType {
     RESET_REAPPLY_EXCLUDE_TYPE_UPDATE = 2;
     // Exclude nexus events when reapplying events beyond the reset point.
     RESET_REAPPLY_EXCLUDE_TYPE_NEXUS = 3;
-    // Exclude cancel request events when reapplying events beyond the reset point.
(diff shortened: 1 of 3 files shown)


temporal/api/enums/v1/reset.proto
```diff
@@ -41,7 +41,7 @@ enum ResetReapplyExcludeType {
     // Exclude nexus events when reapplying events beyond the reset point.
     RESET_REAPPLY_EXCLUDE_TYPE_NEXUS = 3;
     // Exclude cancel request events when reapplying events beyond the reset point.
-    RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST = 4;
+    RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST = 4 [deprecated=true];
```
bergundy, line 44: I think it's still safe to mark as reserved at this point since it has not been used in any client facing code and the only released API version that includes it was released a few days ago.

hai719: updated.

bergundy (approved):

dandavison:
I haven't quite been able to understand the PR explanation:

> Cancel request should be reapplied only in conflict resolution case. There is no option to exclude it for reset.

Is this essentially saying

> The server will decide whether to reapply cancel requests. Therefore we do not allow the user to specify that they should be excluded from reapply: they may in fact be excluded or they may not, but the client gets no say in the matter.

yycptt:
Server can not reapply cancel requests by default and add an new option for excluding it. Otherwise it's a breaking change (because we are not reapplying it now).

We also don't see use cases right now where cancel requests needs to be reapplied, so server will just never reapply cancel requests for ResetWorkflow flow.

yycptt (commented):
LGTM. But calling out that this is technically a breaking change if anyone is already referencing the generated enum constants. 

(not approving here since the second approval needs to come from SDK team)

temporal/api/enums/v1/reset.proto
```diff
@@ -40,8 +40,8 @@ enum ResetReapplyExcludeType {
     RESET_REAPPLY_EXCLUDE_TYPE_UPDATE = 2;
     // Exclude nexus events when reapplying events beyond the reset point.
     RESET_REAPPLY_EXCLUDE_TYPE_NEXUS = 3;
-    // Exclude cancel request events when reapplying events beyond the reset point.
-    RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST = 4;
+    reserved 4;
+    reserved "RESET_REAPPLY_EXCLUDE_TYPE_CANCEL_REQUEST";
```
cretz, line 44: This is tagged and released code. Can you just leave the generated code untouched and deprecate it via comments? This is low-level API, so it is mostly harmless to just deprecate it. If we have released API, we should need a really good reason to break compatibility. Even if the implementation is not released, this API is released.

(also this is why we usually want to complete (at least most of) implementation _before_ merging API to main)

cretz (commented):
Can someone look into why this is not failing CI as an incompatible change? Is there a bug in the buf code that does not properly consider this wire JSON incompatible?

temporal/api/enums/v1/reset.proto
hai719: Previously we used this option in OSS (this [PR](https://github.com/temporalio/temporal/pull/6888), not released yet). Now we decided that the server should not provide the option to exclude the cancel request during reapplying on reset operation. This is the [PR](https://github.com/temporalio/temporal/pull/6926) to remove the option in OSS. There is no released code to use this option.

hai719:
> I haven't quite been able to understand the PR explanation:
> 
> > Cancel request should be reapplied only in conflict resolution case. There is no option to exclude it for reset.
> 
> Is this saying
> 
> > The server will decide whether to reapply cancel requests. Therefore we do not allow the user to specify that they should be excluded from reapply: they may in fact be excluded or they may not, but the client gets no say in the matter.
> 
> ?

Yes. As Yichao commented, server will never reapply cancel request for ResetWorkflow. So this option is no longer useful.

temporal/api/enums/v1/reset.proto
```diff
@@ -41,7 +41,7 @@ enum ResetReapplyExcludeType {
     // Exclude nexus events when reapplying events beyond the reset point.
     RESET_REAPPLY_EXCLUDE_TYPE_NEXUS = 3;
     // Exclude cancel request events when reapplying events beyond the reset point.
```
cretz, line 43:
```suggestion
    // Deprecated, unimplemented option.
```

cretz (commented):
One minor suggestion, then LGTM

cretz (approved):

PR #503, opened 2024-12-05
Add Failure to query result

bergundy:
**What changed?**

Add a `Failure` field to:
- QueryFailedFailure
- WorkflowQueryResult
- RespondQueryTaskCompletedRequest

**Why?**

Support encoding failure attributes for E2E encryption of messages and stack traces.

**Server PR**

Coming soon but I've already validated this change locally.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -914,10 +914,22 @@ message GetSearchAttributesResponse {
...
     bytes task_token = 1;
     temporal.api.enums.v1.QueryResultType completed_type = 2;
+    // The result of the query.
+    // Mutually exclusive with `error_message` and `failure`. Set when the query succeeds.
     temporal.api.common.v1.Payloads query_result = 3;
+    // A plain error message that must be set if completed_type is QUERY_RESULT_TYPE_FAILED.
+    // SDKs should also fill in the more complete `failure` field to provide the full context and support encryption of
+    // the failure message.
+    // The error message will be duplicated if the `failure` field is present to support callers that pre-date the
+    // addition of that field.
```
Sushisource, line 924:
```suggestion
    // The error message may be totally or partially redundant with the `failure` field, if present, to support callers that pre-date the
    // addition of that field.
```

I can imagine that, if this feature is supported, SDK might allow for a sensitive-information-free message separate from the full details.

Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
bergundy: Yeah, in which case the failure message will be the same as `error_message` (e.g. "Encoded...") and the sensitive information will be in the encoded details.

cretz: :+1: For SDK implementers worker side, `error_message` is expected to always be set to `failure.message`. And for SDK implementers client side, `error_message` should be ignored if `failure` is non-null.

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
Sushisource: What I meant is that it might make sense to allow setting `error_message` to something different, if we want to allow for specifying some un-encoded message.

But, regardless, if the expectation is that, any time `failure` is used, the message is always just "Encoded failure" or w/e, then we should specify that in the docstrings too.

PR #505, opened 2024-12-10
Add workflow expiration and run timestamps to execution info

ychebotarev:

**What changed?**
Add new structure -WorkflowExecutionExtendedInfo

That structure has few fields:
* execution timeout time
* run timeout time
* last reset time
* cancel requested

(the last one from https://github.com/temporalio/api/pull/339)

**Why?**
Per customer request.
Execution timeout timestamp may change after workflow reset, and in this case customers have no idea when it will fire.
WorkflowExecutionInfo is reused in other server calls, like ListWorkflow, etc.
Because of that we either add any new field to ES, or it will be nil which is bad user experience.

**Breaking changes**
No

**Server PR**
N/A

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -110,10 +110,10 @@ message WorkflowExecutionInfo {
 
     // Workflow execution expiration time is defined as workflow start time plus expiration timeout.
     // Workflow start time may change after workflow reset.
-    google.protobuf.Timestamp workflow_execution_expiration_time = 23;
+    google.protobuf.Timestamp execution_expiration_time = 23;
 
     // Workflow run expiration time is defined as current workflow run start time plus workflow run timeout.
-    google.protobuf.Timestamp workflow_run_expiration_time = 24;
+    google.protobuf.Timestamp run_expiration_time = 24;
 }
 
 // Holds all the information about versioning for a workflow execution.
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -43,6 +43,9 @@ import "temporal/api/failure/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
 import "temporal/api/sdk/v1/user_metadata.proto";
 
+
+// Hold basic information about a workflow execution.
+// This structure is a part of visibility, and thus contain a limited subset of information.
 message WorkflowExecutionInfo {
     temporal.api.common.v1.WorkflowExecution execution = 1;
     temporal.api.common.v1.WorkflowType type = 2;
@@ -109,7 +112,7 @@ message WorkflowExecutionInfo {
     WorkflowExecutionVersioningInfo versioning_info = 22;
 }
 
-// Holds all the extra information about workflow execution.
+// Holds all the extra information about workflow execution that is not part of Visibility.
 message WorkflowExecutionExtendedInfo {
     // Workflow execution expiration time is defined as workflow start time plus expiration timeout.
     // Workflow start time may change after workflow reset.
@@ -123,6 +126,9 @@ message WorkflowExecutionExtendedInfo {
 
     // Last workflow reset time. Nil if the workflow was never reset.
     google.protobuf.Timestamp last_reset_time = 4;
+
+    // Original workflow start time.
+    google.protobuf.Timestamp original_start_time = 5;
 }
 
 // Holds all the information about versioning for a workflow execution.
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -107,13 +107,22 @@ message WorkflowExecutionInfo {
     // be versioned or unversioned, depending on `versioning_info.behavior` and `versioning_info.versioning_override`.
     // Experimental. Versioning info is experimental and might change in the future.
     WorkflowExecutionVersioningInfo versioning_info = 22;
+}
 
+// Holds all the extra information about workflow execution.
+message WorkflowExecutionExtendedInfo {
     // Workflow execution expiration time is defined as workflow start time plus expiration timeout.
     // Workflow start time may change after workflow reset.
-    google.protobuf.Timestamp execution_expiration_time = 23;
+    google.protobuf.Timestamp execution_expiration_time = 1;
 
     // Workflow run expiration time is defined as current workflow run start time plus workflow run timeout.
-    google.protobuf.Timestamp run_expiration_time = 24;
+    google.protobuf.Timestamp run_expiration_time = 2;
+
+    // indicates if the workflow received a cancel request
+    bool cancel_requested = 3;
+
+    // Last workflow reset time. Nil if the workflow was never reset.
+    google.protobuf.Timestamp last_reset_time = 4;
 }
 
 // Holds all the information about versioning for a workflow execution.
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -107,6 +107,13 @@ message WorkflowExecutionInfo {
     // be versioned or unversioned, depending on `versioning_info.behavior` and `versioning_info.versioning_override`.
     // Experimental. Versioning info is experimental and might change in the future.
     WorkflowExecutionVersioningInfo versioning_info = 22;
+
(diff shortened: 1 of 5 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -107,6 +107,13 @@ message WorkflowExecutionInfo {
     // be versioned or unversioned, depending on `versioning_info.behavior` and `versioning_info.versioning_override`.
     // Experimental. Versioning info is experimental and might change in the future.
     WorkflowExecutionVersioningInfo versioning_info = 22;
+
+    // Workflow execution expiration time is defined as workflow start time plus expiration timeout.
+    // Workflow start time may change after workflow reset.
+    google.protobuf.Timestamp workflow_execution_expiration_time = 23;
+
+    // Workflow run expiration time is defined as current workflow run start time plus workflow run timeout.
+    google.protobuf.Timestamp workflow_run_expiration_time = 24;
```
alexshtin, line 116: I am in doubt about word `workflow` here. All other fields don't have it although it can be added to almost every field in this message. I would remove it.

alexshtin (approved):

temporal/api/workflow/v1/message.proto
yycptt, line 116: Will those fields be populated for ListWorkflow as well?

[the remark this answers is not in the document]
ychebotarev: They can't
We have workflow start time (not in this structure), but that time is fixed (probably should be "originalStartTime". )
So if reset happen start_time+execution_timeout != execution_expiration_time

ychebotarev: not a part of original feature request. But I can add it if you think there is a value.

ychebotarev: (discussing this offline, and whatever the outcome it will not affect this PR)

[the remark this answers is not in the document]
cretz: I wonder if we should have that original start time in here then for others to use. Sounds like it may have value including but not limited to deriving these timestamps. Could be convinced to leave these shortcuts for the user, though IMO you can just document these above `original_start_time`. Similarly, I think we should clarify in the docs above `start_time` what it is and that it may not be the start time used for timeouts if a reset occurred.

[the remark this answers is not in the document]
ychebotarev: we decide that we will add the those explicit fields.
this way it will be the same as we have in server WorkflowExecutionInfo, and customer wouldn't need to do math.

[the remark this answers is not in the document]
cretz: But you are making the user do math if they want the original start time

yycptt: Unresolving the comment here. I want to understand the expected server work here before this PR is landed. I am under the impression that any field in WorkflowExecutionInfo should be populated by both Describe and List.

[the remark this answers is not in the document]
ychebotarev: Users want to know the exact expiration time. In this case I don't see how doing anything else could be better then giving them exactly what they ask.
Even if they don't - it is still better to provide the exact time. This way we can be sure we don't have bugs/weird code path in our code. (the initial feature request was a part of conversation where (as I understand) users are not sure we extend expiration time after reset in some cases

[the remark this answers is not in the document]
cretz: > I don't see how doing anything else could be better then giving them exactly what they ask.

Giving all users the ability to derive this value is better than only giving them exactly what they ask and no more IMO.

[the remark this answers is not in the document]
cretz: I think we should intentionally add it, but if others don't, no problem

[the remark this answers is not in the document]
ychebotarev: right now we don't. 
But (per discussion from Yichao comment below) we may endup adding it, though no in this structure :)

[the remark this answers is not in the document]
Sushisource: I'd agree that, unless this isn't feasible for some performance reason or something (which wouldn't seem to make much sense, since you need these values to perform the addition in the first place) that it'd be nicer to just have the underlying factors.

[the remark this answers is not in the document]
ychebotarev: I'm adding actual time because:
* this is exactly what user asks for
* this eliminates the possibility of mistake (when ->our<- actual expiration time is different for whatever reason, we already has this problem) 

So unless I see very good argument why I shouldn't do that I will continue with this approach.

ychebotarev: those fields are moved into separate structure.

temporal/api/workflow/v1/message.proto
```diff
@@ -109,6 +109,22 @@ message WorkflowExecutionInfo {
     WorkflowExecutionVersioningInfo versioning_info = 22;
 }
 
+// Holds all the extra information about workflow execution.
+message WorkflowExecutionExtendedInfo {
```
ychebotarev, line 113: btw, it will be a good time to ask for more info.
I'm adding cancel_requested - we want to do that for a long time, but also don't want to add it to ES

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: "possibility of mistake" - in a sense that if we make mistake - it will be visible

temporal/api/workflow/v1/message.proto
cretz: > it will be a good time to ask for more info

Can I have `original_start_time`?

cretz, line 113: It is a bit unclear that "info" is in visibility and "extended info" is not. Maybe at least clarify at the top of the proto so people know the difference between the two and why some fields are here and some are there.

ychebotarev: You can.
But start time in workflowExecutionInfo ->is<- the original start time.

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: closing this

temporal/api/workflow/v1/message.proto
ychebotarev: Added some comments.

cretz: Oh, my mistake. From the thread before at https://github.com/temporalio/api/pull/505#discussion_r1878735876, I was under the impression that whatever time was needed to add to `execution_timeout` to make `execution_expiration_time` was not available. Can we add whatever that time is?

ychebotarev: >whatever time was needed to add to execution_timeout to make execution_expiration_time was not available

It is NOT available right now in DescribeWorkflow response. Only "StartTime", which is basically "original_start_time"
That is why I added "last_reset_time".

In theory "execution_expiration_time" = max("start_time", "last_reset_time") + "execution_expiration_timeout"

On practice - we have a case when we still manage to make a bug with this seemingly simple expiration time calculations... :) (for activities, but in a somewhat tricky manner)

So if we have some bug, or change the way expiration time is calculated - without providing exact value it will not be visible to the user.

Thus I prefer to explicitly return "execution_expiration_time". This way we return whatever we use, right or wrong, old or new.

cretz: > In theory "execution_expiration_time" = max("start_time", "last_reset_time") + "execution_expiration_timeout"

Great, this works for me. Nothing else blocking from my POV.

> On practice - we have a case when we still manage to make a bug with this seemingly simple expiration time calculations...

I am a bit curious about this. Are you saying we have a bug today, or are you saying it's possible that "execution_expiration_time" = max("start_time", "last_reset_time") + "execution_expiration_timeout" may not always be true?

> Thus I prefer to explicitly return "execution_expiration_time"

This is fine, I don't mind the redundant data, though a bit concerned about the idea that it may change or that we may introduce bugs.

cretz (approved):
Can you have @alexshtin or someone from server re-approve? This changed quite a bit since original approval.

alexshtin (approved):

alexshtin:
Yes, it looks very different from what it was. I wish we started that new `WorkflowExecutionExtendedInfo` message earlier.

PR #507, opened 2024-12-20
Add original_start_time to the workflow execution started event

ychebotarev:
**What changed?**

Add `original_execution_start_time` field to `WorkflowExecutionStartedEventAttributes`.

**Why?**

To preserve original execution start time. It will be later used to populate extended info for workflow execution during `DescribeWorkflow` calls.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -77,6 +77,10 @@ message WorkflowExecutionStartedEventAttributes {
     // This is the run id when the WorkflowExecutionStarted event was written.
     // A workflow reset changes the execution run_id, but preserves this field.
     string original_execution_run_id = 14;
+    // This is the start time when the WorkflowExecutionStarted event was written.
+    // A workflow reset will change execution start time, but will preserve this field.
+    google.protobuf.Timestamp original_execution_start_time = 34;
+
     // Identity of the client who requested this execution
     string identity = 15;
     // This is the very first runId along the chain of ContinueAsNew, Retry, Cron and Reset.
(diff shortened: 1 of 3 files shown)


alexshtin (approved):

temporal/api/history/v1/message.proto
```diff
@@ -133,6 +133,10 @@ message WorkflowExecutionStartedEventAttributes {
     string inherited_build_id = 32;
     // Versioning override applied to this workflow when it was started.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 33;
+
+    // This is the start time when the WorkflowExecutionStarted event was written.
+    // A workflow reset will change execution start time, but will preserve this field.
+    google.protobuf.Timestamp original_execution_start_time = 34;
```
alexshtin, line 139: I would move it closer to `original_execution_run_id` same way as `namespace` and `namespace_id` are already grouped here.

ychebotarev: done

temporal/api/history/v1/message.proto
```diff
@@ -77,6 +77,10 @@ message WorkflowExecutionStartedEventAttributes {
     // This is the run id when the WorkflowExecutionStarted event was written.
     // A workflow reset changes the execution run_id, but preserves this field.
     string original_execution_run_id = 14;
+    // This is the start time when the WorkflowExecutionStarted event was written.
+    // A workflow reset will change execution start time, but will preserve this field.
+    google.protobuf.Timestamp original_execution_start_time = 34;
```
yycptt, line 82: Why do we need this? The event is reused across reset so start event time is the original execution start time.

ychebotarev: ok great. then we don't need this. closing

PR #508, opened 2025-01-07
[Requires repository secret addition] Add Publish to Buf Github Action

nikki-dag:

**What changed?**
Adds a GitHub Actions workflow to automate the process of pushing protocol buffers to the Buf registry on every push to the main branch.

We need to set the `BUF_TOKEN` secret before merging this PR.

**Why?**
We want to be able to depend on server protos from other repositories (e.g. api-cloud) using the buf package manager.

**Breaking changes**
None.

**Server PR**
NA

### DIFF
--- a/.github/workflows/push-to-buf.yml
+++ b/.github/workflows/push-to-buf.yml
@@ -3,7 +3,7 @@ name: Push to Buf Registry
 on:
   push:
     tags:
-      - '**'
+      - 'v**'
     branches:
       - master
 permissions:
@@ -17,4 +17,4 @@ jobs:
       - uses: bufbuild/buf-action@v1
         with:
           version: 1.49.0
-          token: ${{ secrets.BUF_TOKEN }}
+          token: ${{ secrets.BUF_TEMPORALIO_TOKEN }}
--- a/.github/workflows/push-to-buf.yml
+++ b/.github/workflows/push-to-buf.yml
@@ -2,6 +2,8 @@ name: Push to Buf Registry
 
 on:
   push:
+    tags:
+      - '**'
     branches:
       - master
 permissions:
@@ -14,4 +16,5 @@ jobs:
         uses: actions/checkout@v4
       - uses: bufbuild/buf-action@v1
         with:
+          version: 1.49.0
           token: ${{ secrets.BUF_TOKEN }}
--- a/.github/workflows/push-to-buf.yml
+++ b/.github/workflows/push-to-buf.yml
@@ -2,6 +2,10 @@ name: Push to Buf Registry
 
 on:
   push:
+    tags:
+      - 'v**'
+    branches:
+      - master
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/push-to-buf.yml
+++ b/.github/workflows/push-to-buf.yml
@@ -1,0 +1,24 @@
+name: Push to Buf
+
+on:
+  push:
+    branches:
+      - master
+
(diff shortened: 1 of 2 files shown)


buf.yaml
```diff
@@ -1,4 +1,5 @@
 version: v1
+name: buf.build/temporal/api
```
cretz, line 2: Note, with basically every SDK language, package manager, GH org, Docker hub, etc, etc we have used `temporalio`. Do we want to consider that here instead of `temporal` (ignoring the fact that there are already registered repositories on `temporal`). It seems that `temporalio` org was created, do we own it?

.github/workflows/push-to-buf.yml
```diff
@@ -0,0 +1,24 @@
+name: Push to Buf
+
+on:
+  push:
+    branches:
+      - master
```
cretz, line 6: Pardon my lack of `buf` knowledge, can I get some clarity on how versioning is handled? This isn't doing anything to buf when we make a new tag in this repository, should it?

nikki-dag: I added a new `tags` trigger and successfully tagged my personal buf registry:

image


This way other repositories can depend on a specific release:
https://buf.build/docs/bsr/module/dependency-management/

buf.yaml
nikki-dag: Updated the reference to `temporalio` and brought it up on Slack 🤞

nikki-dag: Would you like me to contact support@buf.build to see if we can settle the ownership?

.github/workflows/push-to-buf.yml
cretz: Great, I see you have now added the `tags:` trigger, all good.

buf.yaml
nikki-dag: Update: we own it and will have access shortly.

cretz: Separate from this PR, can we also move the cloud API in there and stop using the `temporal` org?

.github/workflows/push-to-buf.yml
cretz, line 1: Should we test this before merging? You can request internally for the secret to be set (it should be set GH org wide, not just in this repo IMO). You may need to temporarily enable this to push on PR to test.

nikki-dag: Sounds good 👍

.github/workflows/push-to-buf.yml
```diff
@@ -0,0 +1,20 @@
+name: Push to Buf Registry
+
+on:
+  push:
+    tags:
+      - '**'
```
bergundy, line 6: Do we want this for every tag? Should it be only `v**` tags?

.github/workflows/push-to-buf.yml
bergundy: Very good point.

bergundy (approved):

.github/workflows/push-to-buf.yml
nikki-dag: Updated ✅

buf.yaml
nikki-dag: Draft PR for cloud API: https://github.com/temporalio/api-cloud/pull/56/files

cretz (approved):

.github/workflows/push-to-buf.yml
cretz: Tested, LGTM

PR #510, opened 2025-01-14
Add OnConflictOptions to StartWorkflowExecutionRequest

rodrigozhou:

**What changed?**
Add `OnConflictOptions` to `StartWorkflowExecutionRequest`

**Why?**
When the workflow id conflict policy is `USE_EXISTING`, `OnConflictOptions` add the ability to change the existing running workflow.

**Breaking changes**

**Server PR**
https://github.com/temporalio/temporal/pull/7080

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -746,10 +746,10 @@ message WorkflowExecutionOptionsUpdatedEventAttributes {
     temporal.api.workflow.v1.VersioningOverride versioning_override = 1;
     // Versioning override removed in this event.
     bool unset_versioning_override = 2;
-    // Attached request ID to the running workflow execution so that additional requests with same
+    // Request ID attachedto the running workflow execution so that subsequent requests with same
     // request ID will be deduped.
     string attached_request_id = 3;
-    // Attached completion callbacks to the running workflow execution.
+    // Completion callbacks attached to the running workflow execution.
     repeated temporal.api.common.v1.Callback attached_completion_callbacks = 4;
 }
 
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -746,6 +746,11 @@ message WorkflowExecutionOptionsUpdatedEventAttributes {
     temporal.api.workflow.v1.VersioningOverride versioning_override = 1;
     // Versioning override removed in this event.
     bool unset_versioning_override = 2;
+    // Attached request ID to the running workflow execution so that additional requests with same
+    // request ID will be deduped.
+    string attached_request_id = 3;
+    // Attached completion callbacks to the running workflow execution.
+    repeated temporal.api.common.v1.Callback attached_completion_callbacks = 4;
 }
 
 // Not used anywhere. Use case is replaced by WorkflowExecutionOptionsUpdatedEventAttributes
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -440,3 +440,11 @@ message VersioningOverride {
     // Identifies the worker deployment to pin the workflow to.
     temporal.api.deployment.v1.Deployment deployment = 2;
 }
+
+// When StartWorkflowExecution uses the WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING conflict policy and
+// there is already an existing running workflow, OnConflictOptions defines actions to be taken on
+// the existing running workflow.
+message OnConflictOptions {
+    bool attach_completion_callbacks = 1;
+    bool attach_links = 2;
+}
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -441,10 +441,15 @@ message VersioningOverride {
     temporal.api.deployment.v1.Deployment deployment = 2;
 }
 
-// When StartWorkflowExecution uses the WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING conflict policy and
+// When StartWorkflowExecution uses the conflict policy WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING and
 // there is already an existing running workflow, OnConflictOptions defines actions to be taken on
-// the existing running workflow.
+// the existing running workflow. In this case, it will create a WorkflowExecutionOptionsUpdatedEvent
+// history event in the running workflow with the changes requested in this object.
 message OnConflictOptions {
-    bool attach_completion_callbacks = 1;
-    bool attach_links = 2;
+    // Attaches the request ID to the running workflow.
+    bool attach_request_id = 1;
+    // Attaches the completion callbacks to the running workflow.
+    bool attach_completion_callbacks = 2;
+    // Attaches the links to the WorkflowExecutionOptionsUpdatedEvent history event.
+    bool attach_links = 3;
 }
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -206,7 +206,7 @@ message StartWorkflowExecutionRequest {
     // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
     // To unset the override after the workflow is running, use UpdateWorkflowExecutionOptions.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 25;
-    // It defines actions to be done to the existing running workflow when the conflict policy
+    // Defines actions to be done to the existing running workflow when the conflict policy
     // WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING is used. If not set (ie., nil value) or set to a
     // empty object (ie., all options with default value), it won't do anything to the existing
     // running workflow. If set, it will add a history event to the running workflow.
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -206,10 +206,10 @@ message StartWorkflowExecutionRequest {
     // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
     // To unset the override after the workflow is running, use UpdateWorkflowExecutionOptions.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 25;
-    // It defines behaviors to be executed when the WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING
-    // conflict policy is used. If set to non-nil object, it will attach the RequestID to the
-    // existing running workflow (for deduping), and perform actions specified in the object.
-    // If not set (ie., nil value), it won't do anything to the existing running workflow.
+    // It defines actions to be done to the existing running workflow when the conflict policy
+    // WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING is used. If not set (ie., nil value) or set to a
+    // empty object (ie., all options with default value), it won't do anything to the existing
+    // running workflow. If set, it will add a history event to the running workflow.
     temporal.api.workflow.v1.OnConflictOptions on_conflict_options = 26;
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -206,6 +206,11 @@ message StartWorkflowExecutionRequest {
     // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
     // To unset the override after the workflow is running, use UpdateWorkflowExecutionOptions.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 25;
+    // It defines behaviors to be executed when the WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING
+    // conflict policy is used. If set to non-nil object, it will attach the RequestID to the
+    // existing running workflow (for deduping), and perform actions specified in the object.
+    // If not set (ie., nil value), it won't do anything to the existing running workflow.
+    temporal.api.workflow.v1.OnConflictOptions on_conflict_options = 26;
 }
 
 message StartWorkflowExecutionResponse {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -6536,7 +6536,7 @@
         },
         "onConflictOptions": {
           "$ref": "#/definitions/v1OnConflictOptions",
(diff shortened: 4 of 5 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -743,6 +743,11 @@ message ChildWorkflowExecutionTerminatedEventAttributes {
 message WorkflowExecutionOptionsUpdatedEventAttributes {
     // Versioning override in the mutable state after event has been applied.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 1;
+    // Attach request ID to the running workflow execution so that additional requests with same
+    // request ID will be deduped.
+    string attach_request_id = 2;
+    // Add completion callbacks to the running workflow execution.
+    repeated temporal.api.common.v1.Callback add_completion_callbacks = 3;
```
bergundy, line 750: Keep the past tense for event fields for consistency.

```suggestion
    string attached_request_id = 2;
    // Add completion callbacks to the running workflow execution.
    repeated temporal.api.common.v1.Callback attached_completion_callbacks = 3;
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -206,6 +206,11 @@ message StartWorkflowExecutionRequest {
     // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
     // To unset the override after the workflow is running, use UpdateWorkflowExecutionOptions.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 25;
+    // It defines behaviors to be executed when the WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING
+    // conflict policy is used. If set to non-nil empty object, it will attach the completion
+    // callbacks to the existing running workflow. If not set (ie., nil value), it won't do anything
```
cretz, line 211: > If set to non-nil empty object, it will attach the completion callbacks to the existing running workflow

Why not make this an explicit option on the object? The code has to opt-in to this anyways by providing a non-null object, might as well opt-in to which things they want to attach. It seems strange to have some that are attached by default and maybe later some that are affected by options.

temporal/api/workflow/v1/message.proto
```diff
@@ -440,3 +440,9 @@ message VersioningOverride {
     // Identifies the worker deployment to pin the workflow to.
     temporal.api.deployment.v1.Deployment deployment = 2;
 }
+
+// When StartWorkflowExecution uses the WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING conflict policy and
+// there is already an existing running workflow, OnConflictOptions defines actions to be taken on
+// the existing running workflow.
+message OnConflictOptions {
```
cretz, line 447: Will this message be used anywhere else except start workflow?

temporal/api/workflow/v1/message.proto
```diff
@@ -440,3 +440,9 @@ message VersioningOverride {
     // Identifies the worker deployment to pin the workflow to.
     temporal.api.deployment.v1.Deployment deployment = 2;
 }
+
+// When StartWorkflowExecution uses the WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING conflict policy and
+// there is already an existing running workflow, OnConflictOptions defines actions to be taken on
+// the existing running workflow.
```
bergundy, line 446: I'd mention that we plan to add options here as needed.

bergundy (approved):
Approved with a couple of small comments.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -206,6 +206,11 @@ message StartWorkflowExecutionRequest {
...
+    // to the existing running workflow.
+    temporal.api.workflow.v1.OnConflictOptions on_conflict_options = 26;
```
cretz, line 213: Is this needed for signal-with-start?

bergundy: I don't think so, because for signal with start we're thinking that you care more about the signal than the start and we don't plan on attaching the Nexus operation to the workflow's lifecycle.

temporal/api/workflow/v1/message.proto
bergundy: Maybe in schedules?

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: Sure, I'm ok with being explicit. cc: @bergundy

temporal/api/workflow/v1/message.proto
cretz: There is no ID conflict policy in schedules (by intention as I understand it due to how schedules set IDs)

temporal/api/workflow/v1/message.proto
rodrigozhou: I'm adding explicitly the behaviors we want to implement at first already, so I think no need to update comment.

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: I've added `attach_completion_callbacks` and `attach_links`. I'm not adding `attach_request_id` because request id is used for deduping, so I think it make sense that if `OnConflictOptions` is set, then we need to always attach the request id to dedup the request, and not redo work.

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: In the case of signal with start, the goal is to signal the workflow if it's already running, not update it.
Supporting `OnConflictOptions` in this case would mean that the behavior of signal with start would be to update and signal the workflow? Would this make sense?

cretz: I think the general idea of this being paired with a conflict policy makes sense for consistency reasons. I think your question could be rephrased as "does conflict policy make sense on signal with start", but since it was decided it does however that was decided, I think these options should go alongside them.

temporal/api/workflow/v1/message.proto
```diff
@@ -440,3 +440,11 @@ message VersioningOverride {
...
+message OnConflictOptions {
+    bool attach_completion_callbacks = 1;
+    bool attach_links = 2;
```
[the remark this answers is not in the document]
bergundy: This works for me although I prefer the term attach over append but don't have a strong opinion on that.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I don't see a use case for attaching callbacks and links to a running workflow on signal-with-start.
We plan to attach a link on the WorkflowExecutionSignaled event (and probably on the WorkflowExecutionStarted event if the current request started a new workflow). But we do not have plans to attach callbacks on signal-with-start.
I'd rather only add fields that we plan on using.

bergundy (approved):

temporal/api/workflowservice/v1/request_response.proto
cretz: On conflict options is about what do to on conflict. Whether a use case exists for these first couple of options I don't think affects the fact that everywhere a conflict policy is applied, its conflict options should apply.

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
cretz: I think what each of these do to existing values may need to be documented here because it is clearly confusing. What does "attach" mean for links? If links are set in the history event and immutable, then where in history are these new ones set? Or are there now two kinds of links, start links and post-start links, and they manifest in different ways?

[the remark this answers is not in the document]
bergundy: Agree we should document these options.

The link would essentially be linked to the new event but conceptually they're considered attached to the workflow.

```mermaid
flowchart LR
    CallerWorkflowA --> HandlerWorkflow
    CallerWorkflowB --> HandlerWorkflow
```

The callback is added to the list of completion callbacks.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I would still rather add this later. We don't plan on adding server support for this quite yet.

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
cretz: Which new event? I am under the impression that a use-existing conflict policy will not make any new events if the workflow already exists.

temporal/api/workflowservice/v1/request_response.proto
cretz: I do think conflict policy and options need to be married (arguably if we needed both at the same time they would not have been separate objects). Supporting different conflict behaviors on different start calls is confusing.

temporal/api/workflowservice/v1/request_response.proto
bergundy: That's fine with me.

temporal/api/workflow/v1/message.proto
bergundy: Hmm... I know that users have requested a way to set the workflow ID as-is for schedules and maybe that would make conflict options will come into play. In any case, for now it seems that's not necessary.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I don't think this should be blocking the PR. We can add this later if needed.

One thing to consider though is how multi-ops (update-with-start) works with these conflict options. We haven't put much thought into that. I'd agree that if conflict options work on update-with-start, they should work on signal-with-start as well.

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
bergundy: It creates a WorkflowExecutionOptionsUpdated event (see the added attributes in this PR).
We have to create the event for linking and for reapply in case of reset.

temporal/api/workflowservice/v1/request_response.proto
rodrigozhou: The way I was thinking was that the concept of "request ID" is not user facing. It's a mechanism that Temporal introduced to dedup requests. As such, I didn't think it would make sense for users to decide if they wanna attach "request ID" (do users even know what is a "request ID"? the SDK doesn't expose it in `StartWorkflowOptions`).

And the reason for only attaching if the object is not nil is because if the object is nil, then there's no action that the user wants to perform on the running workflow. Attaching the request ID for deduping would introduce a event unnecessarily since deduping in this case is unnecessary since the request itself is no-op (and deduping arguably costs more?).

However, I do see a case that always attaching request ID even the conflict options is nil can make sense. Suppose a request with conflict policy USE_EXISTING and reuse policy ALLOW_DUPLICATE. If, when the request was first sent, there was workflow running, we don't attach request ID, and for some reason the client retries the request, but at the time of the retry, the running workflow already completed. In this case, currently, Temporal would start a new workflow.

So, maybe we should always attach request ID?

bergundy: That would make every USE_EXISTING request perform a write, we should avoid that.
Let's just make it explicit and consider if we want to expose this to the SDK later.

cretz: > The way I was thinking was that the concept of "request ID" is not user facing

Right, what is user-facing is part of the SDK piece and that's where we can make user-facing decisions. Request ID is part of the API.

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
cretz: I see, I do think a more generic update options approach in API could be present that on-conflict-options leans on, but not that important

cretz (commented):
Everything LGTM (even though I may have differing ideas on how updating workflow options on conflict may look). Once the request ID attachment conflict option is present, will approve.

temporal/api/workflow/v1/message.proto
dnr: Schedules have the "overlap policy" that includes things like terminate if running. I think even after we add the ability to use the id as-is, we would still want to encapsulate all the options for schedules in overlap policy.

Of course, the implementation could decide to translate overlap policy "terminate if running" into conflict policy "terminate" when using workflow id as-is (or it could not bother since it's a tiny rare optimization).

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
cretz: :+1: Note I think it may be too late once this lands in `master` (especially if we start using it in an SDK)

rodrigozhou:
@cretz Updated the PR. I kept the verb as "attach", and added comments. I also re-wrote pretty much all comments.

temporal/api/workflow/v1/message.proto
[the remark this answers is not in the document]
rodrigozhou: This PR is only making changes to the `WorkflowExecutionOptionsUpdatedEvent` which is a history event, and the `StartWorkflowExecution` API. I'm not making any modification to the update options API in this PR.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -206,6 +206,11 @@ message StartWorkflowExecutionRequest {
     // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
     // To unset the override after the workflow is running, use UpdateWorkflowExecutionOptions.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 25;
+    // It defines actions to be done to the existing running workflow when the conflict policy
```
Sushisource, line 209:
```suggestion
    // Defines actions to be done to the existing running workflow when the conflict policy
```

Sushisource (commented):
I've got no real opinion on the substantive topics. Just one minor grammar change.

cretz (approved):

temporal/api/history/v1/message.proto
```diff
@@ -746,6 +746,11 @@ message WorkflowExecutionOptionsUpdatedEventAttributes {
     temporal.api.workflow.v1.VersioningOverride versioning_override = 1;
     // Versioning override removed in this event.
     bool unset_versioning_override = 2;
+    // Attached request ID to the running workflow execution so that additional requests with same
+    // request ID will be deduped.
+    string attached_request_id = 3;
+    // Attached completion callbacks to the running workflow execution.
```
dandavison, line 752:
```suggestion
    // Completion callbacks attached to the running workflow execution.
```

temporal/api/history/v1/message.proto
```diff
@@ -746,6 +746,11 @@ message WorkflowExecutionOptionsUpdatedEventAttributes {
     temporal.api.workflow.v1.VersioningOverride versioning_override = 1;
     // Versioning override removed in this event.
     bool unset_versioning_override = 2;
+    // Attached request ID to the running workflow execution so that additional requests with same
+    // request ID will be deduped.
```
dandavison, line 750:
```suggestion
    // Request ID attached to the running workflow execution so that subsequent requests with same
    // request ID will be deduped.
```

PR #513, opened 2025-01-14
Initial protos for priority

dnr:
*Note this is a feature branch PR*

**What changed?**
Add initial protos for priority, enough to attach priority metadata to workflows and activities.

Later, we'll add APIs to update priority of existing workflows and activities, to override values in task queue configuration, and observability APIs.

This has the full planned `Priority` message to provide context for reviewers, even though the implementation will come slowly over time. Specifically, in the first phase only the `priority` field will be meaningful. I'll comment out/delete the rest before merging and add them back later.

**Why?**
Control task ordering.


temporal/api/common/v1/message.proto
```diff
@@ -246,3 +246,114 @@ message Link {
...
+    // a queue should be processed in close to priority order, although small
+    // deviations are possible.
+    //
+    // The maximum priority value (minimum priority) is determined by server
+    // configuration, and defaults to 5.
+    //
+    // If priority is not present (or zero), then the effective priority will be
+    // the default priority, which is is calculated by (min+max)/2. With the
+    // default max of 5, and min of 1, that comes out to 3.
+    int32 priority = 1;
```
Sushisource, line 295: I know opinions differ on this matter (thanks, Go) but, I'd be for a uint here.

temporal/api/common/v1/message.proto
```diff
@@ -246,3 +246,114 @@ message Link {
...
+    // 1. Weights for a small set of keys can be overridden in task queue
+    //    configuration with an API.
+    // 2. It can be attached to the workflow/activity/task in this field.
+    // 3. The default weight of 1.0 will be used.
+    // Note that if the weight for a key is attached in this field, for best
+    // results, the same weight should be used for the same key for a reasonable
+    // amount of time (minutes). It may change, but it may take some time for
+    // the change to be reflected.
+    //
+    // The recommended range of usable weights is [1e-3, 1e3].
```
Sushisource, line 334:
```suggestion
    // The recommended range of usable weights is [0.001, 1000].
```

I think most people will find that easier to read.

Sushisource (commented):
Neat! Exciting!

temporal/api/common/v1/message.proto
```diff
@@ -246,3 +246,114 @@ message Link {
...
+    float fairness_weight = 3;
+
+    // The fairness mechanism can also enforce rate limits per fairness key.
+    // Rate limits are specified in tasks dispatched per second.
+    // As with weights, rate limits can come from different sources:
+    // 1. Rate limits for a small set of keys can be overridden in task queue
+    //    configuration with an API.
+    // 2. It can be attached to the workflow/activity/task in this field.
```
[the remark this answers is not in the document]
dnr: That's supposed to be "attached to the workflow" or "attached to the activity". If I just remove "/task" is that clearer?

I added "/task" because it could eventually be attached to nexus tasks and other things, but maybe we just shouldn't even mention that at this point.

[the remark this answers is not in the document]
dnr: The problem we're trying to solve is:

- There's no limit on cardinality of fairness keys.
- We're willing to store overrides for only a limited number (e.g. 100) fairness keys in task queue configuration.
- Users may want to set a distinct weight or rate limit on more than 100 keys.

So we allow just including the weight+rate limit in the priority message itself, essentially using workflow/activity/task metadata to store more than 100 keys, except they're distributed around the system so it's scalable. We don't promise to remember and act on an unlimited number of distinct values, but we can try to act on the ones that we see, as we see them, as best we can.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -478,6 +480,8 @@ message PollActivityTaskQueueResponse {
     // (or not) during activity scheduling. The service can override the provided one if some
     // values are not specified or exceed configured system limits.
     temporal.api.common.v1.RetryPolicy retry_policy = 17;
+    // Priority metadata
+    temporal.api.common.v1.Priority priority = 18;
```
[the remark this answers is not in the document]
dnr: Yeah that was my thought, the sdk can track the priority metadata of the workflow from the started event (plus workflow properties modified events when we add it there later)

temporal/api/history/v1/message.proto
```diff
@@ -133,6 +133,8 @@ message WorkflowExecutionStartedEventAttributes {
     string inherited_build_id = 32;
     // Versioning override applied to this workflow when it was started.
     temporal.api.workflow.v1.VersioningOverride versioning_override = 33;
+    // Priority metadata
```
bergundy, line 136: nit: Can you please be consistent with your use of punctuation?

```suggestion
    // Priority metadata.
```

temporal/api/common/v1/message.proto
```diff
@@ -246,3 +246,114 @@ message Link {
...
+// 2. Next, consider fairness: try to dispatch tasks for different fairness keys
+//    in proportion to their weight.
+// 3. Finally, tasks may be ordered by an additional ordering key.
+//
+// Applications may use any subset of mechanisms that are useful to them and
+// leave the other fields to use default values.
+//
+// Not all queues in the system may support the "full" semantics of all priority
+// fields. (Currently only support in matching task queues is planned.)
+message Priority {
```
bergundy, line 283: Wondering if there's a better name for this than `Priority` since that's just one aspect of this message.
Maybe something like `SchedulingPolicy`?

temporal/api/common/v1/message.proto
```diff
@@ -246,3 +246,114 @@ message Link {
...
+    // fairness mechanisms are applied, tasks will be finally ordered by
+    // ordering_key. Note that fine-grained values are allowed here, as opposed
+    // to priority.
+    //
+    // For example, applications might use the start time of a workflow (in
+    // seconds since the unix epoch) as ordering key for the workflow and all
+    // activities in it. This will have the effect of prioritizing activities of
+    // workflows that were started earlier, encouring completing older workflows
+    // over making progress in newer ones.
+    int64 ordering_key = 5;
```
bergundy, line 358: Do we want to add all of these definitions before they're available?

I think that's fine if you're not merging to main but generally, we shouldn't expose fields that aren't yet supported to avoid confusing consumers of the API and have the ability to iterate on the API as the implementation progresses.

bergundy (approved):

temporal/api/common/v1/message.proto
ShahabT: +1

temporal/api/common/v1/message.proto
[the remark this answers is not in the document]
Sushisource: Yeah I think removing task makes it more clear. Or just elaborate a bit to say that the value _originates_ from the workflow/activition/operation _invocation_ and thus ends up attached to the task.

temporal/api/common/v1/message.proto
dnr: Yeah, I went through a few names, and I agree it's confusing that "priority" the name of the message refers to a larger concept than "priority" the field. But I also want a concise name here. "PriorityFairnessMetadata" is accurate but too long.

SchedulingPolicy suggests something slightly different to me. This isn't a "policy", it's metadata to inform prioritization decisions. Let's keep coming up with ideas though.

temporal/api/common/v1/message.proto
dnr: I said in both the PR description and on slack that things will only be merged to main as implemented :)

The full version is presented here to give context to sdk team when considering how to expose this in sdks.

Question, though: for the parts that aren't implemented yet, is commenting out the fields enough, or should we delete everything?

temporal/api/history/v1/message.proto
dnr: The rule I use (very consistently) is: comments that contain full sentences get periods, comments that are only sentence fragments don't.

I didn't just make this up, it basically matches the recommendation here, for example: https://google.github.io/styleguide/go/decisions#comment-sentences

However, I can use a different rule if you like.

temporal/api/common/v1/message.proto
cretz, line 358: Minor nit, would consider calling this `ordering_value` or something as `_key` has a map-key feeling akin to `fairness_key`

temporal/api/history/v1/message.proto
cretz: FWIW, this is the rule I follow quite religiously in code and is a good one, it differentiates the "phrases" from the multi-sentence paragraphs.

temporal/api/common/v1/message.proto
```diff
@@ -246,3 +246,114 @@ message Link {
...
+//
+// Despite being named "Priority", this message also contains fields that
+// control "fairness" mechanisms.
+//
+// For all fields, the field not present or equal to zero/empty string means to
+// inherit the value from the calling workflow. If there is no calling workflow,
+// then use the default value.
+//
+// Note that for all fields other than fairness_key, the not-present/zero value
+// is invalid, but for fairness_key, the empty string is a valid key. This means
```
cretz, line 267:
```suggestion
// Note that for all fields other than fairness_key, the not-present/zero value
// is invalid when returned from the server, but for fairness_key, the empty string is a valid key. This means
```

I was a bit confused seeing absent/zero as invalid here and the paragraph above talking about absent/zero fields

temporal/api/workflow/v1/message.proto
```diff
@@ -110,6 +110,9 @@ message WorkflowExecutionInfo {
     // be versioned or unversioned, depending on `versioning_info.behavior` and `versioning_info.versioning_override`.
     // Experimental. Versioning info is experimental and might change in the future.
     WorkflowExecutionVersioningInfo versioning_info = 22;
+
+    // Priority metadata
+    temporal.api.common.v1.Priority priority = 23;
```
cretz, line 115: To confirm, this is intentionally in visibility as opposed to putting in `WorkflowExecutionExtendedInfo`?

temporal/api/common/v1/message.proto
```diff
@@ -246,3 +246,114 @@ message Link {
         BatchJob batch_job = 2;
     }
 }
+
+// Priority contains metadata that controls relative ordering of task processing
+// when tasks are backed up in a queue. Initially, Priority will be used in
+// matching (workflow and activity) task queues. Later it may be used in history
+// task queues and in rate limiting decisions.
+//
+// Priority is attached to workflows, activities, and Nexus tasks. By default,
```
cretz, line 255: > and Nexus tasks

I don't see that in this PR, is that maybe coming later?

cretz (approved):
Nothing blocking, especially since this is going to a feature branch

temporal/api/workflow/v1/message.proto
dnr: Good question. It was intentional but I'm not totally confident. I do think it can be valuable to have this info available through visibility, but I'm not sure of all the implications

temporal/api/common/v1/message.proto
dnr: I don't think that quite captures my meaning. It's more about commands, actually: say you start an activity (or child wf) and want to override the priority, but not the fairness key. I'm imagining the sdk would attach a Priority message that includes a non-zero priority, and a zero (empty string) fairness key. The server would interpret that as: override the priority, inherit the fairness key from the workflow.

That is, you can override any individual field(s) by setting it to a non-zero value. That works well in all cases except one: you can't override the fairness key to the empty string. I think that limitation is fine, since that's a pretty weird thing to want to do: if you're using fairness keys at all, you want to set it explicitly in all cases and not mix the default (empty string) with other values.

temporal/api/common/v1/message.proto
bergundy: I would delete. I realize that the target of this PR isn't the main branch but still wanted to ensure my position was clear.

temporal/api/history/v1/message.proto
bergundy: I tend to prefer to use punctuation and proper capitalization in all doc strings. It's more consistent and prevents cases where the author adds more sentences later and forgets to add punctuation to the initial comment.

I switched from y'all's convention to this one but I'm obviously not blocking the PR for my personal preference.
Wondering what others think.

temporal/api/common/v1/message.proto
cretz: Could also be `TaskDispatchStrategy` or something similarly obtuse heh

temporal/api/common/v1/message.proto
cretz: > I'm imagining the sdk would attach a Priority message that includes a non-zero priority, and a zero (empty string) fairness key

Right, which makes the existing statement "for all fields other than fairness_key, the not-present/zero value is invalid" incorrect because clearly it is valid for commands, but adding "when returned from the server" or something else to indicate the rules here about what makes the field invalid don't apply to when it is given by a client

cretz: > That is, you can override any individual field(s) by setting it to a non-zero valu

So you can provide this object as partial not-present/zero values to inherit them? The statement "for all fields other than fairness_key, the not-present/zero value is invalid" seems to say I _have_ to provide every field but fairness key each time I make this object?

temporal/api/common/v1/message.proto
cretz, line 283: I wonder if this needs a qualifier since it's in the `common` package, e.g. `TaskPriority`, or if it should move to the `taskqueue` package.

temporal/api/common/v1/message.proto
dnr: By "the zero value is invalid", I mean that zero is not meaningful if interpreted as a value, and therefore we can be absolutely sure that the intention was "default/inherit". Not that it's not acceptable to send this message with missing/zero values. On the contrary, it's always acceptable to leave out the whole message or any subset of fields. (It has to be so for backwards compatibility)

For fairness key, the zero value _is_ meaningful if interpreted as a value, but I'm saying we don't interpret it that way, we interpret it as "inherit".

There may be some confusion here because in some contexts (schedule activity command) there is a "parent" to inherit from, and in other contexts (start workflow execution) there is no "parent", so you can't "inherit", you just get the default values. The same message is used in both cases though, since it's basically the same thing, it's just that the "parent" is the default values.

temporal/api/common/v1/message.proto
dnr: I want a short simple name, not an obtuse one. In the standup today we discussed this and one popular option was to keep this named "Priority" and change the name of the field to "priority key" to differentiate it.

dnr: The initial application will be in matching task queues, but we can eventually use this in other contexts, for example attaching priorities to any api to influence load shedding at capacity, or attaching priorities to signals or timers (which do turn into tasks at some point but conceptually aren't really just tasks).

temporal/api/common/v1/message.proto
dnr: I'll merge this as-is to keep a reference to point to, and delete on the feature branch before merging

temporal/api/workflow/v1/message.proto
dnr: After some discussion we tentatively agreed to keep this here

temporal/api/common/v1/message.proto
cretz: > By "the zero value is invalid", I mean that zero is not meaningful if interpreted as a value

Ah, I misinterpreted "invalid" to mean "wouldn't pass validation" and would therefore error, meaning it's an invalid thing for SDK side to ever set (or leave unset).

temporal/api/common/v1/message.proto
cretz: Yeah, my concern is pedantic, but so often these assumptions of generic reuse don't materialize do to some minor mismatch need for the other use case. But can keep it as generically named for all concepts of Temporal "priority" ever if we want.

temporal/api/common/v1/message.proto
dnr: I'll reword

temporal/api/common/v1/message.proto
dnr: I had a conversation yesterday about how this can fit into the current flow control work and it seems promising, so I'm really hoping we can get some reuse here. We can reconsider at the feature branch merge.

PR #514, opened 2025-01-14
Add data plane fields w/ renamed Worker Versioning concepts

ShahabT:

**What changed?**
The following concepts in Worker Versioning V3 is being renamed. This PR adds the data plane fields with new names so SDK can populate them like the old ones.
`DeploymentSeries` -> `WorkerDeployment`
`Deployment` -> `WorkerDeploymentVersion`

Also `WorkerVersionCapabilities` and its `use_versioning` boolean are replaced with `WorkerDeploymentOptions` and  `WorkflowVersioningMode` enum to be more compatible with V3 changes and also extensible in the future.

`RespondActivityTask*Request` used to send `deployment` but that is now deprecated. Since Server does not have any anticipated use for these values, I did not add a replacement for them at the moment.

**Why?**
See ^^

**Breaking changes**
Not breaking, some things are deprecated that will be cleaned up at a later time with other old Versioning stuff.

**Server PR**
Server PR will follow soon. But these changes should not break server compile because methods are not touched.

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -35,6 +35,7 @@ import "google/protobuf/duration.proto";
 import "google/protobuf/empty.proto";
 
 import "temporal/api/enums/v1/common.proto";
+import "temporal/api/enums/v1/deployment.proto";
 import "temporal/api/enums/v1/event_type.proto";
 import "temporal/api/enums/v1/reset.proto";
 
@@ -140,6 +141,7 @@ message WorkerVersionStamp {
 // Identifies the version that a worker is compatible with when polling or identifying itself,
 // and whether or not this worker is opting into the build-id based versioning feature. This is
 // used by matching to determine which workers ought to receive what tasks.
+// Deprecated. Use WorkerDeploymentOptions instead.
 message WorkerVersionCapabilities {
     // An opaque whole-worker identifier
     string build_id = 1;
@@ -152,6 +154,17 @@ message WorkerVersionCapabilities {
     string deployment_series_name = 4;
 
     // Later, may include info like "I can process WASM and/or JS bundles"
+}
+
+// Worker Deployment options set in SDK that need to be sent to server in every poll.
+message WorkerDeploymentOptions {
+    // Required. Worker deployment name.
+    string name = 1;
+    // Required. Worker deployment version.
+    string version = 2;
+    // Required. Workflow versioning mode for this deployment version. Must be the same for all
+    // pollers in a given deployment version, across all task queues.
+    temporal.api.enums.v1.WorkflowVersioningMode workflow_versioning_mode = 3;
 }
 
 // Describes where and how to reset a workflow, used for batch reset currently
--- a/temporal/api/enums/v1/deployment.proto
+++ b/temporal/api/enums/v1/deployment.proto
@@ -47,3 +47,23 @@ enum DeploymentReachability {
     // deployment went out of retention period. The deployment can be decommissioned safely.
     DEPLOYMENT_REACHABILITY_UNREACHABLE = 3;
 }
+
+// Workflow Versioning Mode for a particular Worker Deployment Version. This value is
+// configured by the app developer in the worker code.
+enum WorkflowVersioningMode {
+    WORKFLOW_VERSIONING_MODE_UNSPECIFIED = 0;
+    // Workflows processed by this Deployment Version will be unversioned and user
+    // needs to use Patching to keep the new code compatible with prior versions.
+    // This mode is recommended to be used along with Rolling Upgrade deployment
+    // strategies.
+    // Deployment Versions with this mode can not be set as the Current or Ramping
+    // Version of their Deployment, and are not distinguished from each other for
+    // task routing.
+    WORKFLOW_VERSIONING_MODE_UNVERSIONED = 1;
+    // Workflow Versioning Behaviors are enabled in this Deployment Version. Each
+    // workflow type must choose between the Pinned and AutoUpgrade behaviors.
+    // Deployment Versions with this mode can be set as the Current or Ramping
+    // Version of their Deployment and hence are the best option for Blue/Green
+    // and Rainbow strategies (but typically not suitable for Rolling upgrades.)
+    WORKFLOW_VERSIONING_MODE_VERSIONING_BEHAVIORS = 2;
+}
--- a/temporal/api/enums/v1/workflow.proto
+++ b/temporal/api/enums/v1/workflow.proto
@@ -139,24 +139,33 @@ enum TimeoutType {
     TIMEOUT_TYPE_HEARTBEAT = 4;
 }
 
+// Versioning Behavior specifies if and how a workflow execution moves between Worker Deployment
+// Versions. The Versioning Behavior of a workflow execution is typically specified by the worker
+// who completes the first task of the execution, but is also overridable manually for new and
+// existing workflows (see VersioningOverride).
 enum VersioningBehavior {
-    // Workflow execution is unversioned. This is the legacy behavior. An unversioned workflow's
-    // task may go to any unversioned worker who is polling for the task queue.
+    // Workflow execution does not have a Versioning Behavior and is called Unversioned. This is the
+    // legacy behavior. An Unversioned workflow's task may go to any worker who is polling for the
+    // Task Queue and has WorkflowVersioningMode == UNVERSIONED (or has not specified a
+    // WorkflowVersioningMode).
     VERSIONING_BEHAVIOR_UNSPECIFIED = 0;
-    // Workflow will be pinned to the current deployment until completion. Can be overridden
-    // explicitly via `UpdateWorkflowExecutionOptions` API.
-    // Activities of `PINNED` workflows are sent to the same deployment. Exception to this would be
-    // when the activity task queue workers are not present in the workflows deployment, in which
-    // case the activity will be sent to the current deployment of its own task queue.
+    // Workflow will be pinned to their Deployment Version (as specified in versioning_info
+    // .deployment_version) until completion. Can be overridden explicitly via
+    // `UpdateWorkflowExecutionOptions` API to move the execution to another Deployment Version.
+    // Activities of `PINNED` workflows are sent to the same Deployment Version. Exception to this
+    // would be when the activity Task Queue workers are not present in the workflow's Deployment
+    // Version, in which case the activity will be sent to a different Deployment Version according
+    // to the activity's Task Queue situation.
     VERSIONING_BEHAVIOR_PINNED = 1;
-    // Workflow will automatically move to the current deployment of its task queue when the next
-    // workflow task is dispatched.
-    // Activities of `AUTO_UPGRADE` workflows are sent to the current deployment of the workflow
-    // execution based on the last completed workflow task. Exception to this would be when the
-    // activity task queue workers are not present in the workflow's deployment, in which case the
-    // activity will be sent to the current deployment of its own task queue.
-    // Workflows stuck on a backlogged activity will still auto-upgrade if the default deployment
-    // of their task queue changes, without having to wait for the backlogged activity to complete
-    // on the old deployment.
+    // Workflow will automatically move to the Current Deployment Version of its Task Queue when the
+    // next workflow task is dispatched.
+    // Activities of `AUTO_UPGRADE` workflows are sent to the Deployment Version of the workflow
+    // execution (as specified in versioning_info.deployment_version based on the last completed
+    // workflow task). Exception to this would be when the activity Task Queue workers are not
+    // present in the workflow's Deployment Version, in which case, the activity will be sent to a
+    // different Deployment Version according to the activity's Task Queue situation.
+    // Workflows stuck on a backlogged activity will still auto-upgrade if the Current Deployment
+    // Version of their Task Queue changes, without having to wait for the backlogged activity to
+    // complete on the old Version.
     VERSIONING_BEHAVIOR_AUTO_UPGRADE = 2;
 }
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -251,11 +251,20 @@ message WorkflowTaskCompletedEventAttributes {
     // The deployment that completed this task. May or may not be set for unversioned workers,
     // depending on whether a value is sent by the SDK. This value updates workflow execution's
     // `versioning_info.deployment`.
+    // Deprecated. Replaced with `deployment_name` and `deployment_version`.
     temporal.api.deployment.v1.Deployment deployment = 7;
     // Versioning behavior sent by the worker that completed this task for this particular workflow
     // execution. UNSPECIFIED means the task was completed by an unversioned worker. This value
     // updates workflow execution's `versioning_info.behavior`.
     temporal.api.enums.v1.VersioningBehavior versioning_behavior = 8;
+    // The deployment name of the worker that completed this task. Must be set if
+    // `versioning_behavior` is set. This value updates workflow execution's
+    // `versioning_info.deployment_name`.
+    string deployment_name = 9;
+    // The deployment version of the worker that completed this task. Must be set if
+    // `versioning_behavior` is set. This value updates workflow execution's
+    // `versioning_info.deployment_version`.
+    string deployment_version = 10;
 }
 
 message WorkflowTaskTimedOutEventAttributes {
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -148,10 +148,21 @@ message WorkflowExecutionVersioningInfo {
     // a deployment value which will be stored here, so the right way to check if an execution is
     // versioned if an execution is versioned or not is via the `behavior` field.
     // Note that `deployment` is overridden by `versioning_override` if the latter is present.
+    // Deprecated. Use deployment_version and deployment_name.
     temporal.api.deployment.v1.Deployment deployment = 2;
+    // The deployment name of worker that completed the last workflow task of this workflow
+    // execution. Absent value means no workflow task is completed, or the last workflow task was
+    // completed by a legacy worker who does not pass deployment_name.
+    string deployment_name = 5;
+    // The deployment version of worker that completed the last workflow task of this workflow
+    // execution. Must be present if `behavior` is set. Absent value means no workflow task is
+    // completed, or the last workflow task was completed by a legacy worker who does not pass
+    // deployment_version.
+    // Note that if `versioning_override.behavior`.
+    string deployment_version = 6;
     // Present if user has set an execution-specific versioning override. This override takes
-    // precedence over SDK-sent `behavior` (and `deployment` when override is PINNED). An override
-    // can be set when starting a new execution, as well as afterwards by calling the
+    // precedence over SDK-sent `behavior` (and `deployment_version` when override is PINNED). An
+    // override can be set when starting a new execution, as well as afterwards by calling the
     // `UpdateWorkflowExecutionOptions` API.
     VersioningOverride versioning_override = 3;
     // When present, indicates the workflow is transitioning to a different deployment. Can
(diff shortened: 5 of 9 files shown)


cretz (commented):
General concerns about deprecating so many things a few weeks after merging it. I just want to make sure we understand the consequences.

temporal/api/history/v1/message.proto
```diff
@@ -251,11 +251,20 @@ message WorkflowTaskCompletedEventAttributes {
     // The deployment that completed this task. May or may not be set for unversioned workers,
     // depending on whether a value is sent by the SDK. This value updates workflow execution's
     // `versioning_info.deployment`.
+    // Deprecated. Replaced with `deployment_name` and `deployment_version`.
```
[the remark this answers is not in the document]
carlydf: I'm a bit surprised by this take because our [Temporal product release stages guide](https://docs.temporal.io/evaluate/development-production-features/release-stages) guidance on API stability is "Experimental; API is subject to change" for Pre-release and "API breaking changes are kept to a minimum" for Public Preview (and then "API is stable" for GA). It seems appropriate for all parties to refer to that as their source of truth for how to plan around API changes.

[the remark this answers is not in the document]
cretz: There is a difference between SDK API stability, API stability behavior, and proto stability. They are all different forms of API stability. Regardless of what we're _technically_ allowed to do, we should always _try_ for stability at every level.

temporal/api/enums/v1/workflow.proto
```diff
@@ -139,24 +139,33 @@ enum TimeoutType {
...
+    // WorkflowVersioningMode).
     VERSIONING_BEHAVIOR_UNSPECIFIED = 0;
-    // Workflow will be pinned to the current deployment until completion. Can be overridden
-    // explicitly via `UpdateWorkflowExecutionOptions` API.
-    // Activities of `PINNED` workflows are sent to the same deployment. Exception to this would be
-    // when the activity task queue workers are not present in the workflows deployment, in which
-    // case the activity will be sent to the current deployment of its own task queue.
+    // Workflow will be pinned to their Deployment Version (as specified in versioning_info
+    // .deployment_version) until completion. Can be overridden explicitly via
+    // `UpdateWorkflowExecutionOptions` API to move the execution to another Deployment Version.
```
carlydf, line 154:
```suggestion
    // Workflow will start on the Current Deployment Version of its Task Queue, and then
    // will be pinned to that same Deployment Version until completion (the version that
    // a Workflow is pinned to is specified in `versioning_info.deployment_version`).
    // Can be overridden explicitly via `UpdateWorkflowExecutionOptions` API to move the
    // execution to another Deployment Version.
```

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
drewhoskins-temporal: We're intentionally taking an iterative approach to worker versioning.  Insisting that there are no breaking changes for pre-releases ever is a recipe for stasis and overthinking because every decision will seem to carry years of weight, even at the pre-release stage.
I'm okay with very cheap mechanisms to ensure backward compatibility here but overall it's a non-goal and we should take any broader strategic discussions out of band.

temporal/api/enums/v1/workflow.proto
```diff
@@ -139,24 +139,33 @@ enum TimeoutType {
...
+    // Activities of `PINNED` workflows are sent to the same Deployment Version. Exception to this
+    // would be when the activity Task Queue workers are not present in the workflow's Deployment
+    // Version, in which case the activity will be sent to a different Deployment Version according
+    // to the activity's Task Queue situation.
```
antlai-temporal, line 160: This is too vague, the previous comment talked about the "current deployment", giving a hint how the Deployment version is chosen in that case...

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
cretz: I am just bringing up that there are consequences for removing something in the public API and we should strive to avoid this. Sometimes for others making API changes to these protos, the consequences are non-obvious until complaints come about. Granted this PR doesn't remove anything, so this comment doesn't block approval (I was waiting for the other one to be addressed).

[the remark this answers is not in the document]
ShahabT: @cretz I understand the churn has been high in the Versioning project because of many iterations, and I will keep your point about striving for stability in mind more than before. However, it'd not be fair to think we have been ignorant of it so far.

I have hard time imagining a world in which protos don't see such changes and we're able to have pre-release/experimental features! Complex projects such as this need revising some parts during pre-release and it's inevitable that those changes impact the proto API contract in breaking ways... it's not gonna be always additive changes.

As far as I can think to reduce impact of deletion of these fields we maybe can do the following:
1) Keep them around long enough after disabling the deprecated APIs so that majority of wfs with such histories are deleted and it reduced the probability of the fields being set. (only relevant if user has enabled the feature)
2) Remove all the deprecated fields (from all the iterations) at once so it affects proxy users / proto consumers only once.

temporal/api/workflow/v1/message.proto
```diff
@@ -148,10 +148,21 @@ message WorkflowExecutionVersioningInfo {
...
+    // The deployment name of worker that completed the last workflow task of this workflow
+    // execution. Absent value means no workflow task is completed, or the last workflow task was
+    // completed by a legacy worker who does not pass deployment_name.
+    string deployment_name = 5;
+    // The deployment version of worker that completed the last workflow task of this workflow
+    // execution. Must be present if `behavior` is set. Absent value means no workflow task is
+    // completed, or the last workflow task was completed by a legacy worker who does not pass
+    // deployment_version.
+    // Note that if `versioning_override.behavior`.
+    string deployment_version = 6;
```
antlai-temporal, line 162: This comment is incomplete

temporal/api/workflow/v1/message.proto
```diff
@@ -184,6 +215,17 @@ message DeploymentTransition {
     // Later: safe transition info
 }
 
+// Holds information about ongoing transition of a workflow execution from one worker
+// deployment version to another.
+// Experimental. Might change in the future.
+message DeploymentVersionTransition {
+    // The target version of the transition. Absent means a so-far-versioned workflow is
+    // transitioning to unversioned workers.
+    string target_version = 1;
```
antlai-temporal, line 224: Do we need the `deployment_name` here?

temporal/api/workflow/v1/message.proto
```diff
@@ -148,10 +148,21 @@ message WorkflowExecutionVersioningInfo {
     // a deployment value which will be stored here, so the right way to check if an execution is
     // versioned if an execution is versioned or not is via the `behavior` field.
     // Note that `deployment` is overridden by `versioning_override` if the latter is present.
+    // Deprecated. Use deployment_version and deployment_name.
     temporal.api.deployment.v1.Deployment deployment = 2;
+    // The deployment name of worker that completed the last workflow task of this workflow
+    // execution. Absent value means no workflow task is completed, or the last workflow task was
+    // completed by a legacy worker who does not pass deployment_name.
```
carlydf, line 155:
```suggestion
    // The deployment name of the worker that completed the last workflow task of this workflow
    // execution. An absent value means no workflow task is completed, or the last workflow task was
    // completed by a legacy worker who did not pass deployment_name.
```

temporal/api/workflow/v1/message.proto
```diff
@@ -148,10 +148,21 @@ message WorkflowExecutionVersioningInfo {
...
+    string deployment_name = 5;
+    // The deployment version of worker that completed the last workflow task of this workflow
+    // execution. Must be present if `behavior` is set. Absent value means no workflow task is
+    // completed, or the last workflow task was completed by a legacy worker who does not pass
+    // deployment_version.
```
carlydf, line 160:
```suggestion
    // The deployment version of the worker that completed the last workflow task of this workflow
    // execution. Must be present if `behavior` is set. An absent value means no workflow task is
    // completed, or the last workflow task was completed by a legacy worker who did not pass
    // deployment_version.
```

temporal/api/workflow/v1/message.proto
```diff
@@ -428,15 +473,21 @@ message WorkflowExecutionOptions {
...
 
     // Required if behavior is `PINNED`. Must be null if behavior is `AUTO_UPGRADE`.
     // Identifies the worker deployment to pin the workflow to.
+    // Deprecated. Use `pinned_version`.
     temporal.api.deployment.v1.Deployment deployment = 2;
+
+    // Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.
+    // Identifies the worker deployment version to pin the workflow to.
+    string pinned_version = 3;
 }
```
antlai-temporal, line 493: Do we need the deployment_name here, and not just the version?

temporal/api/workflow/v1/message.proto
```diff
@@ -148,10 +148,21 @@ message WorkflowExecutionVersioningInfo {
...
     temporal.api.deployment.v1.Deployment deployment = 2;
+    // The deployment name of worker that completed the last workflow task of this workflow
+    // execution. Absent value means no workflow task is completed, or the last workflow task was
+    // completed by a legacy worker who does not pass deployment_name.
+    string deployment_name = 5;
+    // The deployment version of worker that completed the last workflow task of this workflow
+    // execution. Must be present if `behavior` is set. Absent value means no workflow task is
+    // completed, or the last workflow task was completed by a legacy worker who does not pass
+    // deployment_version.
+    // Note that if `versioning_override.behavior`.
```
carlydf, line 161: did you mean to write more information here?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -415,7 +425,11 @@ message RespondWorkflowTaskFailedRequest {
     temporal.api.common.v1.WorkerVersionStamp worker_version = 8 [deprecated = true];
     // Deployment info of the worker that completed this task. Must be present if user has set
     // `WorkerDeploymentOptions` regardless of versioning being enabled or not.
+    // Deprecated. use `deployment_version`.
     temporal.api.deployment.v1.Deployment deployment = 9;
+    // Deployment version of the worker that completed this task. Must be present if user has set
+    // `WorkerDeploymentOptions`.
+    string deployment_version = 10;
```
antlai-temporal, line 432: Do we need the deployment_name?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -535,6 +552,7 @@ message RespondActivityTaskCompletedRequest {
     temporal.api.common.v1.WorkerVersionStamp worker_version = 5 [deprecated = true];
     // Deployment info of the worker that completed this task. Must be present if user has set
     // `WorkerDeploymentOptions` regardless of versioning being enabled or not.
+    // Deprecated.
     temporal.api.deployment.v1.Deployment deployment = 6;
```
antlai-temporal, line 556: I wonder if for future-proofing we should add deployment_options on the activity task responses/failures even if
the current server does not use it...

carlydf (approved):
approved with a few wording suggestions

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
ShahabT: done

antlai-temporal (commented):
before we had a temporal.api.deployment.v1.Deployment struct that encapsulated the name and the "version".
Now we are splitting in two separate fields deployment_name/deployment_version and we are inconsistent when we pass one or both, i.e., making assumptions that version is all it matters in some places.
I think we should use a DeploymentVersion object (or something like that) that always passes both, because in the future we may decide that the deployment_name is also important for routing too...

temporal/api/workflowservice/v1/request_response.proto
ShahabT: We've had this since V1 always thinking "it might become useful" but it only added churn to the APIs during the iterations. I think if want to add them in future for a known purpose we'd then add them to the next SDK release. Better not to be speculative.

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
antlai-temporal: I'm also concerned about the API churn. I understand an iterative process in which based on customer feedback we do changes on experimental features, but this has not been the case. Most changes are internally motivated without any real customer feedback. If we look at what has been removed from the API in the past, it is very little, the API just keeps on growing... and the main reason is what Chad mentioned, non-sdk users that may use automated tools to generate proxies, UI interfaces that interact directly with the API, and so on...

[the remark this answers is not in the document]
cretz: There is not a world where it churn never happens (but there is a never-remove-only-deprecate world, but even that is too strict).

Let's just make sure we call it out clearly in the API release notes if/when we remove the fields, and let's also make sure that we have a Go release at about the same time as that API release so at least our latest Go SDK works with the latest Go API (which is released at the same time as this repo). We have had issues in the past where compat-breaking changes to this repo were released in `api-go` with no associated Go SDK release causing those doing a `go get -u` to have compile breaks.

cretz (approved):

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
ShahabT: >  let's also make sure that we have a Go SDK release at about the same time as that API release so at least our latest Go SDK works with the latest Go API 
@cretz to clarify, you mean the release in which the deprecated fields are removed, not this one because it is not breaking anything, right?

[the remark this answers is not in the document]
ShahabT: > Let's just make sure we call it out clearly in the API release notes if/when we remove the fields

For if/when I'd like to wait until we're close to GA. We can then add a heads up to one of the release notes and specify a plan, and when the specified time comes after that, delete the fields with proper release note.

temporal/api/workflow/v1/message.proto
carlydf: if we allow transitions between deployments, which it sounds like we do:
```
// When present, indicates the workflow is transitioning to a different deployment version
// (which may belong to the same deployment name or another).
```
Then it seems like we should have deployment_name there?

temporal/api/workflow/v1/message.proto
```diff
@@ -428,15 +473,21 @@ message WorkflowExecutionOptions {
...
+// specific workflow execution. If set, takes precedence over the worker-sent values. See
+// `WorkflowExecutionInfo.VersioningInfo` for more information. To remove the override, call
+// `UpdateWorkflowExecutionOptions` with a null `VersioningOverride`, and use the `update_mask`
+// to indicate that it should be mutated.
 message VersioningOverride {
     // Required.
     temporal.api.enums.v1.VersioningBehavior behavior = 1;
 
     // Required if behavior is `PINNED`. Must be null if behavior is `AUTO_UPGRADE`.
     // Identifies the worker deployment to pin the workflow to.
```
carlydf, line 486:
```suggestion
    // Identifies the worker deployment version to pin the workflow to.
```

temporal/api/workflow/v1/message.proto
carlydf: I haven't thought of a use case where someone would want to override the deployment, but I think if we only allow overriding the version within the same deployment (as we do here), then the comment above VersioningOverride needs to reflect that, so I suggested a change there

ShahabT: > we only allow overriding the version within the same deployment

@carlydf Is this what we want though? I was thinking we'd allow to pin to any version in case users want to move pinned workflows to other deployments.

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
drewhoskins-temporal: > non-sdk users that may use automated tools to generate proxies, UI interfaces that interact directly with the API

Doing this on pre-release APIs in such a way that they cannot change should not be a supported scenario.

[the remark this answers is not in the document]
cretz: > to clarify, you mean the release in which the deprecated fields are removed, not this one because it is not breaking anything, right?

Correct

> Doing this on pre-release APIs in such a way that they cannot change should not be a supported scenario.

The issue is that _we_ are doing this on pre-release APIs in our SDKs. So even for people that don't care about versioning, once we remove the fields, if they upgrade `go.temporal.io/api` to latest but choose not to upgrade their Go SDK (which they can totally do with our version constraints), all of their compilations start breaking. So we just tell them "sorry, even though that SDK says it's compatible with API >= some-version, it's not" and ask them to upgrade both SDK and API.

cretz:
Can we wait on the implementation being basically complete and nearer to a server release (even RC/cloud release) before merging this to `master`? Just want to make sure we give as long of time needed to change our minds during implementation.

temporal/api/common/v1/message.proto
```diff
@@ -154,6 +156,17 @@ message WorkerVersionCapabilities {
     // Later, may include info like "I can process WASM and/or JS bundles"
 }
 
+// Worker Deployment options set in SDK that need to be sent to server in every poll.
+message WorkerDeploymentOptions {
+    // Required. Worker deployment name.
+    string name = 1;
```
Shivs11, line 162: should we call this `deployment_name`? Might be too pedantic, I fear, but I like the extra clarity it brings - just my two cents!

temporal/api/enums/v1/deployment.proto
```diff
@@ -47,3 +47,23 @@ enum DeploymentReachability {
...
+    // Deployment Versions with this mode can not be set as the Current or Ramping
+    // Version of their Deployment, and are not distinguished from each other for
+    // task routing.
+    WORKFLOW_VERSIONING_MODE_UNVERSIONED = 1;
+    // Workflow Versioning Behaviors are enabled in this Deployment Version. Each
+    // workflow type must choose between the Pinned and AutoUpgrade behaviors.
+    // Deployment Versions with this mode can be set as the Current or Ramping
+    // Version of their Deployment and hence are the best option for Blue/Green
+    // and Rainbow strategies (but typically not suitable for Rolling upgrades.)
+    WORKFLOW_VERSIONING_MODE_VERSIONING_BEHAVIORS = 2;
```
Shivs11, line 68: In the case of `WORKFLOW_VERSIONING_MODE_UNVERSIONED`, the comments speak about the use-case of users requiring to use patching to keep things non-breaking. For long running unpinned workflows going through changes, users still are required to use patching to keep things non-breaking. The lack of a mention of patching requirements for `WORKFLOW_VERSIONING_MODE_VERSIONING_BEHAVIORS` almost made me think if I don't need to patch my workflows if I use this behaviour - Should we mention the fact that patching is still required for AutoUpgrade long running wfs?

temporal/api/enums/v1/workflow.proto
```diff
@@ -139,24 +139,35 @@ enum TimeoutType {
...
-    // when the activity task queue workers are not present in the workflows deployment, in which
-    // case the activity will be sent to the current deployment of its own task queue.
+    // Workflow will start on the Current Deployment Version of its Task Queue, and then
+    // will be pinned to that same Deployment Version until completion (the version that
+    // a Workflow is pinned to is specified in `versioning_info.deployment_version`).
+    // Can be overridden explicitly via `UpdateWorkflowExecutionOptions` API to move the
+    // execution to another Deployment Version.
+    // Activities of `PINNED` workflows are sent to the same Deployment Version. Exception to this
+    // would be when the activity Task Queue workers are not present in the workflow's Deployment
+    // Version, in which case the activity will be sent to a different Deployment Version according
```
Shivs11, line 159: nit: in which case the activity will be sent to the Deployment Version containing the activity's task-queues

temporal/api/enums/v1/workflow.proto
```diff
@@ -139,24 +139,35 @@ enum TimeoutType {
...
-    // Workflows stuck on a backlogged activity will still auto-upgrade if the default deployment
-    // of their task queue changes, without having to wait for the backlogged activity to complete
-    // on the old deployment.
+    // Workflow will automatically move to the Current Deployment Version of its Task Queue when the
+    // next workflow task is dispatched.
+    // Activities of `AUTO_UPGRADE` workflows are sent to the Deployment Version of the workflow
+    // execution (as specified in versioning_info.deployment_version based on the last completed
+    // workflow task). Exception to this would be when the activity Task Queue workers are not
+    // present in the workflow's Deployment Version, in which case, the activity will be sent to a
+    // different Deployment Version according to the activity's Task Queue situation.
```
Shivs11, line 168: nit: in which case the activity will be sent to the Deployment Version containing the activity's task-queues

temporal/api/workflow/v1/message.proto
Shivs11: Since `versions` are unique across namespaces, I don't think specifying the `deployment_name` becomes necessary - I also am not sure if we are allowing deployment transitions to happen across namespaces but if we are not, I think having just the `target_version` should suffice

happy to be pointed in the right direction if my understanding is wrong!

Shivs11 (approved):
no blockers - just some nit comments

temporal/api/workflow/v1/message.proto
carlydf: yeah I think moving to other deployments is good! I forgot that version was unique within a namespace and was just trying to think flexibly

temporal/api/workflow/v1/message.proto
carlydf: oops I forgot that version is unique within a namespace, nevermind what I said

ShahabT:
@cretz OK, I changed the base to a feature branch for now.

ShahabT:
@antlai-temporal I added the WorkerDeploymentVersion struct and addressed you comments.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Added depoyment_version now.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: we don't strictly need it but now I'm passing a struct containing both version and name.

temporal/api/common/v1/message.proto
ShahabT: I think the host message name makes it clear: WorkerDeploymentOptions.

temporal/api/enums/v1/workflow.proto
ShahabT: updated

temporal/api/enums/v1/workflow.proto
ShahabT: updated

temporal/api/enums/v1/workflow.proto
ShahabT: updated

temporal/api/enums/v1/deployment.proto
ShahabT: good point I clarified

PR #522, opened 2025-01-19
add ManageActivity API

ychebotarev:
**What changed?**
Add "ManageActivity" API. See comments for the details.

**Why?**
1. Contains changes related to the feedback for single activity API.
2. Support "jitter" for unpausing the activities - part of batch activity work.
3. Support multiple operations at once.

Note: becase "ManageActivity" supports multiple operations, each operation support only flags relevant to that operation.

**Breaking changes**
technically yes, but this is a separate branch.

**Server PR**
Irrelevant.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1949,3 +1949,75 @@ message GetDeploymentReachabilityResponse {
     // was actually calculated.
     google.protobuf.Timestamp last_update_time = 3;
 }
+
+message ManageActivityRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
+
+    // Used to de-dupe requests.
+    string request_id = 4;
+
+    // ID of the activity we're pausing.
+    string activity_id = 5;
+
+    // Type of the activity.
+    string activity_type = 6;
+
+    // Pausing the activity. Either id or type should be provided
+    // If type is provided, all activities of this type will be paused
+    message PauseActivityRequest {
+    }
+
+    // Unpausing the activity. If activity was not paused this will be a no-op.
+    // If activity was waiting for retry - it will be scheduled immediately.
+    message UnpauseActivityRequest {
+        // reset the number of attempts
+        bool reset_attempts = 1;
+
+        // Indicates that activity should reset heartbeat details.
+        // This flag will be applied only to the new instance of the activity.
+        bool reset_heartbeat = 2;
+
+        // If set, the activity will start at a random time within the specified jitter
+        // duration, introducing variability to the start time.
+        google.protobuf.Duration jitter = 4;
+    }
+
+    // Resetting the activity. This operation will reset the number of attempts.
+    // If activity is paused - activity will be unpaused.
+    // If activity is currently waiting for retry - it will be scheduled immediately.
+    message ResetActivityRequest {
+        // If this flag is provided and activity is currently paused - it will stay paused.
+        bool keep_paused = 1;
+        // Indicates that activity should reset heartbeat details.
+        // This flag will be applied only to the new instance of the activity.
+        bool reset_heartbeat = 2;
+    }
+
+    message UpdateActivityOptionsRequest {
+        // Activity options. Partial updates are accepted and controlled by update_mask
+        temporal.api.activity.v1.ActivityOptions activity_options = 6;
+
+        // Controls which fields from `activity_options` will be applied
+        google.protobuf.FieldMask update_mask = 7;
+    }
+
+    message Operation {
+        oneof operation_type {
+            PauseActivityRequest pause = 9;
+            UnpauseActivityRequest unpause = 10;
+            ResetActivityRequest reset = 11;
+            UpdateActivityOptionsRequest update_options = 12;
+        }
+    }
+
+    repeated Operation operations = 8;
+}
+
+message ManageActivityResponse {
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1976,13 +1976,6 @@ message ManageActivityRequest {
     // Unpausing the activity. If activity was not paused this will be a no-op.
     // If activity was waiting for retry - it will be scheduled immediately.
     message UnpauseActivityRequest {
-        // reset the number of attempts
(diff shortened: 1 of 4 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -1019,5 +1019,37 @@ service WorkflowService {
...
+    // 3. Unpause operation. Unpauses the activity. If activity was not paused this will be a no-op.
+    // If activity was waiting for retry - it will be scheduled immediately.
+    // Unpause operation supports the following flags:
+    // * jitter - if set, the activity will start at a random time within the specified jitter duration.
+    // 4. Update operation - updates the activity options.
+    // for details see UnpauseActivityById.
+    // Returns a `NotFound` error if there is no pending activity with the provided ID or type.
+    // (-- api-linter: core::0136::prepositions=disabled
+    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
+    rpc ManageActivity (ManageActivityRequest) returns (ManageActivityResponse) {
```
cretz, line 1043: Can we reuse `ExecuteMultiOperation`? It is not named to be specific to workflows, and we don't surface it to the users as raw multi-op. Rather we make specific SDK things that use the generic multi-op. And server-side the multi-op can only support certain combinations if we want.

ychebotarev: This is already big move for me - I'm not a fan of "one function to rule them all" approach:)
But more important -  I would rather have _some_ separation of concerns. 
I want to keep logical separation - it is hard to explain why we may want to bundle "unpause activity" and "start workflow".  Also routing is different.  Etc.
But separation of concerns is the main objection.

That said - if name is concern how about `ExecuteMultipleActivityOperations`? 
(though ChatGPD doesn't like it, it thinks ManageActivity is concise and descriptive )

cretz: > I'm not a fan of "one function to rule them all" approach

But it was decided before I think. I do like `ExecuteMultipleActivityOperations` more than `ManageActivity`, but please run by server team. I think `ExecuteMultipleOperations` was meant for general purpose use despite personal preferences. But if team decides that it should have been named `ExecuteMultipleWorkflowOperations` and they made a mistake, ok.

ychebotarev: no strong option on name though.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1949,3 +1949,68 @@ message GetDeploymentReachabilityResponse {
...
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
+
+    // Used to de-dupe requests.
+    string request_id = 4;
+
+    // ID of the activity we're pausing.
```
ychebotarev, line 1965: may be this two parameters should be part of each operation

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1949,3 +1949,68 @@ message GetDeploymentReachabilityResponse {
...
 }
+
+message ManageActivityRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+    // ID of the workflow which scheduled this activity.
+    string workflow_id = 2;
+    // Run ID of the workflow which scheduled this activity.
+    // If empty - latest workflow is used.
+    string run_id = 3;
```
alexshtin, line 1960: This is `WorkflowExecution` in the majority of other APIs.

temporal/api/workflowservice/v1/request_response.proto
alexshtin: which two?

temporal/api/workflowservice/v1/service.proto
cretz: > All current use-cases are workflow related

It is clearly not named for only workflows unlike RPCs that are clearly named for workflows. Regardless of reason for avoiding "one single API", this was decided when that API was made. Not sure we should re-litigate for every use case.

ychebotarev: From my conversation with the team - no, we don't think about activity API. 
I put a lot of reasons "why" I think this is better then 'throw them all into single API". 
So far haven't seen any "con" beside "it was decided".

cretz: Inconsistency and disregarding past decisions are major cons. The next PR that comes about wanting to do it yet a third way (then fourth, then fifth) are effectively now encouraged to do so.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1949,3 +1949,68 @@ message GetDeploymentReachabilityResponse {
...
+
+    // Used to de-dupe requests.
+    string request_id = 4;
+
+    // ID of the activity we're pausing.
+    string activity_id = 5;
+
+    // Type of the activity.
+    string activity_type = 6;
```
alexshtin, line 1969: Can it use `oneof`?

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: changed in next PR

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: we can. Changed in the next PR

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: activity_id and activity_type.
I decide agains it.

temporal/api/workflowservice/v1/service.proto
ychebotarev: Inconsistency can't be an argument here - that existing function doesn't support any activity related operation. And was not designed to support. So we are not in "some activity operations are here, some are there" state.

"disregarding past decisions" can't be an argument "per se". If we make a decision to walk using our hands - proposal to work using legs will obviously disregard that decision, but that is not an argument agains it. 
Our situation is even worth - we are not walking using our hands. Yet. But we are trying to enforce walking using hands, because we have some other decision, and "walking" somehow fits into it "by analogy". Thus is we decide to walk - we should walk using hands. 

That said - I confirmed that we never design that function for activity operations. And in its current state it will not fit my use case (but this argument is already in the list of my other arguments).

cretz: > Inconsistency can't be an argument here - that existing function doesn't support any activity related operation. And was not designed to support

Disagree, it was clearly designed to add any number of calls to the `oneof` and was clearly not made to be workflow specific

PR #524, opened 2025-01-21
Add SetWorkerDeploymentRampingVersion and `versioning_info` to DescribeTaskQueue

Shivs11:

**What changed?**
- Title

**Why?**
- Versioning-3.1

**Breaking changes**

**Server PR**
- soon to come (developing `SetWorkerDeploymentCurrentVersion` and Ramping functionality :) )

### DIFF
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -55,6 +55,26 @@ message TaskQueueMetadata {
     google.protobuf.DoubleValue max_tasks_per_second = 1;
 }
 
+message TaskQueueVersioningInfo {
+    // Current Version receives all the tasks except the portion that is routed to the Ramping
+    // Version according to `ramping_version_percentage`.
+    // It is possible that Current Version is not present when user is ramping from unversioned to
+    // versioned workers. In that case, remaining tasks after `ramping_version_percentage` go to
+    // unversioned workers. (Unversioned workers are those with UNVERSIONED (or unspecified)
+    // WorkflowVersioningMode.)
+    temporal.api.deployment.v1.WorkerDeploymentVersion current_version = 1;
+    // Ramping Version is set during gradual rollout of a new Version. It takes a share of tasks
+    // according to `ramping_version_percentage`.
+    temporal.api.deployment.v1.WorkerDeploymentVersion ramping_version = 2;
+    // Percentage of tasks that are routed to the Ramping Version instead of the Current Version.
+    // Special case: it is possible that `ramping_version_percentage` is non-zero while
+    // `ramping_version` is absent. That case represents gradual migration out of the Current
+    // Version to unversioned workers.
+    float ramping_version_percentage = 3;
+    // Last time versioning information of this Task Queue changed.
+    google.protobuf.Timestamp update_time = 4;
+}
+
 // Used for specifying versions the caller is interested in.
 message TaskQueueVersionSelection {
     // Include specific Build IDs.
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -61,9 +61,11 @@ import "temporal/api/batch/v1/message.proto";
 import "temporal/api/sdk/v1/task_complete_metadata.proto";
 import "temporal/api/sdk/v1/user_metadata.proto";
 import "temporal/api/nexus/v1/message.proto";
+
 import "google/protobuf/duration.proto";
 import "google/protobuf/field_mask.proto";
 import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
 
 message RegisterNamespaceRequest {
     string namespace = 1;
@@ -1052,6 +1054,17 @@ message DescribeTaskQueueResponse {
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
     // Only set in `ENHANCED` mode.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3;
+
+    // Specifies which Worker Deployment Version(s) Server routes this Task Queue's tasks to.
+    // When not present, it means the tasks are routed to "unversioned" workers. Unversioned workers
+    // are those with UNVERSIONED (or unspecified) WorkflowVersioningMode.
+    // Task Queue Versioning info is updated indirectly by calling SetWorkerDeploymentCurrentVersion
+    // and SetWorkerDeploymentRampingVersion on Worker Deployments.
+    // Note: This information is not relevant to Pinned workflow executions and their activities as
+    // they are always routed to their Pinned Deployment Version. However, new workflow executions
+    // are typically not Pinned until they complete their first task (unless they are started with
+    // a Pinned VersioningOverride or are Child Workflows of a Pinned parent).
+    temporal.api.taskqueue.v1.TaskQueueVersioningInfo versioning_info = 4;
 }
 
 message GetClusterInfoRequest {
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
     temporal.api.deployment.v1.DeploymentInfo previous_deployment_info = 2;
 }
 
-message SetCurrentDeploymentVersionRequest {
+// Set/unset the Current Version of a Worker Deployment.
+message SetWorkerDeploymentCurrentVersionRequest {
     string namespace = 1;
     string deployment_name = 2;
     // Must be a Versioned Build. Pass empty string to unset.
-    string version = 3;
+    google.protobuf.StringValue version = 3;
     // Optional. The identity of the client who initiated this request.
     string identity = 4;
 }
 
-message SetCurrentDeploymentVersionResponse {
-    // Info of the version that was current before executing this operation.
+message SetWorkerDeploymentCurrentVersionResponse {
+    // The version that was current before executing this operation.
     string previous_version = 1;
+}
+
+// Set/unset the Ramping Version of a Worker Deployment and its ramp percentage.
+// To unset, pass zero `percentage`.
+// To ramp to unversioned workers (from Deployment's Current Version), pass `version=""` with
+// non-zero `percentage`.
+message SetWorkerDeploymentRampingVersionRequest {
+    string namespace = 1;
+    string deployment_name = 2;
+
+    // The Version to be set as the Ramping Version of the Deployment. Optional when `percentage` is
+    // zero (unset case), but if given, it must match the current Ramping Version to be unset.
+    // Can be an empty string and paired with a non-zero `percentage`, in which case it will set the ramp
+    // percentage without a Ramping Version which means unversioned workers are being ramped.
+    // (Useful for migrating to unversioned workers from Deployment's Current Version.)
+    google.protobuf.StringValue version = 3;
+    // Valid range: [0,100]. Zero value will unset the Ramping Version.
+    float percentage = 6;
+
+    // Optional. The identity of the client who initiated this request.
+    string identity = 7;
+}
+
+message SetWorkerDeploymentRampingVersionResponse {
+    // The ramping version before executing this operation.
+    string previous_version = 1;
+    // The ramping version percentage before executing this operation.
+    float previous_percentage = 2;
 }
 
 // Returns the Current Deployment of a deployment series.
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -851,8 +851,9 @@ service WorkflowService {
         };
     }
 
-    // Sets a deployment version as the current version for its deployment.
-    rpc SetCurrentDeploymentVersion (SetCurrentDeploymentVersionRequest) returns (SetCurrentDeploymentVersionResponse) {
+    // Set/unset the Current Version of a Worker Deployment. Automatically unsets the Ramping
+    // Version if it is the Version being set as Current.
+    rpc SetWorkerDeploymentCurrentVersion (SetWorkerDeploymentCurrentVersionRequest) returns (SetWorkerDeploymentCurrentVersionResponse) {
         option (google.api.http) = {
             post: "/namespaces/{namespace}/worker-deployment/{deployment_name}/set-current-version"
             body: "*"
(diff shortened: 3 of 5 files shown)


ShahabT (approved):

temporal/api/workflowservice/v1/service.proto
```diff
@@ -851,13 +851,14 @@ service WorkflowService {
         };
     }
 
-    // Sets a deployment version as the current version for its deployment.
-    rpc SetCurrentDeploymentVersion (SetCurrentDeploymentVersionRequest) returns (SetCurrentDeploymentVersionResponse) {
+    // Set/unset the Current Version of a Worker Deployment. Automatically unsets the Ramping
+    // Version if it is the Version being set as Current.
+    rpc SetWorkerDeploymentCurrentVersion (SetWorkerDeploymentCurrentVersionRequest) returns (SetWorkerDeploymentCurrentVersionResponse) {
         option (google.api.http) = {
-            post: "/namespaces/{namespace}/worker-deployment/{deployment_name}/set-current-version"
```
Shivs11, line 857: @cretz - following your advice from one of my previous PR's, I decided to make these into plurals too - let me know if this is agreeable

cretz: Definitely, looks great

cretz (approved):

temporal/api/taskqueue/v1/message.proto
```diff
@@ -55,6 +55,26 @@ message TaskQueueMetadata {
...
+    // Current Version receives all the tasks except the portion that is routed to the Ramping
+    // Version according to `ramping_version_percentage`.
+    // It is possible that Current Version is not present when user is ramping from unversioned to
+    // versioned workers. In that case, remaining tasks after `ramping_version_percentage` go to
+    // unversioned workers. (Unversioned workers are those with UNVERSIONED (or unspecified)
+    // WorkflowVersioningMode.)
+    temporal.api.deployment.v1.WorkerDeploymentVersion current_version = 1;
+    // Ramping Version is set during gradual rollout of a new Version. It takes a share of tasks
+    // according to `ramping_version_percentage`.
+    temporal.api.deployment.v1.WorkerDeploymentVersion ramping_version = 2;
```
antlai-temporal, line 68: Can the current version and the ramping version belong to different deployments? If not, you need to be explicit in the comments, and describe what would happen if we try to ramp a task queue within a different deployment...

temporal/api/taskqueue/v1/message.proto
```diff
@@ -55,6 +55,26 @@ message TaskQueueMetadata {
...
+    // Percentage of tasks that are routed to the Ramping Version instead of the Current Version.
+    // Special case: it is possible that `ramping_version_percentage` is non-zero while
+    // `ramping_version` is absent. That case represents gradual migration out of the Current
+    // Version to unversioned workers.
+    float ramping_version_percentage = 3;
+    // Last time versioning information of this Task Queue changed.
```
antlai-temporal, line 74: we are using a float instead of an int, what is the range? if it can be 100.0 what does it mean for "current"? Explain...

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1052,6 +1054,17 @@ message DescribeTaskQueueResponse {
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
     // Only set in `ENHANCED` mode.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3;
```
antlai-temporal, line 1056: Is this deprecated?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1052,6 +1054,17 @@ message DescribeTaskQueueResponse {
...
+
+    // Specifies which Worker Deployment Version(s) Server routes this Task Queue's tasks to.
+    // When not present, it means the tasks are routed to "unversioned" workers. Unversioned workers
+    // are those with UNVERSIONED (or unspecified) WorkflowVersioningMode.
```
antlai-temporal, line 1060: I thought that unversioned workers can also belong to a deployment, is there a way to know which deployment this TQ belongs to when they have unversioned workers?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
     temporal.api.deployment.v1.DeploymentInfo previous_deployment_info = 2;
 }
 
-message SetCurrentDeploymentVersionRequest {
+// Set/unset the Current Version of a Worker Deployment.
+message SetWorkerDeploymentCurrentVersionRequest {
```
antlai-temporal, line 1990: We need conflict tokens

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
...
+    google.protobuf.StringValue version = 3;
     // Optional. The identity of the client who initiated this request.
     string identity = 4;
 }
 
-message SetCurrentDeploymentVersionResponse {
-    // Info of the version that was current before executing this operation.
+message SetWorkerDeploymentCurrentVersionResponse {
+    // The version that was current before executing this operation.
+    string previous_version = 1;
```
antlai-temporal, line 2001: We need previous deployment_name and conflict token

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
...
 }
 
-message SetCurrentDeploymentVersionRequest {
+// Set/unset the Current Version of a Worker Deployment.
+message SetWorkerDeploymentCurrentVersionRequest {
     string namespace = 1;
     string deployment_name = 2;
     // Must be a Versioned Build. Pass empty string to unset.
-    string version = 3;
+    google.protobuf.StringValue version = 3;
```
antlai-temporal, line 1994: The empty string means "unversioned" but there is a different meaning of unset that is "do not route anything" If we need the latter we will need to have an "unset" method...
In general, using an empty string to unset when empty  already has a different meaning is confusing

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: What about GetWorkerDeploymentCurrentVersion?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
...
+    // The version that was current before executing this operation.
+    string previous_version = 1;
+}
+
+// Set/unset the Ramping Version of a Worker Deployment and its ramp percentage.
+// To unset, pass zero `percentage`.
+// To ramp to unversioned workers (from Deployment's Current Version), pass `version=""` with
+// non-zero `percentage`.
+message SetWorkerDeploymentRampingVersionRequest {
+    string namespace = 1;
```
antlai-temporal, line 2009: We need conflict tokens, In particular, to control interleavings between setCurrent and setRamping.

It should be an error to make inconsistent the deployment_name of ramping vs current.

It may be easier to make the changes atomic if we have a single call that sets current and ramp at the same time.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
...
+    string deployment_name = 2;
+
+    // The Version to be set as the Ramping Version of the Deployment. Optional when `percentage` is
+    // zero (unset case), but if given, it must match the current Ramping Version to be unset.
+    // Can be an empty string and paired with a non-zero `percentage`, in which case it will set the ramp
+    // percentage without a Ramping Version which means unversioned workers are being ramped.
+    // (Useful for migrating to unversioned workers from Deployment's Current Version.)
+    google.protobuf.StringValue version = 3;
+    // Valid range: [0,100]. Zero value will unset the Ramping Version.
```
antlai-temporal, line 2018: Zero value unsets ramping but 100.0 does not unset current. This is inconsistent...

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
...
+    float percentage = 6;
+
+    // Optional. The identity of the client who initiated this request.
+    string identity = 7;
+}
+
+message SetWorkerDeploymentRampingVersionResponse {
+    // The ramping version before executing this operation.
```
antlai-temporal, line 2026: deployment_name?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1973,18 +1986,47 @@ message SetCurrentDeploymentResponse {
...
+message SetWorkerDeploymentRampingVersionRequest {
+    string namespace = 1;
+    string deployment_name = 2;
+
+    // The Version to be set as the Ramping Version of the Deployment. Optional when `percentage` is
+    // zero (unset case), but if given, it must match the current Ramping Version to be unset.
+    // Can be an empty string and paired with a non-zero `percentage`, in which case it will set the ramp
+    // percentage without a Ramping Version which means unversioned workers are being ramped.
+    // (Useful for migrating to unversioned workers from Deployment's Current Version.)
+    google.protobuf.StringValue version = 3;
```
antlai-temporal, line 2017: StringValues are a bit of a pain in languages like go that give default values to everything. I would prefer if you use a struct with {deployment_name, version} that can be easily passed as null. I think we already have one WorkerDeploymentVersion?

ShahabT: This is just a wrapper to make receiving "" strings possible (as differentiated from not-given). I want to stay away from WorkerDeploymentVersion in here because:
- The deployment_name should be given as a top-level field because this API is calling on that deployment name (and not the version), i.e. it is manipulating the state (current_version) of the deployment resource.
- WorkerDeploymentVersion is meant to represent a version, it feels awkward to have it also represent unversioned and it having an empty `version` value.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Hmm... I don't see it as inconsistent. Zero is the empty value. 100% ramping version means the version is taking all the traffic but still not promoted to become Current because user is still doing final validations on the full load. So there is a value for it, IMO. Is there a value for 0 ramp?

temporal/api/workflowservice/v1/request_response.proto
ShahabT: deployment name is the deployment name that user called this API on.

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: What is your thinking for unversioned workers within a deployment in the future? Are you keeping the data structure? Do you want to update the comment...
If they are different, I'm assuming Current wins, right? Or you are assuming a TQ can be in two deployments (one ramping) and we split traffic accordingly

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: I was thinking about the conflict token. Are you adding the token to DescribeWorkerDeployment?  If not, we need a way to start and get a first token, so we will need a Getter...

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: I was  thinking that for autoUpgrade workflows there could be a different meaning between "move them to unversioned" for next task, and do not change anything, leave them were they are...

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: Could be a way of pausing a ramp to set it to zero? If zero means ramp version disappears, people will assume that 100% will change current, and be surprised when it doesn't...

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: Yes, you are right, conflict can only happen in the TQ if there is a change of deployment for that TQ

temporal/api/workflowservice/v1/request_response.proto
ShahabT: > TQ can be in two deployments (one ramping) and we split traffic accordingly

This one. it's better in that it allows gradual ramp to the other deployment.

> What is your thinking for unversioned workers within a deployment in the future?

There has not been enough discussion about that I believe. Whether we want to restrict unversioned TQs to a single deployment or not I don't know. It seems we need to consider what happens when user wants to move the unversioned TQs between deployments and it seems restricting TQ to a single deployment name can add friction to existing rolling deployments of users, so it needs to have a compelling use case.

But if we decide to add the restriction, then I imagine the deployment name would be added to this struct as a separate field, something like `unversioned_deployment_name` maybe.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: > Are you adding the token to DescribeWorkerDeployment?

yes.

temporal/api/workflowservice/v1/request_response.proto
antlai-temporal: I think an empty string meaning "ramp to unversioned" can be very error prone. We have never used before StringValue in the temporal APIs, preferring to add an extra enum or bool field, or a struct,  to indicate a special value in a more explicit way. I see your first point of deployment_name being the primary key. The issue of representing unversioned within a deployment is going to be awkward anyway, it may be easier to have a global "unversioned"...
@drewhoskins-temporal what do you think?

temporal/api/workflowservice/v1/request_response.proto
ShahabT: No, there we won't have that.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Fair, I'm making that possible in the follow up PR so user can set ramp to 0 without removing ramping version.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: @antlai-temporal I can change it to a different format if you prefer, something like 
```
oneof ramping_version {
  string set_ramping_version
  bool unset_ramping_version
}
```

antlai-temporal: that is a bit more confusing, if you set unset_ramping_version=false, what does it should do? Nothing?
I think the root problem is that unversioned should not be just "" in a setter, since it could have all sort of side-effects. Maybe use an Enum for it, with just one entry. Or reserve a string "__temporal_unversioned_", or add a struct with a field bool isUnversioned and force the "version" to be "" if isUnversioned=true...

PR #525, opened 2025-01-22
Add DrainageStatus

carlydf:

Add DrainageStatus

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -139,12 +139,8 @@ message WorkerDeploymentVersionInfo {
         temporal.api.enums.v1.TaskQueueType type = 2;
     }
 
-    // True in one of the following conditions:
-    //  - This is the Current Version of the Deployment.
-    //  - This is the Ramping Version of the Deployment.
-    //  - This is Unversioned and the Deployment does not have a Current
-    //    Deployment.
-    bool accepts_new_executions = 5;
+    // Describes how workflow tasks are or are not routed to this version.
+    enums.v1.VersionRoutingState routing_state = 5;
 
     // Helps user determine when it is safe to decommission the workers of this
     // Version. Not present when accepts_new_executions is true.
@@ -169,8 +165,8 @@ message WorkerDeploymentVersionInfo {
 message VersionDrainageInfo {
     // Set to DRAINING when accepts_new_executions first becomes false.
     // Set to DRAINED when no more open pinned workflows exist on this version.
-    enums.v1.VersionDrainageState state = 1;
-    // Last time the drainage state changed.
+    enums.v1.VersionDrainageStatus status = 1;
+    // Last time the drainage status changed.
     google.protobuf.Timestamp last_changed_time = 2;
     // Last time the system checked for drainage of this version.
     google.protobuf.Timestamp last_checked_time = 3;
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -131,16 +131,24 @@ message WorkerDeploymentVersionInfo {
     string deployment_name = 2;
 
     google.protobuf.Timestamp create_time = 3;
+
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not current.
+    google.protobuf.Timestamp current_since_time = 4;
+
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not ramping.
+    google.protobuf.Timestamp ramping_since_time = 5;
+
     // All the Task Queues that have ever polled from this Deployment version.
-    repeated DeploymentVersionTaskQueueInfo task_queue_infos = 4;
+    repeated DeploymentVersionTaskQueueInfo task_queue_infos = 6;
 
     message DeploymentVersionTaskQueueInfo {
         string name = 1;
         temporal.api.enums.v1.TaskQueueType type = 2;
     }
-
-    // Describes how workflow tasks are or are not routed to this version.
-    enums.v1.VersionRoutingState routing_state = 5;
 
     // Helps user determine when it is safe to decommission the workers of this
     // Version. Not present when accepts_new_executions is true.
@@ -155,7 +163,7 @@ message WorkerDeploymentVersionInfo {
     //   This means if the Version is "drained" but new workflows are sent to it via
     //   Pinned Versioning Override, the status does not account for those Pinned-override
     //   executions and remains "drained".
-    VersionDrainageInfo drainage_info = 6;
+    VersionDrainageInfo drainage_info = 7;
 
     // TODO (Shivam): + Poller_status + WorkerDeploymentVersionMetadata
 }
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -139,7 +139,40 @@ message WorkerDeploymentVersionInfo {
         temporal.api.enums.v1.TaskQueueType type = 2;
     }
 
-    // TODO (Shivam): Accepts_new_executions + Poller_status + Drainage_status + WorkerBuildMetadata
+    // True in one of the following conditions:
+    //  - This is the Current Version of the Deployment.
+    //  - This is the Ramping Version of the Deployment.
+    //  - This is Unversioned and the Deployment does not have a Current
+    //    Deployment.
+    bool accepts_new_executions = 5;
+
+    // Helps user determine when it is safe to decommission the workers of this
+    // Version. Not present when accepts_new_executions is true.
+    // Current limitations:
+    // - Not supported for Unversioned mode.
+    // - Periodically refreshed, may have delays up to few minutes (consult the
+    //   last_checked_time value).
+    // - Refreshed only when accepts_new_executions is false AND the status is not
+    //   "drained" yet.
+    // - Once the status is changed to "drained", it is not changed until the Version
+    //   becomes Current or Ramping again, at which time the drainage info is cleared.
+    //   This means if the Build is "drained" but new workflows are sent to it via
+    //   Pinned Versioning Override, the status does not account for those Pinned-override
+    //   executions and remains "drained".
+    DrainageStatus drainage_status = 6;
+
+    // Information about workflow drainage to help the user determine when it is safe
+    // to decommission a Build.
+    message DrainageStatus {
+        // True means the Build does not have any Pinned OPEN workflows.
+        bool is_drained = 1;
+        // Approximate time when the build drained.
+        google.protobuf.Timestamp drained_time = 2;
+        // Last time the system checked for drainage of this Build.
+        google.protobuf.Timestamp last_checked_time = 3;
+    }
+
+    // TODO (Shivam): + Poller_status + WorkerBuildMetadata
 }
 
 // A Worker Deployment (Deployment, for short) represents all workers serving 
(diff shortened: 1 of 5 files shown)


temporal/api/deployment/v1/message.proto
```diff
@@ -139,7 +139,40 @@ message WorkerDeploymentVersionInfo {
...
+    // - Not supported for Unversioned mode.
+    // - Periodically refreshed, may have delays up to few minutes (consult the
+    //   last_checked_time value).
+    // - Refreshed only when accepts_new_executions is false AND the status is not
+    //   "drained" yet.
+    // - Once the status is changed to "drained", it is not changed until the Version
+    //   becomes Current or Ramping again, at which time the drainage info is cleared.
+    //   This means if the Build is "drained" but new workflows are sent to it via
+    //   Pinned Versioning Override, the status does not account for those Pinned-override
+    //   executions and remains "drained".
```
carlydf, line 161:
```suggestion
    // - Once the status is changed to "drained", it is not changed until the Version
    //   becomes Current or Ramping again, or if new workflows are sent to it via 
    //   Pinned Versioning Override, at which time the drainage info is cleared.
```
I think that it is confusing for Pinned Override workflows to not count towards drainage status. Users would be upset if they accidentally decommissioned a worker that was handling pinned-override workflows because it said drained, when in reality it was still necessary for those workflows to make progress. Since we expect workflow dispatch via `VersioningOverride` to happen mainly:
- on canary until we have a nexus canary endpoint
- in non-routine bug / incident situations

It won't happen very often, so I think it's alright to send an update to the `WorkerDeploymentVersion` in those two cases and make sure `DrainageStatus` is cleared so that it can track the pinned override workflows moving forward. Since those workflows are pinned, they won't be around for long, and the periodic refresh takes care of the load on visibility.

carlydf: I think most use-cases of pinned override will be starting/moving workflows before a version is drained. During that time, I think it makes sense to track pinned-override workflows for the drainage status.

However, if we don't count Pinned Override executions towards drainage status _after_ something is drained, it would be hard to explain to customers that we _do_ count Pinned Overrides towards drainage status _as long as they start before the version became drained_. IMO, we would have to never track them, or always track them. And IMO, never tracking them could easily lead customers to a place of stuck workflows, whereas clearing drainage status when a workflow is dispatched with override adds a rejected-update latency to the override operation.

One trick to avoid that latency piling up during batch `UpdateWorkflowOptions` could be to only clear the drainage status once (at the beginning and/or end of the batch) and not on every operation..

Another option for tracking pinned override workflows in drainage status could be to expose a "clear drainage status" API that users can call after doing a bunch of overrides, but I think that would confuse them too, especially since in most cases it would not need to be used.

cretz (approved):

Shivs11 (approved):

temporal/api/deployment/v1/message.proto
Shivs11: well put out - I do agree that we should remain consistent as well as complete when using the term `drained` for a version workflow. 

From the two options you have listed here, creating a new API to expose the current drainage API would definitely place extra cognitive load on users who now have to remember to query the API and check the drainage status before decommissioning any of their workers.

Having said that, I do also worry about the rejected-update latency in the earlier proposed idea - if we then go ahead and conduct the update in a batch operation, we need to relay to the customers that there could be delays (not sure what the upper bound now would be) when it comes to "updating" the version workflow's status.

temporal/api/deployment/v1/message.proto
```diff
@@ -139,7 +139,41 @@ message WorkerDeploymentVersionInfo {
...
+    // - Not supported for Unversioned mode.
+    // - Periodically refreshed, may have delays up to few minutes (consult the
+    //   last_checked_time value).
+    // - Refreshed only when accepts_new_executions is false AND the status is not
+    //   "drained" yet.
+    // - Once the status is changed to "drained", it is not changed until the Version
+    //   becomes Current or Ramping again, at which time the drainage info is cleared.
+    //   This means if the Version is "drained" but new workflows are sent to it via
+    //   Pinned Versioning Override, the status does not account for those Pinned-override
+    //   executions and remains "drained".
```
antlai-temporal, line 161: Is this a bug? If I override and start sending workflows I would expect to be draining until these workflows finish, so that I don't remove the workers...

antlai-temporal (commented):
Before we had a separate API because of the cost of doing reachability. Are we assuming that this is no longer a problem because we added caching with minutes of refresh delay? Is rate limiting no longer needed?

temporal/api/deployment/v1/message.proto
ShahabT: This is only a temporary limitation. We'll improve it later, but don't want to spend time on it while working on P0 items because this edge case is not expected to be frequent.

(Note that the limitation is only when the version is already drained and THEN use starts new wfs on it with pinned override. If it's not yet drained and user starts wfs with pinned override, the status will count those wfs.)

temporal/api/enums/v1/deployment.proto
```diff
@@ -48,6 +48,21 @@ enum DeploymentReachability {
     DEPLOYMENT_REACHABILITY_UNREACHABLE = 3;
 }
 
+// Specify the drainage state for a Worker Deployment Version so users can decide whether they
+// can safely decommission the version.
+enum VersionDrainageState {
```
ShahabT, line 53: I know the linter we use suggests naming enums as `xyzState` and not `xyzStatus`, but it seems to me "status" would fit here better because it conveys that this is a read-only value that is only reported to user, as opposed to a state that can be mutable by user directly.

ShahabT:
> Before we had a separate API because of the cost of doing reachability. Are we assuming that this is no longer a problem because we added caching with minutes of refresh delay? Is rate limiting no longer needed?

@antlai-temporal, correct. We wont rate limit DescribeWorkerDeploymentVersion aggressively anymore, we'd just return the latest drainage info. It's server's business how often it'll refresh them (within reasonable intervals), but the last refresh timestamp is given to user.

temporal/api/deployment/v1/message.proto
```diff
@@ -139,7 +139,41 @@ message WorkerDeploymentVersionInfo {
         temporal.api.enums.v1.TaskQueueType type = 2;
     }
 
-    // TODO (Shivam): Accepts_new_executions + Poller_status + Drainage_status + WorkerBuildMetadata
+    // True in one of the following conditions:
+    //  - This is the Current Version of the Deployment.
+    //  - This is the Ramping Version of the Deployment.
+    //  - This is Unversioned and the Deployment does not have a Current
+    //    Deployment.
```
ShahabT, line 146: Based on the latest UI work, it seems useful to be able to differentiate between these from the Version POV. So I'm considering to have the following enum instead of the bool:
```
enum VersionRoutingStatus {
  UNSPECIFIED,
  // No routing status, hence new executions are not routed to this Version (unless they are explicitly pinned to this version).
  NONE,
  // This is the Current Version of the Deployment, hence it receives new executions.
  CURRENT,
  // This is the Ramping Version of the Deployment, hence it receives a share of new executions.
  RAMPING,
}
```
(maybe we'll add more routing statuses particularly for unversioned mode later when we add support for them, but right now we want to keep those parts out of the API because they are not implemented)

carlydf: In the framework of State = writeable / Status = readable consequence of stuff, would this be better named as VersionRoutingState?

carlydf: I just read a bit online, and I like this definition of State vs Status. I feel like it fits with what Max has said

> State is used to describe a stage in a process (e.g. pending/dispatched).
Status is used to describe an outcome of an operation (e.g. success/fail).

carlydf: Wait lol actually it does not work with what Max said. According to the above definition, the enums in this PR should be called `DrainageState` (since draining and drained are stages in the drainage process) and `VersionRoutingStatus` (since current / ramping are outcomes of operations SetCurrent and SetRamp). `VersionRoutingState` could also be valid if current/ramping are considered stages in the rollout process.

ShahabT: hmm... its sort of in the grey area because "current_version" is an state from deployment POV but a version being current or not is a status from Version POV.

But now thinking more, maybe we don't need this enum. Because we'd have to have timestamps such as `current_since` and `ramping_since` for each version and if they have value it means the version is current or ramping.

temporal/api/enums/v1/deployment.proto
```diff
@@ -48,6 +48,36 @@ enum DeploymentReachability {
...
+    VERSION_DRAINAGE_STATUS_DRAINING = 1;
+    // The Worker Deployment Version is not used by new or open workflows, but might be still needed by
+    // Queries sent to closed workflows. The version can be decommissioned safely if user does
+    // not query closed workflows. If the user does query closed workflows for some time x after
+    // workflows are closed, they should decommission the version after it has been drained for that duration.
+    VERSION_DRAINAGE_STATUS_DRAINED = 2;
+}
+
+// Describes the routing state that the user has put this version in.
+enum VersionRoutingState {
```
ShahabT, line 69: I think there may be some controversy around this based on previous design discussions. So, let's not add it just yet. Instead the two timestamps of `current_since` and `ramping_since` should be enough for now. Once unversioned workers are supported we'd have a better idea about these.

temporal/api/deployment/v1/message.proto
```diff
@@ -131,15 +131,53 @@ message WorkerDeploymentVersionInfo {
...
+
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not current.
+    google.protobuf.Timestamp current_since_time = 4;
+
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not ramping.
+    google.protobuf.Timestamp ramping_since_time = 5;
```
Shivs11, line 143: sorry for the late review but how does the thought of encapsulating this under a message sound? Both these time stamps are dealing with routing information - looking below, the timestamps are under a single message 

wdyt @carlydf ?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -345,6 +345,7 @@ message RespondWorkflowTaskCompletedRequest {
     // one would not otherwise have been generated. This is used when the worker knows it is doing
     // something useful, but cannot complete it within the workflow task timeout. Local activities
     // which run for longer than the task timeout being the prime example.
+    // slg: originally created for heartbeat, but may be used to paginate requests with big payloads
```
Shivs11, line 348: im confused/don't understand this comment - was this meant to be here?

temporal/api/deployment/v1/message.proto
```diff
@@ -131,15 +131,52 @@ message WorkerDeploymentVersionInfo {
...
     google.protobuf.Timestamp create_time = 3;
+
+    // Last time `is_current` or `ramp_percentage` of this version changed.
+    google.protobuf.Timestamp routing_update_time = 4;
+
+    // If this version is the current version of its deployment.
+    bool is_current = 5;
+
+    // Range: [0, 100]. Must be zero if is_current is true. Must be non-zero if version is ramping.
+    float ramp_percentage = 6;
```
ShahabT, line 142: For the following reasons, I kinda prefet the previous format which had two separate timestamps better (but yeah, we should add `ramp_percentage` too):
- Based on @antlai-temporal 's feedback, we're going to support ramping version with 0% ramp. This format does not allow distinguishing that from "not ramping version".
- In the UI, we plan to tell user "ramping since x" where x is the time from start of the ramp, not the last ramp percentage change. (that info seems less valuable from user's POV)

Shivs11: so I guess what this would also mean is that we need server to have these separate timestamps present (maybe instead of routingUpdateTime) to keep things less-confusing 

cc - @carlydf

temporal/api/workflowservice/v1/request_response.proto
carlydf: removed

PR #528, opened 2025-01-24
remove ManageActivity, cleanup/extend existing activity API

ychebotarev:
**What changed?**
1. Removed added "ManageActivity"
2. Remove multiple fields from various activity related API, per feedback.
3. Add some other fields per feedback, in particular "jitter"
4. Add the ability to unpause to reset api.

**Why?**
1. Don't want to "win by technicality". I think separation of concerns is a deal breaker, and if I can't convince with it - I can't convince.

but more important
2. Removing pause/unpause/update/reset activity API completely will be bad user experience once we (finally) add http rest api support. The more I think about this the more I convinced that this is a price I'm not ready to pay. Now requests are a bit "bloated", but its better then not to have them at all.

**Breaking changes**
Nope.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1717,6 +1717,7 @@ message ExecuteMultiOperationRequest {
     // - The list of operations must not be empty.
     // - The workflow ids must match across operations.
     // - The only valid list of operations at this time is [StartWorkflow, UpdateWorkflow], in this order.
+    // Only this order is supported:
     //
     // Note that additional operation-specific restrictions have to be considered.
     repeated Operation operations = 2;
@@ -1726,7 +1727,7 @@ message ExecuteMultiOperationRequest {
             // Additional restrictions:
             // - setting `cron_schedule` is invalid
             // - setting `request_eager_execution` is invalid
-            // - setting `workflow_start_delay` is invalid
+            // - setting `workflow_start_delay` is invalid, but only if UpdateWorkflowExecutionRequest present
             StartWorkflowExecutionRequest start_workflow = 1;
 
             // Additional restrictions:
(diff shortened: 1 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1791,27 +1791,14 @@ message PauseActivityByIdRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe requests.
-    string request_id = 6;
+    reserved 6;
+    reserved "request_id";
```
cretz, line 1795: Why are things like `request_id` removed? If this is still a request for a side-effecting top-level call, it should have request ID

ychebotarev: what for?

ychebotarev: you can make the same call twice, the result will be the same. In this situation I don't see how/why we should store request id.

alexshtin: It is not used by server and there is no easy way to support it on server. We don't want another map in MS (as we have for Signals) or in-memory storage (as we have for Updates).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1825,14 +1812,21 @@ message UnpauseActivityByIdRequest {
...
     string identity = 5;
 
-    // Used to de-dupe requests.
-    string request_id = 6;
+    reserved 6;
+    reserved 7; // was resume/reset. Flatten this.
+    reserved 8; // was resume/reset. Flatten this.
+    reserved "request_id";
+    reserved "reset";
+    reserved "resume";
```
alexshtin, line 1820: My personal preference is not do it for experimental APIs, but I will leave it up to you.

temporal/api/workflowservice/v1/request_response.proto
cretz: There is a difference between two separate user calls and one user call that is retried implicitly on network issue. The `request_id` is for the latter. I don't think "you can make the same call twice, the result will be the same" is true if there was another call in between the same call twice, but if `request_id` will not help that situation anyways, then no problem leaving off.

temporal/api/workflowservice/v1/request_response.proto
cretz: Concur from a user POV if we never used these in SDKs (and if we did, we would have to discuss other options to not break compile)

ychebotarev: this API is not used by SDK, only in CLI.

alexshtin: So the only thing which will break is older CLI working with newer server. And only for those who opt-in with experimental APIs. I believe it is ok to ask them to upgrade CLI to continue to use it.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: >but if request_id will not help that situation
This brings us back to "what for?". 
There is no situation to help with, there is nothing to protect, request_id is not needed. 
(and yes, it is hard to support it on server, but IF it will be needed it will be another story)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ok will remove them in the next commit
I have one more field to deprecate after another round with Max.

PR #529, opened 2025-01-24
Add completion callbacks to SignalWithStart request

pdoerner:

**What changed?**
Adding a new field for completion callbacks to `SignalWithStartWorkflowExecutionRequest`

**Why?**
So that Nexus operations which use SignalWithStart can be informed of the workflow's completion.

**Breaking changes**

**Server PR**


bergundy (commented):
We don't need this for Nexus FTR, SignalWithStartOperation is about sending the signal and immediately completing, we don't want to tie the operation completion to the workflow completion.

pdoerner:
Decided this is not required for Nexus, so no need to add it right now.

PR #530, opened 2025-01-24
ListWorkerDeployments

Shivs11:

**What changed?**
- title

**Why?**
- versioning-3.1

**Breaking changes**

**Server PR**
- developing :)

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -2043,13 +2043,9 @@ message ListWorkerDeploymentsResponse {
     // (-- api-linter: core::0123::resource-annotation=disabled --)
     message WorkerDeploymentSummary {
         string name = 1;
-        string current_version = 2;
-        string ramping_version = 3;
-        float ramp_version_percentage = 4;
-        google.protobuf.Timestamp create_time = 5;
-        google.protobuf.Timestamp current_version_update_time = 6;
-        google.protobuf.Timestamp ramping_version_update_time = 7;
-      }
+        google.protobuf.Timestamp create_time = 2;
+        temporal.api.deployment.v1.RoutingInfo routing_info = 3;
+    }
 }
 
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -2029,6 +2029,30 @@ message SetWorkerDeploymentRampingVersionResponse {
     float previous_percentage = 2;
 }
 
+message ListWorkerDeploymentsRequest {
+    string namespace = 1;
+    int32 page_size = 2;
+    bytes next_page_token = 3;
+}
+
+message ListWorkerDeploymentsResponse {
+    bytes next_page_token = 1;
+    // The list of worker deployments.
+    repeated WorkerDeploymentSummary worker_deployments = 2;
+  
+    // (-- api-linter: core::0123::resource-annotation=disabled --)
+    message WorkerDeploymentSummary {
+        string name = 1;
+        string current_version = 2;
+        string ramping_version = 3;
+        float ramp_version_percentage = 4;
+        google.protobuf.Timestamp create_time = 5;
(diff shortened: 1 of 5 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2029,6 +2029,30 @@ message SetWorkerDeploymentRampingVersionResponse {
...
+    // (-- api-linter: core::0123::resource-annotation=disabled --)
+    message WorkerDeploymentSummary {
+        string name = 1;
+        string current_version = 2;
+        string ramping_version = 3;
+        float ramp_version_percentage = 4;
+        google.protobuf.Timestamp create_time = 5;
+        google.protobuf.Timestamp current_version_update_time = 6;
+        google.protobuf.Timestamp ramping_version_update_time = 7;
+      }
```
cretz, line 2052:
```suggestion
    }
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2029,6 +2029,30 @@ message SetWorkerDeploymentRampingVersionResponse {
...
+    bytes next_page_token = 3;
+}
+
+message ListWorkerDeploymentsResponse {
+    bytes next_page_token = 1;
+    // The list of worker deployments.
+    repeated WorkerDeploymentSummary worker_deployments = 2;
+  
+    // (-- api-linter: core::0123::resource-annotation=disabled --)
+    message WorkerDeploymentSummary {
```
cretz, line 2044: This looks like `RoutingInfo`, is it possible to reuse types? (if not `RoutingInfo` directly, have it be the first-and-for-now-only field in this message)

carlydf: My vote would be for something like this (also them `WorkerDeploymentSummary` has a subset of the fields in `WorkerDeploymentInfo` which is nice. You could even comment defining Summary as a subset of info and then there is a clear semantic about what should/should not be in each message.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2029,6 +2029,30 @@ message SetWorkerDeploymentRampingVersionResponse {
...
+    int32 page_size = 2;
+    bytes next_page_token = 3;
+}
+
+message ListWorkerDeploymentsResponse {
+    bytes next_page_token = 1;
+    // The list of worker deployments.
+    repeated WorkerDeploymentSummary worker_deployments = 2;
+  
+    // (-- api-linter: core::0123::resource-annotation=disabled --)
```
carlydf, line 2043:
```suggestion
    // (-- api-linter: core::0123::resource-annotation=disabled --)
    // A subset of WorkerDeploymentInfo
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2029,6 +2029,30 @@ message SetWorkerDeploymentRampingVersionResponse {
...
+    message WorkerDeploymentSummary {
+        string name = 1;
+        string current_version = 2;
+        string ramping_version = 3;
+        float ramp_version_percentage = 4;
+        google.protobuf.Timestamp create_time = 5;
+        google.protobuf.Timestamp current_version_update_time = 6;
+        google.protobuf.Timestamp ramping_version_update_time = 7;
```
carlydf, line 2051:
```suggestion
        string name = 1;
        google.protobuf.Timestamp create_time = 2;
        RoutingInfo routing_info 3;
```

carlydf (commented):
I agree with Chad about reusing RoutingInfo instead of repeating the fields, but other than that this looks good!

temporal/api/workflowservice/v1/request_response.proto
Shivs11: I agree that `RoutingInfo` should come here - my bad for not catching this - done!

carlydf (approved):

antlai-temporal:
LGTM

PR #532, opened 2025-01-27
Discuss: namespace rules prototype

ychebotarev:
**What changed?**
This is a proposal for "namespace level rules". First implementation will be "activity pause".
Each rule is defined by:
* predicate - SQL like (or "visibility-like") filtering conditions. Format is specific for provided type.
* type. Type defines both triggering condition and available columns. Right now the only available one is "activity".
* action. Define the action. Currently mostly activity related. Will be expanded to "workflow pause", and then for other rules.

**Why?**
Part of the operational efforts. In case of ongoing issues users don't only want to affect (pause for example) running workflows/activities, but also prevent incoming workflows/activities from running into the same issues. 
But based on the feedback the scope can be wider, and this can be solved in more generable way. Thus this proposal.


temporal/api/enums/v1/namespace.proto
```diff
@@ -49,3 +49,11 @@ enum ReplicationState {
     REPLICATION_STATE_NORMAL = 1;
     REPLICATION_STATE_HANDOVER = 2;
 }
+
+enum RuleActionType {
+    RULE_ACTION_TYPE_UNSPECIFIED = 0;
+    RULE_ACTION_TYPE_PAUSE_ACTIVITY = 1;
+    RULE_ACTION_TYPE_EMIT_METRIC = 2;
+    RULE_ACTION_TYPE_EMIT_VISIBILITY = 3;
+    RULE_ACTION_TYPE_PAUSE_WORKFLOW = 4;
```
cretz, line 58: Pause workflow?

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +108,42 @@ message NamespaceFilter {
...
+
+message RuleAction {
+    // Type of the rule action.
+    temporal.api.enums.v1.RuleActionType type = 1;
+    // Payload of the rule action.
+    temporal.api.common.v1.Payload payload = 2;
+}
+
+// NamespaceRule describes a rule that can be to any workflow in this namespace.
+message NamespaceRule {
```
cretz, line 120: That sounds more like `NamespaceWorkflowRule`

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +108,42 @@ message NamespaceFilter {
     // Note: Namespace is in NAMESPACE_STATE_DELETED state when it was deleted from the system but associated data is not deleted yet.
     bool include_deleted = 1;
 }
+
+message RuleAction {
+    // Type of the rule action.
+    temporal.api.enums.v1.RuleActionType type = 1;
+    // Payload of the rule action.
+    temporal.api.common.v1.Payload payload = 2;
```
cretz, line 116: Why is this a payload? What can it contain? Payloads are (usually) user-encryptable byte arrays only.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +108,42 @@ message NamespaceFilter {
...
+    google.protobuf.Timestamp expiration_time = 3;
+
+    // Predicate is a query that is used to match against workflow data, based on rule type.
+    string predicate = 4;
+    // Type of the rule. Specifies how the rule should be triggered and evaluated.
+    // Currently, only "activity" type is supported.
+    // Types:
+    // - activity: Rule is triggered when a activity is about to start.
+    // - workflow: TBD.
+    // The following activity attributes are supported as part of the predicate:
```
cretz, line 138: What are activity attributes?

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +108,42 @@ message NamespaceFilter {
...
+    // Can be empty if the rule should never expire.
+    google.protobuf.Timestamp expiration_time = 3;
+
+    // Predicate is a query that is used to match against workflow data, based on rule type.
+    string predicate = 4;
+    // Type of the rule. Specifies how the rule should be triggered and evaluated.
+    // Currently, only "activity" type is supported.
+    // Types:
+    // - activity: Rule is triggered when a activity is about to start.
+    // - workflow: TBD.
```
cretz, line 137: Why is this a string? Shouldn't it be better typed/specified with a `oneof` or an enum?

temporal/api/namespace/v1/message.proto
ychebotarev: for now I think it will be restricted to workflow.

temporal/api/enums/v1/namespace.proto
ychebotarev: It can wait for implementation, but can't wait for "prototype" :)

temporal/api/namespace/v1/message.proto
ychebotarev: I assume some actions may have parameters. Like "emit metric" - either pre-defined metrics always, or user defined.
Again, technically for now (activities) we may remove it and add it later if needed, but I want this to be available for discussion.

temporal/api/namespace/v1/message.proto
ychebotarev: mmm right below?  workflowType is not "activity attribute" of cause, but the reset are.

ychebotarev: right below, beside "workflowType".

temporal/api/namespace/v1/message.proto
ychebotarev: yeah, will convert to enum

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1019,5 +1019,56 @@ service WorkflowService {
...
+            post: "/namespaces/{namespace}/rules/{rule.name}"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/rules/{rule.name}"
+                body: "*"
+            }
+        };
+    }
+
+    rpc ReadNamespaceRule (ReadNamespaceRuleRequest) returns (ReadNamespaceRuleResponse) {
```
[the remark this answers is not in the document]
ychebotarev: ok

temporal/api/namespace/v1/message.proto
ychebotarev: more like:
* I can't imagine this to be on a higher level then workflows
* even if it is I probably want them to be separated

ychebotarev: I will rename it.

temporal/api/namespace/v1/message.proto
ychebotarev: Another usage - "send signal" as an action.
I will remove it when going to main, but keep it as a message around action type so we can add it later.

temporal/api/namespace/v1/message.proto
ychebotarev: I actually will remove workflowType from here

temporal/api/enums/v1/namespace.proto
ychebotarev: anyway - removed (I will clean comments when going to main)

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1019,5 +1019,56 @@ service WorkflowService {
...
+    rpc ListNamespaceRules (ListNamespaceRulesRequest) returns (ListNamespaceRulesResponse) {
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/rules"
+            additional_bindings {
+                get: "/api/v1/namespaces/{namespace}/rules"
+            }
+        };
+    }
+
+    rpc ApplyNamespaceRule(ApplyNamespaceRuleRequest) returns (ApplyNamespaceRuleResponse) {
```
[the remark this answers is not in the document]
ychebotarev: done

temporal/api/namespace/v1/message.proto
ychebotarev: Overall - yes, NamespaceRule fills like scope is too big. But NamespaceWorkflowRule feels like scope is too narrow.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,50 @@ message NamespaceFilter {
...
+    // temporal.api.common.v1.Payload payload = 2;
+}
+
+// NamespaceRule describes a rule that can be to any workflow in this namespace.
+message NamespaceRule {
+    // Name of the rule.
+    string name = 1;
+
+    // ID must be unique within the namespace
+    string id = 2;
```
alexshtin, line 126: I think you should follow the same approach as we do for namespaces themselves. `Id` is hidden from the user (mostly). When user creates rule, it specifies only `name` and gets `Id` back in response. Then user can address rule by name (and by `Id` too). Yes, you can't guarantee uniqueness of the name. We need to discuss what we can do in this case.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1948,3 +1949,63 @@ message GetDeploymentReachabilityResponse {
...
+}
+
+message ListNamespaceRulesResponse {
+    repeated temporal.api.namespace.v1.NamespaceRule rules = 1;
+}
+
+message ApplyNamespaceRuleRequest {
+    string namespace = 1;
+
+    temporal.api.workflow.v1.WorkflowExecutionInfo workflow_execution_info = 2;
```
alexshtin, line 2004:
```suggestion
    temporal.api.workflow.v1.WorkflowExecution workflow_execution = 2;
```
It took me a while to figure this out :-)

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1019,5 +1019,58 @@ service WorkflowService {
...
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/rules"
+            additional_bindings {
+                get: "/api/v1/namespaces/{namespace}/rules"
+            }
+        };
+    }
+
+    // ApplyNamespaceRule allows to apply rule to a certain workflow instance without creating a rule.
+    // This is useful for one-off operations.
```
alexshtin, line 1064: and for tests too. I mean users can test how feature works with test workflows.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1948,3 +1949,63 @@ message GetDeploymentReachabilityResponse {
...
+
+message CreateNamespaceRuleResponse {
+    // id of the created rule. Unique within the namespace.
+    string rule_id = 1;
+}
+
+message DescribeNamespaceRuleRequest {
+    string namespace = 1;
+
+    string rule_id = 2;
```
alexshtin, line 1976: It would be great if you use `oneof` for both `id` and `name`. And for `Delete` too.

temporal/api/namespace/v1/message.proto
cretz: We should update the doc that says it's only intended for workflows (unsure if we can expect such a thing when the expected scope-expanse occurs)

temporal/api/namespace/v1/message.proto
cretz: :+1: We can use payload for arbitrary user-defined data as needed, but it's not used for _our_ need of arbitrary data

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,50 @@ message NamespaceFilter {
...
+
+    // Type of the rule trigger. Right now only activity is supported.
+    // Trigger type defines when the rule should be evaluated.
+    // It also defines the columns available in the predicate.
+    // - activity: Rule will be triggered when an activity is about to start.
+    temporal.api.enums.v1.RuleTriggerType trigger_type = 3;
```
cretz, line 132: I think the concept of "action" and the "trigger" should be merged. Some actions make no sense with some triggers I assume. I think an action should also define when it should take place alongside its other options.

temporal/api/enums/v1/namespace.proto
```diff
@@ -49,3 +49,19 @@ enum ReplicationState {
     REPLICATION_STATE_NORMAL = 1;
     REPLICATION_STATE_HANDOVER = 2;
 }
+
+// I will remove it once we agree on the overall design
+// RULE_ACTION_TYPE_EMIT_VISIBILITY = 3;
+// RULE_ACTION_TYPE_PAUSE_WORKFLOW = 4;
+// RULE_ACTION_TYPE_SIGNAL = 5;
+enum RuleActionType {
```
cretz, line 57:
```suggestion
enum NamespaceRuleActionType {
```

Not sure it's safe to assume every use of the term "rule" will apply to namespace rules (same in other places where not qualified)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1948,3 +1949,63 @@ message GetDeploymentReachabilityResponse {
...
+    string namespace = 1;
+
+    // The rule to create as a part of namespace config.
+    // Rule name must be unique namespace-wide.
+    temporal.api.namespace.v1.NamespaceRule rule = 2;
+
+    // If true, the rule will be applied immediately via batch job.
+    bool force_apply = 3;
+
+    // the workflow query to use to apply the rule
```
cretz, line 1964: Pedantic, but there is a lot of inconsistent with capitalization of comments, punctuation, etc

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1948,3 +1949,63 @@ message GetDeploymentReachabilityResponse {
...
+    string workflow_query = 4;
```
cretz, line 1965: Isn't this already in the rule?

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,50 @@ message NamespaceFilter {
...
+
+    // Predicate is a query that is used to match against workflow data, based on rule type.
+    // The following activity attributes are supported as part of the predicate:
+    // - activityType: The type of the activity.
+    // - activityId: The ID of the activity.
+    // - activityAttempt: The number attempts of the activity.
+    // - backoffInterval: The current backoff interval of the activity.
+    // - activityStatus: The status of the activity. Can be one of "scheduled", "started", "paused".
+    // - taskQueue: The name of the task queue of the activity.
+    string predicate = 4;
```
cretz, line 142: What is a "predicate"? Is this some kind of new query format? Is it possible to use well-typed structured representations of this? If not, I would at least request a grammar (we didn't have it for the longest time for workflow query and people were confused for a long time), but ideally we are stronger typed and more structural than that.

temporal/api/namespace/v1/message.proto
ychebotarev: there is no 1:1 mapping between "action" and "trigger". For example workflow pause can be triggered by activity rule.

temporal/api/enums/v1/namespace.proto
```diff
@@ -49,3 +49,19 @@ enum ReplicationState {
...
+    RULE_ACTION_TYPE_UNSPECIFIED = 0;
+    RULE_ACTION_TYPE_PAUSE_ACTIVITY = 1;
+    RULE_ACTION_TYPE_EMIT_METRIC = 2;
```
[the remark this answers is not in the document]
ychebotarev: I was thinking about emitting metrics with predefined name, with dimensions like namespace, "type". May be with "rule name" as well.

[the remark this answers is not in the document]
cretz: Can you document the exact behavior of this enum? It was unclear to me and I suspect other users of the API what this means exactly? I do think this and other actions make sense as structured messages instead of just enumerates.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I go back and force on this a lot...
we will discuss it offline

temporal/api/enums/v1/namespace.proto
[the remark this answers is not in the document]
ychebotarev: yeah, in the doc I moved from enum to "oneof" and messages, so we can have structured action payload later if we need to.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,50 @@ message NamespaceFilter {
     // Note: Namespace is in NAMESPACE_STATE_DELETED state when it was deleted from the system but associated data is not deleted yet.
     bool include_deleted = 1;
 }
+
+message RuleAction {
+    // Type of the rule action.
+    temporal.api.enums.v1.RuleActionType type = 1;
```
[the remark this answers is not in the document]
ychebotarev: yeah, that is exactly what I did, PR is not updated - I want to update it after the discussion.

PR #533, opened 2025-01-27
Add activty type as an option for activity API

ychebotarev:
**What changed?**
remove acitvity_id as a single option for Activity API.
Now callers can chose to pause/undate/unpause/reset all the pending activities of a certain type.

**Why?**
Design review feedback.
If multiple activities are running in parallel they are usually of the same type, and experience the same problem.

**Breaking changes**
No with current implementation.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1756,8 +1756,6 @@ message UpdateActivityOptionsByIdRequest {
     // Run ID of the workflow which scheduled this activity
     // if empty - latest workflow is used
     string run_id = 3;
-    // ID of the activity we're updating
-    string activity_id = 4;
 
     // The identity of the client who initiated this request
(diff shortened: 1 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1768,8 +1766,18 @@ message UpdateActivityOptionsByIdRequest {
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 7;
 
-    // Used to de-dupe requests
-    string request_id = 8;
+    // either activity id or activity type must be provided
```
cretz, line 1769: > or type

This doesn't make sense in an `UpdateActivityOptionsByIdRequest` IMO. This API call as named only makes sense for a single activity _by ID_ as named. If there are batch needs or a non-ID-specific update, maybe a different API call is needed or this one is improperly named.

PR #534, opened 2025-01-28
Single Activity API changes

ychebotarev:
**What changed?**
This PR  contains  a commutative feedback for single activity API:
* add 'jitter' fields to Reset and Unpause
* 'Flatten' unpause. Previous approach was marked as overcomplicated
* Add 'keep_unpause' flag to Reset. By default reset will also unpause the activity.
* Remove no-wait flag. By default, if activity is in retry, they will be scheduled immediately (*jitter)
* Add "activity type" as a routing parameter to every activity operation. If activity type is provided - every pending activity of this type will be paused/reset/updated/unpaused
* Update API descriptions to reflect those changes.

**Why?**
Working on single activity API feedback from the design review.

**Breaking changes**
Yes

**Server PR**
[Add batch activity unpause. Single activity commulative changes](https://github.com/temporalio/temporal/pull/7169)

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1814,16 +1814,18 @@ message UnpauseActivityRequest {
         string id = 4;
         // Unpause all running activities with of this type.
         string type = 5;
+        // Unpause all running activities.
+        bool unpause_all = 6;
     }
 
     // Providing this flag will also reset the number of attempts.
-    bool reset_attempts = 6;
+    bool reset_attempts = 7;
 
     // Providing this flag will also reset the heartbeat details.
-    bool reset_heartbeat = 7;
+    bool reset_heartbeat = 8;
 
     // If set, the activity will start at a random time within the specified jitter duration.
-    google.protobuf.Duration jitter = 8;
+    google.protobuf.Duration jitter = 9;
 }
 
 message UnpauseActivityResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1765,9 +1765,9 @@ message UpdateActivityOptionsRequest {
 
     // either activity id or activity type must be provided
     oneof activity {
-        // ID of the activity we're pausing.
+        // Only activity with this ID will be updated.
         string id = 6;
-        // Type of the activity we're pausing.
+        // Update all running activities of this type.
         string type = 7;
     }
 }
@@ -1788,9 +1788,9 @@ message PauseActivityRequest {
 
     // either activity id or activity type must be provided
     oneof activity {
-        // ID of the activity we're pausing.
+        // Only activity with this ID will be paused.
         string id = 4;
-        // Type of the activity we're pausing.
+        // Pause all running activities of this type.
         string type = 5;
     }
 
@@ -1810,16 +1810,16 @@ message UnpauseActivityRequest {
 
     // either activity id or activity type must be provided
     oneof activity {
-        // ID of the activity we're pausing.
+        // Only activity with this ID will be unpause.
         string id = 4;
-        // Type of the activity we're pausing.
+        // Unpause all running activities with of this type.
         string type = 5;
     }
 
-    // unpause can also reset the number of attempts
+    // Providing this flag will also reset the number of attempts.
     bool reset_attempts = 6;
 
-    // unpause can also reset the heartbeat details
+    // Providing this flag will also reset the heartbeat details.
     bool reset_heartbeat = 7;
 
     // If set, the activity will start at a random time within the specified jitter duration.
@@ -1840,9 +1840,9 @@ message ResetActivityRequest {
 
     // either activity id or activity type must be provided
     oneof activity {
-        // ID of the activity we're pausing.
+        // Only activity with this ID will be reset.
         string id = 4;
-        // Type of the activity we're pausing.
+        // Reset all running activities with of this type.
         string type = 5;
     }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1756,8 +1756,6 @@ message UpdateActivityOptionsByIdRequest {
     // Run ID of the workflow which scheduled this activity
     // if empty - latest workflow is used
     string run_id = 3;
-    // ID of the activity we're updating
-    string activity_id = 4;
 
     // The identity of the client who initiated this request
     string identity = 5;
@@ -1768,8 +1766,18 @@ message UpdateActivityOptionsByIdRequest {
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 7;
 
-    // Used to de-dupe requests
-    string request_id = 8;
+    // either activity id or activity type must be provided
+    oneof activity {
+        // ID of the activity we're pausing.
+        string id = 9;
+        // Type of the activity we're pausing.
+        string type = 10;
+    }
+
+    reserved 4;
+    reserved 8;
+    reserved "activity_id";
+    reserved "request_id";
 }
 
 message UpdateActivityOptionsByIdResponse {
@@ -1785,33 +1793,29 @@ message PauseActivityByIdRequest {
     // Run ID of the workflow which scheduled this activity.
     // If empty - latest workflow is used.
     string run_id = 3;
-    // ID of the activity we're updating.
-    string activity_id = 4;
+
 
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe requests.
-    string request_id = 6;
+    // either activity id or activity type must be provided
+    oneof activity {
+        // ID of the activity we're pausing.
+        string id = 7;
+        // Type of the activity we're pausing.
+        string type = 8;
+    }
+
+    reserved 4;
+    reserved 6;
+    reserved "activity_id";
+    reserved "request_id";
 }
 
 message PauseActivityByIdResponse {
 }
 
 message UnpauseActivityByIdRequest {
-    message ResumeOperation {
-        // Indicates that if the activity is waiting to retry, it will  be scheduled immediately.
-        bool no_wait = 1;
-    }
-
-    message ResetOperation {
-        // Indicates that the activity should be scheduled immediately.
-        // Note that this may run simultaneously with any existing executions of the activity.
-        bool no_wait = 1;
-        // If set, the Heartbeat Details will be cleared out to make the Activity start over from the beginning
-        bool reset_heartbeat = 2;
-    }
-
     // Namespace of the workflow which scheduled this activity.
     string namespace = 1;
     // ID of the workflow which scheduled this activity.
@@ -1819,20 +1823,36 @@ message UnpauseActivityByIdRequest {
     // Run ID of the workflow which scheduled this activity.
     // If empty - latest workflow is used.
     string run_id = 3;
-    // ID of the activity we're updating.
-    string activity_id = 4;
 
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Used to de-dupe requests.
-    string request_id = 6;
-
-    // There are two options to resume the activity - with 'resume' or with 'reset'.
-    oneof operation{
-        ResumeOperation resume = 7;
-        ResetOperation reset = 8;
+    // either activity id or activity type must be provided
+    oneof activity {
+        // ID of the activity we're pausing.
+        string id = 12;
+        // Type of the activity we're pausing.
+        string type = 13;
     }
+
+    // unpause can also reset the number of attempts
+    bool reset_attempts = 9;
+
+    // unpause can also reset the heartbeat details
+    bool reset_heartbeat = 10;
+
+    // If set, the activity will start at a random time within the specified jitter duration.
+    google.protobuf.Duration jitter = 11;
+
+    reserved 4;
+    reserved 6;
+    reserved 7;
+    reserved 8;
+    reserved "activity_id";
+    reserved "request_id";
+    reserved "reset";
+    reserved "resume";
+
 }
 
 message UnpauseActivityByIdResponse {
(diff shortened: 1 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1767,8 +1765,18 @@ message UpdateActivityOptionsByIdRequest {
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 7;
 
-    // Used to de-dupe requests
-    string request_id = 8;
+    // either activity id or activity type must be provided
```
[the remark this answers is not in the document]
ychebotarev: Being able to provide activity type was part of the feedback. If you have multiple activities of the same type running in parallel (the most common scenario) - you want to pause all of them. Doing this 1 by 1 providing activity ids is tiresome.

You right, now "ByID" make less sense.
As I remember, I added "ByID" because of other "ByID" functions that has the same routings (as opposite to "task token")

I'm totally ok with removing it.  

I also will not call it "batch", it still a single workflow, and we use term "batch" for multiple workflows.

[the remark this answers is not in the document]
ychebotarev: since we are doing this - I also change "workflowID+runID" to execution_info, like in every other API.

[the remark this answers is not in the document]
ychebotarev: looks much cleaner now

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1747,123 +1747,117 @@ message ExecuteMultiOperationResponse {
...
-    // Used to de-dupe requests
-    string request_id = 8;
+    google.protobuf.FieldMask update_mask = 5;
+
+    // either activity id or activity type must be provided
+    oneof activity {
+        // ID of the activity we're pausing.
+        string id = 6;
+        // Type of the activity we're pausing.
+        string type = 7;
```
Sushisource, line 1770: Type could apply to multiple activities running concurrently.

Assuming that's the case, this should be something like "Pause all running activities with of this type"

(applies everywhere with this field)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1747,123 +1747,117 @@ message ExecuteMultiOperationResponse {
...
+        // ID of the activity we're pausing.
+        string id = 4;
+        // Type of the activity we're pausing.
+        string type = 5;
     }
+
+    // unpause can also reset the number of attempts
+    bool reset_attempts = 6;
+
+    // unpause can also reset the heartbeat details
```
Sushisource, line 1821: "Can" or "will"? 

It sounds like these should be something like "If set, this unpause will also reset..."

temporal/api/workflowservice/v1/service.proto
```diff
@@ -960,64 +959,68 @@ service WorkflowService {
...
-    //    * If the activity is currently running, Unpause with 'reset' will reset the number of attempts.
-    //    * if 'no_wait' flag is set, the activity will be scheduled immediately.
-    //    * if 'reset_heartbeats' flag is set, the activity heartbeat timer and heartbeats will be reset.
-    // If the activity is in waiting for retry and past it retry timeout, it will be scheduled immediately.
-    // Once the activity is unpaused, all timeout timers will be regenerated.
-    // (-- api-linter: core::0136::prepositions=disabled
-    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
-    rpc UnpauseActivityById (UnpauseActivityByIdRequest) returns (UnpauseActivityByIdResponse) {
+    // UnpauseActivity unpauses the execution of an activity specified by its ID or type.
+    // If there are multiple pending activities of the provided type - all of them will be unpause.
```
Sushisource, line 976:
```suggestion
    // If there are multiple pending activities of the provided type - all of them will be unpaused.
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -960,64 +959,68 @@ service WorkflowService {
...
+    //
+    // If activity is not paused, this call will have no effect.
+    // If the activity is waiting for retry, it will be scheduled immediately (* see 'jitter' flag).
+    // Once the activity is unpause, all timeout timers will be regenerated.
```
Sushisource, line 980:
```suggestion
    // If activity is not paused, this call will have no effect.
    // If the activity is waiting for retry, it will be scheduled immediately (* see 'jitter' flag).
    // Once the activity is unpaused, all timeout timers will be regenerated.
```

The first two lines seem in conflict. If the activity isn't paused, but is waiting to retry, will or will it not be scheduled immediately?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -946,12 +945,12 @@ service WorkflowService {
         };
     }
 
-    // PauseActivityById pauses the execution of an activity specified by its ID.
-    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    // PauseActivity pauses the execution of an activity specified by its ID or type.
+    // If there are multiple pending activities of the provided type - all of them will be paused
```
Sushisource, line 949: Ah, this line clarifies the behavior of `type` -- I think it's still good to have this on the field directly though. Shows up easier in docstrings.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1747,123 +1747,117 @@ message ExecuteMultiOperationResponse {
...
 }
 
-message UpdateActivityOptionsByIdResponse {
+message UpdateActivityOptionsResponse {
     // Activity options after an update
     temporal.api.activity.v1.ActivityOptions activity_options = 1;
 }
 
-message PauseActivityByIdRequest {
+message PauseActivityRequest {
```
Sushisource, line 1779: Will we be adding a `should_interrupt` or `no_interrupt` flag? Or is that coming later? Not blocking this PR by any means.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: will add better comment.

temporal/api/workflowservice/v1/service.proto
ychebotarev: it will not. I will make a clear distinction.

temporal/api/workflowservice/v1/service.proto
ychebotarev: Yes, I change the comment in all those "requests"

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I personally still think we shouldn't add anything at all. 
That said - if we do then yes, it will be in this request.
That said - we should come with a good default.

temporal/api/workflowservice/v1/service.proto
ychebotarev: One more clarification - "reset" now will also unpause by default, so if user want to "trigger" activity not matter if it was paused or not they can use "reset".
Basically this was the main motivation behind adding "keep_unpause" to reset - so users can "trigger" without bothering if activity is paused or not.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1747,123 +1747,117 @@ message ExecuteMultiOperationResponse {
...
+        string id = 4;
+        // Unpause all running activities with of this type.
+        string type = 5;
     }
+
+    // Providing this flag will also reset the number of attempts.
+    bool reset_attempts = 6;
+
+    // Providing this flag will also reset the heartbeat details.
+    bool reset_heartbeat = 7;
```
alexshtin, line 1822: Why it is not synced with batch?
```
  // If set, the number of attempts will not be reset.
  bool keep_attempts = 3;

  // If set, the Heartbeat Details will be cleared out to make the Activity start over from the beginning.
  bool reset_heartbeat = 4;
```

alexshtin (commented):
Yes, this looks much better now.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: before this PR they have different set of flags

ychebotarev: or you are talking about comments? Yeah I will sync those.

alexshtin (approved):

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ah...it is 'keep_attempts' in batch. Yeah, we switch it to `reset_attempts` at the last review

temporal/api/workflowservice/v1/service.proto
```diff
@@ -920,15 +920,14 @@ service WorkflowService {
     rpc RespondNexusTaskFailed(RespondNexusTaskFailedRequest) returns (RespondNexusTaskFailedResponse) {
     }
 
-    // UpdateActivityOptionsById is called by the client to update the options of an activity
-    // (-- api-linter: core::0136::prepositions=disabled
-    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
-    rpc UpdateActivityOptionsById (UpdateActivityOptionsByIdRequest) returns (UpdateActivityOptionsByIdResponse) {
```
cretz, line 926: While I understand this is not GA, this is a backwards incompatible alteration of a released API, which means it is consumed by proxies. Also, even though it is not GA, it is in the CLI and this incompatible alteration means that the current version of CLI cannot be used for this feature even after the feature is GA. It also means, once tagged, we cannot upgrade the API version in the CLI project (which we may want to do for other reasons) because compile will break until the code is altered to match the new API, and then that may not work with the server.

This is unfortunate but technically doable. As mentioned when these APIs were first getting to `master`, this is the cost of not getting further along on implementation.

I would like to request we wait to land this in `master` until we have most of the implementation done (maybe it already is), we are ready to tag the API, and we have a PR ready for the CLI to support these incompatible alterations. Then we can merge this and tag the API and update CLI (ignoring that it may not even work with the latest server).

cretz:
I think we should investigate why our buf-breaking didn't cause a CI failure here

temporal/api/workflowservice/v1/service.proto
ychebotarev: >this is the cost of not getting further along on implementation
we fully implement those API before adding it to CLI. 

>we wait to land this in master until we have most of the implementation done
All (re)implementation is done - see the link to the Server PR in description.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1764,9 +1764,9 @@ message UpdateActivityOptionsRequest {
 
     // either activity id or activity type must be provided
     oneof activity {
-        // ID of the activity we're pausing.
+        // Only activity with this ID will be updated.
```
Sushisource, line 1767: Should be "Only the activity..."

Applies elsewhere too

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1809,16 +1809,16 @@ message UnpauseActivityRequest {
 
     // either activity id or activity type must be provided
     oneof activity {
-        // ID of the activity we're pausing.
+        // Only activity with this ID will be unpause.
```
Sushisource, line 1812: Should be "unpaused" (and "the activity")

For some reason GH not letting me do suggestions here. Weird.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -960,64 +959,68 @@ service WorkflowService {
...
-    // Once the activity is unpaused, all timeout timers will be regenerated.
-    // (-- api-linter: core::0136::prepositions=disabled
-    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
-    rpc UnpauseActivityById (UnpauseActivityByIdRequest) returns (UnpauseActivityByIdResponse) {
+    // UnpauseActivity unpauses the execution of an activity specified by its ID or type.
+    // If there are multiple pending activities of the provided type - all of them will be unpaused.
+    //
+    // If activity is not paused, this call will have no effect.
+    // If the activity was paused while waiting for retry, it will be scheduled immediately (* see 'jitter' flag).
+    // Once the activity is unpause, all timeout timers will be regenerated.
```
Sushisource, line 980:
```suggestion
    // Once the activity is unpaused, all timeout timers will be regenerated.
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -946,12 +945,12 @@ service WorkflowService {
...
 
-    // PauseActivityById pauses the execution of an activity specified by its ID.
-    // Returns a `NotFound` error if there is no pending activity with the provided ID.
+    // PauseActivity pauses the execution of an activity specified by its ID or type.
+    // If there are multiple pending activities of the provided type - all of them will be paused
     //
     // Pausing an activity means:
     // - If the activity is currently waiting for a retry or is running and subsequently fails,
-    //   it will not be rescheduled until it is unpaused.
+    //   it will not be rescheduled until it is unpause.
```
Sushisource, line 953:
```suggestion
    //   it will not be rescheduled until it is unpaused.
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -960,64 +959,68 @@ service WorkflowService {
...
-    //     aip.dev/not-precedent: "By" is used to indicate request type. --)
-    rpc ResetActivityById (ResetActivityByIdRequest) returns (ResetActivityByIdResponse) {
+    // * activity timeouts will be reset.
+    // * if the activity is waiting for retry, and it is not paused or 'keep_paused' is not provided:
+    //    it will be scheduled immediately (* see 'jitter' flag),
+    //
+    // Flags:
+    //
+    // 'jitter': the activity will be scheduled at a random time within the jitter duration.
+    // If the activity currently paused it will be unpause, unless 'keep_paused' flag is provided.
```
Sushisource, line 1011:
```suggestion
    // If the activity currently paused it will be unpaused, unless 'keep_paused' flag is provided.
```

Sushisource (commented):
A few more grammar fixes.

I agree with Chad about not merging until we're satisfied with the implementation

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: "unpaused" is marked as "typo" for me. but ok

PR #535, opened 2025-01-28
Add batch activity unpause operation

ychebotarev:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/batch/v1/message.proto
+++ b/temporal/api/batch/v1/message.proto
@@ -31,6 +31,7 @@ option java_outer_classname = "MessageProto";
 option ruby_package = "Temporalio::Api::Batch::V1";
 option csharp_namespace = "Temporalio.Api.Batch.V1";
 
+import "google/protobuf/duration.proto";
 import "google/protobuf/field_mask.proto";
 import "google/protobuf/timestamp.proto";
 import "temporal/api/common/v1/message.proto";
@@ -116,4 +117,23 @@ message BatchOperationUpdateWorkflowExecutionOptions {
   // Controls which fields from `workflow_execution_options` will be applied.
   // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
   google.protobuf.FieldMask update_mask = 3;
+}
+
+// BatchOperationPauseActivities sends unpause requests to batch workflows.
+message BatchOperationUnpauseActivities {
+  // The identity of the worker/client.
+  string identity = 1;
+
+  // The activity type to unpause.
+  string activity_type = 2;
+
+  // If set, the number of attempts will not be reset.
+  bool keep_attempts = 3;
+
+  // If set, the Heartbeat Details will be cleared out to make the Activity start over from the beginning.
+  bool reset_heartbeat = 4;
+
+  // If set, the activity will start at a random time within the specified jitter
+  // duration, introducing variability to the start time.
+  google.protobuf.Duration jitter = 5;
 }
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -6453,6 +6453,9 @@
         },
         "updateWorkflowOptionsOperation": {
           "$ref": "#/definitions/v1BatchOperationUpdateWorkflowExecutionOptions"
+        },
+        "unpauseActivitiesOperation": {
+          "$ref": "#/definitions/v1BatchOperationUnpauseActivities"
         }
       }
     },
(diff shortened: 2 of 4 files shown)


temporal/api/batch/v1/message.proto
```diff
@@ -117,3 +118,22 @@ message BatchOperationUpdateWorkflowExecutionOptions {
...
   google.protobuf.FieldMask update_mask = 3;
 }
+
+// BatchOperationPauseActivities sends unpause requests to batch workflows.
+message BatchOperationUnpauseActivities {
+  // The identity of the worker/client.
+  string identity = 1;
+
+  // The activity type to unpause.
+  string activity_type = 2;
```
cretz, line 128: Is this the only activity filter we will support today? Is this required or will its absence mean unpause all activities?

ychebotarev: * Yes, for now type is the only option. I don't foresee anything else (especially given that workflow pause is coming ).
* I will add "if empty, every currently paused activity will be unpaused". Though may be "wildcard" will be a better option to avoid random mistakes?

ychebotarev: if we are doing this - how about I will also replace "workflowID + runID" with execution_info, like in every other workflow-level API?

ychebotarev: This is batch operation, not namespace rule. There is no workflow filter (I think what you talk about is a query with is part of startBatchOperation request).

temporal/api/batch/v1/message.proto
```diff
@@ -117,3 +118,23 @@ message BatchOperationUpdateWorkflowExecutionOptions {
...
+message BatchOperationUnpauseActivities {
+  // The identity of the worker/client.
+  string identity = 1;
+
+  // The activity type to unpause.
+  // If empty, every currently paused activity will be unpause
+  string activity_type = 2;
+
+  // Providing this flag will also reset the number of attempts.
+  bool keep_attempts = 3;
```
alexshtin, line 132:
```suggestion
  bool reset_attempts = 3;
```

temporal/api/batch/v1/message.proto
alexshtin: I am not against `*` too. I agree that leaving it empty to "catch all" is dangerous.

alexshtin (approved):

temporal/api/batch/v1/message.proto
```diff
@@ -117,3 +118,23 @@ message BatchOperationUpdateWorkflowExecutionOptions {
...
   google.protobuf.FieldMask update_mask = 3;
 }
+
+// BatchOperationPauseActivities sends unpause requests to batch workflows.
+message BatchOperationUnpauseActivities {
+  // The identity of the worker/client.
+  string identity = 1;
+
+  // The activity type to unpause.  Can't be empty.
+  // Specify "*" to unpause all activities.
```
cretz, line 128: Hrmm, I am not sure we need a concept of wildcard here. A user can definitely have an activity named `*my-activity*` and even though we only support single-asterisk-means-all, often these stringly typed wildcards expand to prefix/suffix matching. I think empty meaning "all" for paused activities on matching workflows of the batch is acceptable, but if there is concern there, a `oneof` with a bool for `all` can be made.

temporal/api/batch/v1/message.proto
```diff
@@ -117,3 +118,23 @@ message BatchOperationUpdateWorkflowExecutionOptions {
   // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
   google.protobuf.FieldMask update_mask = 3;
 }
+
+// BatchOperationPauseActivities sends unpause requests to batch workflows.
+message BatchOperationUnpauseActivities {
```
cretz, line 123: Can we confirm implementation is mostly done before merge here?

temporal/api/batch/v1/message.proto
```diff
@@ -117,3 +118,23 @@ message BatchOperationUpdateWorkflowExecutionOptions {
   // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
   google.protobuf.FieldMask update_mask = 3;
 }
+
+// BatchOperationPauseActivities sends unpause requests to batch workflows.
```
cretz, line 122:
```suggestion
// BatchOperationUnpauseActivities sends unpause requests to batch workflows.
```

temporal/api/batch/v1/message.proto
cretz: Hrmm, I am not sure we need a concept of wildcard here. A user can definitely have an activity named `*my-activity*` and even though we only support single-asterisk-means-all, often these stringly typed wildcards expand to prefix/suffix matching. I think empty meaning "all" for paused activities on matching workflows of the batch is acceptable, but if there is concern there, a oneof with a bool for all can be made.

temporal/api/batch/v1/message.proto
ychebotarev: Yes. There is a link to Server PR in description.

temporal/api/batch/v1/message.proto
```diff
@@ -117,3 +118,25 @@ message BatchOperationUpdateWorkflowExecutionOptions {
   // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
   google.protobuf.FieldMask update_mask = 3;
 }
+
+// BatchOperationUnpauseActivities sends unpause requests to batch workflows.
+message BatchOperationUnpauseActivities {
+  // The identity of the worker/client.
+  string identity = 1;
+
+  // The activity to unpause. If match_all is set to true, all activities will be unpause.
```
Sushisource, line 127:
```suggestion
  // The activity to unpause. If match_all is set to true, all activities will be unpaused.
```

Sushisource (approved):

PR #536, opened 2025-01-28
Versioning 3 API refinements

ShahabT:

**What changed?**
- `version` string fields in `WorkerDeploymentOptions` and `WorkerDeploymentVersion` changed to `build_id`.
- `WorkerDeploymentVersionInfo` and `WorkerDeploymentInfo` messages completed with all the fields that will be implemented before Replay.
- Added `conflict_token` to relevant APIs.
- Now using special `__unversioned__` value instead of empty string to represent unversioned workers.

Note: the following APIs are still to be added:
- `UpdateWorkerDeploymentVersionMetadata`
- `DeleteWorkerDeploymentVersion`
- `DeleteWorkerDeployment`

**Why?**
Refinements based on latest design discussions within the crew.

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -171,7 +171,7 @@ message WorkerDeploymentVersionInfo {
 // Information about workflow drainage to help the user determine when it is safe
 // to decommission a Version. Not present while the version is current or ramping.
 message VersionDrainageInfo {
-    // Set to DRAINING when accepts_new_executions first becomes false.
+    // Set to DRAINING when the version stops accepting new executions (is no longer current or ramping).
     // Set to DRAINED when no more open pinned workflows exist on this version.
     enums.v1.VersionDrainageStatus status = 1;
     // Last time the drainage status changed.
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -145,7 +145,7 @@ message WorkerDeploymentVersionInfo {
     repeated VersionTaskQueueInfo task_queue_infos = 7;
 
     // Helps user determine when it is safe to decommission the workers of this
-    // Version. Not present when accepts_new_executions is true.
+    // Version. Not present when the workflow is accepting new workflow executions because it's current or ramping.
     // Current limitations:
     // - Not supported for Unversioned mode.
     // - Periodically refreshed, may have delays up to few minutes (consult the
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -221,7 +221,7 @@ message VersionMetadata {
 }
 
 message RoutingInfo {
-    // Always present. Specifies what Deployment Version should should receive new workflow
+    // Always present. Specifies which Deployment Version should should receive new workflow
     // executions and tasks of existing unversioned or AutoUpgrade workflows.
     // Can be one of the following:
     // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -208,7 +208,7 @@ message WorkerDeploymentInfo {
     string last_modifier_identity = 5;
 
     message WorkerDeploymentVersionSummary {
-        string version = 1;
+        string build_id = 1;
         temporal.api.enums.v1.WorkflowVersioningMode workflow_versioning_mode = 2;
         google.protobuf.Timestamp create_time = 3;
         enums.v1.VersionDrainageStatus drainage_status = 4;
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -132,9 +132,9 @@ message WorkerDeploymentVersionInfo {
 
     google.protobuf.Timestamp create_time = 3;
     // All the Task Queues that have ever polled from this Deployment version.
-    repeated WorkerBuildTaskQueueInfo task_queue_infos = 4;
+    repeated DeploymentVersionTaskQueueInfo task_queue_infos = 4;
 
-    message WorkerBuildTaskQueueInfo {
+    message DeploymentVersionTaskQueueInfo {
         string name = 1;
         temporal.api.enums.v1.TaskQueueType type = 2;
     }
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -169,7 +169,7 @@ message WorkerDeploymentVersionInfo {
 }
 
 // Information about workflow drainage to help the user determine when it is safe
-// to decommission a Version. Not present while accepts_new_executions is true.
+// to decommission a Version. Not present while the version is current or ramping.
 message VersionDrainageInfo {
     // Set to DRAINING when accepts_new_executions first becomes false.
     // Set to DRAINED when no more open pinned workflows exist on this version.
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -138,6 +138,39 @@ message WorkerDeploymentVersionInfo {
         string name = 1;
         temporal.api.enums.v1.TaskQueueType type = 2;
     }
+
+    // TODO (Shivam): Accepts_new_executions + Poller_status + Drainage_status + WorkerBuildMetadata
+}
+
+// A Worker Deployment (Deployment, for short) represents all workers serving 
+// a shared set of Task Queues. Typically, a Deployment represents one service or 
+// application.
+// A Deployment contains multiple Deployment Versions, each representing a different 
+// version of workers. (see documentation of WorkerDeploymentVersionInfo)
+// Deployment records are created in Temporal server automatically when their 
+// first poller arrives to the server.
+message WorkerDeploymentInfo {
+    // Identifies a Worker Deployment. Must be unique within the namespace.
+    string name = 1;
+
+    // Deployment Versions that are currently tracked in this Deployment. A DeploymentVersion will be
+    // cleaned up automatically if all the following conditions meet:
+    // - It does not receive new executions (see WorkerDeploymentVersionInfo.accepts_new_executions) 
+    // - It has no active pollers (see WorkerDeploymentVersionInfo.pollers_status) 
+    // - It is drained (see WorkerDeploymentVersionInfo.drainage_status) 
+    repeated WorkerDeploymentVersionSummary version_summaries = 2;
+
+    google.protobuf.Timestamp create_time = 3;
+
+    message WorkerDeploymentVersionSummary {
+        string version = 1;
+        temporal.api.enums.v1.WorkflowVersioningMode workflow_versioning_mode = 2;
+        google.protobuf.Timestamp create_time = 3;
+        bool accepts_new_executions = 4;
+    }
+    
+    // TODO (Shivam): WorkerDeploymentVersionSummary.PollerStatus + TaskQueueInfo + RampingInfo +
+    // AggregatedPollerStatus + DrainageStatus 
 }
 
 
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -162,6 +162,8 @@ message WorkerDeploymentInfo {
 
     google.protobuf.Timestamp create_time = 3;
 
+    RoutingInfo routing_info = 4;
+
     message WorkerDeploymentVersionSummary {
         string version = 1;
         temporal.api.enums.v1.WorkflowVersioningMode workflow_versioning_mode = 2;
@@ -174,3 +176,24 @@ message WorkerDeploymentInfo {
 }
 
 
+message RoutingInfo {
+    // Current Version receives all the tasks except the portion that is routed to the Ramping
+    // Version according to `ramping_version_percentage`.
+    // It is possible that Current Version is not present when user is ramping from unversioned to
+    // versioned workers. In that case, remaining tasks after `ramping_version_percentage` go to
+    // unversioned workers. (Unversioned workers are those with UNVERSIONED (or unspecified)
+    // WorkflowVersioningMode.)
+    string current_version = 1;
+    // Ramping Version is set during gradual rollout of a new Version. It takes a share of tasks
+    // according to `ramping_version_percentage`.
+    string ramping_version = 2;
+    // Percentage of tasks that are routed to the Ramping Version instead of the Current Version.
+    // Special case: it is possible that `ramping_version_percentage` is non-zero while
+    // `ramping_version` is absent. That case represents gradual migration out of the Current
+    // Version to unversioned workers.
+    float ramping_version_percentage = 3;
+    // Last time current version was updated.
+    google.protobuf.Timestamp current_version_update_time = 4;
+    // Last time ramping version was updated.
+    google.protobuf.Timestamp ramping_version_update_time = 5;
+}
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -150,8 +150,8 @@ message WorkerDeploymentVersionInfo {
     // - Not supported for Unversioned mode.
     // - Periodically refreshed, may have delays up to few minutes (consult the
     //   last_checked_time value).
-    // - Refreshed only when accepts_new_executions is false AND the status is not
-    //   "drained" yet.
+    // - Refreshed only when the version is neither current nor ramping AND 
+    // the status is not "drained" yet.
     // - Once the status is changed to "drained", it is not changed until the Version
     //   becomes Current or Ramping again, at which time the drainage info is cleared.
     //   This means if the Version is "drained" but new workflows are sent to it via
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -39,12 +39,13 @@ import "temporal/api/common/v1/message.proto";
 
 // Worker Deployment options set in SDK that need to be sent to server in every poll.
 message WorkerDeploymentOptions {
-    // Required. Worker deployment name.
-    string name = 1;
-    // Required. Worker deployment version.
-    string version = 2;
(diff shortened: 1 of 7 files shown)


temporal/api/deployment/v1/message.proto
```diff
@@ -164,33 +202,48 @@ message WorkerDeploymentInfo {
...
+    // Always present. Specifies what Deployment Version should should receive new workflow
+    // executions and tasks of existing unversioned or AutoUpgrade workflows.
+    // Can be one of the following:
+    // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
+    // `WorkflowVersioningMode==VERSIONING_BEHAVIORS`.)
+    // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
+    // with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
+    // Note: Current Version is overridden by the Ramping Version for a portion of traffic when a ramp
+    // is set (see `ramping_version`.)
     string current_version = 1;
```
Shivs11, line 233: this change would now have to mean that by default, the `current_version` of a deployment would be `__unversioned__` unless user changes it, right?

temporal/api/deployment/v1/message.proto
```diff
@@ -121,25 +125,59 @@ message DeploymentListInfo {
...
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not ramping.
+    google.protobuf.Timestamp ramping_since_time = 5;
+    // Zero if the version is not ramping (i.e. `ramping_since_time` is nil).
+    float ramp_percentage = 6;
 
-    message DeploymentVersionTaskQueueInfo {
+    // All the Task Queues that have ever polled from this Deployment version.
+    repeated VersionTaskQueueInfo task_queue_infos = 7;
```
Shivs11, line 145: are we including this feature, of having unique `VersionTaskQueueInfo`s, for pre-release? IMO we should only add this when we complete server work related to this, right?

cretz (approved):
Approving under the assumption that this is not making backwards incompatible alterations with `main`.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1990,43 +1990,58 @@ message SetCurrentDeploymentResponse {
...
+    // Ramp percentage to set. Valid range: [0,100].
     float percentage = 6;
 
+    // Optional. This can be the value of conflict_token from a Describe, or another Worker
+    // Deployment API. Passing conflict token which will cause this request to fail if the
+    // Deployment's configuration has been modified between the previous API call and
+    // this one.
+    bytes conflict_token = 4;
     // Optional. The identity of the client who initiated this request.
     string identity = 7;
```
Shivs11, line 2035: nit: don't see the fifth attribute for this message

temporal/api/deployment/v1/message.proto
```diff
@@ -164,33 +202,48 @@ message WorkerDeploymentInfo {
 
     RoutingInfo routing_info = 4;
 
+    // Identity of the last client who modified the configuration of this Deployment. Set to the
+    // `identity` value sent by APIs such as `SetWorkerDeploymentCurrentVersion` and
+    // `SetWorkerDeploymentRampingVersion`.
+    string last_modifier_identity = 5;
+
     message WorkerDeploymentVersionSummary {
         string version = 1;
```
Shivs11, line 211: lets make this to `string buildID = 1`?;

temporal/api/deployment/v1/message.proto
ShahabT: yes

temporal/api/deployment/v1/message.proto
```diff
@@ -164,33 +202,48 @@ message WorkerDeploymentInfo {
...
         temporal.api.enums.v1.WorkflowVersioningMode workflow_versioning_mode = 2;
         google.protobuf.Timestamp create_time = 3;
-        bool accepts_new_executions = 4;
+        enums.v1.VersionDrainageStatus drainage_status = 4;
```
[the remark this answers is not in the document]
carlydf: I don't think we need `NOT_DRAINING`. If the version is not draining, then `DrainageInfo == nil`. If either `IsCurrent || RampPercentage > 0` (or `CurrentSinceTime != nil || RampingSinceTime != nil` if we go with that) then the version is accepting new executions, so I agree with getting rid of `bool accepts_new_executions`, and we should remove references to it

temporal/api/deployment/v1/message.proto
```diff
@@ -121,25 +125,59 @@ message DeploymentListInfo {
...
+    google.protobuf.Timestamp ramping_since_time = 5;
+    // Zero if the version is not ramping (i.e. `ramping_since_time` is nil).
+    float ramp_percentage = 6;
 
-    message DeploymentVersionTaskQueueInfo {
+    // All the Task Queues that have ever polled from this Deployment version.
+    repeated VersionTaskQueueInfo task_queue_infos = 7;
+
+    // Helps user determine when it is safe to decommission the workers of this
+    // Version. Not present when accepts_new_executions is true.
```
carlydf, line 148:
```suggestion
    // Version. Not present when the workflow is accepting new workflow executions because it's current or ramping.
```

temporal/api/deployment/v1/message.proto
```diff
@@ -121,25 +125,59 @@ message DeploymentListInfo {
...
+    // Current limitations:
+    // - Not supported for Unversioned mode.
+    // - Periodically refreshed, may have delays up to few minutes (consult the
+    //   last_checked_time value).
+    // - Refreshed only when accepts_new_executions is false AND the status is not
+    //   "drained" yet.
```
carlydf, line 154:
```suggestion
    // - Refreshed only when the version is neither current nor ramping AND 
    // the status is not "drained" yet.
```

temporal/api/deployment/v1/message.proto
```diff
@@ -121,25 +125,59 @@ message DeploymentListInfo {
...
+
+    message VersionTaskQueueInfo {
         string name = 1;
         temporal.api.enums.v1.TaskQueueType type = 2;
     }
+}
 
-    // TODO (Shivam): Accepts_new_executions + Poller_status + Drainage_status + WorkerBuildMetadata
+// Information about workflow drainage to help the user determine when it is safe
+// to decommission a Version. Not present while accepts_new_executions is true.
```
carlydf, line 172:
```suggestion
// to decommission a Version. Not present while the version is current or ramping.
```

temporal/api/deployment/v1/message.proto
```diff
@@ -121,25 +125,59 @@ message DeploymentListInfo {
...
+message VersionDrainageInfo {
+    // Set to DRAINING when accepts_new_executions first becomes false.
```
carlydf, line 174:
```suggestion
    // Set to DRAINING when the version stops accepting new executions (is no longer current or ramping).
```

temporal/api/deployment/v1/message.proto
```diff
@@ -121,25 +125,59 @@ message DeploymentListInfo {
...
 
-    google.protobuf.Timestamp create_time = 3;
-    // All the Task Queues that have ever polled from this Deployment version.
-    repeated DeploymentVersionTaskQueueInfo task_queue_infos = 4;
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not ramping.
+    google.protobuf.Timestamp ramping_since_time = 5;
+    // Zero if the version is not ramping (i.e. `ramping_since_time` is nil).
+    float ramp_percentage = 6;
```
carlydf, line 142: After spending some time in the server working with `current_since_time + ramping_since_time + ramp_percentage`, I think it's confusing, and I feel that `routing_update_time + is_current + ramp_percentage` is much easier to reason about, and ultimately contains the same information with less bytes, since `is_current` bool takes up less space than a 64-bit timestamp.

I don't feel strongly enough to block this PR on it -- if others think `current_since_time + ramping_since_time` is better for users then we can do that, but I do think my suggestion will be easier for users to work with.

temporal/api/deployment/v1/message.proto
```diff
@@ -164,33 +202,48 @@ message WorkerDeploymentInfo {
...
+}
 
 message RoutingInfo {
-    // Current Version receives all the tasks except the portion that is routed to the Ramping
-    // Version according to `ramping_version_percentage`.
-    // It is possible that Current Version is not present when user is ramping from unversioned to
-    // versioned workers. In that case, remaining tasks after `ramping_version_percentage` go to
-    // unversioned workers. (Unversioned workers are those with UNVERSIONED (or unspecified)
-    // WorkflowVersioningMode.)
+    // Always present. Specifies what Deployment Version should should receive new workflow
```
carlydf, line 224:
```suggestion
    // Always present. Specifies which Deployment Version should should receive new workflow
```

temporal/api/deployment/v1/message.proto
```diff
@@ -164,33 +202,48 @@ message WorkerDeploymentInfo {
...
+    // with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
+    // Note that it is possible to ramp from one Version to another Version, or from unversioned
+    // workers to a particular Version, or from a particular Version to unversioned workers.
     string ramping_version = 2;
     // Percentage of tasks that are routed to the Ramping Version instead of the Current Version.
-    // Special case: it is possible that `ramping_version_percentage` is non-zero while
-    // `ramping_version` is absent. That case represents gradual migration out of the Current
-    // Version to unversioned workers.
+    // Valid range: [0, 100]. A 100% value means the Ramping Version is receiving full traffic but
+    // not yet "promoted" to be the Current Version, likely due to pending validations.
```
carlydf, line 246: nice, I like your concise explanation of what 100% ramp means!

temporal/api/taskqueue/v1/message.proto
```diff
@@ -65,8 +65,12 @@ message TaskQueueVersioningInfo {
...
     // according to `ramping_version_percentage`.
+    // It is possible that Ramping Version belong to a Deployment other than that of the Current
+    // Version. This happens when a Task Queue is being moved from one Deployment to another.
     temporal.api.deployment.v1.WorkerDeploymentVersion ramping_version = 2;
     // Percentage of tasks that are routed to the Ramping Version instead of the Current Version.
+    // Valid range: [0, 100]. A 100% value means the Ramping Version is receiving full traffic but
+    // not yet "promoted" to be the Current Version, likely due to pending validations.
     // Special case: it is possible that `ramping_version_percentage` is non-zero while
     // `ramping_version` is absent. That case represents gradual migration out of the Current
     // Version to unversioned workers.
```
carlydf, line 76: Is this still true now that we have the special `__unversioned__` string?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1990,43 +1990,58 @@ message SetCurrentDeploymentResponse {
...
     string namespace = 1;
     string deployment_name = 2;
-    // Must be a Versioned Build. Pass empty string to unset.
-    google.protobuf.StringValue version = 3;
+    // Required. Can be one of the following:
+    // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
+    // `WorkflowVersioningMode==VERSIONING_BEHAVIORS`.)
+    // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
+    // with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
+    string version = 3;
```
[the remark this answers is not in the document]
carlydf: I think either this should be:
```
    string namespace = 1;
    // Required. `version.build_id` can be one of the following:
    // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
    // `WorkflowVersioningMode==VERSIONING_BEHAVIORS`.)
    // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
    // with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
    temporal.api.deployment.v1.WorkerDeploymentVersion version = 2;
    bytes conflict_token = 3;
    ...
```

or: 

```
    string namespace = 1;
    string deployment_name = 2;
    // Required. Can be one of the following:
    // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
    // `WorkflowVersioningMode==VERSIONING_BEHAVIORS`.)
    // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
    // with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
    string build_id = 3;
    bytes conflict_token = 4;
    ...
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1990,43 +1990,58 @@ message SetCurrentDeploymentResponse {
...
-    // (Useful for migrating to unversioned workers from Deployment's Current Version.)
-    google.protobuf.StringValue version = 3;
-    // Valid range: [0,100]. Zero value will unset the Ramping Version.
+    // Can be one of the following:
+    // - Absent/empty value to unset the Ramping Version. Must be paired with `percentage=0`.
+    // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
+    // `WorkflowVersioningMode==VERSIONING_BEHAVIORS`) to set it as the Ramping Version.
+    // - Or, the "__unversioned__" special value, to ramp unversioned workers (those with
+    // `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
+    string version = 3;
```
carlydf, line 2025:
```suggestion
    string build_id = 3;
```
See my comment above

temporal/api/deployment/v1/message.proto
carlydf, line 211:
```suggestion
        string build_id = 1;
```

temporal/api/deployment/v1/message.proto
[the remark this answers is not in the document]
carlydf: I went through and removed references to `accepts_new_executions`, so I'm resolving this comment

temporal/api/taskqueue/v1/message.proto
carlydf: I think the comments on all of these fields need to be re-worked to reflect the new `__unversioned__` special build id

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
carlydf: I'll just go with `build_id` for now since I think it makes sense and it's the smaller change

carlydf, line 1998:
```suggestion
    string build_id = 3;
```

temporal/api/deployment/v1/message.proto
carlydf: nevermind, just saw your comment about supporting 0% ramp and `ramping_since` time in the UI! scratch this

temporal/api/deployment/v1/message.proto
ShahabT: @Shivs11, not sure if understand the question, don't we have TQ list per version already?

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
ShahabT: `build_id` sounds good. I hesitate to use `WorkerDeploymentVersion` message with its build ID being "__unversioned__" because that is no longer representing a single Worker Deployment Version.

temporal/api/taskqueue/v1/message.proto
carlydf: pushed a fix that addresses my comment here

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
carlydf: My thinking is that these APIs are split into:
- APIs that act on the Worker Deployment
  - DescribeWorkerDeployment
  - ListWorkerDeployments
  - SetWorkerDeploymentCurrentVersion
  - SetWorkerDeploymentRampingVersion
- APIs that act on the Worker Deployment Version
  - DescribeWorkerDeploymentVersion

and right now the APIs that act on the Worker Deployment take in a `string deployment_name` and a `string build_id` if necessary. The APIs that act on the WorkerDeploymentVersion take in a `WorkerDeploymentVersion version`.

temporal/api/workflowservice/v1/request_response.proto
carlydf: fixed

carlydf (approved):

temporal/api/taskqueue/v1/message.proto
ShahabT: Hmm... but these are WorkerDeploymentVersion objects, not strings. I tried to avoid bringing `__unversioned__` into `WorkerDeploymentVersion.build_id` to keep WorkerDeploymentVersion strictly about a single Version. 

Do you think it's going to be confusing for Deployment routing info to have the special string, but in here nil version values? Note that in this case we cannot use a single string because versions may be from different deployments.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1990,43 +1990,58 @@ message SetCurrentDeploymentResponse {
...
 }
 
 message SetWorkerDeploymentRampingVersionResponse {
-    // The ramping version before executing this operation.
-    string previous_version = 1;
+    bytes conflict_token = 1;
+    // The Build ID of the version that was ramping before executing this operation. Can also be
+    // empty or the `__unversioned__` special value.
+    string previous_version = 2;
     // The ramping version percentage before executing this operation.
```
antlai-temporal, line 2043: Do you want to call this `previous_build_id` ?

antlai-temporal (approved):

temporal/api/workflowservice/v1/request_response.proto
carlydf: yes! fixed for SetCurrent resp and SetRamping resp

PR #540, opened 2025-01-30
Versioning 3.1 APIs

carlydf:

**What changed?**
- Renamed `DeploymentSeries` -> `WorkerDeployment` and `Deployment` -> `WorkerDeploymentVersion`.
- Added the following experimental APIs:
  - ListWorkerDeployments
  - DescribeWorkerDeployment
  - DescribeWorkerDeploymentVersion
  - SetWorkerDeploymentCurrentVersion
  - SetWorkerDeploymentRampingVersion
  - UpdateWorkerVersionMetadata
  - DeleteWorkerDeployment
  - DeleteWorkerDeploymentVersion
- Deprecated the following experimental APIs:
  - DescribeDeployment
  - ListDeployments
  - GetDeploymentReachability
  - GetCurrentDeployment
  - SetCurrentDeployment

**Why?**
These are changes and improvements we made based on user feedback from the previous iteration.

**Breaking changes**
No breaking changes, the old APIs still work until cleaned up at a later point with all other old Versioning APIs.

**Server PR**
https://github.com/temporalio/temporal/pull/7233


temporal/api/deployment/v1/message.proto
```diff
@@ -81,9 +100,153 @@ message UpdateDeploymentMetadata {
...
+// A Worker Deployment Version (Version, for short) represents all workers of the same 
+// code and config within a Deployment. Workers of the same Version are expected to 
+// behave exactly the same so when executions move between them there are no 
+// non-determinism issues.
+// Worker Deployment Versions are created in Temporal server automatically when 
+// their first poller arrives to the server.
+// Experimental. Worker Deployments are experimental and might significantly change in the future.
+message WorkerDeploymentVersionInfo {
+    // The fully-qualified string representation of the version, in the form "<deployment_name>/<build_id>".
+    string version = 1;
```
antlai-temporal, line 121: Using `/` in comments, do we change this before merge?

temporal/api/deployment/v1/message.proto
```diff
@@ -81,9 +100,153 @@ message UpdateDeploymentMetadata {
...
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not current.
+    google.protobuf.Timestamp current_since_time = 5;
+
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not ramping. Updated when the version first starts ramping, not on each ramp change.
+    google.protobuf.Timestamp ramping_since_time = 6;
+
```
antlai-temporal, line 137: @carlydf did you want to add `percentange_since_time` or not anymore?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -434,7 +444,10 @@ message PollActivityTaskQueueRequest {
     temporal.api.taskqueue.v1.TaskQueueMetadata task_queue_metadata = 4;
     // Information about this worker's build identifier and if it is choosing to use the versioning
     // feature. See the `WorkerVersionCapabilities` docstring for more.
+    // Deprecated. Replaced by deployment_options.
     temporal.api.common.v1.WorkerVersionCapabilities worker_version_capabilities = 5;
+    // Worker deployment options that user has set in the worker.
```
antlai-temporal, line 449: can we add [deprecated = true]

antlai-temporal (approved):

temporal/api/deployment/v1/message.proto
carlydf: So in this case the timestamps are a bit different from RoutingInfo, because these timestamps are in the context of a single Version, so when this version stops being current, the `current_since_time` will be set to `nil` and the `routing_update_time` will be set to `now`. Then we have `current_since_time == nil` equivalent to `is_current == false` (`current_since_time` captures the info that could have been held in a separate `is_current` field).

Because of that, `routing_changed_time` cannot simply be reconstructed by taking the max of the timestamps.

The main benefit of all this is to not have to include the booleans `is_ramping` and `is_current`, so maybe the confusion is not worth it, and I should add those two booleans so that the timestamps can all be `x_changed_time` instead of `x_since_time`... open to that if folks feel like having different semantics for the "since" timestamps vs "changed" timestamps is confusing.

carlydf: `RoutingConfig` (on the Deployment as a whole) has the ramping and current versions for the Deployment, + these timestamps:
```
    // Last time current version was changed.
    google.protobuf.Timestamp current_version_changed_time = 4;
    // Last time ramping version was changed. Not updated if only the ramp percentage changes.
    google.protobuf.Timestamp ramping_version_changed_time = 5;
    // Last time ramping version percentage was changed.
    // If ramping version is changed, this is also updated, even if the percentage stays the same.
    google.protobuf.Timestamp ramping_version_percentage_changed_time = 6;
```

temporal/api/deployment/v1/message.proto
```diff
@@ -81,9 +100,153 @@ message UpdateDeploymentMetadata {
...
+message RoutingConfig {
+    // Always present. Specifies which Deployment Version should should receive new workflow
+    // executions and tasks of existing unversioned or AutoUpgrade workflows.
+    // Can be one of the following:
+    // - A Deployment Version identifier in the form "<deployment_name>.<build_id>".
+    // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
+    //   with `UNVERSIONED` (or unspecified) `WorkerVersioningMode`.)
+    // Note: Current Version is overridden by the Ramping Version for a portion of traffic when a ramp
+    // is set (see `ramping_version`.)
+    string current_version = 1;
```
Sushisource, line 231: Stylistically I'd make this (and below) a oneof of `version | unversioned` where unversioned is a bool / empty, which saves clients needing to understand the special string.

But, that's not a blocker or a big deal.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1921,28 +1994,192 @@ message SetCurrentDeploymentRequest {
...
+    string version = 3;
+    // Optional. This can be the value of conflict_token from a Describe, or another Worker
+    // Deployment API. Passing a non-nil conflict token will cause this request to fail if the
+    // Deployment's configuration has been modified between the API call that generated the
+    // token and this one.
+    bytes conflict_token = 4;
+    // Optional. The identity of the client who initiated this request.
+    string identity = 5;
+    // Optional. By default this request would be rejected if not all the expected Task Queues are
+    // being polled by the new Version, to protects against accidental removal of Task Queues, or
```
Sushisource, line 2021:
```suggestion
    // Optional. By default this request would be rejected if not all the expected Task Queues are
    // being polled by the new Version, to protect against accidental removal of Task Queues, or
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1921,28 +1994,192 @@ message SetCurrentDeploymentRequest {
...
+
+    // Optional. This can be the value of conflict_token from a Describe, or another Worker
+    // Deployment API. Passing a non-nil conflict token will cause this request to fail if the
+    // Deployment's configuration has been modified between the API call that generated the
+    // token and this one.
+    bytes conflict_token = 5;
+    // Optional. The identity of the client who initiated this request.
+    string identity = 6;
+    // Optional. By default this request would be rejected if not all the expected Task Queues are
+    // being polled by the new Version, to protects against accidental removal of Task Queues, or
```
Sushisource, line 2067:
```suggestion
    // Optional. By default this request would be rejected if not all the expected Task Queues are
    // being polled by the new Version, to protect against accidental removal of Task Queues, or
```

Sushisource (approved):

PR #541, opened 2025-01-31
Update SetWorkerDeploymentCurrentVersionRequest to have ignore_missing_task_queues

Shivs11:

**What changed?**
- Adding new flag inside SetWorkerDeploymentCurrentVersionRequest for poller verification 

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -2016,6 +2016,19 @@ message SetWorkerDeploymentCurrentVersionRequest {
     bytes conflict_token = 4;
     // Optional. The identity of the client who initiated this request.
     string identity = 5;
+    // Optional. By default, false. If true, this operation will succeed even if the new version does not have 
+    // all the task-queues which were present in the previously set current version. If set to false,
+    // any of the following conditions should hold true before the new version is set as current:
+    //
+    // - The new version must have all the task-queues which were present in the previously set current version.
+    //
+    // - If there are any missing task queues from this new version, all of them should have a current deployment version set. In other 
+    // words, these task queues should not be unversioned. (TaskQueueVersioningInfo.current_version should not be empty && 
+    // TaskQueueVersioningInfo.current_version.build_id != "__unversioned__")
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2016,6 +2016,19 @@ message SetWorkerDeploymentCurrentVersionRequest {
...
+    // any of the following conditions should hold true before the new version is set as current:
+    //
+    // - The new version must have all the task-queues which were present in the previously set current version.
+    //
+    // - If there are any missing task queues from this new version, all of them should have a current deployment version set. In other 
+    // words, these task queues should not be unversioned. (TaskQueueVersioningInfo.current_version should not be empty && 
+    // TaskQueueVersioningInfo.current_version.build_id != "__unversioned__")
+    //
+    // - If there are any missing task queues from this new version, all of them should have an add_rate of 0 since
+    // this would signify that there are no backlogged tasks being added in these queues. (see TaskQueueStats.tasks_add_rate)
```
carlydf, line 2030:
```suggestion
    // Optional. By default this protects against accidentally removing a task queue from the current deployment version.
    // If false, one of the following conditions must hold true for the new version to be set as current:
    //   - The new current version must have all the task-queues that are present in the existing current version.
    // or
    //   - If the new current version is missing task queues that are present in the existing current version, 
    //      each missing task queue should have either:
    //         - A current deployment version set, signifying that the task queue has been added to another deployment.
    //       or  
    //         - An add_rate of 0, signifying that there are no backlogged tasks being added in these queues. (see TaskQueueStats.tasks_add_rate)
    // If true, this operation will succeed even if the new current version does not have all the task-queues that were
    // present in the existing current version.
```
What do you think of this structure?

carlydf (approved):

temporal/api/workflowservice/v1/request_response.proto
Shivs11: I quite like this - thanks for proposing!

PR #545, opened 2025-02-02
Clarify semantics of TaskQueueVersioningInfo current_version and ramping_version

carlydf:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**


temporal/api/taskqueue/v1/message.proto
```diff
@@ -58,22 +58,26 @@ message TaskQueueMetadata {
...
     // executions and tasks of existing unversioned or AutoUpgrade workflows.
-    // Can be one of the following:
+    // `current_version.build_id` can be one of the following:
     // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
     //   `WorkflowVersioningMode==VERSIONING_BEHAVIORS`.)
     // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
-    //   with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
+    //   with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`).
+    // If `current_version.build_id == "__unversioned__", then `current_version.deployment_name` will
+    // be empty, because unversioned task queues send tasks to all unversioned workers regardless of deployment.
```
Shivs11, line 67: just want to confirm here: we want to add a new field inside `WorkerDeploymentVersion` called `id` (or something) which will either be `deploymentName/build_id` or BYOK or `__unversioned__` right?

For now, the comment seems to be a temporary work around in the sense that `build_id` can be __unversioned__ but in the future should not be?

temporal/api/taskqueue/v1/message.proto
```diff
@@ -58,22 +58,26 @@ message TaskQueueMetadata {
...
+    // Version. Must always be different from Current Version.
+    // `ramping_version.build_id` can be one of the following:
     // - The Build ID of a Version of this Deployment that supports Versioning Behaviors (i.e. has
     //   `WorkflowVersioningMode==VERSIONING_BEHAVIORS`.)
     // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
     //   with `UNVERSIONED` (or unspecified) `WorkflowVersioningMode`.)
-    // Note that it is possible to ramp from one Version to another Version, or from unversioned
+    // If `current_version.build_id == "__unversioned__", then `current_version.deployment_name` will
+    // be empty, because unversioned task queues send tasks to all unversioned workers regardless of deployment.
+    // Note: It is possible to ramp from one Version to another Version, or from unversioned
```
Shivs11, line 80: nit: Maybe this sounds better?

It is possible to ramp between different versions, including moving to and from unversioned versions.

Shivs11 (approved):

ShahabT:
I'm changing these fields to string in https://github.com/temporalio/api/pull/547.

cretz (approved):

PR #546, opened 2025-02-02
Describe API field Change

Shivs11:

**What changed?**
- I noticed that all other version related API's, apart from DescribeVersion, were accepting in `deploymentName`, `version` as separate parameters. I think DescribeVersion should do the same.

**Why?**
- Might be easier for folks to just copy-paste deployment/buildID strings from UI rather than make version objects (which is how it would be right now)
- Consistency with other API's
- I understand this should not be called on "unversioned" and we can def have safe-guards for that

**Breaking changes**

**Server PR**

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -1944,7 +1944,7 @@
         ]
       }
     },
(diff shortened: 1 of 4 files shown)


Shivs11:
closing this because I think I was wrong - I see the pattern of now Version API's having the version object since we don't want users to have "__unversioned__" as an input

PR #547, opened 2025-02-03
Use string values for deployment version

ShahabT:

**What changed?**
- Using string type in all the places for `version` fields. Removed `WorkerDeploymentVersion` from the public API. This is for two reasons:
   - String type gives ability to specify special values of `__unversioned__` (and maybe later `<deployment_name>/__unversioned__`.
   - We might add ability for users to pass custom version IDs in with case the struct would be hard for users to use (they'd need to write the logic to convert the struct to an ID based on whether it has a custom ID or not).
- Using `WorkerDeploymentOptions` in the task response APIs instead of `WorkerDeploymentVersion` so that server has the full info regardless of the worker participating in versioning or not.
- Added `worker_deployment` field to execution info (top level) so both versioned and unversioned workflows can have this info (and its corresponding SA).
- Renamed `WorkflowVersioningMode` to `WorkerVersioningMode` and its `VERSIONING_BEHAVIORS` enum to `VERSIONED`, for the following reasons:
  - This enum is more about the worker than the workflows, whether or not workflows are versioned or not depends on whether or not the worker enables versioning.
  - We tried to not use the terms "versioned" and "unversioned" for workers for a while but the attempt did work and we ended up even creating special `__unversioned__` value for them. Hence, it'd be very confusing if they don't see some config such as `WorkerVersioningMode` that directly specifies the worker as versioned or unversioned. The previous name, while doing the same thing, twisted the naming and I believe it'll confuse the users.
  - Deployment Versions are not created for the unversioned mode, at least not so far. This naming and updated comments reflects that better.
  - WorkflowVersioningMode is easy to be confused with Workflow VersioningBehavior.

**Why?**

**Breaking changes**

**Server PR**


temporal/api/enums/v1/deployment.proto
```diff
@@ -65,25 +65,30 @@ enum VersionDrainageStatus {
...
+enum WorkerVersioningMode {
+    WORKER_VERSIONING_MODE_UNSPECIFIED = 0;
+    // Workers with this mode are not distinguished from each other for task routing, even if they
+    // have different Build IDs.
+    // Workflows processed by this worker will be unversioned and user needs to use Patching to keep
+    // the new code compatible with prior versions.
+    // This mode is recommended to be used along with Rolling Upgrade deployment strategies.
+    // Workers with this mode may be represented by the special string `__unversioned__` in the APIs.
+    WORKER_VERSIONING_MODE_UNVERSIONED = 1;
+    // Workers with this mode are part of a Worker Deployment Version which is identified as
```
Shivs11, line 82: nit: `Workers with this mode are part of a Worker Deployment Version which *are* identified as`

temporal/api/deployment/v1/message.proto
```diff
@@ -216,9 +206,8 @@ message WorkerDeploymentInfo {
     message WorkerDeploymentVersionSummary {
         // The fully-qualified string representation of the version, in the form "<deployment_name>/<build_id>".
         string version = 1;
-        temporal.api.enums.v1.WorkflowVersioningMode workflow_versioning_mode = 2;
-        google.protobuf.Timestamp create_time = 3;
-        enums.v1.VersionDrainageStatus drainage_status = 4;
+        google.protobuf.Timestamp create_time = 2;
```
Shivs11, line 209: why remove the worker-versioning-mode? Don't we wanna show this to the users?

temporal/api/enums/v1/deployment.proto
```diff
@@ -65,25 +65,30 @@ enum VersionDrainageStatus {
...
+//   tasks to it.
+// - Whether or not the workflows processed by this worker are versioned using the worker's version.
+enum WorkerVersioningMode {
+    WORKER_VERSIONING_MODE_UNSPECIFIED = 0;
+    // Workers with this mode are not distinguished from each other for task routing, even if they
+    // have different Build IDs.
+    // Workflows processed by this worker will be unversioned and user needs to use Patching to keep
+    // the new code compatible with prior versions.
+    // This mode is recommended to be used along with Rolling Upgrade deployment strategies.
+    // Workers with this mode may be represented by the special string `__unversioned__` in the APIs.
```
Shivs11, line 80: "may be represented" sounds a bit vague - should probably be changed to: 

`Workers with this mode are represented by the special string `__unversioned__` in the APIs.`

Shivs11 (approved):

cretz (approved):
LGTM since going to feature branch and doesn't appear to be affecting any `master` APIs

ShahabT (commented):
Change requests that came out of in-person meeting with @antlai-temporal @carlydf @Shivs11 .

temporal/api/deployment/v1/message.proto
```diff
@@ -41,12 +41,14 @@ import "temporal/api/common/v1/message.proto";
...
-    // Required. Build ID of the worker. `deployment_name` and `build_id` together identify this
-    // Worker Deployment Version.
+    // The Build ID of the worker. Required when `versioning_mode==VERSIONED`, in which case, the
+    // worker will be part of a Deployment Version identified by "<deployment_name>/<build_id>".
     string build_id = 2;
-    // Required. Workflow versioning mode for this Deployment Version. Must be the same for all
-    // pollers in a given Deployment Version, across all Task Queues.
-    temporal.api.enums.v1.WorkflowVersioningMode workflow_versioning_mode = 3;
+    // Required. Versioning Mode for this worker. Must be the same for all pollers workers of the
+    // same `deployment_name` and `build_id`, across all Task Queues.
```
carlydf, line 48:
```suggestion
    // Required. Versioning Mode for this worker. Must be the same for all workers with the
    // same `deployment_name` and `build_id` combination, across all Task Queues.
```

carlydf: maybe "tuple" or "pair" instead of "combination"? but I feel like without that word, when I read this for the first time I thought maybe all workers in the same deployment must have the same versioning mode, which is not what we mean.

temporal/api/deployment/v1/message.proto
carlydf: if the mode is unversioned, the version string will be `__unversioned__`, so I think the mode is not needed.

temporal/api/enums/v1/deployment.proto
carlydf: I agree this is vague, but I think maybe it should be, because we might not be able to add "unversioned" representation of a worker to every single unversioned-worker-related thing before thursday. So it depends on the API.

temporal/api/enums/v1/deployment.proto
```diff
@@ -65,25 +65,30 @@ enum VersionDrainageStatus {
...
+    WORKER_VERSIONING_MODE_UNVERSIONED = 1;
+    // Workers with this mode are part of a Worker Deployment Version which is identified as
+    // "<deployment_name>/<build_id>". Such workers are called "versioned" as opposed to
+    // "unversioned".
+    // Each Deployment Version is distinguished from other Versions and users can configure Temporal
```
carlydf, line 85:
```suggestion
    // Each Deployment Version is unique within a Namespace and users can configure the Temporal
```

temporal/api/enums/v1/deployment.proto
```diff
@@ -65,25 +65,30 @@ enum VersionDrainageStatus {
...
+    // Server to send tasks to a particular Version (see `WorkerDeploymentInfo.routing_config`),
```
carlydf, line 86:
```suggestion
    // Server to send tasks to a particular Version (see `WorkerDeploymentInfo.routing_config`).
```

temporal/api/enums/v1/deployment.proto
```diff
@@ -65,25 +65,30 @@ enum VersionDrainageStatus {
...
+    // hence this mode is the best option for Blue/Green and Rainbow strategies (but typically not
```
carlydf, line 87:
```suggestion
    // This mode is the best option for Blue/Green and Rainbow strategies (but typically not
```

temporal/api/enums/v1/deployment.proto
```diff
@@ -65,25 +65,30 @@ enum VersionDrainageStatus {
...
+    // suitable for Rolling upgrades.)
+    // Workflow Versioning Behaviors are enabled in this mode: each workflow type must choose
+    // between the Pinned and AutoUpgrade behaviors. Depending on the chosen behavior user may or
```
carlydf, line 90:
```suggestion
    // between the Pinned and AutoUpgrade behaviors. Depending on the chosen behavior, the user may or
```

temporal/api/enums/v1/workflow.proto
```diff
@@ -145,14 +145,13 @@ enum TimeoutType {
 // existing workflows (see VersioningOverride).
 enum VersioningBehavior {
     // Workflow execution does not have a Versioning Behavior and is called Unversioned. This is the
-    // legacy behavior. An Unversioned workflow's task may go to any worker who is polling for the
-    // Task Queue and has WorkflowVersioningMode == UNVERSIONED (or has not specified a
-    // WorkflowVersioningMode).
+    // legacy behavior. An Unversioned workflow's task may go to any Unversioned worker (see
```
carlydf, line 148:
```suggestion
    // legacy behavior. An Unversioned workflow's task can go to any Unversioned worker (see
```

temporal/api/enums/v1/workflow.proto
```diff
@@ -145,14 +145,13 @@ enum TimeoutType {
...
+    // `WorkerVersioningMode`.)
     // User needs to use Patching to keep the new code compatible with prior versions when dealing
     // with Unversioned workflows.
     VERSIONING_BEHAVIOR_UNSPECIFIED = 0;
     // Workflow will start on the Current Deployment Version of its Task Queue, and then
-    // will be pinned to that same Deployment Version until completion (the version that
+    // will be pinned to that same Deployment Version until completion (the Version that
     // a Workflow is pinned to is specified in `versioning_info.deployment_version`).
```
carlydf, line 155:
```suggestion
    // this Workflow is pinned to is specified in `versioning_info.deployment_version`).
```
We don't want people to think that `versioning_info.deployment_version` represents "the Version that a Workflow is pinned to" for auto_upgrade workflows.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2017,16 +2011,18 @@ message SetWorkerDeploymentCurrentVersionRequest {
...
+    // Optional. By default this request would be rejected if not all the expected Task Queues are
+    // being polled by the new Version, to protects against accidental removal of Task Queues, or
+    // worker health issues. Pass `true` here to bypass this protection.
+    // The set of expected Task Queues equals to all the Task Queues ever polled from the existing
+    // Current Version of the Deployment, with the following exclusions:
+    //   - Task Queues that are not used anymore (inferred by having empty backlog and a task
+    //     add_rate of 0.)
+    //   - Task Queues that are moved to another Worker Deployment (inferred by the Task Queue
+    //     having a different Current Version than the Current Version of this deployment.)
+    // WARNING: Do not set this flag unless you are sure that the missing task queue poller are not
```
carlydf, line 2023:
```suggestion
    // WARNING: Do not set this flag unless you are sure that the missing task queue pollers are not
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2017,16 +2011,18 @@ message SetWorkerDeploymentCurrentVersionRequest {
...
+    // needed. If the request is unexpectedly rejected due to missing pollers, then that means the
+    // pollers have not reached to server yet.
```
carlydf, line 2025:
```suggestion
    // pollers have not reached the server yet. Only set this if you expect those pollers to never arrive.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2062,18 +2056,21 @@ message SetWorkerDeploymentRampingVersionRequest {
...
-    //     - A current deployment version set, signifying that the task queue has been added to another deployment.
-    //     - An add_rate of 0, signifying that there are no backlogged tasks being added in these queues. (see TaskQueueStats.tasks_add_rate)
-    // If true, this operation will succeed even if the new ramping version does not have all the task-queues that were
-    // present in the existing ramping version.
-    // Note: This flag will only be applicable if the version is being set as the ramping version for the first time, signifying
-    // that these checks will be ignored when the ramping version is being unset or on just a percentage change.
+    // Optional. By default this request would be rejected if not all the expected Task Queues are
+    // being polled by the new Version, to protects against accidental removal of Task Queues, or
+    // worker health issues. Pass `true` here to bypass this protection.
+    // The set of expected Task Queues equals to all the Task Queues ever polled from the existing
```
carlydf, line 2062:
```suggestion
    // The set of expected Task Queues is the set of all the Task Queues that ever belonged to the existing
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2062,18 +2056,21 @@ message SetWorkerDeploymentRampingVersionRequest {
...
+    // Current Version of the Deployment, with the following exclusions:
+    //   - Task Queues that are not used anymore (inferred by having empty backlog and a task
+    //     add_rate of 0.)
+    //   - Task Queues that are moved to another Worker Deployment (inferred by the Task Queue
+    //     having a different Current Version than the Current Version of this deployment.)
+    // WARNING: Do not set this flag unless you are sure that the missing task queue poller are not
```
carlydf, line 2068:
```suggestion
    // WARNING: Do not set this flag unless you are sure that the missing task queue pollers are not
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2062,18 +2056,21 @@ message SetWorkerDeploymentRampingVersionRequest {
...
+    // needed. If the request is unexpectedly rejected due to missing pollers, then that means the
+    // pollers have not reached to server yet.
```
carlydf, line 2070:
```suggestion
    // pollers have not reached the server yet.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2105,17 +2102,23 @@ message ListWorkerDeploymentsResponse {
...
-//  - It is drained (see WorkerDeploymentVersionInfo.drainage_info)
+//  - It is not in Draining status (see WorkerDeploymentVersionInfo.drainage_info). This condition
+//    can be skipped by passing `skip-drainage=true`.
 message DeleteWorkerDeploymentVersionRequest {
-    string namespace = 2;
-    temporal.api.deployment.v1.WorkerDeploymentVersion version = 3;
+    string namespace = 1;
+    // Deployment Version identifier in the form "<deployment_name>/<build_id>".
+    string version = 2;
+    // Pass to force deletion even if the Version is not drained. In this case the open pinned
```
carlydf, line 2119:
```suggestion
    // Pass to force deletion even if the Version is draining. In this case the open pinned
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2105,17 +2102,23 @@ message ListWorkerDeploymentsResponse {
...
+        temporal.api.deployment.v1.RoutingConfig routing_config = 3;
     }
 }
 
 // Used for manual deletion of Versions. User can delete a Version only when all the
 // following conditions are met:
+//  - It is not the Current or Ramping Version of its Deployment.
 //  - It has no active pollers (none of the task queues in the Version have pollers)
-//  - It is drained (see WorkerDeploymentVersionInfo.drainage_info)
+//  - It is not in Draining status (see WorkerDeploymentVersionInfo.drainage_info). This condition
```
carlydf, line 2113:
```suggestion
//  - It is not draining (see WorkerDeploymentVersionInfo.drainage_info). This condition
```

temporal/api/enums/v1/deployment.proto
ShahabT: I meant they are distinguished for routing task. I'll make that clear.

PR #550, opened 2025-02-05
Discuss: namespace rules prototype

ychebotarev:
**What changed?**
* Rename NamespaceRule to WorfklowRule (and all that follows)
* Add structure to rule triggering
* Move predicate inside trigger
* Change RESt API to "workflow-rules"
* Make "Action" generic message, with scope
* Move rules into separate package

**Not for merge**
This is a proposal for "namespace level rules". First implementation will be "activity pause".
This PR is based in the latest discussion, and contains all the changes.

Old PR:
https://github.com/temporalio/api/pull/532


temporal/api/batch/v1/message.proto
```diff
@@ -140,3 +141,15 @@ message BatchOperationUnpauseActivities {
   // duration, introducing variability to the start time.
   google.protobuf.Duration jitter = 6;
 }
+
+// BatchOperationApplyNamespaceRule sends ApplyNamespaceRule requests to batch workflows.
+message BatchOperationApplyNamespaceRule {
+  // The identity of the worker/client.
+  string identity = 1;
+
+	oneof rule {
```
bergundy, line 150: nit: please fix indentation.

All of these fields need to be documented.

bergundy, line 150:
```suggestion
	oneof source {
```

In code it'd read better, `rule.source.id` than `rule.rule.id`.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+// namespace rule actions
+message NamespaceRuleActionActivityPause {
+}
+
+message NamespaceRuleActionMetric {
+}
+
+message NamespaceRuleAction {
+    // Rule action.
+    oneof action  {
```
bergundy, line 121:
```suggestion
    oneof variant  {
```

Again, better than `action.action`.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+
+message NamespaceRuleSpec {
+    // Name of the rule.
+    string name = 1;
+
+    // Specifies how the rule should be triggered and evaluated.
+    // Currently, only "activity start" type is supported.
+    // Types:
+    // - "activity start": Rule is triggered when a activity is about to start.
+    // - "workflow task start": Rule is triggered when the new workflow task is about to start
```
bergundy, line 135: Can remove this since it's not supported.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+    temporal.api.enums.v1.NamespaceRuleTriggerType type = 2;
+
+    // Predicate is used to match against workflow data, based on rule trigger type.
+    // The following activity attributes are supported as part of the predicate:
+    // - activityType: The type of the activity.
+    // - activityId: The ID of the activity.
+    // - activityAttempt: The number attempts of the activity.
+    // - backoffInterval: The current backoff interval of the activity.
+    // - activityStatus: The status of the activity. Can be one of "scheduled", "started", "paused".
+    // - taskQueue: The name of the task queue of the activity.
```
bergundy, line 145: These should be in PascalCase to be consistent with search attributes.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+    string predicate = 3;
```
bergundy, line 146: This needs a more specific name to understand what it applies to. I'm tempted to borrow "component" from the chasm terminology. That's what activities would be in the chasm world.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+    // Can be empty if the rule should never expire.
+    google.protobuf.Timestamp expiration_time = 6;
+}
+
+// NamespaceRule describes a rule that can be applied to any workflow in this namespace.
+message NamespaceRule {
+    // Name of the rule.
+    string name = 1;
+
+    // ID must be unique within the namespace
```
bergundy, line 164: Please explain that this can only be set by the server.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+    repeated NamespaceRuleAction action = 5;
+
+    // Expiration time of the rule. After this time, the rule will be deleted.
+    // Can be empty if the rule should never expire.
+    google.protobuf.Timestamp expiration_time = 6;
+}
+
+// NamespaceRule describes a rule that can be applied to any workflow in this namespace.
+message NamespaceRule {
+    // Name of the rule.
```
bergundy, line 161: This also needs to be unique per namespace.
You might also explain what characters are allowed here and possibly character length.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+    string id = 2;
+
+    // The identity of the worker/client.
+	  string identity = 3;
+
+	  // Rule creation date.
+	  google.protobuf.Timestamp create_time = 4;
+
+	  // Rule specification
+	  NamespaceRuleSpec rule_spec = 5;
```
bergundy, line 174: Just call this `spec` IMHO, no need to stutter.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+// NamespaceRule describes a rule that can be applied to any workflow in this namespace.
+message NamespaceRule {
+    // Name of the rule.
+    string name = 1;
+
+    // ID must be unique within the namespace
+    string id = 2;
+
+    // The identity of the worker/client.
+	  string identity = 3;
```
bergundy, line 168: nit: indentation.

bergundy, line 168: I'd call this `created_by_identity` to clarify what it is.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
     // Note: Namespace is in NAMESPACE_STATE_DELETED state when it was deleted from the system but associated data is not deleted yet.
     bool include_deleted = 1;
 }
+
+////////////////////////////////////////////
+// namespace rule actions
```
bergundy, line 112: This will end up in the generated documentation.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+message NamespaceRuleActionActivityPause {
+}
+
+message NamespaceRuleActionMetric {
+}
```
bergundy, line 117: Please document the messages.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1950,3 +1951,71 @@ message GetDeploymentReachabilityResponse {
     // was actually calculated.
     google.protobuf.Timestamp last_update_time = 3;
 }
+
+/////////////////////////////////////////////////////////////////////////////
+// "CR_D" operations for namespace rules (there is no 'update')
```
bergundy, line 1956: Same here, you don't want this ending up in the generated documentation for this message.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1950,3 +1951,71 @@ message GetDeploymentReachabilityResponse {
...
+message CreateNamespaceRuleRequest {
+    string namespace = 1;
+
+    // The rule to create as a part of namespace config.
+    // Rule name must be unique namespace-wide.
```
bergundy, line 1961: You should move this sentence to the spec itself.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1950,3 +1951,71 @@ message GetDeploymentReachabilityResponse {
...
+    temporal.api.namespace.v1.NamespaceRuleSpec rule = 2;
```
bergundy, line 1962: nit: you could call this spec.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1950,3 +1951,71 @@ message GetDeploymentReachabilityResponse {
...
+
+    // If true, the rule will be applied immediately via batch job.
+    // Visibility query in the rule will be used to select the workflows to apply the rule to.
+    bool force_scan = 3;
+}
+
+message CreateNamespaceRuleResponse {
+    // id of the created rule. Unique within the namespace.
```
bergundy, line 1970:
```suggestion
    // ID of the created rule. Unique within the namespace.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1950,3 +1951,71 @@ message GetDeploymentReachabilityResponse {
...
+    string rule_id = 1;
```
bergundy, line 1971: You can probably just call this `id` since it's within the rule crud APIs.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1950,3 +1951,71 @@ message GetDeploymentReachabilityResponse {
...
+}
+
+message ApplyNamespaceRuleRequest {
+    string namespace = 1;
+
+    // Execution info of the workflow which scheduled this activity
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 2;
+
+	  // either provide name/id of existing rule, or rule specification
+	  oneof rule {
```
bergundy, line 2013: `rule_source` or just `source`.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1023,4 +1023,63 @@ service WorkflowService {
...
         };
     }
+
+
+    // Create a new namespace rule.
+    // Returns the unique ID of the new rule.
+    // Note: if the rule with such name and spec exist this call will fail.
+    rpc CreateNamespaceRule (CreateNamespaceRuleRequest) returns (CreateNamespaceRuleResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/rules/{rule.name}"
```
bergundy, line 1033: Hmm... I don't remember and too lazy to check but I think typically our "create" APIs don't take the ID or name in the path.

cretz: I think with user-supplied identifiers (which "name" is), I think it makes sense to be present in the URL

temporal/api/workflowservice/v1/request_response.proto
cretz, line 1971: Note, it is a bit strange that we're using ID as a value a user cannot define goes against what we have done in other places where we have "[thing] ID" where "[thing]" is user created (e.g. workflows and schedules and activities). Are we sure we want to call the thing that a user cannot provide an "ID" unlike our other use of IDs? Just confirming we're ok being a bit different

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1023,4 +1023,63 @@ service WorkflowService {
             }
         };
     }
+
+
+    // Create a new namespace rule.
+    // Returns the unique ID of the new rule.
+    // Note: if the rule with such name and spec exist this call will fail.
+    rpc CreateNamespaceRule (CreateNamespaceRuleRequest) returns (CreateNamespaceRuleResponse) {
```
cretz, line 1031: To be honest, I kinda wish these were called "ActionRule"s or something. They don't even apply at the namespace level if applied ephemerally. If there is any other reason to have a "rule" in a namespace now, we can't, because it's being taken for this purpose. The term rule is quite generic and from an HTTP API POV are just "rules" and therefore are effectively what "rules" are now at Temporal.

It's just a bit generic and doesn't always apply to a namespace, may apply to a single workflow for instance ephemerally. I would not expect to have "NamespaceSchedule" or "NamespaceWorkflow" as all things in this service are namespace-specific implicitly.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1023,4 +1023,63 @@ service WorkflowService {
...
+            get: "/namespaces/{namespace}/rules"
+            additional_bindings {
+                get: "/api/v1/namespaces/{namespace}/rules"
+            }
+        };
+    }
+
+    // ApplyNamespaceRule allows to apply rule to a certain workflow instance without creating a rule.
+    // This is useful for one-off operations.
+    rpc ApplyNamespaceRule(ApplyNamespaceRuleRequest) returns (ApplyNamespaceRuleResponse) {
```
cretz, line 1075:
```suggestion
    rpc ApplyNamespaceRule (ApplyNamespaceRuleRequest) returns (ApplyNamespaceRuleResponse) {
```

Inconsistent spacing (we may have inconsistent spacing in the file, but no need to be inconsistent with yourself)

cretz, line 1075: Arguably this should be called `ApplyWorkflowExecutionNamespaceRule` and is another reason to use a better name than "NamespaceRule" (like "ActionRule").

temporal/api/namespace/v1/message.proto
ychebotarev: renamed

temporal/api/namespace/v1/message.proto
ychebotarev: I think more. I guess this approach is raising too much questions, and not supported by the rest of our API. 
"Id" is used as a "business oriented unique value".
I will remove name, and ID will be unique, provided by a caller.

temporal/api/namespace/v1/message.proto
ychebotarev: it shouldn't. But I remove it anyway.

temporal/api/namespace/v1/message.proto
ychebotarev: renamed

temporal/api/namespace/v1/message.proto
ychebotarev: it is called identity everywhere else. That said - `identity` is usually part of the request, but not part of the entity (like scheduler for example). I will remove it from spec

temporal/api/namespace/v1/message.proto
ychebotarev: sigh. removed.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: removed

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: renamed

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: if I call it `id`, in the API it will be `/namespaces/{namespace}/rules/{id}`
I think `/namespaces/{namespace}/rules/{rule_id}` is better, more expressive, and consistent with other APIs
(like `/namespaces/{namespace}/schedules/{schedule_id}`)

temporal/api/workflowservice/v1/service.proto
ychebotarev: renamed to `spec.id`

temporal/api/namespace/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: yeah, moved inside

temporal/api/namespace/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: we where talking about combining predicate with action, which I was against. But yes, we can combine predicate with trigger in a more structured way.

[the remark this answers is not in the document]
ychebotarev: I will go with `open trigger { ActivityStartTrigger ...}`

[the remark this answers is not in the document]
ychebotarev: unexpected bonus - it also nicely separates 'visibility_query', now they are more clearly two different things.

temporal/api/namespace/v1/message.proto
```diff
@@ -107,3 +107,69 @@ message NamespaceFilter {
...
+    // - backoffInterval: The current backoff interval of the activity.
+    // - activityStatus: The status of the activity. Can be one of "scheduled", "started", "paused".
+    // - taskQueue: The name of the task queue of the activity.
+    string predicate = 3;
+
+    // Restricted Visibility query, supports "WorkflowType", "StartTime" and few more workflow properties
+    string visibility_query = 4;
+
+    // Action to be taken when the rule is triggered and predicate is matched.
+    repeated NamespaceRuleAction action = 5;
```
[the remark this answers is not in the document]
ychebotarev: yep.

temporal/api/enums/v1/namespace.proto
```diff
@@ -49,3 +49,8 @@ enum ReplicationState {
     REPLICATION_STATE_NORMAL = 1;
     REPLICATION_STATE_HANDOVER = 2;
 }
+
+enum NamespaceRuleTriggerType {
+    NAMESPACE_RULE_TRIGGER_TYPE_UNSPECIFIED = 0;
+    NAMESPACE_RULE_TRIGGER_TYPE_START_ACTIVITY = 1;
```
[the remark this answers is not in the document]
ychebotarev: removed since I will go with a more structured form.
as for "whether this is schedule activity, start activity, or start activity attempt" - I don't think it matters for caller. I can't imagine a situation when being specific between "about to be scheduled" and "about to be started" change anything. Adding this expose too much of internal implementation details.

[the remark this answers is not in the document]
ychebotarev: I also want to not make this decision right now...

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I decide that given all the questions about name/id I will go only with ID. Too much confusion over something that doesn't have enough upfront value.

temporal/api/workflowservice/v1/service.proto
ychebotarev: Will discuss more offline, but I'm drifting toward `WorkflowRule`
`NamespaceRule` scope is broader then this efforts.
`ActionRule` - semantically too abstract, narrow.
`NamespaceWorkflowRule` - too verbose.

temporal/api/workflowservice/v1/service.proto
ychebotarev: done

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1023,4 +1023,63 @@ service WorkflowService {
...
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/rules/apply"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/rules/apply"
+                body: "*"
+            }
+        };
```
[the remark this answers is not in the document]
ychebotarev: good point, I will go with "/namespaces/{namespace}/workflows/{workflow_execution.workflow_id}/apply-rule"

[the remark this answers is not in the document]
ychebotarev: I also renamed to `WorfklowRule`, lets see how it feels.

temporal/api/workflowservice/v1/service.proto
ychebotarev: Renamed to ApplyWorkflowRule

temporal/api/batch/v1/message.proto
ychebotarev: added documentation.

ychebotarev: only if you call variable of BatchOperationApplyNamespaceRule rule.
Usually it is called `operation`
operation.rule.id sounds good to me.

temporal/api/namespace/v1/message.proto
ychebotarev: renamed to variant

temporal/api/namespace/v1/message.proto
ychebotarev: removed

temporal/api/namespace/v1/message.proto
ychebotarev: "component" as as "used to match against workflow data, based on rule trigger type"?
predicate sounds better to me.
Also - see new changes, I moved it into more structured form.

temporal/api/namespace/v1/message.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: done

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: It will be `request.rule.id` or `request.rule.spec`, both sounds ok for me. And `request.source.id` is lacking context (it is about rule).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1950,3 +1951,71 @@ message GetDeploymentReachabilityResponse {
...
+    string namespace = 1;
+
+    // ID of the rule to delete. Unique within the namespace.
+    string rule_id = 2;
+}
+
+message DeleteNamespaceRuleResponse {
+}
+
+message ListNamespaceRulesRequest {
```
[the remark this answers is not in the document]
ychebotarev: `Never` is too strong. So far we don't expect to have huge number of rules. Initial limit will be around 10. 
But lets add it anyway.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I will keep this one open.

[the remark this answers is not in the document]
cretz: In these APIs, especially when viewing in HTTP, it is a nice property to return the resource you created. In the future as more fields may appear that are set/defaulted on create, it helps that too. Granted we haven't followed this "resource"ful response that much in our API thus far, but it is nice to have.

cretz: Just a note here, the pattern for HTTP API is just about field extraction, what the field names are does not matter to this HTTP pattern. Having said that I don't mind which approach is taken so long as consistent with the other rule calls.

temporal/api/workflow/v1/message.proto
```diff
@@ -458,3 +458,64 @@ message OnConflictOptions {
...
 }
+
+message WorkflowRuleAction {
+    // Pause the activity.
+    message WorkflowRuleActionActivityPause {
+    }
+
+    // Emit a metric.
+    message WorkflowRuleActionMetric {
+    }
```
cretz, line 469: Feel free to remove the `WorkflowRuleAction` prefix on the message names if you want (don't have to)

temporal/api/workflow/v1/message.proto
```diff
@@ -458,3 +458,64 @@ message OnConflictOptions {
...
+        // - ActivityStatus: The status of the activity. Can be one of "scheduled", "started", "paused".
+        // - TaskQueue: The name of the task queue of the activity.
+        string predicate = 2;
+    }
+
+    // Specifies how the rule should be triggered and evaluated.
+    // Currently, only "activity start" type is supported.
+    oneof trigger {
+        ActivityStartTrigger activity_start = 3;
+    }
```
cretz, line 499:
```suggestion
        string predicate = 1;
    }

    // Specifies how the rule should be triggered and evaluated.
    // Currently, only "activity start" type is supported.
    oneof trigger {
        ActivityStartTrigger activity_start = 2;
    }
```

And renumber below. Field numbers are unique to the message, don't have to keep in order across messages

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1023,4 +1023,64 @@ service WorkflowService {
             }
         };
     }
+
+
+    // Create a new workflow rule.
+    // If the rule with such ID already exist this call will fail.
+    rpc CreateWorkflowRule (CreateWorkflowRuleRequest) returns (CreateWorkflowRuleResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/workflows/rules/{spec.id}"
```
cretz, line 1032:
```suggestion
            post: "/namespaces/{namespace}/workflow-rules/{spec.id}"
```

So as not to be confused with `workflows` (a different thing). Same for other paths.

temporal/api/workflowservice/v1/service.proto
ychebotarev: I will try 'WorkflowRule'

cretz (commented):
Only minor comments, I think overall the updates are great

temporal/api/workflow/v1/message.proto
ychebotarev: good point, but - after compilation that are not "included" into WorkflowRuleAction. So in code it will look like
```
action : &workflowservice.ActivityPause { ...
```
And `ActivityPause` is already taken.

temporal/api/workflowservice/v1/service.proto
ychebotarev: accepted

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
ychebotarev: I search for best practice, and seems like common recommendation is "it is tradeoff between payload size and ease of use".
In this case payload is small, so yeah, lets return it.

temporal/api/batch/v1/message.proto
```diff
@@ -140,3 +140,16 @@ message BatchOperationUnpauseActivities {
   // duration, introducing variability to the start time.
   google.protobuf.Duration jitter = 6;
 }
+
+// BatchOperationApplyWorkflowRule sends ApplyWorkflowRule requests to batch workflows.
+message BatchOperationApplyWorkflowRule {
```
bergundy, line 145: I don't think workflow rule is a good name. Ideally these rules could be applied to other entities in the system that we have planned for 2025 (e.g. standalone activity).

temporal/api/workflow/v1/message.proto
```diff
@@ -458,3 +458,64 @@ message OnConflictOptions {
...
+        string predicate = 1;
+    }
+
+    // Specifies how the rule should be triggered and evaluated.
+    // Currently, only "activity start" type is supported.
+    oneof trigger {
+        ActivityStartTrigger activity_start = 2;
+    }
+
+    // Restricted Visibility query, supports "WorkflowType", "StartTime" and few more workflow properties
```
bergundy, line 501: Document what's supported please.

bergundy (commented):
Not sure what was applied from my previous round of feedback, with this drastic change it's hard to follow.

temporal/api/workflowservice/v1/request_response.proto
bergundy: We have namespace and nexus endpoint names and IDs to support renaming. For rules, we discussed that there's value in knowing what specific incarnation of the rule (e.g. when recreated with the same name) affected which workflows, and having a way to undo that effect.

temporal/api/workflowservice/v1/service.proto
bergundy: Commented on this here: https://github.com/temporalio/api/pull/550#discussion_r1946905537

> I don't think workflow rule is a good name. Ideally these rules could be applied to other entities in the system that we have planned for 2025 (e.g. standalone activity).

temporal/api/workflow/v1/message.proto
cretz: That's why I suggested `ActivityPauseAction` not `ActivityPause`. And in code it's actually something like `&workflowservice.WorkflowRuleAction_Whatever`.

temporal/api/batch/v1/message.proto
cretz: That's the question is whether these are only rules for workflows, which I was told was the case for this project. If not, I suggested "ActionRule". I balked at "NamespaceRule" because everything here is implied to belong to a namespace and the term "rule" was too generic (e.g. in HTTP APIs).

temporal/api/workflow/v1/message.proto
```diff
@@ -458,3 +458,64 @@ message OnConflictOptions {
...
+    message ActivityStartTrigger {
+        // Activity predicate is used to match against workflow data, based on rule trigger type.
+        // The following activity attributes are supported as part of the predicate:
+        // - ActivityType: The type of the activity.
+        // - ActivityId: The ID of the activity.
+        // - ActivityAttempt: The number attempts of the activity.
+        // - BackoffInterval: The current backoff interval of the activity.
+        // - ActivityStatus: The status of the activity. Can be one of "scheduled", "started", "paused".
+        // - TaskQueue: The name of the task queue of the activity.
+        string predicate = 1;
```
cretz, line 492: May need to say this takes the same format as visibility queries just with different attributes. @rodrigozhou - now that we're reusing this format, can we now publish the grammar doc that we maintain internally?

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: What is the exact scenario for name/id?  
* user created rule with some name and ID
* then deleted it
* then created with the same name but different ID

What was the problem this solves for a user? Comparing to 
* user created a rule with some ID
* then deleted it
* then (re)created it with the same ID

I would rather keep it simple. If we will need something like that we can introduce `name` later (or `description` - it better fits into "shouldn't be unique")

cretz: I will say from a user POV if they can't delete + recreate with same user-defined identifier, it will be a UX issue (especially if we don't allow update)

temporal/api/workflowservice/v1/service.proto
ychebotarev: So far every rule we create is in workflow scope - they are triggered based on workflow state transitions, use something in workflow for triggering, etc.
I don't know the details of all those other entities, so not sure how much we could/should generalize here. 
Under the current constraints `WorkflowRule` seems like a best fit.

ychebotarev: I will raise it during review meeting.

temporal/api/workflow/v1/message.proto
ychebotarev: ok.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: if they delete - they will be able to recreate with the same ID.

temporal/api/batch/v1/message.proto
ychebotarev: `ActionRule` - what is the meaning of "action" in this name? 
Are there rules that don't have actions? Are we creating rules that results in action, as opposite to rules that will not have action attached to them? 
I see very clear boundaries for this rules, and currently `what happens in workflow` is the best description of those boundaries.

cretz: > what is the meaning of "action" in this name?
A rule that runes actions. It doesn't have meaning really and any adjective besides "Namespace" or none is fine IMO.

bergundy: I would rather not develop this to be workflow specific. I think there's a very high likelyhood that we're going to want to extend this to other entities.

temporal/api/workflowservice/v1/service.proto
ychebotarev: stick with `WorkflowRule`, but make actions more generic. Also move to the separate package, so later we can reuse it.

temporal/api/batch/v1/message.proto
ychebotarev: stick with WorkflowRule, but make actions more generic. Also move to the separate package, so later we can reuse it.

temporal/api/enums/v1/common.proto
```diff
@@ -104,3 +104,10 @@ enum NexusOperationCancellationState {
     // Cancellation request is blocked (eg: by circuit breaker).
     NEXUS_OPERATION_CANCELLATION_STATE_BLOCKED = 6;
 }
+
+enum Scope {
+    SCOPE_UNSPECIFIED = 0;
+    SCOPE_NAMESPACE = 1;
+    SCOPE_WORKFLOW = 2;
+    SCOPE_ACTIVITY = 3;
+}
```
bergundy, line 113: This needs a qualifier since it's in the shared `enums` package. Consider `RuleActionScope`.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,102 @@
...
+
+message Action {
+  // Pause action.
+  message ActionPause {
+    temporal.api.enums.v1.Scope scope = 1;
+  }
+
+  // Emit a metric.
+  message ActionEmitMetric {
+    temporal.api.enums.v1.Scope scope = 1;
```
bergundy, line 46: Worth explaining the semantics of scope here.

temporal/api/enums/v1/common.proto
bergundy: What's SCOPE_NAMESPACE?
Please document all of these.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,102 @@
...
+  // Action to be taken when the rule is triggered and predicate is matched.
+  repeated Action actions = 4;
+
+  // Expiration time of the rule. After this time, the rule will be deleted.
+  // Can be empty if the rule should never expire.
+  google.protobuf.Timestamp expiration_time = 5;
+}
+
+// WorkflowRule describes a rule that can be applied to any workflow in this namespace.
+message WorkflowRule {
```
bergundy, line 96: This should also include the batch job ID IMHO if there's one associated.

temporal/api/enums/v1/common.proto
ychebotarev: Add a qualifier, remove SCOPE_NAMESPACE - not needed for the nearest future.

temporal/api/rules/v1/message.proto
ychebotarev: comments added.

temporal/api/enums/v1/common.proto
```diff
@@ -104,3 +104,9 @@ enum NexusOperationCancellationState {
     // Cancellation request is blocked (eg: by circuit breaker).
     NEXUS_OPERATION_CANCELLATION_STATE_BLOCKED = 6;
 }
+
+enum RuleActionScope {
+    RULE_ACTION_SCOPE_UNSPECIFIED = 0;
+    RULE_ACTION_SCOPE_WORKFLOW = 1;
+    RULE_ACTION_SCOPE_ACTIVITY = 2;
```
bergundy, line 111: Please document these.

bergundy: A short description is enough.

temporal/api/rules/v1/message.proto
bergundy: I'm not blocking on this, it can always be added later.

bergundy (approved):
I skimmed this again and don't have any blocking comments. Please also get one SDK approver (@cretz).

cretz (commented):
Sorry, started this review and realized I didn't submit it.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+  // Pause action.
+  message ActionPause {
+    // Indicates what should be paused. Depending on the scope value, the action
+    // will pause the entire workflow or a specific activity (additional entities
+    // may be supported in the future).
+    temporal.api.enums.v1.RuleActionScope scope = 1;
+  }
+
+  // Emit a metric.
+  message ActionEmitMetric {
```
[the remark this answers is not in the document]
ychebotarev: it is not expected to be implemented now. It will not be in "merge" PR

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+import "temporal/api/enums/v1/common.proto";
+
+message Action {
+  // Pause action.
+  message ActionPause {
+    // Indicates what should be paused. Depending on the scope value, the action
+    // will pause the entire workflow or a specific activity (additional entities
+    // may be supported in the future).
+    temporal.api.enums.v1.RuleActionScope scope = 1;
+  }
```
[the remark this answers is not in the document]
ychebotarev: there are two ways of modeling this -  have a separate message to encode each valid <action, scope> pair:
```
message ActivityPauseAction {
}
message WorkflowPauseAction {
}
...
```

or encapsulation -  single action - pause - that can be applied in different scope.
```
message ActionPause {
   Scope scope;
}
```
There are multiple pro-con for each of those approaches, but in a context of current WorkflowRule and future rules (according to Roey they will come) generic action with specific scope feels like a cleaner fit.

Yes, its a valid concern,  that sooner or later I would have some action that is not applicable to all scopes, like "activity-specific" action. It will be a situation where not every pair of <action>*<scope> is supported. "pro" (at least for me) are outweighing "con". 

Specifically:
* less upfront complexity. If we decide to encode every cross-product as a single message it will make a bigger surface. For 4 actions 2 scopes it is already 8 different messages. And it is very likely there will be 4.Messy.
* Extensibility. We expect more "scope" and "actions".  and adding new will be easier this way.
* This is "reactive' API, so compile-time guarantee (main "pro" of separate messages for every pair) is not that important. Run-time validation is good enough.

temporal/api/enums/v1/common.proto
ychebotarev: add some comments

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1132,4 +1132,64 @@ service WorkflowService {
             }
         };
     }
+
+
+    // Create a new workflow rule.
+    // If the rule with such ID already exist this call will fail.
+    rpc CreateWorkflowRule (CreateWorkflowRuleRequest) returns (CreateWorkflowRuleResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/workflows-rules/{spec.id}"
```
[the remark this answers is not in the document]
ychebotarev: renames

PR #551, opened 2025-02-06
Add HandlerErrorRetryBehavior

bergundy:


Quinn-With-Two-Ns (approved):

temporal/api/nexus/v1/message.proto
```diff
@@ -211,3 +212,15 @@ message EndpointTarget {
         External external = 2;
     }
 }
+
+// HandlerErrorRetryBehavior allows handlers to explicity set the retry behavior of a HandlerError. If not specified,
+// retry behavior is determined from the error type. For example internal errors are not retryable by default unless
+// specified otherwise.
+enum HandlerErrorRetryBehavior {
```
cretz, line 219: Should this be in `enums` like all other enums in this project?

bergundy: Yeah, you're right, I checked, all other enums are there. I don't love spreading related definitions out like that but I understand the argument for consistency.

pdoerner (approved):

cretz (approved):
I note this is merging to `master` and not a feature branch. I trust the implementation and design is far enough along that we don't expect incompatible changes to this.

bergundy:
Yes, there is a corresponding server and Go SDK PR (Java is waiting on this too).

PR #552, opened 2025-02-06
Deprecate ScanWorkflowExecutions

jmbarzee:

Deprecate ScanWorkflowExecutions

To ultimately remove ScanWorkflowExecutions in favor of ListWorkflowExecutions

No

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -899,6 +899,7 @@ message ListArchivedWorkflowExecutionsResponse {
     bytes next_page_token = 2;
 }
 
+// Deprecated: Use with `ListWorkflowExecutions`.
 message ScanWorkflowExecutionsRequest {
     string namespace = 1;
     int32 page_size = 2;
@@ -906,6 +907,7 @@ message ScanWorkflowExecutionsRequest {
     string query = 4;
 }
 
+// Deprecated: Use with `ListWorkflowExecutions`.
 message ScanWorkflowExecutionsResponse {
     repeated temporal.api.workflow.v1.WorkflowExecutionInfo executions = 1;
     bytes next_page_token = 2;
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -899,13 +899,15 @@ message ListArchivedWorkflowExecutionsResponse {
     bytes next_page_token = 2;
 }
 
+// Deprecated. See with `ListWorkflowExecutions`.
 message ScanWorkflowExecutionsRequest {
-    string namespace = 1;
+    string namespace = 1;
```
rodrigozhou, line 904: nit: Remove trailing space.

rodrigozhou (approved):

rodrigozhou:
You need to run `make proto` to update some auto-generate files.

Quinn-With-Two-Ns (approved):

mjameswh (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -899,13 +899,15 @@ message ListArchivedWorkflowExecutionsResponse {
     bytes next_page_token = 2;
 }
 
+// Deprecated. See with `ListWorkflowExecutions`.
```
bergundy, line 902: If you want golang to mark this as deprecated it needs to have be `Deprecated:`

bergundy: I think it also needs to be in the second paragraph: https://go.dev/wiki/Deprecated

bergundy: Yes, it only affects golang but I figure that's better than nothing.

jmbarzee: Okay works for me. I was following all the deprecations I saw in the repo.

PR #553, opened 2025-02-08
Poller Scaling Decisions

Sushisource:

**What changed?**
Added a proto that is optionally attached to task responses and contains data for the SDK about whether or not pollers should be scaled up or down.

**Why?**
Part of the worker management effort to simplify configuration of workers for users.

**Breaking changes**
Nope

**Server PR**
It's nonbreaking, but the PR is here: https://github.com/temporalio/temporal/pull/7300

### DIFF
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -325,7 +325,7 @@ message TimestampedCompatibleBuildIdRedirectRule {
     google.protobuf.Timestamp create_time = 2;
 }
 
-// Attached to task responsese to give hints to the SDK about how it may adjust its number of
+// Attached to task responses to give hints to the SDK about how it may adjust its number of
 // pollers.
 message PollerScalingDecision {
   // How many poll requests to suggest should be added or removed, if any. As of now, server only
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -324,3 +324,15 @@ message TimestampedCompatibleBuildIdRedirectRule {
     CompatibleBuildIdRedirectRule rule = 1;
     google.protobuf.Timestamp create_time = 2;
 }
+
+// Attached to task responsese to give hints to the SDK about how it may adjust its number of
+// pollers.
+message PollerScalingDecision {
+  // How many poll requests to suggest should be added or removed, if any. As of now, server only
+  // scales up or down by 1. However, SDKs should allow for other values (while staying within
+  // defined min/max).
+  //
+  // The SDK is free to ignore this suggestion, EX: making more polls would not make sense because
+  // all slots are already occupied.
+  int32 poll_request_delta_suggestion = 1;
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -331,6 +331,8 @@ message PollWorkflowTaskQueueResponse {
     map<string, temporal.api.query.v1.WorkflowQuery> queries = 14;
     // Protocol messages piggybacking on a WFT as a transport
     repeated temporal.api.protocol.v1.Message messages = 15;
+    // Server-advised information the SDK may use to adjust its poller count.
+    temporal.api.taskqueue.v1.PollerScalingDecision poller_scaling_decision = 16;
 }
 
 message RespondWorkflowTaskCompletedRequest {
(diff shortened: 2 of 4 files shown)


temporal/api/sdk/v1/poller_scaling.proto
```diff
@@ -0,0 +1,38 @@
...
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "PollerScalingProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+message PollerScalingDecision {
+  // How many pollers should be added or removed, if any. As of now, server only scales up or down
+  // by 1. However, SDKs should allow for other values (while staying within defined min/max).
+  int32 poller_delta = 1;
```
mfateev, line 37: I would recommend an API that is not delta based as it is much less prone to race conditions. Something like target_poller_count.

Sushisource: We could do that, but the problem is it'd require a lot more internal communication among partitions to determine overall load. That can certainly be more accurate, but also has more overhead. This solution has produced some really great results while introducing effectively zero new overhead, which seems like a great place to be.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1733,6 +1738,8 @@ message PollNexusTaskQueueResponse {
     bytes task_token = 1;
     // Embedded request as translated from the incoming frontend request.
     temporal.api.nexus.v1.Request request = 2;
+    // Server-advised information the SDK may use to adjust its poller count.
+    temporal.api.sdk.v1.PollerScalingDecision poller_scaling_decision = 3;
```
bergundy, line 1742: What is this scaling decision based on? Are we looking at the in-memory queue length?

bergundy (approved):

temporal/api/sdk/v1/poller_scaling.proto
cretz, line 1: I am a little afraid of putting in this "sdk" package, because I think this package was meant for "SDK protos that whose internals have no bearing on the server code", meaning SDK team can change the contents of the message at will without the server team being concerned. We even have a special `CODEOWNERS` file rule that excludes this package from server review. But I think _this_ message's body is very much a co-owned/important one. Maybe in the `taskqueue` package instead?

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Check out the server review https://github.com/temporalio/temporal/pull/7300

temporal/api/sdk/v1/poller_scaling.proto
cretz: Forgot to remove this empty file once moved

cretz (approved):
This looks good to me, but I'd like to request we don't merge until https://github.com/temporalio/temporal/pull/7300 is approved by the right people just in case we change our mind on the API in that PR. (feel free to use feature branches all around if you think it may be an involved process)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1733,6 +1737,8 @@ message PollNexusTaskQueueResponse {
     bytes task_token = 1;
     // Embedded request as translated from the incoming frontend request.
     temporal.api.nexus.v1.Request request = 2;
+    // Server-advised information the SDK may use to adjust its poller count.
+    temporal.api.taskqueue.v1.PollerScalingDecision poller_scaling_decision = 3;
```
dnr, line 1741: I didn't look at the implementation yet, but does the server report this? My question is that nexus tasks aren't backlogged, they're only synchronous, so maybe the algorithm has to change slightly? It still makes sense to adjust pollers based on traffic

temporal/api/taskqueue/v1/message.proto
```diff
@@ -323,4 +323,16 @@ message TimestampedBuildIdAssignmentRule {
...
     CompatibleBuildIdRedirectRule rule = 1;
     google.protobuf.Timestamp create_time = 2;
-}
\ No newline at end of file
+}
+
+// Attached to task responsese to give hints to the SDK about how it may adjust its number of
+// pollers.
+message PollerScalingDecision {
+  // How many poll requests to suggest should be added or removed, if any. As of now, server only
```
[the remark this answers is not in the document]
Sushisource: No, there are two separate scaling controllers when this is enabled

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Yeah, it doesn't at the moment. I'll need to see where I can fit that in.

Sushisource: My server PR is properly scaling Nexus polls now

ShahabT: The plan is to count sync-match in-memory "backlog" as part of the ApproximateBacklogCount in which case Nexus will start to have non-zero "backlog count" when there are not enough pollers. So it's good that we have this here.

ShahabT (approved):

PR #556, opened 2025-02-28
add workflow rules to the namespace config

ychebotarev:
**What changed?**
Add workflow rules to the namespace config proto.

**Why?**
So we can return workflow rules as a part of API calls.

### DIFF
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -35,6 +35,7 @@ import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
 
 import "temporal/api/enums/v1/namespace.proto";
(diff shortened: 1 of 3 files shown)


temporal/api/namespace/v1/message.proto
```diff
@@ -74,6 +75,7 @@ message NamespaceConfig {
     string visibility_archival_uri = 6;
     // Map from field name to alias.
     map<string, string> custom_search_attribute_aliases = 7;
+    map<string, temporal.api.rules.v1.WorkflowRule> workflow_rules = 9;
```
bergundy, line 78: Don't put this on `NamespaceInfo` the fact that rules are stored on a namespace record is an implementation detail and deserve their own APIs.

PR #558, opened 2025-03-06
Add workflow rules proto definitions

ychebotarev:
**What changed?**
Note:
->[Previous PR we use to review](https://github.com/temporalio/api/pull/550)<-
It has lots of discussion, and current implementation is mostly what we settle on. With some changes from design review.

Add definitions for the following API:
* CreateWorkflowRule
* DescribeWorkflowRule
* DeleteWorkflowRule
* ListWorkflowRules

Add definitions for Rules, and Actions.
Corresponding Server PR contains implementation of those APIs.

**Why?**
Working on workflow rules.

**Breaking changes**
New API is added.

**Server PR**

[Implementation](https://github.com/temporalio/temporal/pull/7437)

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -6858,7 +6858,7 @@
       "type": "object",
       "properties": {
         "scope": {
-          "$ref": "#/definitions/v1RuleActionScope",
+          "$ref": "#/definitions/v1WorkflowRuleActionScope",
           "description": "Indicates what should be paused. Depending on the scope value, the action\nwill pause the entire workflow or a specific activity (additional entities\nmay be supported in the future)."
         }
       },
@@ -12480,16 +12480,6 @@
         }
       }
     },
-    "v1RuleActionScope": {
-      "type": "string",
-      "enum": [
-        "RULE_ACTION_SCOPE_UNSPECIFIED",
-        "RULE_ACTION_SCOPE_WORKFLOW",
-        "RULE_ACTION_SCOPE_ACTIVITY"
-      ],
-      "default": "RULE_ACTION_SCOPE_UNSPECIFIED",
-      "description": " - RULE_ACTION_SCOPE_UNSPECIFIED: Default value, unspecified scope.\n - RULE_ACTION_SCOPE_WORKFLOW: The action will be applied to the entire workflow.\n - RULE_ACTION_SCOPE_ACTIVITY: The action will be applied to a specific activity."
-    },
     "v1ScanWorkflowExecutionsResponse": {
       "type": "object",
       "properties": {
@@ -15242,6 +15232,16 @@
           "$ref": "#/definitions/WorkflowRuleActionActionPause"
         }
       }
+    },
+    "v1WorkflowRuleActionScope": {
+      "type": "string",
+      "enum": [
+        "WORKFLOW_RULE_ACTION_SCOPE_UNSPECIFIED",
+        "WORKFLOW_RULE_ACTION_SCOPE_WORKFLOW",
+        "WORKFLOW_RULE_ACTION_SCOPE_ACTIVITY"
+      ],
+      "default": "WORKFLOW_RULE_ACTION_SCOPE_UNSPECIFIED",
+      "description": " - WORKFLOW_RULE_ACTION_SCOPE_UNSPECIFIED: Default value, unspecified scope.\n - WORKFLOW_RULE_ACTION_SCOPE_WORKFLOW: The action will be applied to the entire workflow.\n - WORKFLOW_RULE_ACTION_SCOPE_ACTIVITY: The action will be applied to a specific activity."
     },
     "v1WorkflowRuleSpec": {
       "type": "object",
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -2210,6 +2210,50 @@
             "in": "path",
             "required": true,
             "type": "string"
+          },
+          {
+            "name": "nextPageToken",
+            "in": "query",
+            "required": false,
+            "type": "string",
+            "format": "byte"
+          }
+        ],
+        "tags": [
+          "WorkflowService"
+        ]
+      },
+      "post": {
+        "summary": "Create a new workflow rule.\nIf the rule with such ID already exist this call will fail.",
+        "operationId": "CreateWorkflowRule2",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1CreateWorkflowRuleResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpcStatus"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "namespace",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/WorkflowServiceCreateWorkflowRuleBody"
+            }
           }
         ],
         "tags": [
@@ -2284,52 +2328,6 @@
             "in": "path",
             "required": true,
             "type": "string"
-          }
-        ],
-        "tags": [
-          "WorkflowService"
-        ]
-      }
-    },
-    "/api/v1/namespaces/{namespace}/workflow-rules/{spec.id}": {
-      "post": {
-        "summary": "Create a new workflow rule.\nIf the rule with such ID already exist this call will fail.",
-        "operationId": "CreateWorkflowRule2",
-        "responses": {
-          "200": {
-            "description": "A successful response.",
-            "schema": {
-              "$ref": "#/definitions/v1CreateWorkflowRuleResponse"
-            }
-          },
-          "default": {
-            "description": "An unexpected error response.",
-            "schema": {
-              "$ref": "#/definitions/rpcStatus"
-            }
-          }
-        },
-        "parameters": [
-          {
-            "name": "namespace",
-            "in": "path",
-            "required": true,
-            "type": "string"
-          },
-          {
-            "name": "spec.id",
-            "description": "The id of the new workflow rule. Must be unique within the namespace.",
-            "in": "path",
-            "required": true,
-            "type": "string"
-          },
-          {
-            "name": "body",
-            "in": "body",
-            "required": true,
-            "schema": {
-              "$ref": "#/definitions/WorkflowServiceCreateWorkflowRuleBody"
-            }
           }
         ],
         "tags": [
@@ -5641,6 +5639,50 @@
             "in": "path",
             "required": true,
             "type": "string"
+          },
+          {
+            "name": "nextPageToken",
+            "in": "query",
+            "required": false,
+            "type": "string",
+            "format": "byte"
+          }
+        ],
+        "tags": [
+          "WorkflowService"
+        ]
+      },
+      "post": {
+        "summary": "Create a new workflow rule.\nIf the rule with such ID already exist this call will fail.",
+        "operationId": "CreateWorkflowRule",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1CreateWorkflowRuleResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpcStatus"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "namespace",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/WorkflowServiceCreateWorkflowRuleBody"
+            }
           }
         ],
         "tags": [
@@ -5715,52 +5757,6 @@
             "in": "path",
             "required": true,
             "type": "string"
-          }
-        ],
-        "tags": [
-          "WorkflowService"
-        ]
-      }
-    },
-    "/namespaces/{namespace}/workflow-rules/{spec.id}": {
-      "post": {
-        "summary": "Create a new workflow rule.\nIf the rule with such ID already exist this call will fail.",
-        "operationId": "CreateWorkflowRule",
-        "responses": {
-          "200": {
-            "description": "A successful response.",
-            "schema": {
-              "$ref": "#/definitions/v1CreateWorkflowRuleResponse"
-            }
-          },
-          "default": {
-            "description": "An unexpected error response.",
-            "schema": {
-              "$ref": "#/definitions/rpcStatus"
-            }
-          }
-        },
-        "parameters": [
-          {
-            "name": "namespace",
-            "in": "path",
-            "required": true,
-            "type": "string"
-          },
-          {
-            "name": "spec.id",
-            "description": "The id of the new workflow rule. Must be unique within the namespace.",
-            "in": "path",
-            "required": true,
-            "type": "string"
-          },
-          {
-            "name": "body",
-            "in": "body",
-            "required": true,
-            "schema": {
-              "$ref": "#/definitions/WorkflowServiceCreateWorkflowRuleBody"
-            }
           }
         ],
         "tags": [
@@ -6507,16 +6503,6 @@
     }
   },
   "definitions": {
-    "ActionActionPause": {
-      "type": "object",
-      "properties": {
-        "scope": {
-          "$ref": "#/definitions/v1RuleActionScope",
-          "description": "Indicates what should be paused. Depending on the scope value, the action\nwill pause the entire workflow or a specific activity (additional entities\nmay be supported in the future)."
-        }
-      },
-      "description": "Pause action."
-    },
     "CallbackInfoTrigger": {
       "type": "object",
       "properties": {
@@ -6868,12 +6854,22 @@
         }
       }
     },
+    "WorkflowRuleActionActionPause": {
+      "type": "object",
+      "properties": {
+        "scope": {
+          "$ref": "#/definitions/v1RuleActionScope",
+          "description": "Indicates what should be paused. Depending on the scope value, the action\nwill pause the entire workflow or a specific activity (additional entities\nmay be supported in the future)."
+        }
+      },
+      "description": "Pause action."
+    },
     "WorkflowRuleSpecActivityStartTrigger": {
       "type": "object",
       "properties": {
         "predicate": {
           "type": "string",
-          "description": "Activity predicate is used to match against workflow data, based on rule trigger type.\nThe following activity attributes are supported as part of the predicate:\n- ActivityType: The type of the activity.\n- ActivityId: The ID of the activity.\n- ActivityAttempt: The number attempts of the activity.\n- BackoffInterval: The current backoff interval of the activity.\n- ActivityStatus: The status of the activity. Can be one of \"scheduled\", \"started\", \"paused\".\n- TaskQueue: The name of the task queue of the activity."
+          "title": "Activity predicate is an SQL-like string filter parameter.\nIt is used to match against workflow data.\nThe following activity attributes are supported as part of the predicate:\n- ActivityType: The type of the activity.\n- ActivityId: The ID of the activity.\n- ActivityAttempt: The number attempts of the activity.\n- BackoffInterval: The current backoff interval of the activity.\n- ActivityStatus: The status of the activity. Can be one of \"Scheduled\", \"Started\", \"Paused\".\n- TaskQueue: The name of the task queue of the activity.\nActivity predicate support the following operators:\n * =, !=, >, >=, <, <=\n * AND, OR, ()\n * BETWEEN ... AND\n   STARTS_WITH"
         }
       },
       "description": "Activity trigger will be triggered when an activity is about to start."
@@ -6910,35 +6906,16 @@
       "type": "object",
       "properties": {
         "spec": {
-          "type": "object",
-          "properties": {
-            "activityStart": {
-              "$ref": "#/definitions/WorkflowRuleSpecActivityStartTrigger"
-            },
-            "visibilityQuery": {
-              "type": "string",
-              "title": "Restricted Visibility query.\nThe following activity attributes are supported as part of the predicate:\n- WorkflowType\n- WorkflowId\n- StartTime\n- ExecutionStatus"
-            },
-            "actions": {
-              "type": "array",
-              "items": {
-                "type": "object",
-                "$ref": "#/definitions/v1Action"
-              },
-              "description": "Action to be taken when the rule is triggered and predicate is matched."
-            },
-            "expirationTime": {
-              "type": "string",
-              "format": "date-time",
-              "description": "Expiration time of the rule. After this time, the rule will be deleted.\nCan be empty if the rule should never expire."
-            }
-          },
-          "description": "The rule specification .",
-          "title": "The rule specification ."
+          "$ref": "#/definitions/v1WorkflowRuleSpec",
+          "description": "The rule specification ."
         },
         "forceScan": {
           "type": "boolean",
           "description": "If true, the rule will be applied immediately via batch job.\nVisibility query in the rule will be used to select the workflows to apply the rule to."
+        },
+        "requestId": {
+          "type": "string",
+          "description": "Used to de-dupe requests. Typically should be UUID."
         }
       }
     },
@@ -8138,14 +8115,6 @@
             "type": "object",
             "$ref": "#/definitions/protobufAny"
           }
-        }
-      }
-    },
-    "v1Action": {
-      "type": "object",
-      "properties": {
-        "pause": {
-          "$ref": "#/definitions/ActionActionPause"
         }
       }
     },
@@ -15266,6 +15235,14 @@
       },
       "description": "WorkflowRule describes a rule that can be applied to any workflow in this namespace."
     },
+    "v1WorkflowRuleAction": {
+      "type": "object",
+      "properties": {
+        "pause": {
+          "$ref": "#/definitions/WorkflowRuleActionActionPause"
+        }
+      }
+    },
     "v1WorkflowRuleSpec": {
       "type": "object",
       "properties": {
@@ -15278,15 +15255,15 @@
         },
         "visibilityQuery": {
           "type": "string",
-          "title": "Restricted Visibility query.\nThe following activity attributes are supported as part of the predicate:\n- WorkflowType\n- WorkflowId\n- StartTime\n- ExecutionStatus"
+          "title": "Restricted Visibility query.\nThe following workflow attributes are supported:\n- WorkflowType\n- WorkflowId\n- StartTime\n- ExecutionStatus"
         },
         "actions": {
           "type": "array",
           "items": {
             "type": "object",
-            "$ref": "#/definitions/v1Action"
+            "$ref": "#/definitions/v1WorkflowRuleAction"
           },
-          "description": "Action to be taken when the rule is triggered and predicate is matched."
+          "description": "WorkflowRuleAction to be taken when the rule is triggered and predicate is matched."
         },
         "expirationTime": {
           "type": "string",
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -6593,9 +6593,6 @@
     }
   },
   "definitions": {
-    "ActionPauseActivityPause": {
-      "type": "object"
-    },
     "CallbackInfoTrigger": {
       "type": "object",
       "properties": {
@@ -6947,14 +6944,8 @@
         }
       }
     },
-    "WorkflowRuleActionActionPause": {
-      "type": "object",
-      "properties": {
-        "activity": {
-          "$ref": "#/definitions/ActionPauseActivityPause"
-        }
-      },
-      "description": "Pause action."
+    "WorkflowRuleActionActionActivityPause": {
+      "type": "object"
     },
     "WorkflowRuleSpecActivityStartingTrigger": {
       "type": "object",
@@ -15350,8 +15341,8 @@
     "v1WorkflowRuleAction": {
       "type": "object",
       "properties": {
-        "pause": {
-          "$ref": "#/definitions/WorkflowRuleActionActionPause"
+        "activityPause": {
+          "$ref": "#/definitions/WorkflowRuleActionActionActivityPause"
         }
       }
     },
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -2263,7 +2263,7 @@
     },
     "/api/v1/namespaces/{namespace}/workflow-rules/{ruleId}": {
       "get": {
-        "summary": "DescribeWorkflowRule return the rule specification for existing rule id.\nIf there is no rule with such id - NO FOUND error will be returned.",
+        "summary": "DescribeWorkflowRule return the rule specification for existing rule id.\nIf there is no rule with such id - NOT FOUND error will be returned.",
         "operationId": "DescribeWorkflowRule2",
         "responses": {
           "200": {
@@ -2288,7 +2288,7 @@
           },
           {
             "name": "ruleId",
-            "description": "ID of the rule to read. Unique within the namespace.",
+            "description": "User-specified ID of the rule to read. Unique within the namespace.",
             "in": "path",
             "required": true,
             "type": "string"
@@ -5692,7 +5692,7 @@
     },
     "/namespaces/{namespace}/workflow-rules/{ruleId}": {
       "get": {
-        "summary": "DescribeWorkflowRule return the rule specification for existing rule id.\nIf there is no rule with such id - NO FOUND error will be returned.",
+        "summary": "DescribeWorkflowRule return the rule specification for existing rule id.\nIf there is no rule with such id - NOT FOUND error will be returned.",
         "operationId": "DescribeWorkflowRule",
         "responses": {
           "200": {
@@ -5717,7 +5717,7 @@
           },
           {
             "name": "ruleId",
-            "description": "ID of the rule to read. Unique within the namespace.",
+            "description": "User-specified ID of the rule to read. Unique within the namespace.",
             "in": "path",
             "required": true,
             "type": "string"
@@ -6869,7 +6869,7 @@
       "properties": {
         "predicate": {
           "type": "string",
-          "title": "Activity predicate is an SQL-like string filter parameter.\nIt is used to match against workflow data.\nThe following activity attributes are supported as part of the predicate:\n- ActivityType: The type of the activity.\n- ActivityId: The ID of the activity.\n- ActivityAttempt: The number attempts of the activity.\n- BackoffInterval: The current backoff interval of the activity.\n- ActivityStatus: The status of the activity. Can be one of \"Scheduled\", \"Started\", \"Paused\".\n- TaskQueue: The name of the task queue of the activity.\nActivity predicate support the following operators:\n * =, !=, >, >=, <, <=\n * AND, OR, ()\n * BETWEEN ... AND\n   STARTS_WITH"
+          "title": "Activity predicate is a SQL-like string filter parameter.\nIt is used to match against workflow data.\nThe following activity attributes are supported as part of the predicate:\n- ActivityType: The type of the activity.\n- ActivityId: The ID of the activity.\n- ActivityAttempt: The number attempts of the activity.\n- BackoffInterval: The current amount of time between scheduled attempts of the activity.\n- ActivityStatus: The status of the activity. Can be one of \"Scheduled\", \"Started\", \"Paused\".\n- TaskQueue: The name of the task queue the workflow specified that the activity should run on.\nActivity predicate support the following operators:\n * =, !=, >, >=, <, <=\n * AND, OR, ()\n * BETWEEN ... AND\n   STARTS_WITH"
         }
       },
       "description": "Activity trigger will be triggered when an activity is about to start."
@@ -6911,7 +6911,7 @@
         },
         "forceScan": {
           "type": "boolean",
-          "description": "If true, the rule will be applied immediately via batch job.\nVisibility query in the rule will be used to select the workflows to apply the rule to."
+          "description": "If true, the rule will be applied to the currently running workflows via batch job.\nIf not set , the rule will only be applied when triggering condition is satisfied.\nvisibility_query in the rule will be used to select the workflows to apply the rule to."
         },
         "requestId": {
           "type": "string",
@@ -9374,7 +9374,7 @@
         },
         "jobId": {
           "type": "string",
-          "description": "Job ID if force-scan flag was provided. Otherwise empty."
+          "description": "Batch Job ID if force-scan flag was provided. Otherwise empty."
         }
       }
     },
@@ -15226,7 +15226,7 @@
         "createTime": {
           "type": "string",
           "format": "date-time",
-          "description": "Rule creation date."
+          "description": "Rule creation time."
         },
         "spec": {
           "$ref": "#/definitions/v1WorkflowRuleSpec",
@@ -15255,7 +15255,7 @@
         },
         "visibilityQuery": {
           "type": "string",
-          "title": "Restricted Visibility query.\nThis query will be applied to the currently running workflow for every triggering event.\nThis query will be applied before trigger predicate is evaluated.\nThe following workflow attributes are supported:\n- WorkflowType\n- WorkflowId\n- StartTime\n- ExecutionStatus"
+          "title": "Restricted Visibility query.\nThis query is used to filter workflows in this namespace to which this rule should apply.\nIt is applied to any running workflow each time a triggering event occurs, before the trigger predicate is evaluated.\nThe following workflow attributes are supported:\n- WorkflowType\n- WorkflowId\n- StartTime\n- ExecutionStatus"
         },
         "actions": {
           "type": "array",
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -2469,51 +2469,6 @@
         ]
       }
     },
-    "/api/v1/namespaces/{namespace}/workflows/{execution.workflowId}/apply-rule": {
-      "post": {
-        "summary": "ApplyWorkflowRule allows to:\n * apply existing rule to a specific workflow execution;\n * apply rule to a specific workflow execution without creating a rule;\nThis is useful for one-off operations.",
-        "operationId": "ApplyWorkflowRule2",
-        "responses": {
-          "200": {
-            "description": "A successful response.",
-            "schema": {
-              "$ref": "#/definitions/v1ApplyWorkflowRuleResponse"
-            }
-          },
-          "default": {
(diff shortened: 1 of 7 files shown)


temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,99 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Rules::V1";
+option csharp_namespace = "Temporalio.Api.Rules.V1";
+
+
+import "google/protobuf/timestamp.proto";
+import "temporal/api/enums/v1/common.proto";
+
+message Action {
```
cretz, line 38: Sorry if this was discussed previously, but I do not believe this should be a top-level concept. What is the use case for this not being a `WorkflowRuleAction` given it's only used in a `WorkflowRule`? Or it can move under there. And then, once it does move under there, there's no longer a need to squat on the generic term/package "rules" for "workflow rules".

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,99 @@
...
+  message ActivityStartTrigger {
+    // Activity predicate is used to match against workflow data, based on rule trigger type.
+    // The following activity attributes are supported as part of the predicate:
+    // - ActivityType: The type of the activity.
+    // - ActivityId: The ID of the activity.
+    // - ActivityAttempt: The number attempts of the activity.
+    // - BackoffInterval: The current backoff interval of the activity.
+    // - ActivityStatus: The status of the activity. Can be one of "scheduled", "started", "paused".
+    // - TaskQueue: The name of the task queue of the activity.
+    string predicate = 1;
```
cretz, line 67: Is this the same format as https://docs.temporal.io/list-filter? May need to be noted here (same as visibility query, though it's used in several spots in proto and is kinda known). Also please make a note to update those docs as they make people assume that search attributes are available for all uses of list filtering.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,99 @@
...
+  oneof trigger {
+    ActivityStartTrigger activity_start = 2;
+  }
+
+  // Restricted Visibility query.
+  // The following activity attributes are supported as part of the predicate:
+  // - WorkflowType
+  // - WorkflowId
+  // - StartTime
+  // - ExecutionStatus
```
cretz, line 81: This mentions "activity" attributes and "predicate", is there a copy/paste issue here? Also to help me understand, why can't I use all search attributes at https://docs.temporal.io/search-attribute (including my own) like I can any other visibility query in our API?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2195,3 +2196,52 @@ message GetDeploymentReachabilityResponse {
     // was actually calculated.
     google.protobuf.Timestamp last_update_time = 3;
 }
+
+message CreateWorkflowRuleRequest {
```
cretz, line 2200: I think there should be some kind of idempotency key here (e.g. `request_id`). Our SDK clients automatically retry what seem like transient failures, so we make sure to make all calls idempotent, but at first glance it doesn't seem a request with `force_scan` is safe for retrying, same for delete.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1132,4 +1132,48 @@ service WorkflowService {
             }
         };
     }
+
+    // Create a new workflow rule.
+    // If the rule with such ID already exist this call will fail.
```
cretz, line 1137: How do I update a rule? If we may support that in the future, can we change this API binding from `/namespaces/{namespace}/workflow-rules/{spec.id}` to `/namespaces/{namespace}/workflow-rules`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2195,3 +2196,52 @@ message GetDeploymentReachabilityResponse {
...
+message DeleteWorkflowRuleResponse {
+}
+
+message ListWorkflowRulesRequest {
+    string namespace = 1;
+}
+
+message ListWorkflowRulesResponse {
+    repeated temporal.api.rules.v1.WorkflowRule rules = 1;
+    bytes next_page_token = 2;
```
cretz, line 2246: Why does this return a next page token if the request does not accept the token? Also, for the PR in general, can you confirm that at least most of the implementation is done/tested?

temporal/api/rules/v1/message.proto
ychebotarev: It was discussed, at design review. Action is "generic" enough to be a top-level concept. 
For use cases it will be better to talk to Roey, I can say that he think we can reuse them with Nexus/Schedulers/etc.

Rules in its current implementation are not, so they will be scoped to workflow - WorkflowRules.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: the rule ID is a natural idempotency key. Rule ID is unique per namespace, and if it is present code will not create new/start new batch. 

Why you think we may need another one?

temporal/api/rules/v1/message.proto
ychebotarev: I add more verbose help, but without direct link.

temporal/api/rules/v1/message.proto
ychebotarev: >is there a copy/paste issue here?
Nope. This is the activity trigger, which is triggering when activity is about to start, and that checks the activity attributes to make an action.

> Also to help me understand, why can't I use all search attributes
This is activity trigger, and it passed based on activity properties mentioned in the description. No need to use anything else.

As for "why can't I use all search attributes":
* not needed for activity trigger specifically. 
* still can think about using them in visibility_query (few lines below), but...
* not needed for MVP
* (I'n not 100% sure) search attributes are not always available.  We can load them of cause from DB, but that is possible one extra DB load for every task processing operation. We need to be sure that benefit outweighs this cost.

temporal/api/workflowservice/v1/service.proto
ychebotarev: We don't update workflow rules, and I don't think we will (or more like I don't foresee any real reason why would we do that), but I see where you coming from, and in case we add "update" for whatever reason I will change the binding.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: >Why does this return a next page token if the request does not accept the token? 
ah right, added.
>Also, for the PR in general, can you confirm that at least most of the implementation is done/tested?
Implementation for this API? Yes, is is done and tested, including _some_ func tests. But in general there are more work. It just that work (some subtle cases, more func tests) in not related to the API.

temporal/api/rules/v1/message.proto
cretz: I don't believe a pause action, the only action that is supported here, should be generic. A workflow pause is not the same as an activity pause and neither are the same as Nexus operation pause, and we may want different fields for each. I think rule actions need to be specific to their rule type. I cannot think of a single benefit of trying to have a shared data type here when it has to go into a specific rule data type anyways.

temporal/api/rules/v1/message.proto
cretz: > Nope. This is the activity trigger, which is triggering when activity is about to start, and that checks the activity attributes to make an action.

This is talking about "activity attributes" in a workflow "visibility query" which is confusing

> still can think about using them in visibility_query (few lines below), but...

I am very specifically commenting on visibility query which is this doc

temporal/api/workflowservice/v1/request_response.proto
cretz: > Why you think we may need another one?

Because this is not idempotent. Our client automatically retries requests that get a transient failure. What if that transient failure happened after a `force_scan=true` call succeeded from a server POV (e.g. client-side network failure)? That would make a second batch job which I don't think we want.

temporal/api/workflowservice/v1/service.proto
cretz: :+1: So we should change _this_ binding for "create" to be `POST /namespaces/{namespace}/workflow-rules`, not `POST /namespaces/{namespace}/workflow-rules/{spec.id}`

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: We have rule id. it is stored as a part of the rule. If request with the same rule id comes - it is rejected.
You are asking to add 'request_id`. In order to check for it it should also be stored as a part of the rule. If request with the same `request id` comes - it is rejected.

So behavior is the same, in both cases we reject the request if id is the same.
Why you think `rule id` is not enough?

cretz: > If request with the same rule id comes - it is rejected.

Ah, that it is rejected is _technically_ good enough to prevent double-run, but that is still not idempotent (changes outcome). We have traditionally tried _not_ to fail implicit retries but rather just dedupe them. In the scenario I laid out, a user that did nothing wrong will now get an error because the implicit-retry-on-transient-error is treated as if _they_ retried it. This is why we usually have a request ID for these things.

temporal/api/rules/v1/message.proto
ychebotarev: ah, I see,  that is wrong comment.
should be  `The following workflow attributes are supported`

temporal/api/rules/v1/message.proto
ychebotarev: lets talk offline. But we specifically design it with workflow/nexus pause in mind.
To add workflow pause one should create the same action, but with different `scope`
Same for Nexus.
For other possible actions it will either be the same, or there will be `oneof`

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,105 @@
...
+  // Activity trigger will be triggered when an activity is about to start.
+  message ActivityStartTrigger {
+    // Activity predicate is an SQL-like string filter parameter.
+    // It is used to match against workflow data.
+    // The following activity attributes are supported as part of the predicate:
+    // - ActivityType: The type of the activity.
+    // - ActivityId: The ID of the activity.
+    // - ActivityAttempt: The number attempts of the activity.
+    // - BackoffInterval: The current backoff interval of the activity.
+    // - ActivityStatus: The status of the activity. Can be one of "scheduled", "started", "paused".
```
alexshtin, line 66: `ExecutionStatus` uses capitalized values. Even if this is case insensitive, they need to be documented as "Scheduled", "Started", and "Paused".

temporal/api/workflowservice/v1/request_response.proto
alexshtin: Please link server PR in this PR description.

temporal/api/workflowservice/v1/request_response.proto
alexshtin: I think you should add `request_id` here. Improve DX and sacrifice some storage.

As for delete, I believe it is idempotent and doesn't return error if rule is not found.

temporal/api/rules/v1/message.proto
ychebotarev: ok

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ok, talked offline. request_id will be added.

temporal/api/rules/v1/message.proto
ychebotarev: talked offline, and start a discussion thread.

cretz (approved):

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+    ActionPause pause = 1;
+  }
+}
+
+message WorkflowRuleSpec {
+  // The id of the new workflow rule. Must be unique within the namespace.
+  string id = 1;
+
+  // Activity trigger will be triggered when an activity is about to start.
+  message ActivityStartTrigger {
```
Sushisource, line 59:
```suggestion
  message ActivityStartingTrigger {
```

IMO this would be a slightly more accurate name since the activity is _about_ to start

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+    ActivityStartTrigger activity_start = 2;
+  }
+
+  // Restricted Visibility query.
+  // The following workflow attributes are supported:
+  // - WorkflowType
+  // - WorkflowId
+  // - StartTime
+  // - ExecutionStatus
+  string visibility_query = 3;
```
Sushisource, line 89: I assume this is the query to use to determine which workflows to apply this rule to, but the comment doesn't actually state that. Would be good to add that

openapi/openapiv3.yaml
```diff
@@ -1951,6 +1951,131 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
+    post:
+      tags:
+        - WorkflowService
+      description: |-
+        Create a new workflow rule.
+         If the rule with such ID already exist this call will fail.
```
drewhoskins, line 1989: From `CreateWorkflowRuleRequest`, it looks like we have a `RequestID`.  I'm guessing that it's an idempotency key such that this statement is not true because we'll return success?  
If so, can you suggest using it here?
If not, can we reconsider?  I'd like to support client-side retries after network timeouts, for example.

openapi/openapiv3.yaml
```diff
@@ -1951,6 +1951,131 @@ paths:
...
+          description: Default error response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
+    post:
+      tags:
+        - WorkflowService
+      description: |-
+        Create a new workflow rule.
```
drewhoskins, line 1988: I'd also want to know whether it will go into effect immediately upon creation.  (as opposed to being created in a disabled state)
I think we should allow enabling/disabling these rules.  Imagine in the UI listing rules.  Somebody turns a rule off, but then wants to turn it back on without re-authoring the query/filters.

alexshtin (approved):
Please address @drewhoskins and @Sushisource comments before merging.

openapi/openapiv3.yaml
```diff
@@ -1951,6 +1951,131 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/namespaces/{namespace}/workflow-rules:
+    get:
+      tags:
+        - WorkflowService
+      description: Return all namespace rules
```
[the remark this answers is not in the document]
alexshtin: image

temporal/api/rules/v1/message.proto
ychebotarev: Ok, renamed.

openapi/openapiv3.yaml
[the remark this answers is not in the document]
ychebotarev: Not sure what I'm missing here. This function literally return all namespace rules. I don't understand how explaining what function does is unhelpful. Example:
```
    // List all schedules in a namespace.
    rpc ListSchedules (ListSchedulesRequest) returns (ListSchedulesResponse) {
```

I think we should do that in the documentation. This is the best place, since conceptions often involve explaining the reasoning, use cases, logic, examples, etc.

API proto definitions is not the best place to explain Temporal conceptions, be it Rules or Activities or Workflow or Scheduler. 

API Proto definitions for RPC calls are expected to be precise, machine-readable contracts and method signatures. They focus on "what", not on "why".

openapi/openapiv3.yaml
ychebotarev: I add more comments. 

>I think we should allow enabling/disabling these rules.
...
>but then wants to turn it back on 

This imply that we add few more API, specifically to enable/disable existing rules
We can add some API later if needed/requested.

temporal/api/rules/v1/message.proto
Sushisource: OK, this helped because it definitely wasn't doing what I thought it was at first read. I thought it would be scanning over all workflows. I think it could still be a bit more clear.

I'll write what I think this is doing and you can use it if I'm right or correct me:

```
This query is used to filter workflows in this namespace to which this rule should apply. 
It is applied to any running workflow each time a triggering event occurs, before the trigger predicate is evaluated.
```

However, writing that leaves me with the question: Why can `ExecutionStatus` be specified in the query, if rules are only ever applied to running workflows? Seemingly specifying any value for it other than `Running` means the query will never apply to anything.

openapi/openapiv3.yaml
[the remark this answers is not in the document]
alexshtin: I believe @drewhoskins comment is mostly about "Rule" it self. You introduce new entity with a very broad name without explaining what are those rules about. I am not sure where is the best place to describe it though. Probably where `Rule` is defined. Not here.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+
+  // WorkflowRuleAction to be taken when the rule is triggered and predicate is matched.
+  repeated WorkflowRuleAction actions = 4;
+
+  // Expiration time of the rule. After this time, the rule will be deleted.
```
drewhoskins, line 96: I can't remember a slam-dunk use case for this, and it seems dangerous to use properly as it requires people to predict the future.  Should we remove it for now until we get user requests?
And expiring a rule won't unpause paused activities, right?

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+  // Can be empty if the rule should never expire.
+  google.protobuf.Timestamp expiration_time = 5;
+}
+
+// WorkflowRule describes a rule that can be applied to any workflow in this namespace.
+message WorkflowRule {
+  // Rule creation date.
```
drewhoskins, line 103: nit: not just a date

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2195,3 +2196,56 @@ message GetDeploymentReachabilityResponse {
...
 }
+
+message CreateWorkflowRuleRequest {
+    string namespace = 1;
+
+    // The rule specification .
+    temporal.api.rules.v1.WorkflowRuleSpec spec = 2;
+
+    // If true, the rule will be applied immediately via batch job.
+    // Visibility query in the rule will be used to select the workflows to apply the rule to.
```
drewhoskins, line 2207: nit: `visibility_query` in the rule

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2195,3 +2196,56 @@ message GetDeploymentReachabilityResponse {
...
     google.protobuf.Timestamp last_update_time = 3;
 }
+
+message CreateWorkflowRuleRequest {
+    string namespace = 1;
+
+    // The rule specification .
+    temporal.api.rules.v1.WorkflowRuleSpec spec = 2;
+
+    // If true, the rule will be applied immediately via batch job.
```
drewhoskins, line 2206: Needs to say what happens if it's false.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2195,3 +2196,56 @@ message GetDeploymentReachabilityResponse {
...
+
+    // Used to de-dupe requests. Typically should be UUID.
+    string request_id = 4;
+}
+
+message CreateWorkflowRuleResponse {
+    // Created rule.
+    temporal.api.rules.v1.WorkflowRule  rule = 1;
+
+    // Job ID if force-scan flag was provided. Otherwise empty.
```
drewhoskins, line 2218: nit: Say "Batch Job ID".

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2195,3 +2196,56 @@ message GetDeploymentReachabilityResponse {
...
+
+message CreateWorkflowRuleRequest {
+    string namespace = 1;
+
+    // The rule specification .
+    temporal.api.rules.v1.WorkflowRuleSpec spec = 2;
+
+    // If true, the rule will be applied immediately via batch job.
+    // Visibility query in the rule will be used to select the workflows to apply the rule to.
+    bool force_scan = 3;
```
drewhoskins, line 2208: What does a user do if they _only_ want to do a batch job to remediate existing workflows, but don't want to create an ongoing rule for new workflows?

drewhoskins (changes requested):

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+    string predicate = 1;
+  }
+
+  // Specifies how the rule should be triggered and evaluated.
+  // Currently, only "activity start" type is supported.
+  oneof trigger {
+    ActivityStartingTrigger activity_start = 2;
+  }
+
+  // Restricted Visibility query.
```
drewhoskins, line 83: Comment for batch is:
> defines the the group of workflows to apply the batch operation to.

Try something like that.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: out of scope for this PR.
But plan is - there will be "ApplyRule" batch operatoin

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: renamed

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: added

openapi/openapiv3.yaml
```diff
@@ -1951,6 +1951,134 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
+  /api/v1/namespaces/{namespace}/workflow-rules/{ruleId}:
+    get:
+      tags:
+        - WorkflowService
+      description: |-
+        DescribeWorkflowRule return the rule specification for existing rule id.
+         If there is no rule with such id - NO FOUND error will be returned.
```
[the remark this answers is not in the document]
ychebotarev: renamed.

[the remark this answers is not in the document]
ychebotarev: >And please note my reply to the earlier thread.

I don't know what comment you are referencing to here.

openapi/openapiv3.yaml
```diff
@@ -1951,6 +1951,134 @@ paths:
...
+      operationId: DescribeWorkflowRule
+      parameters:
+        - name: namespace
+          in: path
+          required: true
+          schema:
+            type: string
+        - name: ruleId
+          in: path
+          description: ID of the rule to read. Unique within the namespace.
```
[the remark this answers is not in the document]
ychebotarev: renamed

openapi/openapiv3.yaml
```diff
@@ -12261,6 +12564,89 @@ components:
...
+             will pause the entire workflow or a specific activity (additional entities
+             may be supported in the future).
+          format: enum
+      description: Pause action.
+    WorkflowRuleSpec:
+      type: object
+      properties:
+        id:
+          type: string
+          description: The id of the new workflow rule. Must be unique within the namespace.
```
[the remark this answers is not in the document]
ychebotarev: It is a string, so it most likely will be GUID. I don't think adding 'human-readable' is needed here.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+
+  // Activity trigger will be triggered when an activity is about to start.
+  message ActivityStartingTrigger {
+    // Activity predicate is an SQL-like string filter parameter.
+    // It is used to match against workflow data.
+    // The following activity attributes are supported as part of the predicate:
+    // - ActivityType: The type of the activity.
+    // - ActivityId: The ID of the activity.
+    // - ActivityAttempt: The number attempts of the activity.
+    // - BackoffInterval: The current backoff interval of the activity.
```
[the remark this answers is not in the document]
ychebotarev: renamed.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+    // - ActivityStatus: The status of the activity. Can be one of "Scheduled", "Started", "Paused".
+    // - TaskQueue: The name of the task queue of the activity.
```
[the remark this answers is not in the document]
ychebotarev: renamed.

temporal/api/rules/v1/message.proto
ychebotarev: I don't see any dangerous here.

temporal/api/rules/v1/message.proto
ychebotarev: renamed.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+message WorkflowRuleSpec {
+  // The id of the new workflow rule. Must be unique within the namespace.
+  string id = 1;
+
+  // Activity trigger will be triggered when an activity is about to start.
+  message ActivityStartingTrigger {
+    // Activity predicate is an SQL-like string filter parameter.
+    // It is used to match against workflow data.
+    // The following activity attributes are supported as part of the predicate:
+    // - ActivityType: The type of the activity.
```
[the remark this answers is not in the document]
ychebotarev: Technically you may have 2 different workers with different activities, but with the same name. 
You can attach those workers to the different task queues.
And when you call ExecuteActivity you specify tasks queue.
So you can route activities to the specific workers, and have activities with the same name, but different implementation

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+    // - ActivityId: The ID of the activity.
+    // - ActivityAttempt: The number attempts of the activity.
```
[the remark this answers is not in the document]
ychebotarev: We currently only using Attempts, I don't think we will rename it anytime soon. We may add support for Failures once we introduce them

temporal/api/rules/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: >but I hope it wouldn't mean the user would have to specify the task queue when they name it here...

They don't need to. But they can. So in this hypothetical situation, to block only activities on a certain task queue, users will need to specify TaskQueue name in a predicate.

temporal/api/rules/v1/message.proto
ychebotarev: I'm not sure I got this comment - what are you referencing to as "batch" in this file?

temporal/api/enums/v1/common.proto
```diff
@@ -104,3 +104,12 @@ enum NexusOperationCancellationState {
     // Cancellation request is blocked (eg: by circuit breaker).
     NEXUS_OPERATION_CANCELLATION_STATE_BLOCKED = 6;
 }
+
+enum RuleActionScope {
```
cretz, line 108: I forgot to mention, this should be called `WorkflowRuleActionScope` in line with changing the thing it applies to from `Action` to `WorkflowRuleAction`.

temporal/api/rules/v1/message.proto
ychebotarev: I updated the comments. I will discus ExecutionStatus offline (it doesn't affect API)

temporal/api/enums/v1/common.proto
ychebotarev: renamed

Sushisource (approved):

openapi/openapiv3.yaml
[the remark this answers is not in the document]
drewhoskins-temporal: Agreed, they don't have to make them human-readable, but we can suggest it.  Because we want to list them in the UI and CLI.

temporal/api/rules/v1/message.proto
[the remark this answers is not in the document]
drewhoskins-temporal: Makes sense.    WDYT of my initial suggestion?

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+  // Specifies how the rule should be triggered and evaluated.
+  // Currently, only "activity start" type is supported.
+  oneof trigger {
+    ActivityStartingTrigger activity_start = 2;
+  }
+
+  // Restricted Visibility query.
+  // This query is used to filter workflows in this namespace to which this rule should apply.
+  // It is applied to any running workflow each time a triggering event occurs, before the trigger predicate is evaluated.
+  // The following workflow attributes are supported:
```
drewhoskins-temporal, line 86: This list doesn't satisfy all our requirements.  We need to support other tags.  I can easily see scenarios for:
* The version tag on the worker
* The worker version it's running on.
* User specified metadata

Curious why the list is smaller and if we have line of sight to making it bigger.

temporal/api/rules/v1/message.proto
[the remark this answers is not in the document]
drewhoskins-temporal: 👍, sounds good.  FYI @THardy98

temporal/api/rules/v1/message.proto
drewhoskins-temporal: Danger is if the rule disappears and the system reverts to the bad state automatically.

There are many automatic pop operations users might want:
* expiration time
* after we detect failure is resolved
* when there aren't any affected workflows left
* Custom logic.

The others seem strictly better.  Expiration time is easiest to build, but it's the most dangerous option.  I'd rather leave it out until we think through the problem more holistically or until we get a user request.

temporal/api/workflowservice/v1/request_response.proto
drewhoskins-temporal: I don't understand.  
I thought the CLI experience was gonna be, roughly,
`temporal worktivity pause [--create-rule|--scan|both]` 
and my question was how does this proto support the case where they only pass `--scan` ?

openapi/openapiv2.json
```diff
@@ -2210,22 +2210,27 @@
...
         ]
-      }
-    },
-    "/api/v1/namespaces/{namespace}/workflow-rules/{ruleId}": {
-      "get": {
-        "summary": "DescribeWorkflowRule return the rule specification for existing rule id.\nIf there is no rule with such id - NO FOUND error will be returned.",
-        "operationId": "DescribeWorkflowRule2",
+      },
+      "post": {
+        "summary": "Create a new workflow rule. The rules are used to control the workflow execution.\nThe rule will be applied to all running and new workflows in the namespace.\nIf the rule with such ID already exist this call will fail\nNote: the rules are part of namespace configuration and will be stored in the namespace config.\nNamespace config is eventually consistent.",
```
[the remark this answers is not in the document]
ychebotarev: add more detailed comment.
May be return the whole list of rules?

temporal/api/rules/v1/message.proto
drewhoskins-temporal: List of documented searc attributes: https://docs.temporal.io/search-attribute

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1132,4 +1132,67 @@ service WorkflowService {
...
+                get: "/api/v1/namespaces/{namespace}/workflow-rules"
+            }
+        };
+    }
+
+    // ApplyWorkflowRule allows to:
+    //  * apply existing rule to a specific workflow execution;
+    //  * apply rule to a specific workflow execution without creating a rule;
+    // This is useful for one-off operations.
+    rpc ApplyWorkflowRule (ApplyWorkflowRuleRequest) returns (ApplyWorkflowRuleResponse) {
```
cretz, line 1187: Could also consider naming this `TriggerWorkflowRule` akin to our concept of manually triggering a schedule, but apply is fine too

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2201,3 +2202,73 @@ message GetDeploymentReachabilityResponse {
...
+    bytes next_page_token = 2;
+}
+
+message ApplyWorkflowRuleRequest {
+    string namespace = 1;
+
+    // Execution info of the workflow which scheduled this activity
+    temporal.api.common.v1.WorkflowExecution execution = 2;
+
+    // either provide id of existing rule, or rule specification
```
cretz, line 2266: Inconsistent with whether you capitalize comments

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2201,3 +2202,73 @@ message GetDeploymentReachabilityResponse {
...
+message ListWorkflowRulesResponse {
+    repeated temporal.api.rules.v1.WorkflowRule rules = 1;
+    bytes next_page_token = 2;
+}
+
+message ApplyWorkflowRuleRequest {
+    string namespace = 1;
+
+    // Execution info of the workflow which scheduled this activity
+    temporal.api.common.v1.WorkflowExecution execution = 2;
```
cretz, line 2264: It's about 50/50 whether we call this workflow_execution or just execution. I'm fine with either, I wish we were more consistent in the past.

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+import "google/protobuf/timestamp.proto";
+import "temporal/api/enums/v1/common.proto";
+
+message WorkflowRuleAction {
+  // Pause action.
+  message ActionPause {
+    // Indicates what should be paused. Depending on the scope value, the action
+    // will pause the entire workflow or a specific activity (additional entities
+    // may be supported in the future).
+    temporal.api.enums.v1.WorkflowRuleActionScope scope = 1;
```
cretz, line 44: I still believe that "pause activity" and "pause workflow" are distinct action variants that may, in the future, get differing options. There is no benefit to combining these into one `ActionPause` message. Why make a user set an enum to differentiate instead of just a different `variant` below?

temporal/api/workflowservice/v1/service.proto
cretz, line 1187: Can you confirm that this RPC does not return until the applying is finished, and therefore if there is a problem applying the rule it will return an error?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2201,3 +2202,73 @@ message GetDeploymentReachabilityResponse {
...
+
+    // either provide id of existing rule, or rule specification
+    oneof rule {
+        string id = 4;
+        temporal.api.rules.v1.WorkflowRuleSpec spec = 6;
```
cretz, line 2269: If I provide this spec inline, are there any changes into what is allowed? For instance, do we disallow the `trigger` since it makes no sense? What about expiration and visibility_query?

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,108 @@
...
+    //  * =, !=, >, >=, <, <=
+    //  * AND, OR, ()
+    //  * BETWEEN ... AND
+    //    STARTS_WITH
+    string predicate = 1;
+  }
+
+  // Specifies how the rule should be triggered and evaluated.
+  // Currently, only "activity start" type is supported.
+  oneof trigger {
```
cretz, line 79: Coming back to this, is it by intention that a rule can only have one triggering condition? So if I want a rule to trigger on two distinct things I have to make separate rules that are exact copies of each other except for this one field? Are we sure we are ok with that limitation?

cretz (commented):
Reviewed the new `ApplyWorkflowRule by request, also a couple of other unrelated comments

temporal/api/rules/v1/message.proto
ychebotarev: Difference is - are those variants linear or hierarchical? 
From the perspective we chose its hierarchical - "pause" is an action, and "activity" or "workflow" is a subject of that action. What action we perform? "Pause". What we apply this action to? "Activity".

I do see a slight benefit of being more structural. And don't see much benefit of being "linear". Also if (if! :) ) the number of action increased linear "ActionSubject" enum will become big and hard to navigate

temporal/api/workflowservice/v1/service.proto
ychebotarev: Condition will be evaluated immediately. Then, is action is sync (all actions we have right now in mind) it will also be applied instantly.
I can imagine that there are actions that are "async".  But even in that case it will be something "action was performed, job was scheduled".

temporal/api/rules/v1/message.proto
cretz: > Difference is - are those variants linear or hierarchical? From the perspective we chose its hierarchical

To me, activity and workflow pause are _completely_ unrelated except they share the English word "pause". There is no benefit to users by forcing them to be the same proto message.

Say we want to support the option of whether to "interrupt" activities or not on activity pause. Where will you put that option? That has no workflow pause equivalent. Will you clutter workflow pause with properties for activity pause and just ask user to ignore them? Might as well be separate, they are completely separate things.

temporal/api/rules/v1/message.proto
ychebotarev: yes. I want to avoid "pause the activity if workflow task failed" rules. There is only one trigger, but there can be multiple actions.

temporal/api/rules/v1/message.proto
ychebotarev: this is the initial list we came with. The beauty of this - adding support of new fieds mean just change the comments (or we may add "see more in the documentation')

temporal/api/rules/v1/message.proto
ychebotarev: discussed offline. It is requested, expetded, "off" by default - so if users add it they should know what they are doing. What I agree with - we need a proper auditing mechanisc (like a special log/visibility where users can go and see "rule is expired and was deleted" message)

temporal/api/rules/v1/message.proto
ychebotarev: I was hoping to avoid it until workflow pause... Scenario you described will come with worklfow pause, expected parameter - should we propagate workflow pause to child workflows?
This is how I was planning to implement it (see commented section with WorkflowPause definition)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: fixed

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: trigger and expiration doesn't make sense in this case

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: actually I added new APIs to this PR.
If users want to do a batch job to remediate existing workflows - they call  batch with TriggerWorkflowRule operation (which will call TriggerWorkflowRule API).
If the users want to apply existing rule and rule spec to a signle workflow - they can call TriggerWorkflowRule.

temporal/api/workflowservice/v1/request_response.proto
cretz: Right, so is it allowed or will it fail if you try to set them? Can you at least document whether it's just ignored vs disallowed?

temporal/api/rules/v1/message.proto
cretz: :+1: I think you could just as easily treat these as separate actions instead of the same action with separate scopes. They are not the same action and the only thing they have in common is the English word "pause". (also let's make sure not to merge to `master` with commented out code)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: trigger has a predicate, which will be evaluated, so it is not ignored. visibility_query also will be evaluated

openapi/openapiv3.yaml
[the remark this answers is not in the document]
ychebotarev: I add more comments

temporal/api/rules/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+  message ActionPause {
+
+    message ActivityPause {
+    }
+
+    // Indicates what should be paused. Depending on the scope value, the action
+    // will pause the entire workflow or a specific activity (additional entities
+    // may be supported in the future).
+    oneof scope  {
+        ActivityPause activity = 2;
```
cretz, line 48: Should this be index 1?

temporal/api/rules/v1/message.proto
cretz: If we have committee decision on this point, sure. I am not sure this specific point was visited when the whole was. Is it possible to get them to confirm on this point specifically?

Specifically, can this committee weigh in on which is preferred, this:

```proto
message WorkflowRuleAction {
  message ActionPause {
    message ActivityPause {
    }

    message WorkflowPause {
        bool pause_child_workflow = 1;
    }

    oneof scope  {
        ActivityPause activity = 1;
        ActivityPause activity = 2;
    }
  }

  oneof variant  {
    ActionPause pause = 1;
  }
}
```
or this:
```proto
message WorkflowRuleAction {
  message ActivityPause {
  }

  message WorkflowPause {
    bool pause_child_workflow = 1;
  }

  oneof variant  {
    ActivityPause activity_pause = 1;
    WorkflowPause workflow_pause = 2;
  }
}
```

understanding of course the former means future actions are hierarchical and latter is flat. If we can get confirmation on this point specifically, then I can approve even though I disagree.

temporal/api/rules/v1/message.proto
cretz, line 48: Inconsistent 4-space indent

temporal/api/rules/v1/message.proto
ychebotarev: I already said we discuss specifically that - hierarchical vs flat.

>If we can get confirmation on this point specifically
you welcome to do that. I don't feel I need more confirmations that I already have.

mfateev: I like flat. I don't see any real benefit from hierarchical structure here.

Sushisource: +1 - On a practical level, very-nested protos like this are a pain to work with. If there were fields inside `ActionPause` that were shared, then there'd be some point, but as of now it's just more layers that aren't providing anything.

ychebotarev: Real benefit from hierarchical structure is the same benefit hierarchy and structure provides in general. Which is "better understanding the code, improved readability, easier to find information, data sharing, ...", etc. After some point "flat" will become a mess.

>. If there were fields inside ActionPause that were shared

We should look forward, otherwise there was no point in generalization at all. 
Shared fields are likely to appear in other actions. In fact they will appear in the upcoming EmitMetric action (which is coming next, with workflow pause). 
Same is true for other next potential action - adding to visibility.
With "flat" approach, we might end up duplicating the same fields across multiple structures.

That said - I appreciate the discussion on this topic. But I suggest we reserve this kind of debates for our design reviews and "design PRs", where we can fully explore the surface.  
In "implementation" PR it is better to focus our PR discussions on implementation details. This will make process more efficient.

ychebotarev: @mfateev - I provided a reasons to be hierarchical

ychebotarev: discussed offline. This will be detrimental, but ok.

PR #561, opened 2025-03-18
Merge initial protos for priority

dnr:
**What changed?**
Add initial protos for priority, enough to attach simple priority metadata to workflows and activities.

See #513 for the full future plans, only the first field (priority_key) is included so far.

**Why?**
Control task ordering.

**Server PR**
Implementation on `priority` branch: https://github.com/temporalio/temporal/compare/main...priority


stephanos (approved):

temporal/api/common/v1/message.proto
```diff
@@ -247,3 +247,40 @@ message Link {
         BatchJob batch_job = 2;
     }
 }
+
+// Priority contains metadata that controls relative ordering of task processing
+// when tasks are backlogged in a queue. Initially, Priority will be used in
+// matching (workflow and activity) task queues. Later it may be used in history
+// task queues and in rate limiting decisions.
```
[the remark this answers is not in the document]
dnr: Ok.. I don't want to get into too much server architecture here, but let me try again...

temporal/api/common/v1/message.proto
```diff
@@ -247,3 +247,41 @@ message Link {
...
+    // correspond to higher priorities (tasks run sooner). In general, tasks in
+    // a queue should be processed in close to priority order, although small
+    // deviations are possible.
+    //
+    // The maximum priority value (minimum priority) is determined by server
+    // configuration, and defaults to 5.
+    //
+    // If priority is not present (or zero), then the effective priority will be
+    // the default priority, which is is calculated by (min+max)/2. With the
+    // default max of 5, and min of 1, that comes out to 3.
```
cretz, line 285: > If priority is not present (or zero), then the effective priority will be the default priority, which is is calculated by (min+max)/2. With the default max of 5, and min of 1, that comes out to 3.

This is a bit in conflict (or at least confusing) with the comment above the message which states:

> For all fields, the field not present or equal to zero/empty string means to inherit the value from the calling workflow, or if there is no calling workflow, then use the default value.

Should the first quoted statement change "If priority is not present (or zero)" to "If priority is not present (or zero) and there is no calling workflow"?

cretz (approved):

temporal/api/common/v1/message.proto
dnr: I see the confusion: the above comment is talking about how to merge these messages if there's one on the workflow and also an activity call (or child wf), the one here is talking from the perspective of a post-merged message and saying what happens if it's still zero. I'll try to reword.

PR #563, opened 2025-03-19
Add request ID reference in links

rodrigozhou:

**What changed?**
Add request ID reference to `Link`.

**Why?**
Request ID can be used to associate links to history event ID.

**Breaking changes**

**Server PR**


temporal/api/common/v1/message.proto
```diff
@@ -224,6 +224,10 @@ message Link {
             temporal.api.enums.v1.EventType event_type = 2;
         }
 
+        message RequestIdReference {
```
cretz, line 227: Can you add the `event_type` in here?

temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +137,9 @@ message WorkflowExecutionExtendedInfo {
 
     // Original workflow start time.
     google.protobuf.Timestamp original_start_time = 5;
+
+    // Request ID information (eg: history event information associated with the request ID).
+    map<string, RequestIdInfo> request_id_infos = 6;
```
cretz, line 142: Will this include the request ID for the start event?

temporal/api/common/v1/message.proto
rodrigozhou: I could, but it won't be populated anyway because the SDK won't know if the workflow started or not.

temporal/api/workflow/v1/message.proto
rodrigozhou: Yes, it does.

rodrigozhou: At this moment, only the request ID in the `StartWorkflowExecution` request is stored in this map. So, for signal with start, if it starts a workflow, then the request ID would be there as the starter of the workflow.
I think we could add the other request IDs as well if needed.

temporal/api/common/v1/message.proto
rodrigozhou: Added event_type as it will be useful.

temporal/api/workflow/v1/message.proto
cretz: May want to clarify in docs that this is not all request ID infos

cretz (approved):

temporal/api/common/v1/message.proto
```diff
@@ -224,6 +224,11 @@ message Link {
...
```
bergundy, line 227: Still prefer just putting the request ID in `EventReference` but not blocking this PR since @cretz seems to be more opinionated here.

Could you please document this message and `EventReference` to highlight the differences?

temporal/api/workflow/v1/message.proto
```diff
@@ -139,6 +140,9 @@ message WorkflowExecutionExtendedInfo {
 
     // Reset Run ID points to the new run when this execution is reset. If the execution is reset multiple times, it points to the latest run.
     string reset_run_id = 6;
+
+    // Request ID information (eg: history event information associated with the request ID).
```
bergundy, line 144:
```suggestion
    // Request ID information (eg: history event information associated with the request ID).
    // Some request IDs may not appear here if they correspond to (unflushed) buffered events.
```

bergundy (approved):
Approved with a couple of small comments.

temporal/api/workflow/v1/message.proto
rodrigozhou: Oh, actually, all request IDs would be present. If the event is buffered, it will contain an invalid event ID. We could remove them when building the DescribeWorkflow response though.

bergundy: We should, that would be better than providing invalid data back. Alternatively, we can have a status field that says the request is still buffered so consumers don't wonder why there requests aren't showing up.

PR #564, opened 2025-04-01
Add history events for NexusCancelRequest[Completed|Failed]

pdoerner:
**What changed?**
Added two new history events: `NexusOperationCancelRequestCompleted` and `NexusOperationCancelRequestFailed`

**Why?**
To support Nexus operation cancellations that should wait only until the cancellation has been delivered to the handler.

**Breaking changes**
No

**Server PR**


temporal/api/history/v1/message.proto
```diff
@@ -1038,6 +1056,8 @@ message HistoryEvent {
         NexusOperationTimedOutEventAttributes nexus_operation_timed_out_event_attributes = 58;
         NexusOperationCancelRequestedEventAttributes nexus_operation_cancel_requested_event_attributes = 59;
         WorkflowExecutionOptionsUpdatedEventAttributes workflow_execution_options_updated_event_attributes = 60;
+        NexusOperationCancelRequestCompletedEventAttributes nexus_operation_cancel_request_completed_event_attributes = 61;
+        NexusOperationCancelRequestFailedEventAttributes nexus_operation_cancel_request_failed_event_attributes = 62;
```
cretz, line 1060: Will these be marked as `worker_may_ignore`? I am curious about the backwards-compatible properties from an SDK POV.

cretz, line 1060: Can you clarify what happens on current server version on Nexus cancel requested? Will any changes we make server side still work with older SDKs?

Quinn-With-Two-Ns: Yes it will need to be

pdoerner: Currently we record a `NEXUS_OPERATION_CANCEL_REQUESTED` event when we receive the command. Then we actually send the request to the handler. We do not record the operation as cancelled until the completion callback is invoked with a cancelled error. None of this behavior will change. We will just start recording a history event after the handler responds to the cancel request.

Quinn-With-Two-Ns (approved):

temporal/api/history/v1/message.proto
cretz: Sounds good, so this means that we are effectively just sending more events that we usually would for the same situation, and if we properly mark as `worker_may_ignore`, we're good.

My only concern now is from the SDK side, how is an SDK that is supposed to wait for this event supposed to know it is never coming on an older server? So at least for Core-based SDKs, the default cancel type for activities is "try cancel" (i.e. fire and forget) and the default cancel type for child workflows is "wait cancel completed". I think Nexus operations are going to unfortunately have to be "try cancel" by default to make "wait cancel completed" an opt-in behavior since it is not supported on all GA Nexus servers.

cretz (approved):

temporal/api/history/v1/message.proto
pdoerner: Currently SDKs are doing "wait cancel completed" which would continue to be the default. These new events are for supporting the "try cancel" case. When a newer SDK is talking to an older server, it will never receive these new events and will effectively fall back to "wait cancel completed". I think this behavior is okay, do you see an issue with it? I believe this is the same as how child workflows are handled.

cretz: I see...

> I think this behavior is okay, do you see an issue with it?

No issue if we make the SDK-side default of this new enum to be "wait cancel completed" which we've done for child workflows in Core-based SDKs, but never have done in Go and Java based on my quick reading.

cretz: :+1: Whatever is the default today I think we need to make the default when we make it configurable in SDKs for compat reasons.

temporal/api/history/v1/message.proto
```diff
@@ -954,6 +954,24 @@ message NexusOperationCancelRequestedEventAttributes {
     int64 workflow_task_completed_event_id = 2;
 }
 
+message NexusOperationCancelRequestCompletedEventAttributes {
+    // The ID of the `NEXUS_OPERATION_CANCEL_REQUESTED` event.
+    int64 requested_event_id = 1;
+    // The `WORKFLOW_TASK_COMPLETED` event that the corresponding RequestCancelNexusOperation command was reported
+    // with.
```
rodrigozhou, line 961:
```suggestion
    // The `WORKFLOW_TASK_COMPLETED` event that the corresponding RequestCancelNexusOperation command was reported.
```

temporal/api/history/v1/message.proto
```diff
@@ -954,6 +954,24 @@ message NexusOperationCancelRequestedEventAttributes {
...
+    int64 workflow_task_completed_event_id = 2;
+}
+
+message NexusOperationCancelRequestFailedEventAttributes {
+    // The ID of the `NEXUS_OPERATION_CANCEL_REQUESTED` event.
+    int64 requested_event_id = 1;
+    // The `WORKFLOW_TASK_COMPLETED` event that the corresponding RequestCancelNexusOperation command was reported
+    // with.
```
rodrigozhou, line 969:
```suggestion
    // The `WORKFLOW_TASK_COMPLETED` event that the corresponding RequestCancelNexusOperation command was reported.
```

rodrigozhou (approved):
nit comments.

PR #566, opened 2025-04-03
Add reset_run_id field in api/workflow/v1/WorkflowExecutionInfo

gow:
Added `reset_run_id` field in WorkflowExecutionInfo message. This field will be populated in the DescribeWorkflow API calls (coming up in another PR in server repo)

This is needed so that we can expose this field to the UI. The UI would use this information to point the user to the new execution when viewing the old execution that was reset.

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -137,7 +137,7 @@ message WorkflowExecutionExtendedInfo {
     // Original workflow start time.
     google.protobuf.Timestamp original_start_time = 5;
 
-    // Reset Run ID points to the new nun when this execution is reset. If the execution is reset multiple times, it points to the latest run.
+    // Reset Run ID points to the new run when this execution is reset. If the execution is reset multiple times, it points to the latest run.
     string reset_run_id = 25;
 }
 
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -117,9 +117,6 @@ message WorkflowExecutionInfo {
 
     // Priority metadata
     temporal.api.common.v1.Priority priority = 24;
-
-    // Reset Run ID points to the new nun when this execution is reset. If the execution is reset multiple times, it points to the latest run.
-    string reset_run_id = 25;
 }
 
 // Holds all the extra information about workflow execution that is not part of Visibility.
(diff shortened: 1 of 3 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -117,6 +117,9 @@ message WorkflowExecutionInfo {
 
     // Priority metadata
     temporal.api.common.v1.Priority priority = 24;
+
+    // Reset Run ID points to the new nun when this execution is reset. If the execution is reset multiple times, it points to the latest run.
+    string reset_run_id = 25;
```
cretz, line 122: From PR description:

> This field will be populated in the DescribeWorkflow API calls

Will this also be set in visibility calls, e.g. `ListWorkflowExecutions`? If not, we should move it to `WorkflowExecutionExtendedInfo`

gow: No. These fields are not indexed in Visibility service. So they won't be in those response.
Moved it to `WorkflowExecutionExtendedInfo`

temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +136,9 @@ message WorkflowExecutionExtendedInfo {
 
     // Original workflow start time.
     google.protobuf.Timestamp original_start_time = 5;
+
+    // Reset Run ID points to the new nun when this execution is reset. If the execution is reset multiple times, it points to the latest run.
```
Sushisource, line 140:
```suggestion
    // Reset Run ID points to the new run when this execution is reset. If the execution is reset multiple times, it points to the latest run.
```

temporal/api/workflow/v1/message.proto
```diff
@@ -136,6 +136,9 @@ message WorkflowExecutionExtendedInfo {
 
     // Original workflow start time.
     google.protobuf.Timestamp original_start_time = 5;
+
+    // Reset Run ID points to the new run when this execution is reset. If the execution is reset multiple times, it points to the latest run.
+    string reset_run_id = 25;
```
ychebotarev, line 141: nit: 6

ychebotarev (approved):

temporal/api/workflow/v1/message.proto
gow: Fixed.

cretz (approved):

PR #568, opened 2025-04-10
Add reason to the PauseActivityRequest

ychebotarev:

**What changed?**
Add "reason" to PauseActivityRequest

**Why?**
Product request.
Unlike other activity operation, pause is reversible. For manually paused activities operator may need to know the reason for pausing.

**Breaking changes**
No.

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1860,6 +1860,8 @@ message PauseActivityRequest {
         string type = 5;
     }
 
+    // Reason to pause the activity.
+    string reason = 6;
 }
 
 message PauseActivityResponse {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -7049,6 +7049,10 @@
         "type": {
           "type": "string",
           "description": "Pause all running activities of this type."
+        },
+        "reason": {
+          "type": "string",
+          "description": "Reason to pause the activity."
         }
       }
     },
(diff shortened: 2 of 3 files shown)


Sushisource (approved):

yiminc (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1860,6 +1860,8 @@ message PauseActivityRequest {
         string type = 5;
     }
 
+    // Reason to pause the activity.
+    string reason = 6;
```
cretz, line 1864: Is this surfaced anywhere?

cretz, line 1864: Do I need to be able to set this when I use workflow rules to pause activities?

ychebotarev: later it will be surfaced in DescribeWorkflow reponse, in pendingActivityInfo

ychebotarev: not at the moment. If we will need "reason" for rules (or rather something like "description") it should be part of rule itself.

cretz: Why not add it now?

ychebotarev: probably this require a bit more explanation, though it will be clear in the next PR.
there will be PauseInfo, and it will contain "identity", and for rules it will be "rule_id". 
My though process is - if later we will need to get more details -we can use that rule_id to get rule info, and read a description. There is no need to duplicate rule specific information here.

PR #569, opened 2025-04-10
Add PauseInfo to the pending activity info

ychebotarev:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -8794,6 +8794,26 @@ components:
           allOf:
             - $ref: '#/components/schemas/Priority'
           description: Priority metadata
+        pauseInfo:
+          $ref: '#/components/schemas/PendingActivityInfo_PauseInfo'
+    PendingActivityInfo_PauseInfo:
+      type: object
+      properties:
+        pauseTime:
+          type: string
+          description: The time when the activity was paused.
+          format: date-time
+        identity:
+          type: string
+          description: |-
+            The identity of the actor that paused the activity.
+             Will be identity for manual pause triggering or rule_id if it was paused by the rule.
+        triggerType:
+          type: string
+          description: Can be either "manual" or "rule".
+        reason:
+          type: string
+          description: Reason for pausing the activity. Only populated for 'manual' trigger type.
     PendingChildExecutionInfo:
       type: object
       properties:
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -304,6 +304,20 @@ message PendingActivityInfo {
 
     // Priority metadata
     temporal.api.common.v1.Priority priority = 22;
+
+    message PauseInfo {
+        // The time when the activity was paused.
+        google.protobuf.Timestamp pause_time = 1;
+        // The identity of the actor that paused the activity.
+        // Will be identity for manual pause triggering or rule_id if it was paused by the rule.
+        string identity = 2;
+        // Can be either "manual" or "rule".
+        string trigger_type = 3;
+        // Reason for pausing the activity. Only populated for 'manual' trigger type.
+        string reason = 4;
+    }
+
+    PauseInfo pause_info = 23;
 }
 
 message PendingChildExecutionInfo {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -6756,6 +6756,28 @@
         }
       }
     },
+    "PendingActivityInfoPauseInfo": {
+      "type": "object",
+      "properties": {
+        "pauseTime": {
+          "type": "string",
+          "format": "date-time",
+          "description": "The time when the activity was paused."
+        },
+        "identity": {
+          "type": "string",
+          "description": "The identity of the actor that paused the activity.\nWill be identity for manual pause triggering or rule_id if it was paused by the rule."
+        },
+        "triggerType": {
+          "type": "string",
+          "description": "Can be either \"manual\" or \"rule\"."
+        },
+        "reason": {
+          "type": "string",
+          "description": "Reason for pausing the activity. Only populated for 'manual' trigger type."
+        }
+      }
+    },
     "StartOperationResponseAsync": {
       "type": "object",
       "properties": {
@@ -11603,6 +11625,9 @@
         "priority": {
           "$ref": "#/definitions/v1Priority",
           "title": "Priority metadata"
+        },
+        "pauseInfo": {
+          "$ref": "#/definitions/PendingActivityInfoPauseInfo"
         }
       }
     },


temporal/api/workflow/v1/message.proto
```diff
@@ -304,6 +304,20 @@ message PendingActivityInfo {
 
     // Priority metadata
     temporal.api.common.v1.Priority priority = 22;
+
+    message PauseInfo {
```
cretz, line 308: Do we want similar on activity task started (or completed) event?

temporal/api/workflow/v1/message.proto
```diff
@@ -304,6 +304,20 @@ message PendingActivityInfo {
...
+        // The time when the activity was paused.
+        google.protobuf.Timestamp pause_time = 1;
+        // The identity of the actor that paused the activity.
+        // Will be identity for manual pause triggering or rule_id if it was paused by the rule.
+        string identity = 2;
```
cretz, line 313: I think it is a bit confusing to have identity doesn't mean just identity, sometimes it can have a new meaning (rule ID). I'd make rule ID a separate field if it's really required. A manual triggering of the rule can have an identity and a rule ID (or at least we should accept an identity on `TriggerWorkflowRuleRequest`). Alternatively, we can remove this altogether until we're sure it's needed.

temporal/api/workflow/v1/message.proto
```diff
@@ -304,6 +304,20 @@ message PendingActivityInfo {
...
+        // Can be either "manual" or "rule".
+        string trigger_type = 3;
```
cretz, line 315: Use well-typed proto enumerates instead strings here. Also do we call the act of pausing "triggering"? It seems that may be lingo we only use in rules? Alternatively, we can remove this altogether until we're sure it's needed.

openapi/openapiv3.yaml
```diff
@@ -8794,6 +8794,26 @@ components:
...
+          description: The time when the activity was paused.
+          format: date-time
+        identity:
+          type: string
+          description: |-
+            The identity of the actor that paused the activity.
+             Will be identity for manual pause triggering or rule_id if it was paused by the rule.
+        triggerType:
+          type: string
+          description: Can be either "manual" or "rule".
```
drewhoskins-temporal, line 8813: I found this confusing to think through.
So if I do 
`temporal activity pause --create-rule --scan ...`
it's rule but 
`temporal activity pause --scan ...`
it's manual?
Even, in both cases, if it was applied via a "backfill" of existing workflows?

Maybe more straightforward to just have a rule ID if a rule was created and otherwise  have all the fields be the same?

openapi/openapiv3.yaml
```diff
@@ -8794,6 +8794,26 @@ components:
...
+        reason:
+          type: string
+          description: Reason for pausing the activity. Only populated for 'manual' trigger type.
```
drewhoskins-temporal, line 8816: Product scenario is: I'm in the UI seeing a paused activity and wondering what/who caused that?  

To code that, one would do this?
```
if trigger type == rule
  query rule
  render rule.reason, rule.identity
else
  render identity, reason
``` 

I imagine it might be more convenient if we denormalize info here.

If we do keep this pattern, maybe use `oneof` to make it more clear/straightforward?
@stevekinney will be more authoritative here.

temporal/api/workflow/v1/message.proto
ychebotarev: not at the moment

temporal/api/workflow/v1/message.proto
ychebotarev: what will be a better word? we need it.

temporal/api/workflow/v1/message.proto
drewhoskins-temporal: 👍 I wonder if we can combine chad's suggestion with a `oneof` to express the constraints better.  I wrote a scenario above.

openapi/openapiv3.yaml
ychebotarev: with "temporal activity pause --create-rule --scan" you create a rule, and then apply it immediately. 
Same as calling TriggerWorkflowRule and provide rule_id.
In this case if activity is paused - it is paused by the rule.

`temporal activity pause --scan ...`

it is a batch version of `call TriggerWorkflowRule and provide rule spec` or just PauseActivity.
In this second case we don't have rule_id, and we don't have an audit trail like "who created the rule".
In the first case we do have that trail, so there is no need to provide rule author identity.

openapi/openapiv3.yaml
ychebotarev: `oneof` is good in terms of clarity, but implementation wise (not for me, for consumers) is harder.

temporal/api/workflow/v1/message.proto
ychebotarev: >I think it is a bit confusing to have identity not mean just identity,

Drew has `oneof` proposal. I can use it.

> A manual triggering of the rule can have an identity and a rule ID 
I look at it differently. 

If you manually trigger the existing rule - activity is paused by the rule. "Was a rule triggered by event or manually" is a possible question, but I don't think its important. User can get all the needed context from the rule itself.

If you pause without existing rule (PauseActivity or TriggerWorkflowRule with spec) - such context doesn't exist, and trail should lead to you.

openapi/openapiv3.yaml
ychebotarev: changed to oneof, lets see how it goes.

temporal/api/workflow/v1/message.proto
ychebotarev: irrelevant with oneof

temporal/api/workflow/v1/message.proto
ychebotarev: replaced with oneof.

cretz: > If you manually trigger the existing rule - activity is paused by the rule

It's paused by both. To tell a user that they can only get client identity if manual triggering was done one way and not a another is confusing. We should always give an identity if we have it. Identity and rule ID should be independent optional fields, not related to each other at all, and definitely not mutually exclusive.

And we need to add `identity` on `TriggerWorkflowRuleRequest`, I am surprised we forgot it. We need it on every mutable RPC call IMO (so also on create rule and delete rule).

ychebotarev: I disagree, I see no value for users in mixing those two conceptions. 
Mutable state is not an audit log, lets not bloat it for no practical reason.
If the request to add identity will come from the user - we will add it. 


I will add identity to TriggerWorkflowRuleRequest.

Sushisource (approved):
Personally, I am OK with having either identity or rule id. But I agree with Chad's point that they aren't really mutually exclusive. I also agree with not bloating mut state unless we have a good reason, so I can understand looking up the rule id and then using that to look up the rule's creator.

Unless there's some big rush to get this in, though, I think it would be good to wait until Chad is back tomorrow to ensure he's fine with it.

temporal/api/workflow/v1/message.proto
```diff
@@ -304,6 +304,24 @@ message PendingActivityInfo {
...
+
+        oneof paused_by {
+            // The identity of the actor that paused the activity.
+            string identity = 2;
+
+            // Id of the rule that paused the activity.
+            string rule_id = 3;
+        }
+
+        // Reason for pausing the activity. Only populated for 'manual' trigger type.
```
drewhoskins-temporal, line 320: I think this comment, and others alike, are out of date.

* If reason is for everything now, this can stay here
* If it's only for manual, seems like it should be nested further inside next to `identity`

temporal/api/workflow/v1/message.proto
cretz: > I will add identity to TriggerWorkflowRuleRequest

Where will this identity be visible? Here are the three ways I can see where I can explicitly trigger pause as a user:

1. Provide identity and call `PauseActivityRequest`
2. Provide identity and call `TriggerWorkflowRuleRequest` with a rule ID to a pause rule
3. Provide identity and call `TriggerWorkflowRuleRequest` with a rule spec that pauses

To clarify, which ones of those are we choosing to hide the identity I provide? Why is there no practical use of identity sometimes and not the other? Is having the identity in cases where you are hiding it with the current API really bloat? If they can do this same thing three different ways, why are we trying so hard to inconsistently hide the identity on some?

cretz:
>  I can understand looking up the rule id and then using that to look up the rule's creator.

@Sushisource - No, the rules creator is not the one triggering. Also you can provide a never-persisted rule at trigger time. This is swallowing the identity of the triggerer on purpose with more downside that upside (it is not serious bloat to consistently record the identity).

temporal/api/workflow/v1/message.proto
ychebotarev: >2. Provide identity and call TriggerWorkflowRuleRequest with a rule ID to a pause rule

In this case you are triggering the rule. You may not even know what the rule is doing. 

In other cases you are directly pausing the activities (providing rule spec is just a convenience, you can achieve the same by examining the activity, "running" rule in your head  and calling pause only for those that match the rule)

So in other cases there will be identity.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2271,6 +2271,9 @@ message TriggerWorkflowRuleRequest {
         // Note: Rule ID and expiration date are not used in the trigger request.
         temporal.api.rules.v1.WorkflowRuleSpec spec = 5;
     }
+
+    // The identity of the client who initiated this request
+    string identity = 3;
```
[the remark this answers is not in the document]
ychebotarev: If trigger by rule - nowhere right now. We plan to have better audit log story at some point. 
If by spec - in PauseInfo.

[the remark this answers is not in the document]
cretz: I think regardless of whether rule ID or rule spec given, we should always make it available in pause info

temporal/api/workflow/v1/message.proto
gow: I think we should capture the actor actually performing the action. So if a rule is triggered manually I would expect the identity of the actor triggering that rule. Ex: Person-A manually triggering person-B's rule should trace back to person-A. Is there anything preventing us from doing this?
If the action is automatic then the actor it the original author of the automatic rule.

gow: Discussed it offline with @ychebotarev. Looks like this information is very ephemeral and its main purpose is to be shown only in the UI during the lifetime of the activity. From that perspective, I feel it's ok to just show who paused it or which rule paused it. Further relationship of rule to actor can be a separate lookup in my opinion.

cretz: To clarify, we're not showing identity of who paused it sometimes because it's too hard in the server?

ychebotarev: we are showing identity of who/what paused in a consistent manner.

cretz: It's not consistent if whether the rule was ID based or inline means this value is different. Consistent would be "no matter how the user manually triggered pause, they provided identity, so we make it available in the same consistent way".

ychebotarev: Activity can be paused either by rule or by ID. Consistent would be "provide rule id if it was paused by the rule, or identity if it was paused manually". That is what we are doing.
What you are described "also provide who/what triggered a rule". No needed here.

gow: `PauseInfo.Manual.Identity` always records the identity of the person performing the manual action, no? So, I'm not sure I'm seeing the inconsistency here.
My understand is that when the rule is triggered automatically, we just record the rule ID and leave the further lookup of rule_id -> author for the end user viewing this on UI.

cretz: > PauseInfo.Manual.Identity always records the identity of the person performing the manual action, no?

No. Look at the protos. If you call trigger with a rule ID instead of a rule spec, we are hiding identity.

ychebotarev: This is mixing "manual acton to pause the activity" and "manual action to trigger the rule". It doesn't matter how rule was triggered, we don't share this in pauseInfo. What matter is - did rule actually pause it. If it does - we record the rule id.

cretz: > did rule actually pause it. If it does - we record the rule id.

You mean "did a pre-persisted rule actually pause it" not to be confused with "did rule actually pause it" because if it's not a rule with an ID, then we don't hide the identity.

We should show the rule ID if there is a rule ID and the identity if there is an identity.

ychebotarev: There is no "pre-persisted rules" as opposite to "persisted rule". 

If activity was paused by rule we will show this is pause info. If it was paused manually we will shod it in pause info.
I think I put enough efforts to explain it.

gow (approved):
This is ok from server perspective since we can always add more information later if needed (but cannot take away information once added).
Assuming SDK is team is also ok with this change, it's good to go.

PR #570, opened 2025-04-15
Add ApplicationErrorCategory to ApplicationFailureInfo

THardy98:
**What changed?**
Added an ENUM `ApplicationErrorCategory`, used in `ApplicationFailureInfo` proto.

**Why?**
Give users a way to categorize application failures. Error categories map to different observability behaviours SDK-side (i.e. logging/metrics), part of the work for benign exceptions. Categories could also be used with rules in the future.

**Breaking changes**
No

**Server PR**
Don't believe this breaks server or requires any changes to server

### DIFF
--- a/temporal/api/enums/v1/common.proto
+++ b/temporal/api/enums/v1/common.proto
@@ -113,3 +113,9 @@ enum WorkflowRuleActionScope {
     // The action will be applied to a specific activity.
     WORKFLOW_RULE_ACTION_SCOPE_ACTIVITY = 2;
 }
+
+enum ApplicationErrorCategory {
+    APPLICATION_ERROR_CATEGORY_UNSPECIFIED = 0;
+    // Expected application error with little/no severity.
+    APPLICATION_ERROR_CATEGORY_BENIGN = 1;
+}
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -8535,6 +8535,15 @@
       },
       "description": "Alert contains notification and severity."
     },
+    "v1ApplicationErrorCategory": {
+      "type": "string",
+      "enum": [
+        "APPLICATION_ERROR_CATEGORY_UNSPECIFIED",
+        "APPLICATION_ERROR_CATEGORY_BENIGN"
+      ],
+      "default": "APPLICATION_ERROR_CATEGORY_UNSPECIFIED",
+      "description": " - APPLICATION_ERROR_CATEGORY_BENIGN: Expected application error with little/no severity."
+    },
     "v1ApplicationFailureInfo": {
       "type": "object",
       "properties": {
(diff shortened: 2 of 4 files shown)


temporal/api/enums/v1/failed_cause.proto
```diff
@@ -153,3 +153,8 @@ enum ResourceExhaustedScope {
     RESOURCE_EXHAUSTED_SCOPE_SYSTEM = 2;
 }
 
+enum ApplicationErrorCategory {
```
cretz, line 156: Pedantic, but I think this should move to the `common.proto` file or a new `failure.proto` file, as this file I think it more about lower-level failure causes and not as much about user-facing failures.

cretz (approved):
LGTM, very minor suggestion about which file, not blocking

temporal/api/enums/v1/failed_cause.proto
THardy98: Yeah I was unsure between the two, moved to `common.proto`.

cretz: I would confirm with server team if any explicit code is needed. Of course the tagged `api-go` has to at least land in server before usable by SDKs.

temporal/api/enums/v1/common.proto
```diff
@@ -112,4 +112,10 @@ enum WorkflowRuleActionScope {
     WORKFLOW_RULE_ACTION_SCOPE_WORKFLOW = 1;
     // The action will be applied to a specific activity.
     WORKFLOW_RULE_ACTION_SCOPE_ACTIVITY = 2;
+}
+
+enum ApplicationErrorCategory {
+    APPLICATION_ERROR_CATEGORY_UNSPECIFIED = 0;
```
ychebotarev, line 118: This approach means that "APPLICATION_ERROR_CATEGORY_UNSPECIFIED" is the default "this is bad error" category.
I would rather have an explicit "default" category.

THardy98: > I want to be able to differentiate between "it was not set by mistake" and "it was to the 0 to indicate that this is just an error".

I don't mind changing, but why do you want to do this?

I get that we want to work with values we define, but if a user doesn't specify a category that's equivalent to the existing behavior, which makes sense as a default.

ychebotarev: wait.
scrap that comment, enums are not nullable.

ychebotarev (approved):

temporal/api/enums/v1/common.proto
THardy98: Sounds good.

Just wanted to double-check - are there any server changes necessary for this api change?
I looked through the server code and it looks like we pass the entire failure object when persisting. Looked like no schema changes or anything necessary but wanted to make sure since I haven't read much into the server code generally.

ychebotarev: I can tell mostly from activity perspective - no server changes is needed for that. This structure is used in some other places, but again from what I can see we use as you said entire object.

PR #571, opened 2025-04-16
Deployment and TaskQueue Stats API

stephanos:

**What changed?**

Added two new APIs:

* `GetWorkerDeploymentStats`
* `GetTaskQueueStats`

**Why?**

To return task queue stats to users for making scaling decisions.

**Breaking changes**

No.

**Server PR**

https://github.com/temporalio/temporal/pull/7581 _(feature complete, but not yet production-ready)_

### DIFF
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -336,3 +336,9 @@ message PollerScalingDecision {
   // all slots are already occupied.
   int32 poll_request_delta_suggestion = 1;
 }
+
+message TaskQueueStatsInfo {
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 1;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 2;
+    temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 3;
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1085,6 +1085,16 @@ message DescribeTaskQueueResponse {
     temporal.api.taskqueue.v1.TaskQueueVersioningInfo versioning_info = 4;
 }
 
+message GetTaskQueueStatsRequest {
+    string namespace = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
+}
+
+message GetTaskQueueStatsResponse {
+    temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 1;
+}
+
 message GetClusterInfoRequest {
 }
 
@@ -2205,6 +2215,29 @@ message GetDeploymentReachabilityResponse {
     // Reachability level might come from server cache. This timestamp specifies when the value
     // was actually calculated.
     google.protobuf.Timestamp last_update_time = 3;
+}
+
+message GetWorkerDeploymentStatsRequest {
+    string namespace = 1;
+
+    // Name of the deployment to return stats for.
+    //
+    // (-- api-linter: core::0122::name-suffix=disabled
+    string deployment_name = 2;
+
+    // Optional. Deployment version to return stats for.
+    // If left empty, the current version will be queried.
+    string deployment_version = 3;
+}
+
+message GetWorkerDeploymentStatsResponse {
+    // Collection of stats for all task queues in the requested deployment.
+    // Note that the data is
+    repeated temporal.api.taskqueue.v1.TaskQueueStatsInfo per_queue_metrics = 1;
+
+    // The sum of backlog counts across all task queues in the deployment.
+    // Note that same as the task queue metrics, this value is
+    int64 approximate_total_backlog_count = 2;
 }
 
 message CreateWorkflowRuleRequest {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1087,7 +1087,7 @@ message DescribeTaskQueueResponse {
 
 message GetTaskQueueStatsRequest {
     string namespace = 1;
-    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
+    string task_queue = 2;
     temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
 }
 
@@ -2225,9 +2225,9 @@ message GetWorkerDeploymentStatsRequest {
     // (-- api-linter: core::0122::name-suffix=disabled
     string deployment_name = 2;
 
-    // Optional. Deployment version to return stats for.
-    // If left empty, the current version will be queried.
-    string deployment_version = 3;
+    // Optional. Build ID to return stats for.
+    // If left empty, the deployment's current one will be queried.
+    string build_id = 3;
 }
 
 message GetWorkerDeploymentStatsResponse {
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -585,9 +585,9 @@ service WorkflowService {
     // GetTaskQueueStats returns stats for a single task queue.
     rpc GetTaskQueueStats (GetTaskQueueStatsRequest) returns (GetTaskQueueStatsResponse) {
         option (google.api.http) = {
-            post: "/namespaces/{namespace}/task-queues/{task_queue.name}/type/{task_queue_type}/get-stats"
+            post: "/namespaces/{namespace}/task-queues/{task_queue}/type/{task_queue_type}/get-stats"
             additional_bindings {
-                post: "/api/v1/namespaces/{namespace}/task-queues/{task_queue.name}/type/{task_queue_type}/get-stats"
+                post: "/api/v1/namespaces/{namespace}/task-queues/{task_queue}/type/{task_queue_type}/get-stats"
             }
         };
     }
@@ -891,9 +891,9 @@ service WorkflowService {
     // GetWorkerDeploymentStats returns the stats for all task queues in a deployment version.
     rpc GetWorkerDeploymentStats (GetWorkerDeploymentStatsRequest) returns (GetWorkerDeploymentStatsResponse) {
         option (google.api.http) = {
-            get: "/namespaces/{namespace}/worker-deployments/{deployment_name}/{deployment_version}/stats"
+            get: "/namespaces/{namespace}/worker-deployments/{deployment_name}/stats"
             additional_bindings {
-                get: "/api/v1/namespaces/{namespace}/worker-deployments/{deployment_name}/{deployment_version}/stats"
+                get: "/api/v1/namespaces/{namespace}/worker-deployments/{deployment_name}/stats"
             }
         };
     }
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -582,6 +582,16 @@ service WorkflowService {
         };
     }
 
+    // GetTaskQueueStats returns stats for a single task queue.
+    rpc GetTaskQueueStats (GetTaskQueueStatsRequest) returns (GetTaskQueueStatsResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/task-queues/{task_queue.name}/type/{task_queue_type}/get-stats"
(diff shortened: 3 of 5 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2205,6 +2226,30 @@ message GetDeploymentReachabilityResponse {
...
+    string deployment_version = 3;
+}
+
+message GetDeploymentStatsResponse {
+    // Collection of stats for all task queues in this deployment version.
+    // The list is not sorted.
+    repeated temporal.api.taskqueue.v1.TaskQueueStatDetails per_queue_stats = 1;
+
+    // Aggregated backlog count across all task queues in this deployment version.
+    int64 approximate_total_backlog_count = 2;
```
stephanos, line 2250: Name is derived from ``TaskQueueStatDetails`'s `approximate_backlog_count` field.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1085,6 +1085,16 @@ message DescribeTaskQueueResponse {
     temporal.api.taskqueue.v1.TaskQueueVersioningInfo versioning_info = 4;
 }
 
+message GetTaskQueueStatsRequest {
+    string namespace = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
```
stephanos, line 1090: Should this just be `string` instead?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2207,6 +2217,27 @@ message GetDeploymentReachabilityResponse {
...
+    string namespace = 1;
+
+    // Name of the deployment to return stats for.
+    //
+    // (-- api-linter: core::0122::name-suffix=disabled
+    string deployment_name = 2;
+
+    // Optional. Deployment version to return stats for.
+    // If left empty, the current version will be queried.
+    string deployment_version = 3;
```
stephanos, line 2230: Need to double-check that `build_id` isn't a better field name here.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -878,6 +888,17 @@ service WorkflowService {
         };
     }
 
+    // GetWorkerDeploymentStats returns the stats for all task queues in a deployment version.
+    rpc GetWorkerDeploymentStats (GetWorkerDeploymentStatsRequest) returns (GetWorkerDeploymentStatsResponse) {
+        option (google.api.http) = {
+            // TODO: deployment_version can be empty?
+            get: "/namespaces/{namespace}/worker-deployments/{deployment_name}/{deployment_version}/stats"
```
stephanos, line 895: Since deployment_version can be empty, I'm not sure this is right yet.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -336,3 +336,9 @@ message PollerScalingDecision {
   // all slots are already occupied.
   int32 poll_request_delta_suggestion = 1;
 }
+
+message TaskQueueStatsInfo {
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 1;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 2;
+    temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 3;
+}
```
stephanos, line 344: This message only exists because `TaskQueueStats` doesn't contain the task queue name and task_queue_type.

I considered adding them to the existing `TaskQueueStats`, but was discouraged to do so.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -878,6 +888,16 @@ service WorkflowService {
         };
     }
 
+    // GetWorkerDeploymentStats returns the stats for all task queues in a deployment version.
+    rpc GetWorkerDeploymentStats (GetWorkerDeploymentStatsRequest) returns (GetWorkerDeploymentStatsResponse) {
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/worker-deployments/{deployment_name}/{deployment_version}/stats"
```
stephanos, line 894: Since deployment_version can be empty, I'm not sure this is right yet.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -582,6 +582,16 @@ service WorkflowService {
         };
     }
 
+    // GetTaskQueueStats returns stats for a single task queue.
+    rpc GetTaskQueueStats (GetTaskQueueStatsRequest) returns (GetTaskQueueStatsResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/task-queues/{task_queue.name}/type/{task_queue_type}/get-stats"
```
stephanos, line 588: Does this work with the `task_queue_type` enum in there?

temporal/api/workflowservice/v1/request_response.proto
ShahabT: Yes. here and in other places. The struct lets you specify internal TQs such as sticky ones, rather than user-defined TQs. We've been trying to not use this proto for new APIs that are intended for the user-defined TQs only.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: We're in the middle of that discussion right now in the Versioning crew, we have not fully converged yet. If you want to merge this asap, I think `build_id` is the safer alternative.

temporal/api/workflowservice/v1/service.proto
ShahabT: Copilot says we can and caller would put the numeric values of the enum in the URL.

temporal/api/workflowservice/v1/service.proto
ShahabT: I think we should only put the name in the url.

stephanos: This shows I'm not familiar with how the grpc-http bridge works. I suppose the other args are passed as query parameters?

temporal/api/workflowservice/v1/service.proto
stephanos: Hm, I don't know if the _numeric_ enum representation is a good choice here.

temporal/api/taskqueue/v1/message.proto
cretz: Hrmm, I wonder if we can come up with a better name for this that just wraps stats and has a qualified set. Would you consider making this message a sub-message of `GetWorkerDeploymentStatsResponse` instead of top-level where it is a bit confusing/ambiguous?

Though if you look at my suggestion later concerning `ListTaskQueues`, you may find that this message may not be necessary.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -878,6 +888,16 @@ service WorkflowService {
         };
     }
 
+    // GetWorkerDeploymentStats returns the stats for all task queues in a deployment version.
+    rpc GetWorkerDeploymentStats (GetWorkerDeploymentStatsRequest) returns (GetWorkerDeploymentStatsResponse) {
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/worker-deployments/{deployment_name}/stats"
```
cretz, line 894: #579 seems like it may move away from this `deployment_name` concept. Want to make sure whichever PR is merged second is changed to apply to the first.

temporal/api/workflowservice/v1/request_response.proto
cretz: Is this just a convenience value that is easily derived from the per queue metrics field? If so, is there any need for "get worker deployment stats" if it's just "list task queues with deployment name filter and opt-in to stats"?

I think we should add a `ListTaskQueues` as a multi-task-queue equivalent to `DescribeTaskQueue`, have it accept filters (can limit right now to _requiring_ deployment name at runtime if you don't want full implementation), and allow opting in to getting stats while listing. The `DescribeX` for single + `ListXs` for multiple is a much more consistent API and has better reuse potential and is clearer to use/understand.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -582,6 +582,16 @@ service WorkflowService {
         };
     }
 
+    // GetTaskQueueStats returns stats for a single task queue.
+    rpc GetTaskQueueStats (GetTaskQueueStatsRequest) returns (GetTaskQueueStatsResponse) {
```
cretz, line 586: What does this provide that `DescribeTaskQueue` doesn't?

temporal/api/workflowservice/v1/request_response.proto
stephanos: > Is this just a convenience value that is easily derived from the per queue metrics field?

Yes, it's really just for convenience. There was a _long_ discussion about this at the DevX review (happy to provide details) and this was the compromise that both camps (no aggregation vs lots of aggregation) were okay with.

> I think we should add a `ListTaskQueues`

The major issue with that option is that we cannot list task queues. Since their are "unloaded" when not in active use after a few minutes, the API wouldn't be able to return stats for unloaded task queues. Users are expected to provide the task queues they are interested in for that reason. On the other hands with deployments we actually track the task queues, so that's why that API is more convenient/powerful.

temporal/api/workflowservice/v1/service.proto
stephanos: 👍 I'll check-in with the team on our Thu sync about that.

temporal/api/workflowservice/v1/service.proto
stephanos: I'll send you the DevX review; it details this in depth!

temporal/api/workflowservice/v1/request_response.proto
cretz: > The major issue with that option is that we cannot list task queues [...] On the other hands with deployments we actually track the task queues, so that's why that API is more convenient/powerful.

To confirm, you can list task queues if deployment name is present? This is why I added "can limit right now to requiring deployment name at runtime if you don't want full implementation" to my comment. IMO it is better to have a limited form of `ListTaskQueues` than to have worker deployment have its own "list task queues". I would like to see an issue open to make the full list task queues (i.e. if a user doesn't provide deployment name) work.

temporal/api/workflowservice/v1/service.proto
cretz: :+1: I have disagreements with this assessment (we already did a refresh, this is another refresh with no benefit from a user POV). Can discuss internally.

stephanos: To clarify; any objections towards the Deployment Stats API part?

temporal/api/workflowservice/v1/service.proto
carlydf: I was just figuring out the grpc-http stuff yesterday! This convo with ChatGPT helped (start reading from where I ask "What is the benefit of putting deployment_name in the http path, and what is the problem with not including it"): https://chatgpt.com/share/68113ebc-e300-8004-9732-3b1bd38a1029

Basically, namespace _must_ be included in the path because it is used for both routing and auth. These resource names are not used for routing or auth, so it is technically not required to put them in the path. However, it is more "RESTful" to put resource names in the http path, because it makes using the APIs via curl more idiomatic.

temporal/api/workflowservice/v1/service.proto
carlydf: Hi! In #579 I have removed `deployment_name` from APIs that also take a `WorkerDeploymentVersion` (with it's own `deployment_name` field, so that it will be impossible to create a request object that has `deployment_name != version.deployment_name`.

APIs that do not take a version, such as `DescribeWorkerDeployment` still take a `string deployment_name` and I can't think of any reason that would change.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2207,6 +2217,29 @@ message GetDeploymentReachabilityResponse {
...
+    string build_id = 3;
+}
+
+message GetWorkerDeploymentStatsResponse {
+    // Collection of stats for all task queues in the requested deployment.
+    // Note that the data is
+    repeated temporal.api.taskqueue.v1.TaskQueueStatsInfo per_queue_metrics = 1;
+
+    // The sum of backlog counts across all task queues in the deployment.
+    // Note that same as the task queue metrics, this value is
```
Shivs11, line 2239: nit: is this an unfinished comment?

PR #572, opened 2025-04-16
Add ScheduledEventID to NexusOperationCancelRequest[Completed|Failed]

pdoerner:
**What changed?**
Added ScheduledEventID of the Nexus operation to the new NexusOperationCancelRequestCompleted and NexusOperationCancelRequestFailed events

**Why?**
For efficient lookups. Go SDK stores cancellation state machines keyed on scheduled event ID.

**Server PR**
https://github.com/temporalio/temporal/pull/7612

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -960,6 +960,8 @@ message NexusOperationCancelRequestCompletedEventAttributes {
     // The `WORKFLOW_TASK_COMPLETED` event that the corresponding RequestCancelNexusOperation command was reported
     // with.
     int64 workflow_task_completed_event_id = 2;
+    // The id of the `NEXUS_OPERATION_SCHEDULED` event this cancel request corresponds to.
+    int64 scheduled_event_id = 3;
 }
 
 message NexusOperationCancelRequestFailedEventAttributes {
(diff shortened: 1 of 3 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -960,6 +960,8 @@ message NexusOperationCancelRequestCompletedEventAttributes {
     // The `WORKFLOW_TASK_COMPLETED` event that the corresponding RequestCancelNexusOperation command was reported
     // with.
     int64 workflow_task_completed_event_id = 2;
+    // The id of the `NEXUS_OPERATION_SCHEDULED` event this cancel request corresponds to.
+    int64 scheduled_event_id = 3;
```
cretz, line 964: Hrmm, not technically needed since it's already in the `NexusOperationCancelRequestedEventAttributes` that `requested_event_id` links to. But I guess `ChildWorkflowExecutionCanceledEventAttributes` has `initiated_event_id` which is redundant too because it can be derived via `started_event_id`.

Didn't go over the rest of history, but I assume our posture is that we are ok with redundant event ID fields like these?

pdoerner: We do have some redundant fields elsewhere, yes. Technically it is not needed but in the Go SDK we store cancellation state machines keyed by ScheduledEventID. So the alternative is to do some extra book-keeping there to maintain a map of RequestedEventID -> ScheduledEventID which is updated whenever we get a NexusOperationCancelRequested event. I thought this would be a little cleaner.

cretz: Definitely don't need to work around just the Go SDK's internals (it's only one of our SDKs by many numbers, not even the most popular or most used). Feel free to make the decision on whether to have redundant event ID references in events regardless of the SDK. May prefer to take into account history size (though I know this is a negligible byte or two).

rodrigozhou (approved):

PR #575, opened 2025-04-23
Async SetRampingVersion and SetCurrentVersion

carlydf:

- Tell users that SetRampingVersion and SetCurrentVersion will be async
- Add routing_config_propagation_state so that users can have visibility into whether all task queue partitions have the latest routing config.

Synchronous task queue propagation meant that the lock on the Worker Deployment Manager was held for many seconds, meaning the APIs would be unavailable if there were large numbers of partitions. This allows us to have high availability APIs while being transparent about where tasks are being routed.

**Breaking changes**

**Server PR**


temporal/api/deployment/v1/message.proto
```diff
@@ -201,6 +201,10 @@ message WorkerDeploymentInfo {
 
     RoutingConfig routing_config = 4;
 
+    // Indicates whether the routing_config has been fully propagated to all
+    // relevant task queues and partitions.
+    enums.v1.TaskQueuePropagation routing_config_propagation_state = 6;
```
dnr, line 206: I wonder if even saying "task queue" and "propagation" is too much detail / too specific? Maybe something like "update state": "pending" or "completed"

Shivs11, line 206: `routing_config_propagation_state` is something that's related to the routing information of a worker deployment info and I wonder if it's better suited to be present inside `routing_config`.

PR #577, opened 2025-04-25
Add indentity/description to the rule, propagate then in PauseInfo

ychebotarev:
**What changed?**
1. Add identity/description to the rules.
2. Add new message with "rule_id, identity, reason" to the PauseInfo

**Why?**
Product request - identity/reason should be always populated, for both manual/rule.

**Breaking changes**
Not really. But old "rule_id" in oneof is effectively deprecated.

### DIFF
--- a/temporal/api/rules/v1/message.proto
+++ b/temporal/api/rules/v1/message.proto
@@ -105,7 +105,7 @@ message WorkflowRule {
   //     aip.dev/not-precedent: It is better reflect the intent this way, we will also have updated_by. --)
   // (-- api-linter: core::0142::time-field-names=disabled
   //     aip.dev/not-precedent: Same as above. All other options sounds clumsy --)
-  string created_by = 3;
+  string created_by_identity = 3;
 
   // Rule description.
   string description = 4;
--- a/temporal/api/rules/v1/message.proto
+++ b/temporal/api/rules/v1/message.proto
@@ -100,7 +100,7 @@ message WorkflowRule {
   // Rule specification
   WorkflowRuleSpec spec = 2;
 
-  // Indentity of the actor that created the rule
+  // Identity of the actor that created the rule
   string identity = 3;
 
   // Rule description.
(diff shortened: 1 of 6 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -322,12 +322,26 @@ message PendingActivityInfo {
...
+            string reason = 3;
+        }
+
         oneof paused_by {
             // activity was paused by the manual intervention
             Manual manual = 2;
 
             // Id of the rule that paused the activity.
+            // Deprecated, replaced by Rule
             string rule_id = 3;
```
cretz, line 341: Can you just delete this? Usually we can't or don't want to make backwards-incompatible changes, but I think we can here since nobody is using pause info yet, correct?

ychebotarev: Yes, nobody besides my tests is using pause info yet.
We can't remove with reserving, because "reserve" apparently is not supported in "oneof"
I add it to "ignore", will remove it in the later PR.

cretz (approved):

temporal/api/rules/v1/message.proto
```diff
@@ -99,4 +99,10 @@ message WorkflowRule {
 
   // Rule specification
   WorkflowRuleSpec spec = 2;
+
+  // Identity of the actor that created the rule
+  string identity = 3;
```
gow, line 104: nit: should we call this `actor` (or actor_identity)? Same in other messages as well.
WorkflowRule.identity could be misunderstood if they look at it in isolation.

gow (approved):

temporal/api/rules/v1/message.proto
cretz: The term "identity" is used API wide, no benefit in changing just here. From an SDK POV it's actually "client identity".

ychebotarev: discussed offline.
I will change this to "created_by".
Main motivation - we want "toggle rule" support, and want this information - "who toggles the rule" - to be available.
I think it will be stored in something like "updated_by".
Other options like "Identity2" or "updated_by" and "identity" instead of "created by" feels not as clean.

ychebotarev: per Chad suggestion, renamed to 'created_by_identity'

PR #578, opened 2025-04-25
Add reset info to the activity heartbeat response

ychebotarev:
**What changed?**
Add two new flags to activity heartbeat response:
* reset_requested // indicates that reset happen while activity is still running
* reset_heartbeat //indicates that user asked to also reset heartbeats

**Why?**
Product request.

**Breaking changes**
No

Corresponding Server PR: https://github.com/temporalio/temporal/pull/7677

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -557,6 +557,10 @@ message RecordActivityTaskHeartbeatByIdResponse {
 
     // Will be set to true if the activity is paused.
     bool activity_paused = 2;
+
+    // Will be set to true if the activity was reset.
+    // Applies only to the current run.
+    bool activity_reset = 3;
 }
 
 message RespondActivityTaskCompletedRequest {
@@ -1844,13 +1848,13 @@ message UpdateActivityOptionsRequest {
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 5;
 
-    // either activity id or activity type must be provided
-    oneof activity {
-        // Only activity with this ID will be updated.
-        string id = 6;
-        // Update all running activities of this type.
-        string type = 7;
-    }
+        // either activity id or activity type must be provided
+        oneof activity {
+            // Only activity with this ID will be updated.
+            string id = 6;
+            // Update all running activities of this type.
+            string type = 7;
+        }
 }
 
 message UpdateActivityOptionsResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -529,6 +529,13 @@ message RecordActivityTaskHeartbeatResponse {
 
     // Will be set to true if the activity is paused.
     bool activity_paused = 2;
+
+    // Will be set to true if the activity was reset.
+    // Applyes only to the current run.
+    bool activity_reset = 3;
+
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -526,6 +526,13 @@ message RecordActivityTaskHeartbeatResponse {
 
     // Will be set to true if the activity is paused.
     bool activity_paused = 2;
+
+    // Will be set to true if the activity was reset.
+    // Applyes only to the current run.
```
cretz, line 531:
```suggestion
    // Applies only to the current run.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -526,6 +526,13 @@ message RecordActivityTaskHeartbeatResponse {
...
+    bool activity_reset = 3;
```
cretz, line 532: Can I get confirmation that we want SDKs to interrupt running activities when this is true, same as we do for paused or cancel requested?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -526,6 +526,13 @@ message RecordActivityTaskHeartbeatResponse {
...
+
+    // Will be set to true if the activity heartbeat was reset.
+    bool heartbeat_reset = 4;
```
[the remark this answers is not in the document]
ychebotarev: I want to provide all the information. 
->I think<- SDK should do nothing if this flag is not set. Also ->I think<- SDK should interrupt it if it is set.

We may end up interrupting it in both cases. Or not interrupting in both cases.
Or add another flag "interrupt requested".
Either way I share all the information we have at the moment, so we have all the options before us.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: My goal is to provide all the information. 
I think we need to decide it separately.

cretz: Would rather decide how we want SDK to react before merging this API (we obviously don't provide _all_ information on heartbeat, we only provide what callers need, there is a ton of stuff about an activity we leave off).

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
cretz: Would rather decide how we want SDK to react before merging this API (we obviously don't provide _all_ information on heartbeat, we only provide what callers need, there is a ton of stuff about an activity we leave off).

temporal/api/workflowservice/v1/request_response.proto
drewhoskins-temporal: @cretz that's right.

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
drewhoskins-temporal: Yeah, I don't see an immediate use case for it either.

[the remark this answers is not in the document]
ychebotarev: ok, will remove it then.

cretz (approved):

Sushisource (approved):

gow (approved):

PR #579, opened 2025-04-26
prevent misconfiguration of VersioningOverride and add more pinned options

carlydf:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/enums/v1/workflow.proto
+++ b/temporal/api/enums/v1/workflow.proto
@@ -183,3 +183,12 @@ enum VersioningBehavior {
     // complete on the old Version.
     VERSIONING_BEHAVIOR_AUTO_UPGRADE = 2;
 }
+
+enum PinnedBehaviorOverride {
+    // Keep original pinned behavior which might be pin run or execution
+    PINNED_BEHAVIOR_OVERRIDE_UNSPECIFIED = 0;
+    // Pin run to specific version
+    PINNED_BEHAVIOR_OVERRIDE_PIN_RUN = 1;
+    // Pin execution to specific version
+    PINNED_BEHAVIOR_OVERRIDE_PIN_EXECUTION = 2;
+}
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -531,18 +531,29 @@ message WorkflowExecutionOptions {
 // `UpdateWorkflowExecutionOptions` with a null `VersioningOverride`, and use the `update_mask`
 // to indicate that it should be mutated.
 message VersioningOverride {
+    oneof type {
+        PinnedOverrideBehavior pinned = 3;
+        bool auto_upgrade = 4;
+    }
     // Required.
+    // Deprecated. Use `type`.
     temporal.api.enums.v1.VersioningBehavior behavior = 1;
 
     // Required if behavior is `PINNED`. Must be null if behavior is `AUTO_UPGRADE`.
     // Identifies the worker deployment to pin the workflow to.
-    // Deprecated. Use `pinned_version`.
+    // Deprecated. Use `type.pinned.version`.
     temporal.api.deployment.v1.Deployment deployment = 2 [deprecated = true];
 
     // Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.
     // Identifies the worker deployment version to pin the workflow to, in the format
     // "<deployment_name>.<build_id>".
+    // Deprecated. Use `type.pinned.version`.
     string pinned_version = 9;
+}
+
+message PinnedOverrideBehavior {
+    temporal.api.enums.v1.PinnedBehaviorOverride type = 1;
+    string version = 2;
 }
 
 // When StartWorkflowExecution uses the conflict policy WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING and
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -11832,6 +11832,27 @@
       ],
       "default": "PENDING_WORKFLOW_TASK_STATE_UNSPECIFIED"
     },
+    "v1PinnedBehaviorOverride": {
+      "type": "string",
+      "enum": [
+        "PINNED_BEHAVIOR_OVERRIDE_UNSPECIFIED",
+        "PINNED_BEHAVIOR_OVERRIDE_PIN_RUN",
+        "PINNED_BEHAVIOR_OVERRIDE_PIN_EXECUTION"
+      ],
+      "default": "PINNED_BEHAVIOR_OVERRIDE_UNSPECIFIED",
+      "title": "- PINNED_BEHAVIOR_OVERRIDE_UNSPECIFIED: Keep original pinned behavior which might be pin run or execution\n - PINNED_BEHAVIOR_OVERRIDE_PIN_RUN: Pin run to specific version\n - PINNED_BEHAVIOR_OVERRIDE_PIN_EXECUTION: Pin execution to specific version"
+    },
+    "v1PinnedOverrideBehavior": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "$ref": "#/definitions/v1PinnedBehaviorOverride"
+        },
+        "version": {
+          "type": "string"
+        }
+      }
+    },
     "v1PollActivityTaskQueueResponse": {
       "type": "object",
       "properties": {
@@ -14544,17 +14565,23 @@
     "v1VersioningOverride": {
       "type": "object",
       "properties": {
+        "pinned": {
+          "$ref": "#/definitions/v1PinnedOverrideBehavior"
+        },
+        "autoUpgrade": {
+          "type": "boolean"
+        },
         "behavior": {
           "$ref": "#/definitions/v1VersioningBehavior",
-          "description": "Required."
+          "description": "Required.\nDeprecated. Use `type`."
         },
         "deployment": {
           "$ref": "#/definitions/v1Deployment",
-          "description": "Required if behavior is `PINNED`. Must be null if behavior is `AUTO_UPGRADE`.\nIdentifies the worker deployment to pin the workflow to.\nDeprecated. Use `pinned_version`."
+          "description": "Required if behavior is `PINNED`. Must be null if behavior is `AUTO_UPGRADE`.\nIdentifies the worker deployment to pin the workflow to.\nDeprecated. Use `type.pinned.version`."
         },
         "pinnedVersion": {
           "type": "string",
-          "description": "Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.\nIdentifies the worker deployment version to pin the workflow to, in the format\n\"<deployment_name>.<build_id>\"."
+          "description": "Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.\nIdentifies the worker deployment version to pin the workflow to, in the format\n\"<deployment_name>.<build_id>\".\nDeprecated. Use `type.pinned.version`."
         }
       },
       "description": "Used to override the versioning behavior (and pinned deployment version, if applicable) of a\nspecific workflow execution. If set, takes precedence over the worker-sent values. See\n`WorkflowExecutionInfo.VersioningInfo` for more information. To remove the override, call\n`UpdateWorkflowExecutionOptions` with a null `VersioningOverride`, and use the `update_mask`\nto indicate that it should be mutated."
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -8953,6 +8953,18 @@ components:
         attempt:
           type: integer
           format: int32
+    PinnedOverrideBehavior:
+      type: object
+      properties:
+        type:
+          enum:
+            - PINNED_BEHAVIOR_OVERRIDE_UNSPECIFIED
+            - PINNED_BEHAVIOR_OVERRIDE_PIN_RUN
+            - PINNED_BEHAVIOR_OVERRIDE_PIN_EXECUTION
+          type: string
+          format: enum
+        version:
+          type: string
     PollWorkflowTaskQueueResponse:
       type: object
       properties:
@@ -11845,13 +11857,19 @@ components:
     VersioningOverride:
       type: object
       properties:
+        pinned:
+          $ref: '#/components/schemas/PinnedOverrideBehavior'
+        autoUpgrade:
+          type: boolean
         behavior:
           enum:
             - VERSIONING_BEHAVIOR_UNSPECIFIED
             - VERSIONING_BEHAVIOR_PINNED
             - VERSIONING_BEHAVIOR_AUTO_UPGRADE
           type: string
-          description: Required.
+          description: |-
+            Required.
+             Deprecated. Use `type`.
           format: enum
         deployment:
           allOf:
@@ -11859,13 +11877,14 @@ components:
           description: |-
             Required if behavior is `PINNED`. Must be null if behavior is `AUTO_UPGRADE`.
              Identifies the worker deployment to pin the workflow to.
-             Deprecated. Use `pinned_version`.
+             Deprecated. Use `type.pinned.version`.
         pinnedVersion:
           type: string
           description: |-
             Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.
              Identifies the worker deployment version to pin the workflow to, in the format
              "<deployment_name>.<build_id>".
+             Deprecated. Use `type.pinned.version`.
       description: |-
         Used to override the versioning behavior (and pinned deployment version, if applicable) of a
          specific workflow execution. If set, takes precedence over the worker-sent values. See


temporal/api/enums/v1/workflow.proto
```diff
@@ -183,3 +183,12 @@ enum VersioningBehavior {
     // complete on the old Version.
     VERSIONING_BEHAVIOR_AUTO_UPGRADE = 2;
 }
+
+enum PinnedBehaviorOverride {
```
cretz, line 187: Can you qualify this with `Versioning` since "pinning" is a bit generic?

```suggestion
enum VersioningPinnedBehaviorOverride {
```

temporal/api/workflow/v1/message.proto
```diff
@@ -531,20 +531,31 @@ message WorkflowExecutionOptions {
...
     temporal.api.deployment.v1.Deployment deployment = 2 [deprecated = true];
 
     // Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.
     // Identifies the worker deployment version to pin the workflow to, in the format
     // "<deployment_name>.<build_id>".
+    // Deprecated. Use `type.pinned.version`.
     string pinned_version = 9;
 }
 
+message PinnedOverrideBehavior {
```
cretz, line 554: Can you either qualify this with the term `Versioning` or move the message to inside the `VersioningOverride` message?

temporal/api/workflow/v1/message.proto
```diff
@@ -531,20 +531,31 @@ message WorkflowExecutionOptions {
 // `UpdateWorkflowExecutionOptions` with a null `VersioningOverride`, and use the `update_mask`
 // to indicate that it should be mutated.
 message VersioningOverride {
+    oneof type {
```
cretz, line 534: Can you document what "type" means and what the different values here mean?

temporal/api/workflow/v1/message.proto
Sushisource: +1 to inlining it

temporal/api/workflow/v1/message.proto
```diff
@@ -531,20 +531,31 @@ message WorkflowExecutionOptions {
...
     // Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.
     // Identifies the worker deployment version to pin the workflow to, in the format
     // "<deployment_name>.<build_id>".
+    // Deprecated. Use `type.pinned.version`.
     string pinned_version = 9;
 }
 
+message PinnedOverrideBehavior {
+    temporal.api.enums.v1.PinnedBehaviorOverride type = 1;
+    string version = 2;
```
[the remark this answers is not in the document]
cretz: Especially with the much-more-frequent releases from `master` occurring now, would definitely suggest landing as many things together for a project as possible

temporal/api/enums/v1/workflow.proto
carlydf: I inlined it

temporal/api/workflowservice/v1/service.proto
```diff
@@ -857,7 +857,7 @@ service WorkflowService {
     // Version if it is the Version being set as Current.
     // Experimental. This API might significantly change or be removed in a future release.
     rpc SetWorkerDeploymentCurrentVersion (SetWorkerDeploymentCurrentVersionRequest) returns (SetWorkerDeploymentCurrentVersionResponse) {
-        option (google.api.http) = {
+        option (google.api.http) = { // TODO(carlydf): replace deployment_name with appropriate thing here (how to handle one-of?)
```
carlydf, line 860: @Sushisource Unfortunately, fields within a oneof cannot be referenced in the http spec, so I think we may have to keep the `deployment_name` field at the top level and just say "MUST BE EQUAL TO THE INTERIOR FIELDS" (or something along those lines)

Sushisource: We've got oneofs in bunches of other places? This seems like a totally lame reason to have to compromise the proto

temporal/api/enums/v1/workflow.proto
cretz: While inlining makes sense in many cases, we haven't ever done it for `enum`s. In fact, we have never defined an enum outside of this package. While I don't have a strong reason to continue to do this, I think we need broad consensus that we are ok breaking this trend. I personally am ok with this. @alexshtin or @dnr - any opinions?

temporal/api/workflowservice/v1/service.proto
cretz: To clarify, a oneof that _also_ represents a part of the HTTP _path_ is where the issue lies I believe. Which makes sense in some cases (additional bindings notwithstanding) - an HTTP path match based on a proto field value needs to have that value unambiguously present.

temporal/api/deployment/v1/message.proto
```diff
@@ -207,13 +209,32 @@ message WorkerDeploymentInfo {
...
+        // Deprecated. Use `deployment_version`.
+        string version = 1 [deprecated = true];
+
+        WorkerDeploymentVersion deployment_version = 4;
         google.protobuf.Timestamp create_time = 2;
         enums.v1.VersionDrainageStatus drainage_status = 3;
     }
 }
 
+// A Worker Deployment Version (Version, for short) represents a different
```
Sushisource, line 221:
```suggestion
// A Worker Deployment Version (Version, for short) represents a
```
A version is different from other versions sort of tautologically. No need to call it out.

temporal/api/workflowservice/v1/service.proto
carlydf: Since we don't use the `deployment_name` for routing or authorization, including it in the http path is not a technical necessity. Given that, I propose simply removing it.

temporal/api/workflow/v1/message.proto
```diff
@@ -542,18 +549,43 @@ message WorkflowExecutionOptions {
...
 
     // Required if behavior is `PINNED`. Must be absent if behavior is not `PINNED`.
     // Identifies the worker deployment version to pin the workflow to, in the format
     // "<deployment_name>.<build_id>".
-    string pinned_version = 9;
+    // Deprecated. Use `override.pinned.version`.
+    string pinned_version = 9 [deprecated = true];
+
+    message PinnedOverride {
+        PinnedOverrideBehavior type = 1;
```
Sushisource, line 576:
```suggestion
        PinnedOverrideBehavior behavior = 1;
```

I think this is probably a bit more consistent

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2023,12 +2024,16 @@ message SetCurrentDeploymentResponse {
...
-    string version = 3;
+    // Deprecated. Use `current`.
+    string deployment_name = 2 [deprecated = true];
+    // Deprecated. Use `current`.
+    string version = 3 [deprecated = true];
+
+    // Required.
+    oneof current {
+        temporal.api.deployment.v1.WorkerDeploymentVersion deployment_version = 7;
+        string unversioned_deployment_name = 8;
```
Sushisource, line 2036: Worth explaining (even though the name is pretty clear) that this string needs to be the deployment name portion of a workerdeploymentversion (same for ramping).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2057,21 +2062,27 @@ message SetWorkerDeploymentCurrentVersionResponse {
     // that write to the Worker Deployment state to ensure that the state
     // did not change between this API call and a future write.
     bytes conflict_token = 1;
-    // The version that was current before executing this operation, in the form
-    // "<deployment_name>.<build_id>". Can also be the `__unversioned__` special value.
-    string previous_version = 2;
+    // Deprecated. Use `previous_current`.
+    string previous_version = 2 [deprecated = true];
+    // The version that was current before executing this operation.
+    oneof previous_current {
```
Sushisource, line 2069: Previous current feels a bit oxymoronic.
```suggestion
    oneof replaced_version {
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2106,9 +2117,13 @@ message SetWorkerDeploymentRampingVersionResponse {
     // that write to the Worker Deployment state to ensure that the state
     // did not change between this API call and a future write.
     bytes conflict_token = 1;
-    // The version that was ramping before executing this operation, in the form
-    // "<deployment_name>.<build_id>". Can also be the `__unversioned__` special value.
-    string previous_version = 2;
+    // Deprecated. Use `previous_ramping`.
+    string previous_version = 2 [deprecated = true];
+    // The version that was ramping before executing this operation.
+    oneof previous_ramping {
```
Sushisource, line 2125: If we accept my other suggestion, do this to be consistent. We don't have to do them.
```suggestion
    oneof replaced_ramping {
```

Sushisource (approved):
Nice! Much safer. Having all these deprecated fields is slightly sad, but, that's life.

None of my comments are blocking.

temporal/api/workflow/v1/message.proto
```diff
@@ -542,18 +553,43 @@ message WorkflowExecutionOptions {
...
+    }
+
+    enum PinnedOverrideBehavior {
+        // If original behavior is Pinned or PinnedUntilContinueAsNew, keep that behavior.
+        // If original behavior is AutoUpgrade, override behavior to Pinned.
+        PINNED_OVERRIDE_BEHAVIOR_UNSPECIFIED = 0;
+        // Override workflow behavior to be Pinned.
+        PINNED_OVERRIDE_BEHAVIOR_PINNED = 1;
+        // Override workflow behavior to be PinnedUntilContinueAsNew.
+        PINNED_OVERRIDE_BEHAVIOR_PINNED_UNTIL_CONTINUE_AS_NEW = 2;
```
carlydf, line 591: TODO: remove this before committing, because this "mode" won't be implemented in Public Preview. I'm just including it here so that it's clear why this override structure is needed.

Shivs11 (approved):

temporal/api/workflowservice/v1/service.proto
cretz: Makes sense to me. If a resource is not simply/uniquely identified by a literal `string` in the request, shouldn't be on the path.

temporal/api/workflow/v1/message.proto
```diff
@@ -240,9 +241,15 @@ message DeploymentTransition {
...
 // Experimental. Might change in the future.
 message DeploymentVersionTransition {
-    // Required. The target Version of the transition. May be `__unversioned__` which means a
+    // Deprecated. Use `target`.
+    string version = 1 [deprecated = true];
+
+    // Required. The target Version of the transition. If `unversioned`, a
     // so-far-versioned workflow is transitioning to unversioned workers.
-    string version = 1;
+    oneof target {
```
carlydf, line 249: need to fix this

carlydf: one sec

temporal/api/deployment/v1/message.proto
```diff
@@ -207,41 +209,79 @@ message WorkerDeploymentInfo {
...
+}
+
 message VersionMetadata {
     // Arbitrary key-values.
     map<string, temporal.api.common.v1.Payload> entries = 1;
 }
 
 message RoutingConfig {
-    // Always present. Specifies which Deployment Version should should receive new workflow
+    // When present, specifies which Deployment Version should should receive new workflow
```
carlydf, line 244: Nil version is semantically replacing what we used to call "unversioned current version"

This is because, when Current Version is set to nil (or previously "unversioned") the Deployment does not really know that tasks on the task queues previously present in that Deployment are definitely being sent to unversioned workers. All the Deployment knows is that those task queues are sending tasks somewhere else (could be unversioned workers, if the task queue has no current version, or could be another version, if the task queue became current on another Deployment)

carlydf: Even though a task queue changing from one deployment to another is a rare case, we want to have an invariant that is true all the time. With the new nil current_deployment_version meaning, this statement is true:
 **_If a TQ is not part of any current or ramping version in any deployment, unversioned workers get all tasks_**

That is always true no matter what is happening with the Deployments that the Task Queue is/was in

So, "ramp to unversioned" becomes "deployment is not concerning itself with the target"

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2066,21 +2078,52 @@ message SetWorkerDeploymentCurrentVersionResponse {
...
+        //   ```
+        //   current_deployment_version = nil
+        //   ramping_deployment_version = V
+        //   ramping_version_percentage = N
+        //   ```
+        // Ramping N% of traffic _from_ a particular Version V in this Deployment _to_ outside this Deployment's scope:
+        //   ```
+        //   current_deployment_version = nil
+        //   ramping_deployment_version = V
+        //   ramping_version_percentage = (100 - N)
```
Sushisource, line 2119: I'm guessing there was some discussion on this I missed but this seems needlessly confusing to me. Whether it's within or outside of the current deployment, you want to move some % of traffic to this other version. Why do I have to think about if that % should be inverted or not?

This applies to all these spots on the read side too where the percentage is inverted conditionally.

I'm still not sure I even understand what's even going on here, but nominally it's that you want to shift traffic away from some deployment _without_ having to pass in the current version? I thought there could only be one ramp defined at a time - so how could we need to ramp traffic _away_ from some deployment to the current version? 

Anyway, as per slack convo sounds like we'll want to discuss this. If we could consider adding this later, (not sure what use case requires it?) that's an option too

cretz (approved):
LGTM from a pure structural POV, but will want to wait for @Sushisource approval. It is unfortunate to have so many versioning deprecations overall on the API, I wonder if there are preventative things we can do in the future on other projects.

temporal/api/deployment/v1/message.proto
```diff
@@ -207,41 +209,79 @@ message WorkerDeploymentInfo {
...
+    //   ```
+    // Ramping N% of traffic _from_ a particular Version V in this Deployment _to_ outside this Deployment's scope:
+    //   ```
+    //   current_deployment_version = nil
+    //   ramping_deployment_version = V
+    //   ramping_version_percentage = (100 - N)
+    //   ```
+    temporal.api.deployment.v1.WorkerDeploymentVersion ramping_deployment_version = 8;
+
+    // Deprecated. Use `ramping`.
```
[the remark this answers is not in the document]
carlydf: fixed!

temporal/api/deployment/v1/message.proto
```diff
@@ -207,41 +209,79 @@ message WorkerDeploymentInfo {
...
+// Experimental. Worker Deployment Versions are experimental and might significantly change in the future.
+message WorkerDeploymentVersion {
+    // A unique identifier for this Version within the Deployment it is a part of.
+    // Not necessarily unique within the namespace.
+    // The combination of `deployment_name` and `build_id` uniquely identifies
+    // this Version within the namespace.
+    string build_id = 1;
+
+    // Identifies the Worker Deployment this Version is part of.
+    // Must be unique within the namespace.
```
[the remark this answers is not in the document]
carlydf: I changed the explanation; is this more clear?
```
message WorkerDeploymentVersion {
    // A unique identifier for this Version within the Deployment it is a part of.
    // Not necessarily unique within the namespace.
    // The combination of `deployment_name` and `build_id` uniquely identifies this
    // Version within the namespace, because Deployment names are unique within a namespace.
    string build_id = 1;

    // Identifies the Worker Deployment this Version is part of.
    string deployment_name = 2;
}
```

temporal/api/workflow/v1/message.proto
```diff
@@ -169,17 +169,18 @@ message WorkflowExecutionVersioningInfo {
...
     // versioned if an execution is versioned or not is via the `behavior` field.
     // Note that `deployment` is overridden by `versioning_override` if the latter is present.
-    // Deprecated. Use `version`.
+    // Deprecated. Use `deployment_version`.
     temporal.api.deployment.v1.Deployment deployment = 2 [deprecated = true];
-    // The Worker Deployment Version that completed the last workflow task of this workflow
-    // execution, in the form "<deployment_name>.<build_id>".
+    // Deprecated. Use `deployment_version`.
+    temporal.api.deployment.v1.WorkerDeploymentVersion version = 5 [deprecated = true];
+    // The Worker Deployment Version that completed the last workflow task of this workflow execution.
```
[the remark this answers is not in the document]
carlydf: a mistake!

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2074,21 +2079,24 @@ message SetWorkerDeploymentCurrentVersionResponse {
...
-    // - Or, the "__unversioned__" special value, to represent all the unversioned workers (those
-    //   with `UNVERSIONED` (or unspecified) `WorkerVersioningMode`.)
-    string version = 3;
+    // Deprecated. Use `build_id`.
+    string version = 3 [deprecated = true];
+
+    // The build id of the Version that you want to ramp traffic to.
+    // Pass an empty value to set the Ramping Version to nil.
+    // A nil Ramping Version represents all the unversioned workers (those with `UNVERSIONED` (or unspecified) `WorkerVersioningMode`.)
+    string build_id = 8;
```
Sushisource, line 2098: Arguably we could deprecate `deployment_name` too and use the full type here as well, but, I'm fine either way.

Sushisource (approved):
Great, happy with this and 10x more obvious than the previous iteration

ShahabT (approved):

temporal/api/workflow/v1/message.proto
```diff
@@ -542,18 +553,47 @@ message WorkflowExecutionOptions {
...
+        temporal.api.deployment.v1.WorkerDeploymentVersion version = 2;
+    }
+
+    enum PinnedOverrideBehavior {
+        // If original behavior is Pinned or PinnedUntilContinueAsNew, keep that behavior.
+        // If original behavior is AutoUpgrade, override behavior to Pinned.
+        PINNED_OVERRIDE_BEHAVIOR_UNSPECIFIED = 0;
+
+        // Not yet supported. Reserved for future use.
+        // Override workflow behavior to be Pinned.
```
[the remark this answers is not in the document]
carlydf: My plan was to only support the default `UNSPECIFIED` behavior in this first change, which means 
```
// If original behavior is Pinned or PinnedUntilContinueAsNew, keep that behavior.
// If original behavior is AutoUpgrade, override behavior to Pinned.
```
because I felt that that maps best to what we currently have (in the world without 2 different types of pinned).

but arguably, `// Override workflow behavior to be Pinned.` (in the new world) is more equivalent to "pinned override" in the old world..

I'm fine with either, but "old-world pinned override" should default to exactly one of these three types

temporal/api/history/v1/message.proto
```diff
@@ -256,11 +265,17 @@ message WorkflowTaskCompletedEventAttributes {
...
+    // Deprecated. Replaced with `deployment_version`.
+    string worker_deployment_version = 9 [deprecated = true];
     // The name of Worker Deployment that completed this task. Must be set if `versioning_behavior`
     // is set. This value updates workflow execution's `worker_deployment_name`.
     // Experimental. Worker Deployments are experimental and might significantly change in the future.
-    string worker_deployment_name = 10;
+    // Deprecated. Use `deployment_version`.
+    string worker_deployment_name = 10 [deprecated = true];
+    // The Worker Deployment Version that completed this task. Must be set if `versioning_behavior`
+    // is set. This value updates workflow execution's `versioning_info.worker_deployment_version`.
```
carlydf, line 276:
```suggestion
    // is set. This value updates workflow execution's `versioning_info.deployment_version`.
```

temporal/api/history/v1/message.proto
```diff
@@ -374,8 +392,11 @@ message ActivityTaskCompletedEventAttributes {
     // id of the worker that completed this task
     string identity = 4;
     // Version info of the worker who processed this workflow task.
-    // Deprecated. Use the info inside the corresponding ActivityTaskStartedEvent
```
carlydf, line 377: @ShahabT There is no more versioning info in the ActivityTaskStartedEvent, since we deprecated the version stamp and didn't replace it with anything. 

We can ask the worker that is calling `ActivityTaskCompleted` to send a deployment options, and then turn that into a version, or we could do nothing. AFAIK, we store in the workflow execution info . versioning info the version that completed the last _workflow_ task, but we do not store the version that completed the last activity task. So I forget how we handle versioning of activities, or if we even need to know what worker completed the activity task. But either way the comment "Use the info inside the corresponding ActivityTaskStartedEvent" needs to be corrected.

temporal/api/history/v1/message.proto
```diff
@@ -389,8 +410,11 @@ message ActivityTaskFailedEventAttributes {
     string identity = 4;
     temporal.api.enums.v1.RetryState retry_state = 5;
     // Version info of the worker who processed this workflow task.
-    // Deprecated. Use the info inside the corresponding ActivityTaskStartedEvent
```
carlydf, line 392: same comment as above, ActivityTaskStartedEvent has no version stamp now

temporal/api/history/v1/message.proto
```diff
@@ -424,8 +448,11 @@ message ActivityTaskCanceledEventAttributes {
     // id of the worker who canceled this activity
     string identity = 5;
     // Version info of the worker who processed this workflow task.
-    // Deprecated. Use the info inside the corresponding ActivityTaskStartedEvent
```
carlydf, line 427: same comment as above, ActivityTaskStartedEvent has no version stamp now

temporal/api/history/v1/message.proto
carlydf: There is a `last_deployment_version` in `PendingActivityInfo`

```
    // The Worker Deployment Version this activity was dispatched to most recently.
    // If nil, the activity has not yet been dispatched or was last dispatched to an unversioned worker.
    temporal.api.deployment.v1.WorkerDeploymentVersion last_deployment_version = 24;
```

I'm not yet sure when it's populated or what it's used for

temporal/api/workflow/v1/message.proto
```diff
@@ -93,7 +93,8 @@ message WorkflowExecutionInfo {
 
     // The name of Worker Deployment that completed the most recent workflow task.
     // Experimental. Worker Deployments are experimental and might change in the future.
-    string worker_deployment_name = 23;
+    // Deprecated. Replaced with `versioning_info.deployment_version`.
```
carlydf, line 96: undo this deprecation too

PR #580, opened 2025-04-28
Add `status` to StartWorkflowExecutionResponse

stephanos:

**What changed?**

Added field `WorkflowExecutionStatus status` to `StartWorkflowExecutionResponse`.

**Why?**

To align Update-with-Start's behavior with a regular Update. A regular Update returns its outcome from a closed Workflow, to make Update-with-Start do the same thing, it needs to returns a successful response in that case. But that would tell the user that a running Workflow exists. To help the user distinguish whether UwS operated on a closed or running workflow, this field is introduced.

**Breaking changes**

No.

**Server PR**

https://github.com/temporalio/temporal/pull/7656

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -221,6 +221,8 @@ message StartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 3;
+    // If true, the workflow is running.
+    bool running = 5;
     // When `request_eager_execution` is set on the `StartWorkflowExecutionRequest`, the server - if supported - will
     // return the first workflow task to be eagerly executed.
     // The caller is expected to have a worker available to process the task.
@@ -1810,6 +1812,9 @@ message ExecuteMultiOperationResponse {
 
     message Response {
         oneof response {
+            // Note that for the combination [StartWorkflow, UpdateWorkflow], aka Update-with-Start, when the Update's
+            // outcome for the requested Update ID is available but the workflow has already been closed, the
+            // StartWorkflowExecutionResponse will set `running` to `false`.
             StartWorkflowExecutionResponse start_workflow = 1;
             UpdateWorkflowExecutionResponse update_workflow = 2;
         }
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -221,8 +221,9 @@ message StartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 3;
-    // If true, the workflow is running.
-    bool running = 5;
+    // Status of the workflow. Note that it will be WORKFLOW_EXECUTION_STATUS_RUNNING unless used within an
+    // `ExecuteMultiOperation` (see more details there).
+    temporal.api.enums.v1.WorkflowExecutionStatus status = 5;
     // When `request_eager_execution` is set on the `StartWorkflowExecutionRequest`, the server - if supported - will
     // return the first workflow task to be eagerly executed.
     // The caller is expected to have a worker available to process the task.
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -221,6 +221,8 @@ message StartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 3;
+    // If true, the workflow is running.
```
stephanos, line 224: Not sure what else to say here. I'm not sure I should to go into MultiOps here. But it is (for now) the only reason we're adding this.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -221,6 +221,8 @@ message StartWorkflowExecutionResponse {
...
+    bool running = 5;
```
cretz, line 225: The problem with `running` is that it defaults to `false` for older APIs/SDKs/servers when this field does not exist. We should try to make new booleans on existing APIs have the default value of `false` be the primary one. Would you consider a value where `false` is the default and `true` is the abnormal state, e.g. `not_running` or `used_last_closed` or something?

This is not _that_ important to me FWIW, just bringing it up (will mark approved if we _really_ want to leave it as is)

temporal/api/workflowservice/v1/request_response.proto
cretz: May be worth noting that, up to you. I think it's valuable information to know this is always true except for multi-op. So long as the comment doesn't grow stale if we do something to regular start to cause this to be false too.

temporal/api/workflowservice/v1/request_response.proto
stephanos: @cretz is the `StartWorkflowExecutionResponse` exposed to users directly in SDKs? I've only thought of Go so far where it's wrapped; not even the `started` is exposed there.

The other thought I had was whether we could/should return `WorkflowExecutionStatus` instead. It gives more information - although I'm not sure there's a use case - and defaults to "unspecified".

dandavison: > whether we could/should return WorkflowExecutionStatus

That does seem like it could make sense. It's already queryable via `DescribeWorkflowExecution`, so this would be giving the same thing but "as of the time that the server processed the StartWorkflowRequest"

Sushisource: I think returning the status is a great idea, and the default of unspecified works perfectly when a new SDK targets an old server.

cretz: > @cretz is the StartWorkflowExecutionResponse exposed to users directly in SDKs? I've only thought of Go so far where it's wrapped; not even the started is exposed there.

No, not directly, though it is always technically available. When we add high-level running/started to users and they have older server they will both look like they are false, but probably not a big deal. I too like the status approach.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1810,6 +1812,9 @@ message ExecuteMultiOperationResponse {
 
     message Response {
         oneof response {
+            // Note that for the combination [StartWorkflow, UpdateWorkflow], aka Update-with-Start, when the Update's
+            // outcome for the requested Update ID is available but the workflow has already been closed, the
+            // StartWorkflowExecutionResponse will set `running` to `false`.
             StartWorkflowExecutionResponse start_workflow = 1;
             UpdateWorkflowExecutionResponse update_workflow = 2;
         }
```
stephanos, line 1820: One more thing I just noticed; there's of course also `SignalWithStartWorkflowExecutionResponse`. I'm inclined to _not_add the new field there since there's no use case.

cretz (approved):
Approving with boolean as is (though may want to get consensus with others), though if we do introduce status instead would like a chance to approve that independently

alexshtin (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1805,6 +1808,12 @@ message ExecuteMultiOperationRequest {
     }
 }
 
+// IMPORTANT: For [StartWorkflow, UpdateWorkflow] combination ("Update-with-Start") when both
+//   1. the workflow update for the requested update ID has already completed, and
+//   2. the workflow for the requested workflow ID has already been closed,
+// then you'll receive
+//   - an update response containing the update's outcome, and
+//   - a start response with a `status` field that reflects the workflow's current state.
```
cretz, line 1816: Out of curiosity, what status enums might happen here? Do we reach in to the most-recent-workflow-run for update IDs for all cases of most recent? So if I had a several month old workflow run that failed after accepting an update, we will not start a new update?

cretz (approved):

Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -221,6 +221,9 @@ message StartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 3;
+    // Status of the workflow. Note that it will be WORKFLOW_EXECUTION_STATUS_RUNNING unless used within an
+    // `ExecuteMultiOperation` (see more details there).
```
stephanos, line 225: Actually, I just noticed that this can be not "Running" outside of MultiOps when we dedup the Start (ie same request ID); see https://github.com/temporalio/temporal/pull/7685/files#diff-4cde1cd2e50c75036c3cdbd0732f1f884d1ec81048904c9e90e125e9bd670a55R330

In that case it seems incorrect to return "Running" when in fact, it may not.

stephanos: I've updated the comment accordingly now. I'll go ahead with the merge, but I'm happy to revise the comment again if there's feedback.

PR #581, opened 2025-04-29
Define post reset operations

gow:
We are adding support to performing few specific operations when a workflow is reset. So this PR is adding proto definitions for the supported operations. In this first iteration we'll include updating workflow options and signal operations.

We want to allow users to perform operations like "reset + move_worker_version" or "reset + signal" atomically at the time of reset. We decided to keep the gRPC definition flexible to be able to add more operations in the future.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -820,6 +821,9 @@ message ResetWorkflowExecutionRequest {
     temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 6;
     // Event types not to be reapplied
     repeated temporal.api.enums.v1.ResetReapplyExcludeType reset_reapply_exclude_types = 7;
+    // Operations to perform after the workflow has been reset. These operations will be applied
+    // to the *new* run of the workflow execution in the order they are provided.
```
Sushisource, line 825: I think it's important for this comment to include some information about the atomicity guarantees (or lack thereof) provided by this.

IE (and, I'm guessing at the behavior here, but I hope this is the case):

"All operations are applied to the workflow _before_ the first new workflow task is generated"

gow: Agree. Added.

openapi/openapiv2.json
```diff
@@ -7309,6 +7309,14 @@
...
           },
           "title": "Event types not to be reapplied"
+        },
+        "postResetOperations": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "$ref": "#/definitions/v1PostResetOperation"
+          },
+          "title": "Operations to perform after the workflow has been reset. These operations will be applied\nto the *new* run of the workflow execution in the order they are provided.\nAll operations are applied to the workflow before the first new workflow task is generated"
```
drewhoskins-temporal, line 7319: I'm curious about atomicity here.  Suppose there's a validation failure on update_options--will that be checked before the reset is applied?

openapi/openapiv2.json
```diff
@@ -12094,6 +12102,50 @@
...
+        }
+      },
+      "description": "PostResetSignalOperation represents sending a signal after a workflow reset."
+    },
+    "v1PostResetUpdateWorkflowOptionsOperation": {
+      "type": "object",
+      "properties": {
+        "workflowExecutionOptions": {
+          "$ref": "#/definitions/v1WorkflowExecutionOptions",
+          "description": "Workflow Execution options. Partial updates are accepted and controlled by update_mask."
```
drewhoskins-temporal, line 12140: nit: first part of the description adds no value.  Maybe instead, "Update Workflow options that were originally specified via `StartWorkflowExecution`.
(I'm guessing you copied this--ff to go update the original too!)

temporal/api/reset/v1/post_reset_operations.proto
```diff
@@ -0,0 +1,41 @@
...
+    oneof operation {
+        temporal.api.reset.v1.PostResetSignalOperation signal_operation = 1;
+        temporal.api.reset.v1.PostResetUpdateWorkflowOptionsOperation update_workflow_options_operation = 2;
+    }
+}
+
+// PostResetSignalOperation represents sending a signal after a workflow reset.
+message PostResetSignalOperation {
+    // The workflow author-defined name of the signal to send to the workflow.
+    string signal_name = 1;
```
drewhoskins-temporal, line 27: nit: s/signal_name/name/ ?

temporal/api/reset/v1/post_reset_operations.proto
```diff
@@ -0,0 +1,41 @@
...
+option java_package = "io.temporal.api.reset.v1";
+option java_multiple_files = true;
+option java_outer_classname = "PostResetOperationsProto";
+option ruby_package = "Temporalio::Api::Reset::V1";
+option csharp_namespace = "Temporalio.Api.Reset.V1";
+
+// PostResetOperation represents an operation to be performed on the new workflow execution after a workflow reset.
+message PostResetOperation {
+    oneof operation {
+        temporal.api.reset.v1.PostResetSignalOperation signal_operation = 1;
```
drewhoskins-temporal, line 19: nit: This is cool and has been an occasional feature request, but note that this is a bit lower in priority, so ff to break it out into a separate PR when we're ready to execute it.  

Offline, we can prioritize between signal, pause activity type/update activity type.

openapi/openapiv2.json
gow: Yes. The API params will be validated before applying performing operation. If there is a parameter validation failure, the reset request fails. I'm not completely sure about the error messages returned. Most likely I'll return the error message from the existing param validation code.

openapi/openapiv2.json
gow: Yes. Copy-pasted. Will fix the other ones too.

temporal/api/reset/v1/post_reset_operations.proto
gow: Yes. I'll prioritize update workflow options over this one on server. I'll leave this one here for now since the feature is already there, the effort to support it is relatively minimal.

temporal/api/reset/v1/post_reset_operations.proto
```diff
@@ -0,0 +1,41 @@
...
+
+option go_package = "go.temporal.io/api/reset/v1;reset";
+option java_package = "io.temporal.api.reset.v1";
+option java_multiple_files = true;
+option java_outer_classname = "PostResetOperationsProto";
+option ruby_package = "Temporalio::Api::Reset::V1";
+option csharp_namespace = "Temporalio.Api.Reset.V1";
+
+// PostResetOperation represents an operation to be performed on the new workflow execution after a workflow reset.
+message PostResetOperation {
```
bergundy, line 17: We discussed reusing the requests from WorkflowService and noting that the identifiers on those requests will not be used.

If you do not take that approach, you should add a note to the request objects and these operation definitions to ask authors to keep the definitions in sync.

temporal/api/reset/v1/post_reset_operations.proto
```diff
@@ -0,0 +1,41 @@
...
+    oneof operation {
```
bergundy, line 18: I typically like to call fields like this `variant` to avoid stuttering in the code (`operation.operation` reads worse than `operation.variant`).

temporal/api/reset/v1/post_reset_operations.proto
bergundy, line 19: Can drop the `_operation` suffix since it can be inferred from the context.

temporal/api/reset/v1/post_reset_operations.proto
gow: I misunderstood then. I thought since there are many repeated and mandatory fields (ex: namespace, workflow_execution, request_id, identity etc) in the original request definitions, we wanted to remove them and keep this minimal.
Let me use the `SignalWorkflowExecutionRequest` and `UpdateWorkflowExecutionOptionsRequest` directly here and see how that looks and if causes any issues/confusion.

Curious - why didn't we take this approach in other operations and decided to duplicate proto definitions? (Ex: batch operation)

temporal/api/reset/v1/post_reset_operations.proto
gow: Done.

temporal/api/reset/v1/post_reset_operations.proto
gow: I like this. Done.

temporal/api/reset/v1/post_reset_operations.proto
gow: Apparently `name` is special and there are linter rules that complain about this - https://google.aip.dev/122.
We can override the linter. But for consistency, we should leave it `signal_name` like in other places.

openapi/openapiv2.json
gow: Done.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -820,6 +820,18 @@ message ResetWorkflowExecutionRequest {
...
+    // All operations are applied to the workflow before the first new workflow task is generated
+    repeated PostResetOperation post_reset_operations = 8;
+}
+
+// PostResetOperation represents an operation to be performed on the new workflow execution after a workflow reset.
+message PostResetOperation {
```
cretz, line 830: This needs to be a child of `ResetWorkflowExecutionRequest` IMO. The only top-level messages in this package should be request/response messages.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -820,6 +820,18 @@ message ResetWorkflowExecutionRequest {
     temporal.api.enums.v1.ResetReapplyType reset_reapply_type = 6;
     // Event types not to be reapplied
     repeated temporal.api.enums.v1.ResetReapplyExcludeType reset_reapply_exclude_types = 7;
+    // Operations to perform after the workflow has been reset. These operations will be applied
+    // to the *new* run of the workflow execution in the order they are provided.
+    // All operations are applied to the workflow before the first new workflow task is generated
+    repeated PostResetOperation post_reset_operations = 8;
```
cretz, line 826: What about `temporal.api.batch.v1.BatchOperationReset`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -820,6 +820,18 @@ message ResetWorkflowExecutionRequest {
...
+}
+
+// PostResetOperation represents an operation to be performed on the new workflow execution after a workflow reset.
+message PostResetOperation {
+    oneof variant {
+        SignalWorkflowExecutionRequest signal = 1;
+        UpdateWorkflowExecutionOptionsRequest update_workflow_options = 2;
```
cretz, line 833: It is a bit confusing on what fields are accepted in these requests and what are not. Can you be clear in docs on what is/isn't accepted and can we make sure to strictly validate that certain things are not accepted (e.g. namespace or workflow execution or whatever)?

temporal/api/reset/v1/post_reset_operations.proto
gow: I used `SignalWorkflowExecutionRequest` and `UpdateWorkflowExecutionOptionsRequest`. But that didn't look right since each of them will bring unwanted and potentially ambiguous fields. So reverted that change.

temporal/api/workflowservice/v1/request_response.proto
gow: This is removed from here with the revert. It now lives in `temporal.api.reset.v1.PostResetOperation`

temporal/api/workflowservice/v1/request_response.proto
gow: Yes. This feature is needed in batch operations too. Added it there.

temporal/api/workflowservice/v1/request_response.proto
gow: Yes. I reverted this change to go back to an earlier version where I defined messages with a subset of fields that are necessary for this feature. Please take a look now.

temporal/api/reset/v1/post_reset_operations.proto
cretz, line 1: IMO such a narrow need does not deserve its own new proto package. We have many things that are related to reset in other packages, including `ResetOptions` in `common`.

temporal/api/reset/v1/post_reset_operations.proto
```diff
@@ -0,0 +1,45 @@
...
+message PostResetOperation {
+    oneof variant {
+        temporal.api.reset.v1.PostResetSignalOperation signal = 1;
+        temporal.api.reset.v1.PostResetUpdateWorkflowOptionsOperation update_workflow_options = 2;
+    }
+}
+
+// PostResetSignalOperation represents sending a signal after a workflow reset.
+// Keep the parameter in sync with temporal.api.workflowservice.v1.SignalWorkflowExecutionRequest.
+message PostResetSignalOperation {
```
cretz, line 26: You may move these messages inside the `PostResetOperation` message and remove their prefixing if you'd like, especially since that's the only message they'll ever be used in. This can help justify not using a whole new package for this concept.

cretz, line 26: Pedantic but this message name should be "SignalWorkflow" not just "Signal" IMO (and the oneof name should be `signal_workflow` too IMO)

temporal/api/reset/v1/post_reset_operations.proto
gow: Didn't realize proto package was a thing. Will move it to common. Seems like a good place for it.

temporal/api/reset/v1/post_reset_operations.proto
gow: You mean like this?
```
// PostResetOperation represents an operation to be performed on the new workflow execution after a workflow reset.
message PostResetOperation {
    // SignalWorkflowOperation represents sending a signal after a workflow reset.
    // Keep the parameter in sync with temporal.api.workflowservice.v1.SignalWorkflowExecutionRequest.
    message SignalWorkflowOperation {
        // The workflow author-defined name of the signal to send to the workflow.
        string signal_name = 1;
        // Serialized value(s) to provide with the signal.
        temporal.api.common.v1.Payloads input = 2;
        temporal.api.common.v1.Header header = 3;
        // Links to be associated with the WorkflowExecutionSignaled event.
        repeated temporal.api.common.v1.Link links = 4;
    }

    // UpdateWorkflowOptionsOperation represents updating workflow execution options after a workflow reset.
    // Keep the parameters in sync with temporal.api.workflowservice.v1.UpdateWorkflowExecutionOptionsRequest.
    message UpdateWorkflowOptionsOperation {
        // Update Workflow options that were originally specified via StartWorkflowExecution. Partial updates are accepted and controlled by update_mask.
        temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options = 1;
        // Controls which fields from `workflow_execution_options` will be applied.
        // To unset a field, set it to null and use the update mask to indicate that it should be mutated.
        google.protobuf.FieldMask update_mask = 2;
    }

    oneof variant {
        SignalWorkflowOperation signal_workflow = 1;
        UpdateWorkflowOptionsOperation update_workflow_options = 2;
    }
}
```

temporal/api/reset/v1/post_reset_operations.proto
gow: Moving it to common causes a cyclic import issue. While `ResetOptions` is fairly independent, `PostResetOperation` depends on `temporal/api/workflow/v1/message.proto` which seem to cause a cycle.

temporal/api/reset/v1/post_reset_operations.proto
gow: Done.

temporal/api/workflowservice/v1/request_response.proto
gow: Done.

temporal/api/workflowservice/v1/request_response.proto
gow: Done.

temporal/api/reset/v1/post_reset_operations.proto
gow: Done.

gow: Done.

temporal/api/workflowservice/v1/request_response.proto
gow: This change was reverted.

temporal/api/reset/v1/post_reset_operations.proto
gow: ok. Moved PostResetOperation to `temporal.api.workflow package`. No new package introduced now.

cretz: Thanks! Sorry it has to be in that package, I think if we knew `WorkflowExecutionOptions` would be referenced more generally we would have put it in common.

cretz (approved):

temporal/api/workflow/v1/message.proto
```diff
@@ -581,3 +582,35 @@ message RequestIdInfo {
...
+        // Serialized value(s) to provide with the signal.
+        temporal.api.common.v1.Payloads input = 2;
+        // Headers that are passed with the signal to the processing workflow.
+        temporal.api.common.v1.Header header = 3;
+        // Links to be associated with the WorkflowExecutionSignaled event.
+        repeated temporal.api.common.v1.Link links = 4;
+    }
+
+    // UpdateWorkflowOptionsOperation represents updating workflow execution options after a workflow reset.
+    // Keep the parameters in sync with temporal.api.workflowservice.v1.UpdateWorkflowExecutionOptionsRequest.
```
bergundy, line 602: I would add a note on the request definitions to keep these operations in sync too.

bergundy (approved):

yycptt (approved):

temporal/api/workflow/v1/message.proto
gow: Done.

PR #585, opened 2025-05-07
VersionLocalState in sync with VersionSummary + Activation/Deactivation timestamps

Shivs11:

**What changed?**
- Added two new timestamps to signify when a version first/last receives traffic in the `WorkerDeploymentVersionInfo` and `WorkerDeploymentInfo`.
Added missing fields from VersionLocalState that were not present in VersionSummary.

**Why?**
- Adding the two new timestamps shall allow operators to have more visibility into when a version gets activated/deactivated.
- The reason fields from `VersionLocalState` were added to the `VersionSummary` is because the UI, right now, is querying a bunch of different places to get information about a specific version. This might be annoying and the aim of this PR is to make VersionSummary the source of truth for information related to a specific version. A nice cherry on top is that this shall also reduce the number of API calls to get information about a version.

**Breaking changes**
- None

**Server PR**
- https://github.com/temporalio/temporal/pull/7736)

### DIFF
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -115,7 +115,7 @@ message WorkerDeploymentVersionInfo {
 
     // Timestamp when this version first became current or ramping.
     google.protobuf.Timestamp first_activation_time = 11;
-    // Timestamp when this version last stopped from being current or ramping.
+    // Timestamp when this version last stopped being current or ramping.
     google.protobuf.Timestamp last_deactivation_time = 12;
 
     // Range: [0, 100]. Must be zero if the version is not ramping (i.e. `ramping_since_time` is nil).
@@ -210,7 +210,7 @@ message WorkerDeploymentInfo {
         google.protobuf.Timestamp routing_update_time = 7;
         // Timestamp when this version first became current or ramping.
         google.protobuf.Timestamp first_activation_time = 8;
-        // Timestamp when this version last stopped from being current or ramping.
+        // Timestamp when this version last stopped being current or ramping.
         google.protobuf.Timestamp last_deactivation_time = 9;
     }
 }
--- a/temporal/api/deployment/v1/message.proto
+++ b/temporal/api/deployment/v1/message.proto
@@ -113,6 +113,11 @@ message WorkerDeploymentVersionInfo {
     // Nil if not ramping. Updated when the version first starts ramping, not on each ramp change.
     google.protobuf.Timestamp ramping_since_time = 6;
 
+    // Timestamp when this version first started receiving traffic, either as current or ramping.
+    google.protobuf.Timestamp traffic_first_received_time = 11;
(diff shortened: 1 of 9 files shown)


carlydf (approved):

temporal/api/deployment/v1/message.proto
```diff
@@ -188,7 +193,25 @@ message WorkerDeploymentInfo {
         // The fully-qualified string representation of the version, in the form "<deployment_name>.<build_id>".
         string version = 1;
         google.protobuf.Timestamp create_time = 2;
+        // Deprecated. Use `drainage_info` instead.
         enums.v1.VersionDrainageStatus drainage_status = 3;
+        // Information about workflow drainage to help the user determine when it is safe
+        // to decommission a Version. Not present while version is current or ramping
+        VersionDrainageInfo drainage_info = 4;
+        // Nil if not current.
```
Sushisource, line 201:
```suggestion
        // Unset if not current.
```

Nil is a bit Go-specific.

temporal/api/deployment/v1/message.proto
```diff
@@ -188,7 +193,25 @@ message WorkerDeploymentInfo {
...
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+        google.protobuf.Timestamp current_since_time = 5;
+        // Nil if not ramping. Updated when the version first starts ramping, not on each ramp change.
```
Sushisource, line 205:
```suggestion
        // Unset if not ramping. Updated when the version first starts ramping, not on each ramp change.
```

Sushisource (approved):

temporal/api/deployment/v1/message.proto
```diff
@@ -188,7 +193,25 @@ message WorkerDeploymentInfo {
...
+        // Unset if not current.
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+        google.protobuf.Timestamp current_since_time = 5;
+        // Unset if not ramping. Updated when the version first starts ramping, not on each ramp change.
+        // (-- api-linter: core::0140::prepositions=disabled
+        //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+        google.protobuf.Timestamp ramping_since_time = 6;
+        // Last time `current_since_time`, `ramping_since_time, or `ramp_percentage` of this version changed.
+        google.protobuf.Timestamp routing_update_time = 7;
```
drewhoskins-temporal, line 210: nit: `last_routing_update_time` ?

Shivs11: Thanks for the suggestion! Hmm, I still am leaning towards `routing_update_time` since the word 'update' already implies the latest change, so adding last_ felt a bit redundant to me.

Not making this change right now, but let me know if you still feel strongly about this.

PR #586, opened 2025-05-07
Add activity options to the pending activity info

ychebotarev:

**What changed?**
Add activity options to the pending activity info

**Why?**
If activity options where changed - there is no "official" way to users to know what are the current options.
UI should display current options, so we will provide them as a part of DescribeWorfklow response. 
Thus we add activity options to the pending activity info, which will be a part of DescribeWorfklow response. 

**Breaking changes**
No

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -14,6 +14,7 @@ import "google/protobuf/empty.proto";
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/field_mask.proto";
 
+import "temporal/api/activity/v1/message.proto";
 import "temporal/api/enums/v1/common.proto";
 import "temporal/api/enums/v1/event_type.proto";
 import "temporal/api/enums/v1/workflow.proto";
@@ -322,6 +323,9 @@ message PendingActivityInfo {
     }
 
     PauseInfo pause_info = 23;
+
+    // current activity options. May be different from the one used to start the activity.
+    temporal.api.activity.v1.ActivityOptions activity_options = 24;
 }
 
 message PendingChildExecutionInfo {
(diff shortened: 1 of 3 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -322,6 +323,9 @@ message PendingActivityInfo {
     }
 
     PauseInfo pause_info = 23;
+
+    // current activity options. May be different from the one used to start the activity.
```
cretz, line 327: Inconsistent with whether you capitalize your sentences

temporal/api/workflow/v1/message.proto
```diff
@@ -322,6 +323,9 @@ message PendingActivityInfo {
...
+    temporal.api.activity.v1.ActivityOptions activity_options = 24;
```
cretz, line 328: I wish we had put activity options in `common` tbh. Setting up this proto package dependency cycle may make other things more difficult in the future.

cretz (approved):
Nothing blocking

temporal/api/workflow/v1/message.proto
ychebotarev: >dependency cycle

where is it coming from? "activity" depends on "common", not visa/versa.

cretz: It's not a "cycle" per se, but now having the `workflow` depend on `activity` means that nothing in `activity` can ever depend on anything that depends on `workflow` now (e.g. `batch`). Separating these messages out to single-use small packages makes things difficult for users later.

For example, the fact that `WorkflowExecutionOptions` was put in `workflow` by the versioning team meant we couldn't put `PostResetOperation` in `common` alongside other reset things, ref https://github.com/temporalio/api/pull/581#discussion_r2069526513. Same can happen here. Now everyone that wants `ActivityOptions` has to make sure they are in a package that doesn't cause a cycle with `activity`.

gow (approved):

PR #587, opened 2025-05-08
Nexus WorkflowService APIs

bergundy:
**Why?**

To allow non-workflow Nexus calls.

**Server PR**

Not yet, this is just a draft for initial discussion.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2293,3 +2293,148 @@ message TriggerWorkflowRuleResponse {
...
+    message Async {
+        string operation_token = 1;
+        repeated temporal.api.nexus.v1.Link links = 2;
+    }
+
+    oneof variant {
+        Sync sync_success = 1;
+        Async async_success = 2;
+        // The operation completed unsuccessfully (failed or canceled).
+        temporal.api.nexus.v1.UnsuccessfulOperationError operation_error = 3;
```
bergundy, line 2342: Note that HandlerError is not represented here or in any of the other responses. The idea is that HandlerError would be a new serviceerror but I am on the fence here, we may want to include the handler errors directly in the response.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2293,3 +2293,148 @@ message TriggerWorkflowRuleResponse {
...
+message StartNexusOperationRequest {
+    // The caller namespace. The endpoint below may point to a different namespace.
+    string namespace = 1;
+    // A registered endpoint. Mutually exclusive with task queue.
+    string endpoint = 2;
+    // Dispatch directly to a task queue on this namespace.
+    string task_queue = 3;
+
+    // Request headers to forward to the Handler.
+    map<string, string> header = 4;
```
bergundy, line 2306: Alternatively, we may want headers to be grpc metadata. Trying to figure out what would be better here.

PR #588, opened 2025-05-13
inherit pinned version and override across CaN when pinned

carlydf:

**What changed?**
Add fields to workflow started event to enable inheriting pinned version and override across CaN when the effective behavior of the WF is pinned.

This inheritance will not happen if the behavior is anything else (auto-upgrade, pinned-until-continue-as-new, etc)

**Why?**
To differentiate pinned from trampolining

No

**Server PR**


temporal/api/history/v1/message.proto
```diff
@@ -135,6 +135,13 @@ message WorkflowExecutionStartedEventAttributes {
     // Worker Deployment Version.
     temporal.api.deployment.v1.WorkerDeploymentVersion parent_pinned_deployment_version = 36;
 
+    // When present, it means this is a continued-as-new workflow of a previous execution that is Pinned
+    // to this Worker Deployment Version. In this case, the new workflow will start as Pinned to this Version
+    // instead of starting on the Current Version of its Task Queue.
+    // This is set only if the new workflow is starting on a Task Queue belonging to the same
+    // Worker Deployment Version.
+    temporal.api.deployment.v1.WorkerDeploymentVersion previous_pinned_deployment_version = 37;
```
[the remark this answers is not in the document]
drewhoskins-temporal: (The main related question I have is, have we thought through transition points where the workflow is changing behaviors across code revisions?)

temporal/api/history/v1/message.proto
```diff
@@ -191,8 +198,19 @@ message WorkflowExecutionContinuedAsNewEventAttributes {
     temporal.api.common.v1.SearchAttributes search_attributes = 14;
     // If this is set, the new execution inherits the Build ID of the current execution. Otherwise,
     // the assignment rules will be used to independently assign a Build ID to the new execution.
+    // Only considered for versioning v0.2, which is deprecated.
     bool inherit_build_id = 15;
 
+    // If this is set, the new execution inherits the versioning override of the current execution,
+    // because the effective behavior of the current execution is pinned.
+    temporal.api.workflow.v1.VersioningOverride current_pinned_versioning_override = 16;
```
ShahabT, line 206: since this event belongs to the previous run we might not need it at all. That workflow already knows it's pinned version and override.

temporal/api/history/v1/message.proto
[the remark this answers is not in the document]
carlydf: Here is what I think we want:

CaN: Auto-upgrade -> Pinned
- WF1 is running on version A, where it is labelled auto-upgrade
- WF1 Continues as new, call the new wf WF1.1
- Because WF1 is not Pinned, WF1.1 does not inherit the pinned version, and starts on the current version of its task queue, which is now B. In version B, this WF type is labelled Pinned, so the behavior of WF1.1 becomes Pinned on the first WFT completion send by the version B worker.
- WF1.1 Continues as new, call the new wf WF1.2
- Because WF1.1 is Pinned to B, WF1.2 inherits the pinned version B, and starts on version B*
- The first WFT completion on WF1.2 will come from worker B, which will say this wf type is Pinned, so WF1.2 will remain Pinned.


*if WF1.2 starts on a different task queue from WF1.1, and that other task queue is not in Version B, then WF1.2 will not inherit, and will instead start on the current version of its task queue. This is uncommon.

temporal/api/history/v1/message.proto
carlydf: yes, the important thing is just that we create the new mutable state correctly

ShahabT (approved):

temporal/api/history/v1/message.proto
```diff
@@ -125,18 +125,42 @@ message WorkflowExecutionStartedEventAttributes {
...
-    // Deprecated. Use `parent_pinned_deployment_version`.
+    // Deprecated. Use `parent_versioning_info`.
     string parent_pinned_worker_deployment_version = 34 [deprecated = true];
 
-    // When present, it means this is a child workflow of a parent that is Pinned to this Worker
-    // Deployment Version. In this case, child workflow will start as Pinned to this Version instead
-    // of starting on the Current Version of its Task Queue.
-    // This is set only if the child workflow is starting on a Task Queue belonging to the same
-    // Worker Deployment Version.
-    temporal.api.deployment.v1.WorkerDeploymentVersion parent_pinned_deployment_version = 36;
```
cretz, line 136: Not generally reviewing PR (leaving up to others that know versioning better), but I can I get confirmation that this field is not referenced in any SDK or CLI code? I didn't see it, just wanting to confirm.

carlydf: I asked @Sushisource last week and he told me that it was not. I also don't think that it is. This is used internally in the server.

PR #589, opened 2025-05-14
Fix protobuf syntax that is causing an error in protobufjs

mjameswh:
**What changed?**

- Fix a double semicolon introduced in #579.

**Why?**

- The double semicolon sequence is causing a compilation error in `protobufjs`, even though this is technically valid according to [the official Protobuf 3 spec](https://protobuf.dev/reference/protobuf/proto3-spec/#emptystatement).

  This is a bug in `protobufjs`, for which there's been [an open issue ticket](https://github.com/protobufjs/protobuf.js/issues/1322) since 2019, but it was never acknowledged, and a PR fixing this has been ignored for more than a year.

  Given that there's simply no reason to use a double semicolon sequence anyway, it appears preferable to simply avoid that in our proto definitions.

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -35,7 +35,7 @@ message WorkflowExecutionStartedEventAttributes {
     // Contains information about parent workflow execution that initiated the child workflow these attributes belong to.
     // If the workflow these attributes belong to is not a child workflow of any other execution, this field will not be populated.
     temporal.api.common.v1.WorkflowExecution parent_workflow_execution = 3;
-    // EventID of the child execution initiated event in parent workflow 
+    // EventID of the child execution initiated event in parent workflow
     int64 parent_initiated_event_id = 4;
     temporal.api.taskqueue.v1.TaskQueue task_queue = 5;
     // SDK will deserialize this and provide it as arguments to the workflow function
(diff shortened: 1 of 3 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -440,7 +440,7 @@ message ActivityTaskCanceledEventAttributes {
     string identity = 5;
     // Version info of the worker who processed this workflow task.
     // Deprecated. This field should be cleaned up when versioning-2 API is removed. [cleanup-experimental-wv]
-    temporal.api.common.v1.WorkerVersionStamp worker_version = 6 [deprecated = true];;
```
Quinn-With-Two-Ns, line 443: Can we add some CI step to prevent this in the future?

Quinn-With-Two-Ns (approved):

temporal/api/history/v1/message.proto
mjameswh: It is technically possible, but since this is technically valid according to the spec and the official `protoc` tool, detecting that would require adding `protobufjs` to this repository, or to do some specifically tailored checks (e.g. `grep ';;'` on all `.proto` files, which might trigger false-positives). That would most likely incurs more maintenance hassle than what it would save us.

So after discussion with the team, we agreed to not automate this check.

carlydf: sorry!

PR #590, opened 2025-05-14
Add identity param to reset API

gow:

Adding ~~2~~ 1  new parameter to Reset API request.
1. identity - needed to capture the identity of the worker/client making the operation.
2. ~~terminate_child_workflow_if_necessary - allows operators to optionally terminate child workflows that were started after the reset-point~~

~~terminate_child_workflow_if_necessary is needed to unblock parent after reset in some in cases where WorkflowIdConflictPolicy prevents the new child from being started.~~


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -818,6 +818,16 @@ message ResetWorkflowExecutionRequest {
...
+    // The identity of the worker/client
+    string identity = 9;
+    // Terminate child workflow if necessary.
+    // When a parent workflow is reset, it may run into child workflow ID conflict if it is using stable child workflow IDs.
+    // This can happen when the parent is reset back to a point before starting the child.
+    // In such cases this flag can be used to terminate the child workflow (if necessary) so that the parent can make progress.
+    // When this flag is set to true, the server will track all the child workflow IDs that are about to be started again
+    // by the parent after reset. At the time of starting the child workflow again, it checks if the child workflow is running
+    // and terminates it if so before starting the new child instance.
+    bool terminate_child_workflow_if_necessary = 10;
```
cretz, line 830: This is confusing IMO and adds more state burden on server. I think we should just consider supporting conflict policy on child workflows, and since it's not non-deterministic to change child options, they can update their code on which child workflows they actually want to do this one. That is easier to understand IMO and it gives the flexibility this can never give - the ability to choose _which_ child workflows you want to terminate if existing.

gow: > I am trying to understand the overlap with ID reuse policy of terminate existing on child workflows. That is easier to understand IMO and it gives the flexibility this can never give - the ability to choose which child workflows you want to terminate if existing.
Yes, this is the same as `WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING`. The only few (important) difference are
1. This is a 1-time override in the context of reset and the server automatically applies it to the right children based on what the state of the history at the time of reset.
2. No code change necessary. Changing code means they would have to change the `WorkflowIdConflictPolicy` permanently and it applies in all scenarios (outside reset context as well). This may not be something they desire. If we had separate WorkflowIdConflictPolicy just for reset that would be great. But it doesn't exist.

>If we must put this here, it should be named to basically say this is the new implied ID reuse policy for pre-reset child workflows that are duplicated.

Naming this option has been very hard. Open to suggestions here.

>I am curious if this situation is something users are hitting?

I'm not aware of any reports yet. cc @drewhoskins-temporal 

> Also, where is the server tracking the pre-reset child workflows after reset? Is there a place that the user can see?

The server tracks it in the mutable state of the new run. It is currently not visible outside server.

cretz: > No code change necessary. Changing code means they would have to change the WorkflowIdConflictPolicy permanently and it applies in all scenarios (outside reset context as well). This may not be something they desire [...] I'm not aware of any reports yet.

My concern here is we're guessing at desires. For instance, some people may desire to choose specific child workflow IDs to apply this to instead of assuming all. But maybe not, who knows.

Also there is no workflow ID conflict policy on child workflows, but I assume you mean ID reuse policy's equivalent of that field. Regardless, whether an author chooses to terminate in cases of duplicate IDs running is a workflow author choice I think regardless of _why_ there may be a duplicate.

> Naming this option has been very hard. Open to suggestions here.

> The server tracks it in the mutable state of the new run. It is currently not visible outside server.

This is a bit rough. If a workflow got terminated for this reason, how non-obvious is it?

General question I am curious about too - does this only have value to those explicitly opting in to a non-default parent close policy? Because if they leave the default parent close policy, the children after the reset point are already terminated on reset right? And even if they did opt-in to parent close policy of cancel, it may be considered dangerous to interrupt cancellation with termination, but that's not as big of a deal.

Still, it feels like a very specific case for a very specific hypothetical of a user opting in to a separate close policy, resetting before the child started, and not willing to author code to react to duplicate IDs when starting a workflow (child or otherwise) that we would ask them to in any other case.

If this is really needed, ok, just seems a bit specific.

drewhoskins-temporal: * +1 to the point about consistently reusing the existing ontology.  Framing this as "override the child conflict/reuse policy" seems better.
* Yes, the point is so that you don't have to do a code push.  In particular, reset is often needed for rollbacks when code pushes won't help.
* However, this will only come up when the workflow ID is consistent and:
** The parent  child policy is ABANDON (so the default conflict policy of reject kicks in), OR
** The ID reuse policy is `REJECT_DUPLICATES`

Since this won't come up all that much because it's not the default, the can can be kicked down the road.  But I think the product behavior is likely valid and valuable.

gow: Done. Removed this from the API for now. Will revisit if needed.

cretz (approved):

drewhoskins (approved):

drewhoskins (approved):

PR #592, opened 2025-05-16
Sync //Deprecated comment and [deprecated = true] attribute

alexshtin:
**What changed?**
Sync `//Deprecated` comment and `[deprecated = true]` attribute.

**Why?**
When field is deprecated it should be marked with both comment and attribute to get full support in all languages.

**Breaking changes**
No, only comments/attributes changes.


Sushisource (approved):
I'm happy you did this but by gawd we're going to have to add so many linter ignore annotations in the SDKs

dnr (approved):

PR #593, opened 2025-05-16
Add flag to restore activity options to reset and update

ychebotarev:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1835,13 +1835,18 @@ message UpdateActivityOptionsRequest {
     // Controls which fields from `activity_options` will be applied
     google.protobuf.FieldMask update_mask = 5;
 
-        // either activity id or activity type must be provided
-        oneof activity {
-            // Only activity with this ID will be updated.
-            string id = 6;
-            // Update all running activities of this type.
-            string type = 7;
-        }
+    // either activity id or activity type must be provided
+    oneof activity {
+        // Only activity with this ID will be updated.
+        string id = 6;
+        // Update all running activities of this type.
+        string type = 7;
+    }
+
+    // If set, the activity options will be restored to the default.
+    // Default options are then options activity was created with.
+    // They are part of the first SCHEDULE event.
+    bool restore_original = 8;
 }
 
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1835,13 +1835,18 @@ message UpdateActivityOptionsRequest {
...
+        // Only activity with this ID will be updated.
+        string id = 6;
+        // Update all running activities of this type.
+        string type = 7;
+    }
+
+    // If set, the activity options will be restored to the default.
+    // Default options are then options activity was created with.
+    // They are part of the first SCHEDULE event.
+    bool restore_original = 8;
```
cretz, line 1849: Can you confirm if this is true then `activity_options` and `update_mask` must both be `null` or the call will fail? If so, can we document this?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1932,6 +1937,12 @@ message ResetActivityRequest {
     // If set, and activity is in backoff, the activity will start at a random time within the specified jitter duration.
     // (unless it is paused and keep_paused is set)
     google.protobuf.Duration jitter = 8;
+
+    // If set, the activity options will be restored to the defaults.
+    // Default options are then options activity was created with.
+    // They are part of the first SCHEDULE event.
+    bool restore_original_options = 9;
```
cretz, line 1944: Can you help me understand this a bit? To confirm, we are willing to allow you to do _some_ of `UpdateOptions` but not other parts? Meaning you're only allowed to update options if it's back to original? If there is a need to update options and reset at the same time, is there a more generic approach to doing both atomically we should take?

ychebotarev: We are not updating options in reset per se. It is more like we are reseting them. 
Just like "-reset-heartbeats", "reset attempts".

Now activity reset supports - attempts, heartbeats, options.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: yes, and yes.

gow (approved):

PR #594, opened 2025-05-23
added latest_version, current_version, ramping_version summaries in ListDeployment

Shivs11:

**What changed?**
- WISOTT

**Why?**
- making more information available in one API call rather than via multiple API calls.

**Breaking changes**

**Server PR**
- In progress...

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -6812,6 +6812,15 @@
         },
         "routingConfig": {
           "$ref": "#/definitions/v1RoutingConfig"
+        },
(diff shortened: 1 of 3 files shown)


ShahabT (approved):

Shivs11:
@cretz - ah yes, should have maybe just kept #595 as draft since that is built on top of this one. I would like to get this one in first!

cretz (commented):
(deferring to @Sushisource  who has more background here than I do, nothing syntactically wrong from my POV)

PR #595, opened 2025-05-24
VersionStatus for WorkerDeploymentVersions

Shivs11:

**What changed?**
- add VersionStatus inside `WorkerDeploymentVersionInfo` + `WorkerDeploymentVersionSummary`

**Why?**
- right now, the UI code deduces the status of a version by checking `CurrentSince`, `RampingSince` and other attributes. This can just be simplified by adding this enum and the server doing the heavy lifting instead.

**Breaking changes**
- None

**Server PR**
- [741011](https://github.com/temporalio/temporal/pull/7804)


temporal/api/enums/v1/deployment.proto
```diff
@@ -72,3 +72,27 @@ enum WorkerVersioningMode {
...
 }
+
+// (-- api-linter: core::0216::synonyms=disabled
+//     aip.dev/not-precedent: Call this status because it is . --)
+// Specify the status of a Worker Deployment Version.
+// Experimental. Worker Deployments are experimental and might significantly change in the future.
+enum VersionStatus {
+    // The Worker Deployment Version has been created inside the Worker Deployment but is not used by any
+    // workflow executions.
+    VERSION_STATUS_UNSPECIFIED = 0;
```
ShahabT, line 83: I think we should have a explicit value for INACTIVE. Also, need to clarify that these version can still have executions with override. something like below:

```suggestion
    VERSION_STATUS_UNSPECIFIED = 0;
    // The Worker Deployment Version has been created inside the Worker Deployment but is not automatically used by
    // workflow executions. These Versions can still have workflows if they have an explicit Versioning Override targeting this Version. Such Versioning Override could be set at workflow start time, or at a later time via `UpdateWorkflowExecutionOptions`.
    VERSION_STATUS_INACTIVE = 1;
```

ShahabT (approved):
lgtm otherwise

temporal/api/enums/v1/deployment.proto
Shivs11: ah, I just did a quick slack search and realized that it's a practice followed here to not have the default value of a message correspond to the zero value - changed and updated the comment as well

PR #596, opened 2025-05-26
Add Stats to DescribeTaskQueue

stephanos:

**What changed?**

Added `TaskQueueStats` field to `DescribeTaskQueueResponse`.

**Why?**

To provide non-versioning users with relevant task queue stats.

PS: The long-term plan is to deprecate the ENHANCED mod fields.

**Breaking changes**

N/A

**Server PR**

https://github.com/temporalio/temporal/pull/7816

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1033,7 +1033,7 @@ message DescribeWorkflowExecutionResponse {
 message DescribeTaskQueueRequest {
     string namespace = 1;
 
-    // However, sticky queues are not supported in ENHANCED mode.
+    // Sticky queues are not supported in deprecated ENHANCED mode.
     temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
 
     // If unspecified (TASK_QUEUE_TYPE_UNSPECIFIED), then default value (TASK_QUEUE_TYPE_WORKFLOW) will be used.
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1032,45 +1032,68 @@ message DescribeWorkflowExecutionResponse {
 //     aip.dev/not-precedent: field_behavior annotation not available in our gogo fork --)
 message DescribeTaskQueueRequest {
     string namespace = 1;
-    // Sticky queues are not supported in `ENHANCED` mode.
-    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
-    // Deprecated. Use `ENHANCED` mode with `task_queue_types`. Ignored in `ENHANCED` mode.
-    // If unspecified (TASK_QUEUE_TYPE_UNSPECIFIED), then default value (TASK_QUEUE_TYPE_WORKFLOW) will be used.
-    temporal.api.enums.v1.TaskQueueType task_queue_type = 3 [deprecated = true];
-    // Deprecated. Ignored in `ENHANCED` mode.
-    bool include_task_queue_status = 4 [deprecated = true];
 
-    // All options except `task_queue_type` and `include_task_queue_status` are only available in the `ENHANCED` mode.
+    // Select the API mode to use for this request: DEFAULT mode (if unset) or ENHANCED mode.
+    // Consult the documentation for each field to understand which mode it is supported in.
     temporal.api.enums.v1.DescribeTaskQueueMode api_mode = 5;
 
+    // [BOTH MODES]
+    // However, sticky queues are not supported in ENHANCED mode.
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
+
+    // [DEFAULT MODE ONLY]
+    // If unspecified (TASK_QUEUE_TYPE_UNSPECIFIED), then default value (TASK_QUEUE_TYPE_WORKFLOW) will be used.
+    // Only supported in default mode (use `task_queue_types` in ENHANCED mode instead).
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
+
+    // [DEFAULT MODE ONLY]
+    // If true, the task queue status will be included in the response.
+    // In ENHANCED mode, use `report_stats` instead.
+    bool include_task_queue_status = 4;
+
+    // [ENHANCED MODE ONLY]
     // Optional. If not provided, the result for the default Build ID will be returned. The default Build ID is the one
     // mentioned in the first unconditional Assignment Rule. If there is no default Build ID, the result for the
     // unversioned queue will be returned.
     // (-- api-linter: core::0140::prepositions --)
     temporal.api.taskqueue.v1.TaskQueueVersionSelection versions = 6;
 
+    // [ENHANCED MODE ONLY]
     // Task queue types to report info about. If not specified, all types are considered.
     repeated temporal.api.enums.v1.TaskQueueType task_queue_types = 7;
-    // Report stats for the requested task queue types and versions
+
+    // [BOTH MODES]
+    // Report stats for the requested task queue type(s), and versions (only in ENHANCED mode).
     bool report_stats = 8;
-    // Report list of pollers for requested task queue types and versions
+
+    // [ENHANCED MODE ONLY]
+    // Report list of pollers for requested task queue types and versions.
     bool report_pollers = 9;
+
+    // [ENHANCED MODE ONLY]
     // Report task reachability for the requested versions and all task types (task reachability is not reported
     // per task type).
     bool report_task_reachability = 10;
 }
 
 message DescribeTaskQueueResponse {
-    // Deprecated. Use `versions_info.types_info.pollers` with `ENHANCED` mode instead.
-    // Not set in `ENHANCED` mode.
-    repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1 [deprecated = true];
-    // Deprecated. Not set in `ENHANCED` mode.
+    // [DEFAULT MODE ONLY]
+    repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
+
+    // [DEFAULT MODE ONLY]
+    // Deprecated.
+    // Status of the task queue. Only populated when `include_task_queue_status` is set to true in the request.
     temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2 [deprecated = true];
 
+    // [DEFAULT MODE ONLY]
+    // Statistics for the task queue. Only populated when `report_stats` is set to true in the request.
+    temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 5;
+
+    // [ENHANCED MODE ONLY]
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
-    // Only set in `ENHANCED` mode.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3;
 
+    // [ENHANCED MODE ONLY]
     // Specifies which Worker Deployment Version(s) Server routes this Task Queue's tasks to.
     // When not present, it means the tasks are routed to Unversioned workers (workers with
     // UNVERSIONED or unspecified WorkerVersioningMode.)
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,68 @@ message DescribeWorkflowExecutionResponse {
 //     aip.dev/not-precedent: field_behavior annotation not available in our gogo fork --)
 message DescribeTaskQueueRequest {
     string namespace = 1;
-    // Sticky queues are not supported in `ENHANCED` mode.
+
+    // Select the API mode to use for this request: DEFAULT mode (if unset) or ENHANCED mode.
+    // Consult the documentation for each field to understand which mode it is supported in.
+    temporal.api.enums.v1.DescribeTaskQueueMode api_mode = 5;
```
stephanos, line 1038: I moved this near the top since it influences how any of the other fields work/are understood.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,68 @@ message DescribeWorkflowExecutionResponse {
...
 message DescribeTaskQueueResponse {
-    // Deprecated. Use `versions_info.types_info.pollers` with `ENHANCED` mode instead.
-    // Not set in `ENHANCED` mode.
-    repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1 [deprecated = true];
-    // Deprecated. Not set in `ENHANCED` mode.
+    // [DEFAULT MODE ONLY]
+    repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
+
+    // [DEFAULT MODE ONLY]
+    // Deprecated.
```
stephanos, line 1084: This field remains deprecated. `TaskQueueStats` supersedes it.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,68 @@ message DescribeWorkflowExecutionResponse {
...
+    // [DEFAULT MODE ONLY]
+    // Statistics for the task queue. Only populated when `report_stats` is set to true in the request.
+    temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 5;
+
+    // [ENHANCED MODE ONLY]
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
-    // Only set in `ENHANCED` mode.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3;
 
+    // [ENHANCED MODE ONLY]
```
ShahabT, line 1097: this is only reported in default mode.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,69 @@ message DescribeWorkflowExecutionResponse {
...
+    // [DEFAULT MODE ONLY]
+    repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
+
+    // [DEFAULT MODE ONLY]
+    // Deprecated.
+    // Status of the task queue. Only populated when `include_task_queue_status` is set to true in the request.
     temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2 [deprecated = true];
 
+    // [DEFAULT MODE ONLY]
+    // Statistics for the task queue. Only populated when `report_stats` is set to true in the request.
```
cretz, line 1090: This is a bit confusing. So this says `[DEFAULT MODE ONLY]`, but then says "Only populated when `report_stats` is set to true" and that field is `[ENHANCED MODE ONLY]`. Isn't this contradictory?

ShahabT: I see `report_stats` says `[BOTH MODES]` though

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,69 @@ message DescribeWorkflowExecutionResponse {
...
+    temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 5;
+
+    // [ENHANCED MODE ONLY]
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
-    // Only set in `ENHANCED` mode.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3;
 
+    // [DEFAULT MODE ONLY]
```
cretz, line 1097: So we won't let an enhanced user get this basic versioning info? Is it some kind of aggregated form of `versions_info`?

temporal/api/workflowservice/v1/request_response.proto
stephanos: Right; I marked `report_stats` as `BOTH MODES` for that reason. I'll admit, it's a bit awkward, but the alternatives felt worse to me (specifically, introducing yet another field or re-using the (default mode-only) `include_task_queue_status`).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,68 @@ message DescribeWorkflowExecutionResponse {
...
+    repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
+
+    // [DEFAULT MODE ONLY]
+    // Deprecated.
+    // Status of the task queue. Only populated when `include_task_queue_status` is set to true in the request.
     temporal.api.taskqueue.v1.TaskQueueStatus task_queue_status = 2 [deprecated = true];
 
+    // [DEFAULT MODE ONLY]
+    // Statistics for the task queue. Only populated when `report_stats` is set to true in the request.
+    temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 5;
```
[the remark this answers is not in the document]
cretz: I'm now wondering if this may necessitate at least a high-level idea of what this call may look like in non-Go SDKs. For instance, what parameters might `await my_client.describe_task_queue()` accept and what might the return type look like? I admit I am getting a bit lost with enhanced no longer being a superset of non-deprecated default data.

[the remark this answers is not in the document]
stephanos: I'd expect these to work (Python-ish):

```
await my_client.describe_task_queue("my-task-queue", TaskQueueType.Workflow)
```
(no stats queried)

```
await my_client.describe_task_queue("my-task-queue", TaskQueueType.Workflow, include_task_queue_status=true)
```
(`task_queue_status` populated)

```
await my_client.describe_task_queue("my-task-queue", TaskQueueType.Workflow, report_stats=true)
```
(`task_queue_stats` populated)

Does that help?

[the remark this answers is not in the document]
cretz: I don't see a separation between default and enhanced there. Are those calls not using "enhanced" mode? What is the return type of the functions (meaning what's the return model)? Of course they can ask about multiple task queue types at a time, so we wouldn't limit there, but that's a minor thing.

temporal/api/workflowservice/v1/request_response.proto
cretz: (sorry @stephanos, we submitted our comments here at the same time, but I do have some questions there)

stephanos: I've come into this API only recently; it's definitely confusing :-/

Here's my understanding:

- default ("legacy") mode: query via `task_queue` and `task_queue_type`
- enhanced mode: query via `task_queue`, `task_queue_types` (plural!) and `versions` (allows selecting multiple `build_ids`)

So you can already see that the enhanced mode is for batch requests, and the default is only for a single task queue.

> So do we hide top-level stats for enhanced users? If so, why? 

So the `TaskQueueStats` _currently_, is only included in `TaskQueueTypeInfo`, which is only included in `TaskQueueVersionInfo` which is only used in the `DescribeTaskQueueResponse`'s `versions_info` map (which is enhanced mode only). Meaning users of enhanced mode have access to that data here: `version_info[build_id].types_info[workflow_type].stats`.

> What does this top-level task_queue_stats contain that that one doesn't?

`task_queue_stats` would only be present when using the default mode (`version_info` is only present in enhanced mode). So there's no duplication. The new field "upgrades" the default mode to contain the same information as the enhanced mode already does.

Long-term, the goal is to get rid of "enhanced mode" altogether. Then, any current SDK implementation for DescribeTaskQueue only has to add the new toggle field to populate `task_queue_stats`.

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
stephanos: I don't know the exact state of support of `describe_task_queue` across SDKs, but my understanding is that only the Go SDK added support for the enhanced mode (as a separate API, too). Since the (new) goal is to get rid of enhanced mode altogether, those implementations don't have to chance materially. They just have to add support for the on/off toggle `report_stats` to opt-in to the new stats field in the response.

temporal/api/workflowservice/v1/request_response.proto
ShahabT: @cretz the enhanced mode operates on a "task queue family" which is all task queues of the same name. Officially, a task queue is identified by name + single type. So the right granularity for this API seem to be a single type.

When I proposed and implemented the enhanced mode I did not know that a task queue is a single type. I can see also how most users think of task queue as an entity encompassing all types because the UI and SDK makes it easy to people to think that way. But if one reads the [docs](https://docs.temporal.io/task-queue#task-queue) carefully they would see that it says: _"There are three types of Task Queues: Activity Task Queues, Workflow Task Queues, and Nexus Task Queues."_.

When @mfateev saw the enhanced mode he strongly disliked it because it deals with multiple types meaning it's not describing a single task queue anymore.

Given that, we are thinking to double down on the default/legacy mode. An eventually deprecate the enhanced mode.

If you think it's less confusing for users I don't see an issue with marking enhanced mode as deprecated right now with these changes. After all, all the extra stuff for versioning is not deprecated anyways because they belonged to versioning V2, and the stats part is being covered for unversioned. (stats for versioned V3 can follow later).

temporal/api/workflowservice/v1/request_response.proto
cretz: It's confusing that you get more enhanced versioning information if you don't use enhanced mode. We should deprecate enhanced mode if it's actually no longer the enhanced form with enhanced information.

ShahabT (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,73 @@ message DescribeWorkflowExecutionResponse {
...
```
dnr, line 1094: Maybe too late for this: TaskQueueStats has a `google.protobuf.Duration approximate_backlog_age`, which I've found pretty awkward to work with: it depends on the current time, so by the time you get it it's out of date. (Of course, any time you get a description of the backlog it may be out of date, but this one is _guaranteed_ to be.)

I've changed a bunch of the internal interfaces to work with the absolute timestamp (of the task at the head of the queue) instead, and I think it makes things much nicer. Could we carry that through the api and consider changing the age in TaskQueueStats to a timestamp instead? (i.e. adding a timestamp and deprecating the age)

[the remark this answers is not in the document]
stephanos: Right place? yes. Right time? 😬 trying to get it into the release tmrw.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1032,45 +1032,73 @@ message DescribeWorkflowExecutionResponse {
...
 message DescribeTaskQueueRequest {
     string namespace = 1;
-    // Sticky queues are not supported in `ENHANCED` mode.
+
+    // Deprecated. ENHANCED mode is also being deprecated.
+    // Select the API mode to use for this request: DEFAULT mode (if unset) or ENHANCED mode.
+    // Consult the documentation for each field to understand which mode it is supported in.
+    temporal.api.enums.v1.DescribeTaskQueueMode api_mode = 5 [deprecated = true];;
+
+    // [BOTH MODES]
```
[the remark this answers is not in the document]
stephanos: I like that 👍

temporal/api/workflowservice/v1/request_response.proto
stephanos: It sounds reasonable to me, I just won't be able to make that change in time. I'd be happy to add that as a follow-up, though.

cretz (approved):
The deprecated enhanced form clears things up, thanks!

temporal/api/workflowservice/v1/request_response.proto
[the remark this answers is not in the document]
dnr: "Enhanced" was supposed to be a protocol-only thing, users wouldn't have asked for it directly, the sdk would just always use it. The old mode was still supported just to not break old clients.

PR #597, opened 2025-05-26
remove versioning-0.30 API defs

Shivs11:

- WISOTT

**Why?**
- These API's were added for the pre-release of versioning and are now deprecated. Moreover, no users are also using the pre-release version of versioning so it's safe to remove this.

**Breaking changes**
- None

**Server PR**
- https://github.com/temporalio/temporal/pull/7807


Shivs11:
closing this since it was decided internally that we shall remove the old versioning API's a little later

PR #598, opened 2025-05-27
add create_request_id to child initiated event

hai719:

Add request_id to child initiated event.

**Why**
For event-based replication, if request_id is not replicated to standby clusters, after failover, standby cannot dedup the start workflow request.

**Breaking changes**

**Server PR**
https://github.com/temporalio/temporal/pull/7813

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -665,6 +665,9 @@ message StartChildWorkflowExecutionInitiatedEventAttributes {
     bool inherit_build_id = 19;
     // Priority metadata
     temporal.api.common.v1.Priority priority = 20;
+    // A unique identifier for the request to start this child workflow execution. This is used to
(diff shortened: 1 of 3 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -665,6 +665,9 @@ message StartChildWorkflowExecutionInitiatedEventAttributes {
     bool inherit_build_id = 19;
     // Priority metadata
     temporal.api.common.v1.Priority priority = 20;
+    // A unique identifier for the request to start this child workflow execution. This is used to
+    // deduplicate requests to start a child workflow execution with the same workflow ID.
+    string create_request_id = 21;
```
cretz, line 670: > This is used to deduplicate requests to start a child workflow execution with the same workflow ID.

I assume the tuple of workflow run ID + event ID is not enough?

PR #599, opened 2025-05-28
Add worker visibility API - heartbeat and list worker satus

ychebotarev:
MAYBE FOR MERGE. 
The goal of this PR is to review the proposed APIs.
It will not necessary be merged, treat it as a part of Design review.

**What changed?**
2 new APIs added:
* WorkerHeartbeat
* ListWorkerStatus

**Why?**
Part of the worker visibility work.

Note:
Currently metrics are not supported as a part of List operation query. But it may happen, based on user feedback.

### DIFF
--- a/temporal/api/enums/v1/common.proto
+++ b/temporal/api/enums/v1/common.proto
@@ -102,7 +102,7 @@ enum ApplicationErrorCategory {
 //     aip.dev/not-precedent: It seems we have both state and status, and status is a better fit for workers. --)
 enum WorkerStatus {
     WORKER_STATUS_UNSPECIFIED = 0;
-    WORKER_STATUS_STATUS_RUNNING = 1;
+    WORKER_STATUS_RUNNING = 1;
     WORKER_STATUS_SHUTTING_DOWN = 2;
     WORKER_STATUS_SHUTDOWN = 3;
 }
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -18,14 +18,13 @@ message WorkerPollerInfo {
 
   int32 active_pollers = 1;
 
-  int32 available_pollers = 2;
-
-  google.protobuf.Timestamp last_successful_poll_time = 3;
+  google.protobuf.Timestamp last_successful_poll_time = 2;
 }
 
 message WorkerSlotsInfo {
 
   // Number of slots available for the worker to specific tasks.
+  // May be -1 if the upper bound is not known.
   int32 current_slots_available = 1;
   // Number of slots used by the worker for specific tasks.
   int32 current_slots_used = 2;
@@ -63,7 +62,7 @@ message WorkerHostInfo {
 //     aip.dev/not-precedent: Removing those words make names less clear. --)
 message WorkerInfo {
   // Worker identifier, should be unique for the namespace.
-  // It is different from worker identity, which is usually a combination of host_name and process_id.
+  // It is distinct from worker identity, which is not necessarily namespace-unique.
   string worker_instance_key = 1;
 
   // Worker host information.
@@ -77,7 +76,7 @@ message WorkerInfo {
   string sdk_name = 5;
   string sdk_version = 6;
 
-  // Worker status. Possible values - "running", "shutting_down", "shutdown".
+  // Worker status. Defined by SDK.
   temporal.api.enums.v1.WorkerStatus status = 7;
 
   // Worker start time.
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -15292,7 +15292,7 @@
       "properties": {
         "workerInstanceKey": {
           "type": "string",
-          "description": "Worker identifier, should be unique for the namespace.\nIt is different from worker identity, which is usually a combination of host_name and process_id."
+          "description": "Worker identifier, should be unique for the namespace.\nIt is distinct from worker identity, which is not necessarily namespace-unique."
         },
         "hostInfo": {
           "$ref": "#/definitions/v1WorkerHostInfo",
@@ -15313,7 +15313,7 @@
         },
         "status": {
           "$ref": "#/definitions/v1WorkerStatus",
-          "description": "Worker status. Possible values - \"running\", \"shutting_down\", \"shutdown\"."
+          "description": "Worker status. Defined by SDK."
         },
         "startTime": {
           "type": "string",
@@ -15386,10 +15386,6 @@
           "type": "integer",
           "format": "int32"
         },
-        "availablePollers": {
-          "type": "integer",
-          "format": "int32"
-        },
         "lastSuccessfulPollTime": {
           "type": "string",
           "format": "date-time"
@@ -15402,7 +15398,7 @@
         "currentSlotsAvailable": {
           "type": "integer",
           "format": "int32",
-          "description": "Number of slots available for the worker to specific tasks."
+          "description": "Number of slots available for the worker to specific tasks.\nMay be -1 if the upper bound is not known."
         },
         "currentSlotsUsed": {
           "type": "integer",
@@ -15435,7 +15431,7 @@
       "type": "string",
       "enum": [
         "WORKER_STATUS_UNSPECIFIED",
-        "WORKER_STATUS_STATUS_RUNNING",
+        "WORKER_STATUS_RUNNING",
         "WORKER_STATUS_SHUTTING_DOWN",
         "WORKER_STATUS_SHUTDOWN"
       ],
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -12686,7 +12686,7 @@ components:
           type: string
           description: |-
             Worker identifier, should be unique for the namespace.
-             It is different from worker identity, which is usually a combination of host_name and process_id.
+             It is distinct from worker identity, which is not necessarily namespace-unique.
         hostInfo:
           allOf:
             - $ref: '#/components/schemas/WorkerHostInfo'
@@ -12703,11 +12703,11 @@ components:
         status:
           enum:
             - WORKER_STATUS_UNSPECIFIED
-            - WORKER_STATUS_STATUS_RUNNING
+            - WORKER_STATUS_RUNNING
             - WORKER_STATUS_SHUTTING_DOWN
             - WORKER_STATUS_SHUTDOWN
           type: string
-          description: Worker status. Possible values - "running", "shutting_down", "shutdown".
+          description: Worker status. Defined by SDK.
           format: enum
         startTime:
           type: string
@@ -12767,9 +12767,6 @@ components:
         activePollers:
           type: integer
           format: int32
-        availablePollers:
-          type: integer
-          format: int32
         lastSuccessfulPollTime:
           type: string
           format: date-time
@@ -12778,7 +12775,9 @@ components:
       properties:
         currentSlotsAvailable:
           type: integer
-          description: Number of slots available for the worker to specific tasks.
+          description: |-
+            Number of slots available for the worker to specific tasks.
+             May be -1 if the upper bound is not known.
           format: int32
         currentSlotsUsed:
           type: integer


temporal/api/workflowservice/v1/service.proto
```diff
@@ -1173,4 +1173,25 @@ service WorkflowService {
         };
     }
 
+    // WorkerHeartbeat receive heartbeat request from the worker.
+    rpc WorkerHeartbeat(WorkerHeartbeatRequest) returns (WorkerHeartbeatResponse) {
```
cretz, line 1177: Can you call this `RecordWorkerHeartbeat` or `HeartbeatWorker` instead? Ideally, though we don't in all cases, we can lead with the verb

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1173,4 +1173,25 @@ service WorkflowService {
...
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/worker-heartbeat"
+                body: "*"
+            }
+        };
+    };
+
+    // ListWorkerStatus is a visibility API to list worker status information in a specific namespace.
+    rpc ListWorkerStatus (ListWorkerStatusRequest) returns (ListWorkerStatusResponse) {
```
cretz, line 1189: Can we call this `ListWorkers` instead or can we update the HTTP path?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+  string sdk_version = 8;
+
+  string worker_identity = 9;
+
+  int32 worker_status = 10;
+  google.protobuf.Duration uptime = 11;
+
+  google.protobuf.Timestamp last_heartbeat_time = 12;
+
+  WorkerTaskStatus workflow_task_status = 13;
```
Quinn-With-Two-Ns, line 51: Is this expected to be a combination of sticky and non sticky status?

temporal/api/workflowservice/v1/service.proto
cretz, line 1189: List calls should be plural

temporal/api/worker/v1/message.proto
cretz, line 1: Lots of missing docs. This is especially important if the same models are used for both sending and receiving data (hard to know what's required, what's server populated, etc).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2327,3 +2328,30 @@ message TriggerWorkflowRuleResponse {
...
+
+message WorkerHeartbeatResponse {
+
+}
+
+message ListWorkerStatusRequest {
+    string namespace = 1;
+    int32 page_size = 2;
+    bytes next_page_token = 3;
+    string query = 4;
```
cretz, line 2349: Need docs on what this query exactly is

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+  string host_id = 3;
+  string task_queue = 4;
+  string deployment_name = 5;
+  string build_id = 6;
+  string sdk_name = 7;
+  string sdk_version = 8;
+
+  string worker_identity = 9;
+
+  int32 worker_status = 10;
```
Quinn-With-Two-Ns, line 46: What does this represent?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+  google.protobuf.Duration uptime = 11;
+
+  google.protobuf.Timestamp last_heartbeat_time = 12;
+
+  WorkerTaskStatus workflow_task_status = 13;
+  WorkerTaskStatus activity_task_status = 14;
```
Quinn-With-Two-Ns, line 52: In Go a worker can also have a session worker, conceptually in the SDK the session worker is under the users worker, but is listening on a different task queue so I guess it would heartbeat separately?

temporal/api/workflowservice/v1/service.proto
cretz, line 1189: Is there value in making this part of task queue APIs instead? If there needs to be some way to list task queues (and maybe have a filter to only include ones with workers), I wonder if we should consider that.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+
+  int32 processed_tasks = 5;
+  int32 failed_tasks = 6;
+
+  float process_rate_min = 7;
+  float failure_rate_min = 8;
+}
+
+message WorkerStatus {
+  string namespace_id = 1;
```
cretz, line 33: What is the use of namespace ID here?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+  // Unique identifier for the namespace.
+  string worker_id = 2;
```
cretz, line 35: Not following what is meant here by "unique identifier for the namespace" for a worker ID and how this compares with worker identity.

temporal/api/worker/v1/message.proto
cretz, line 46: What is this integer?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+
+  string host_id = 3;
+  string task_queue = 4;
+  string deployment_name = 5;
+  string build_id = 6;
```
cretz, line 40: I think we have a data type for these

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+  string sdk_name = 7;
+  string sdk_version = 8;
+
+  string worker_identity = 9;
+
+  int32 worker_status = 10;
+  google.protobuf.Duration uptime = 11;
+
+  google.protobuf.Timestamp last_heartbeat_time = 12;
```
cretz, line 49: Who populates this value?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+
+  WorkerTaskStatus workflow_task_status = 13;
+  WorkerTaskStatus activity_task_status = 14;
+  WorkerTaskStatus nexus_task_status = 15;
+
+  float cpu_usage_percent = 16;
+  int64 memory_usage_bytes = 17;
+  float cache_hit_ratio = 18;
+  float cache_size = 19;
```
cretz, line 58: Is cache size a float?

temporal/api/worker/v1/message.proto
Quinn-With-Two-Ns, line 33: What is this field for? The requests that contain `WorkerStatus` already have a namespace right?

temporal/api/worker/v1/message.proto
Quinn-With-Two-Ns, line 58: I think it is worth noting `cache_size` is not a per worker property , in Go it is a per process size and Java it is per factory (a factory just being a collection of workers).

Also does this field mean the max cache size or the current free cache size? I think the latter is more useful to users no?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+}
```
yuandrew, line 59: I know we previously mentioned including kubernetes information, is that something we want to add in the initial design here?

temporal/api/worker/v1/message.proto
ychebotarev: as far as I know worker identity can be provided by a user and doesn't need to be unique in the namespace.
worker_id is something that must be unique in the namespace, and will be provided by SDK.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+  string worker_identity = 9;
+
+  int32 worker_status = 10;
+  google.protobuf.Duration uptime = 11;
+
+  google.protobuf.Timestamp last_heartbeat_time = 12;
+
+  WorkerTaskStatus workflow_task_status = 13;
+  WorkerTaskStatus activity_task_status = 14;
+  WorkerTaskStatus nexus_task_status = 15;
```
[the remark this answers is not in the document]
ychebotarev: Not intentionally. But I guess local activity status should be different. 
What metrics for local activities users may want to see?

temporal/api/workflowservice/v1/service.proto
ychebotarev: ok

temporal/api/worker/v1/message.proto
ychebotarev: removed

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes I think it deserves to be reported and as its own status. Almost everything under `WorkerTaskStatus` is relevant for local activities as well. Except maybe for `task_pollers`. I think we can just add another `WorkerTaskStatus ` for local activities?

temporal/api/worker/v1/message.proto
Quinn-With-Two-Ns: The SDK may have a different number of pollers on the sticky and non sticky queue. It may be useful for a user to know the rates separately as well, like some latency sensitive users like to make sure they are taking advantage of the sticky task queue as much as possible so they could want to monitor the two rates separately.

It is possible for the SDK to be polling on one of the queues but not the other in some scenarios.

Talking about it more is making me think we should have them as separate status's @yuandrew @cretz thoughts?

temporal/api/worker/v1/message.proto
Quinn-With-Two-Ns: Is the status a fixed set of a values? Might be easier for the UI team if we use an enum?

temporal/api/workflowservice/v1/service.proto
cretz: Deployments cross task queues, but workers never do. Workers only ever belong to one task queue, so they can be seen as just a property of a task queue. I understand that it may be too much data to put on each task queue list item, but at the least, a subset of the info may need to be there.

temporal/api/worker/v1/message.proto
cretz: We can't guarantee this any more than we can for worker identity. What happens if it's not unique? And we of course cannot require users start providing some new value they never have before, so what would SDK be expected to set as the default value for this? If it's worker identity, which seems reasonable, we might as well use worker identity as the identifier (which is meant to uniquely identify the worker, even if we can't enforce it in a distributed way).

temporal/api/worker/v1/message.proto
Quinn-With-Two-Ns: Worker sessions are a Go SDK feature https://docs.temporal.io/develop/go/sessions, that let a user schedule multiple activities on the same worker. To achieve this the worker polls on a couple of additional task queues (one task queue for session creation and one to schedule the activities). From the users perspective it is logically under one worker, but there are actually three task queues being used.

temporal/api/worker/v1/message.proto
Sushisource: I would segregate this differently. IMO we should have `SlotsStatus` and `PollerStatus` messages and have, independently:

* `workflow_slots_status`
* `activity_slots_status`
* `local_activity_slots_status`
* `nexus_slots_status`
* `workflow_poller_status`
* `workflow_sticky_poller_status`
* `activity_poller_status`
* `nexus_poller_status`

temporal/api/worker/v1/message.proto
Sushisource: IMO it probably makes sense to just leave this off for now. We don't really have these statii well-defined across languages, or what it means to be healthy yet either. We can always add this in later, after the SDK team has agreed on what those states should be / are

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,59 @@
...
+  float process_rate_min = 7;
+  float failure_rate_min = 8;
+}
+
+message WorkerStatus {
+  string namespace_id = 1;
+  // Unique identifier for the namespace.
+  string worker_id = 2;
+
+  string host_id = 3;
```
Sushisource, line 37: Can add `process_id`

temporal/api/worker/v1/message.proto
ychebotarev: I want to be somewhat agnostic to a specific environment.
How about I will add "environment_info" (string)?

temporal/api/worker/v1/message.proto
Sushisource: If this exists purely to be a primary key, I don't think there's anything we can do to make that consistent across worker restarts (nor do I know we want to).

So, given that, SDK can just make up a UUID - but what's the point? We could just hash together `namespace_id + task_queue + worker_id + host_id + process_id` for example.

temporal/api/worker/v1/message.proto
ychebotarev: tnx, changed

temporal/api/worker/v1/message.proto
ychebotarev: >We can't guarantee this 

new.guild?

temporal/api/worker/v1/message.proto
ychebotarev: ok. but don't expect it to be final/complete. My goal so far to get "this is missing, this is wrong" comments.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: did local activities has slots? can worker have more slots then workflow task slots?

temporal/api/worker/v1/message.proto
yiminc: I feel like we don't need to handle the go session thing for now. Unless we decided to make the session more generally available across SDKs.

temporal/api/worker/v1/message.proto
ychebotarev: yeah, if we add local activity it make sense.
I will need to sync on what needs to be in "xxx_poller_status"

temporal/api/worker/v1/message.proto
ychebotarev: I also don't want to add sdk-specific metrics.

Worth to mention - info in this API is expected to be "top level". 
There will also be an interface to get "low level" metrics/information from workers. example - "last 100 errors", etc.
This will be done via "worker commands". Getting "session" information can be part of that.

temporal/api/worker/v1/message.proto
cretz: I think we should make it final/complete before merging this PR

temporal/api/workflowservice/v1/service.proto
cretz, line 1: Will we support `DescribeWorker`?

cretz, line 1: Will we support `DescribeWorker`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2327,3 +2328,43 @@ message TriggerWorkflowRuleResponse {
...
+//* WorkerId
+//* HostId
+//* TaskQueue
+//* DeploymentName
+//* BuildId
+//* SdkName
+//* SdkVersion
+//* Uptime
+//* LastHeartbeatTime
+//* Status
```
cretz, line 2356: What is "Status" here?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1173,4 +1173,25 @@ service WorkflowService {
         };
     }
 
+    // WorkerHeartbeat receive heartbeat request from the worker.
+    rpc RecordWorkerHeartbeat(RecordWorkerHeartbeatRequest) returns (RecordWorkerHeartbeatResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/worker-heartbeat"
```
cretz, line 1179:
```suggestion
            post: "/namespaces/{namespace}/workers/heartbeat"
```

temporal/api/worker/v1/message.proto
ychebotarev: Chad: add comment that SDK should set it to "now".

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+// Holds everything needed to identify the host/process context
+message WorkerHostInfo {
+  // worker host identifier, should be unique for the namespace.
+  string host_id = 1;
+
+  // worker process identifier, should be unique for the host.
+  string process_id = 2;
+
+  // freeform (e.g. "k8s container abc123", etc.)
+  string host_context = 3;
```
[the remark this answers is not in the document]
ychebotarev: may be we don't need it at all?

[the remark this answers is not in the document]
ychebotarev: memo? (if it needs to be in the list). Unindexed, encryptable.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: lets get more opitions.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+  int32 task_pollers = 2;
+
+  int32 slots_available = 3;
+  int32 slots_used = 4;
+
+  int32 processed_tasks = 5;
+  int32 failed_tasks = 6;
+
+  float process_rate_min = 7;
+  float failure_rate_min = 8;
```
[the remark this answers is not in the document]
ychebotarev: remove "_min", confusing with "minimum". Or - make "_sec".
process - success??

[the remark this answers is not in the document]
ychebotarev: need comment that indicated that this is since last heartbeat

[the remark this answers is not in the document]
ychebotarev: or we may just have "number of successfull/failer task since last heartbeat"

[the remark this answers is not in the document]
ychebotarev: document everty case where values change per heartbeat( gauge vs counter)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I added comments to the status field. Essentially it is "running/shutting down/shutdown".
We may think about adding "Stuck" status, that we will set when extracting the data.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2327,3 +2328,43 @@ message TriggerWorkflowRuleResponse {
...
+
+    temporal.api.worker.v1.WorkerStatus worker_status = 3;
+}
+
+message RecordWorkerHeartbeatResponse {
+
+}
+
+// `query` in ListWorkers is used to filter workers based on worker status info.
+// The following worker status attributes are expected are supported as part of the query:
```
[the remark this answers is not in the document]
ychebotarev: moved

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
Sushisource: Yes, slot metrics are the same for all kinds of slots

temporal/api/worker/v1/message.proto
ychebotarev: I split hoststatus into "slot status" and "poller status". Basically the way Spenser proposed.

temporal/api/worker/v1/message.proto
ychebotarev: not in this version.

temporal/api/worker/v1/message.proto
ychebotarev: we will. I will make clear in the description that this is not for merge.

temporal/api/worker/v1/message.proto
ychebotarev: worker identity is added to queriable fields.

temporal/api/worker/v1/message.proto
ychebotarev: I change all those "cache" fields to match SDK metrics.
Now we will have:

  int32 sticky_cache_hit = 20;
  int32 sticky_cache_miss = 21;
  int32 sticky_cache_size = 22;

temporal/api/workflowservice/v1/service.proto
ychebotarev: Discussed. Yes we will, and DescribeWorker will contain more information, like config/environment/user metadata/etc.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+  // worker host identifier, should be unique for the namespace.
+  string host_id = 1;
+
+  // worker process identifier, should be unique for the host.
+  string process_id = 2;
+
+  // freeform (e.g. "k8s container abc123", etc.)
+  string host_context = 3;
+
+  // Worker identity, set by the worker, may not be unique.
```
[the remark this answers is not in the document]
ychebotarev: fixed

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+  string worker_identity = 7;
```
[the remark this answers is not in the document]
ychebotarev: fixed

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+}
+
+message WorkerStatus {
+  // Worker identifier, should be unique for the namespace. Required.
+  string worker_id = 1;
```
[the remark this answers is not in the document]
ychebotarev: Initially it will be from Heartbeat API. Later (or at the same time, depends how hard it is for SDK) the same info will be coming as a part of Poll.
Cadence is still controlled by SDK (we don't want to receive it on every poll)
Also (from discussion with Chad) status should also come during Shutdown.

>So perhaps it makes sense to call this worker_heartbeat_id

We want eventually to have it as a part of workflow visibility (or even in event history), and worker_heartbeat_id will be weird there.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+
+  // Worker host information. Required.
+  WorkerHostInfo host_info = 2;
```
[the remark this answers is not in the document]
ychebotarev: we are discussing on adding k8s specific info, so lets keep it separated.

temporal/api/worker/v1/message.proto
ychebotarev: changed.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+  string process_id = 2;
+
+  // freeform (e.g. "k8s container abc123", etc.)
+  string host_context = 3;
+
+  // Worker identity, set by the worker, may not be unique.
+  string worker_identity = 7;
+}
+
+message WorkerStatus {
```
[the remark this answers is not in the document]
ychebotarev: renamed to WorkerInfo (and all other XXXStatus to XXXInfo)

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
Sushisource: Then maybe just `worker_uuid` and just be super specific about it

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
Sushisource: Works for me

temporal/api/worker/v1/message.proto
ychebotarev: and bring it back.
We will have "running", "shutting down", "shutdown". Those are well defined.
I will create an enum.

ychebotarev: also changed to enum

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: renamed, and now its absolute values.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,76 @@
...
+  // Worker host information. Required.
+  WorkerHostInfo host_info = 2;
+
+  // Task queue this worker is polling for tasks. Required.
+  string task_queue = 3;
+
+  // Required.
+  temporal.api.deployment.v1.WorkerDeploymentVersion deployment_version = 4;
+
+  string sdk_name = 6;
```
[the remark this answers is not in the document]
ychebotarev: fixed

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+  string sdk_name = 5;
+  string sdk_version = 6;
+
+  // Worker status. Possible values - "running", "shutting_down", "shutdown".
+  temporal.api.enums.v1.WorkerStatus status = 7;
+
+  // Uptime of the worker, since it started.
+  // This is the time since the worker started, not the time since the last heartbeat.
+  // It can be used to determine worker start time. (current time - uptime)
+  google.protobuf.Duration uptime = 8;
```
dnr, line 82: I've been finding it awkward to deal with fields that are a distance to/from now, and much easier to deal with the absolute time instead. Why not just make this
```suggestion
  google.protobuf.Timestamp start_time = 8;
```

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+  WorkerSlotsInfo activity_task_slots_info = 11;
+  WorkerSlotsInfo nexus_task_slots_info = 12;
+  WorkerSlotsInfo local_activity_slots_info = 13;
+
+  WorkerPollerInfo workflow_poller_info = 14;
+  WorkerPollerInfo workflow_sticky_poller_info = 15;
+  WorkerPollerInfo activity_poller_info = 16;
+  WorkerPollerInfo nexus_poller_info = 17;
+
+  float cpu_usage_percent = 18;
```
dnr, line 97: cpu usage is surprisingly tricky to measure and interpret. Over what time range is this measured? Does it include kernel time or just user time? Treating it as relative to some other thing makes it even worse: relative to the machine? the container soft limits? hard limits? To one core? What about burst (frequency scaling or burstable cycles in some cloud vm types)? SMT?

To do this properly you need multiple stats. But to keep it simple, maybe something like:

```suggestion
  // Average number of CPU cores used per second by this worker process (user time only),
  // over the time between the previous heartbeat and this heartbeat.
  float cpu_cores_used = 18;
```

i.e. ignore the machine/container and just count cpu time.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+  int64 memory_usage_bytes = 19;
```
dnr, line 98: Total allocated or in-use? Just heap or stacks too? Counting shared memory or only private?

I'm wondering if cpu/memory shouldn't even be here, maybe leave that to monitoring systems that know how to deal with all the intricacies?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+
+  // A Workflow Task found a cached Workflow Execution to run against.
+  int32 sticky_cache_hit = 20;
```
dnr, line 101: Are these counters since process start or since the last heartbeat?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+
+  // Number of tasks processed in the last minute.
+  int32 processed_tasks_last_minute = 5;
+  // Number of failed tasks processed in the last minute.
+  int32 failure_tasks_last_minute = 6;
+}
+
+// Holds everything needed to identify the worker host/process context
+message WorkerHostInfo {
+  // Worker host identifier, should be unique for the namespace.
```
dnr, line 47: what does "unique for the namespace" mean? obviously multiple workers can run on the same host...

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
dnr: I think `worker_uuid` makes sense here, otherwise `worker_identity` and `worker_id` is too confusing (id is just short for identity so it's the same word). At least the "uu" makes it clear that this is something different and it should be in uuid form.

temporal/api/worker/v1/message.proto
ychebotarev: Ok, not in https://docs.temporal.io/references/sdk-metrics.
I will start the discussion.

temporal/api/worker/v1/message.proto
ychebotarev: Now this is something we already have: https://docs.temporal.io/references/sdk-metrics
So I guess since process start.

temporal/api/worker/v1/message.proto
cretz: I too prefer the fixed start time. I can subtract as necessary, but you can't index uptime. And I think you should consider indexing this value similar to other when listing. It is reasonable to want a list of workers started before a certain time.

temporal/api/worker/v1/message.proto
cretz: I would be ok just dropping all of these system metrics for the initial version. I think worker status is not necessarily system/process status. If we want to have a system/process metrics section (maybe later?), we should more clearly define exactly how the values are obtained.

Sushisource: Well, we do have them in SDK - we need them for the resource based tuner. David is right that properly measuring them is extremely tricky, though (and there are some known problems with my existing implementation). This would be a forcing function to fix those.

I'm OK with adding them in later, but I do think we want them, for the same reason we want all this other stuff, which is that people simply _don't_ set up other forms of monitoring.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: yeah, lets keep it separated. It is nice to have all host related info in one place. Are there any concerns?

temporal/api/worker/v1/message.proto
ychebotarev: ok, start_time it is

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2327,3 +2328,45 @@ message TriggerWorkflowRuleResponse {
...
+    // The following worker status attributes are expected are supported as part of the query:
+    //* WorkerId
+    //* WorkerIdentity
+    //* HostId
+    //* TaskQueue
+    //* DeploymentName
+    //* BuildId
+    //* SdkName
+    //* SdkVersion
+    //* Uptime
```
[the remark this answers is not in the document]
ychebotarev: will change to StartTime.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+
+message WorkerSlotsInfo {
+
+  // Number of slots available for the worker to specific tasks.
+  int32 slots_available = 1;
+  // Number of slots used by the worker for specific tasks.
+  int32 slots_used = 2;
+
+  // Total number of tasks processed by the worker so far.
+  int32 processed_tasks = 3;
```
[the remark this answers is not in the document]
ychebotarev: Usually "total" indicates both successful and failed. Like "total workflows".
And for me "processed" implies that they are completed.
Otherwise I will put it as "currently processing". I will try to make it more clear.

temporal/api/workflowservice/v1/service.proto
cretz: My concern is this PR can kinda be affected by both 1) how you will uniquely identify a worker and 2) what's in the worker info model vs live-queried model. Maybe we can just have a high level discussion on those. Mainly for number 1, I just want to know whether I can look up a worker by its identity (the thing we put on events even though it technically can be non-unique), and for number 2, I just want to confirm everything in this PR we are sure isn't better served live-obtaining from the worker.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
cretz: "completed" often can mean "completed successfully" and "completed unsuccessfully". Yeah, can just clarify. And also this means that we aren't showing how many are being processed at this time? That can kinda be obtained from slots used (though that includes poller count), so not necessarily required.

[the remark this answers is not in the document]
ychebotarev: >And also this means that we aren't showing how many are being processed at this time? That can kinda be obtained from slots used (though that includes poller count), so not necessarily required. 

I though "occupied slots" has the information only on currently processing (not processed) tasks?
If task is completed slot is free, isn't that the case?

Either way the intention is - total (commutative) value from the start. For more recent we will have "processed_tasks_last_minute" (or whatever the name will be)

temporal/api/workflowservice/v1/service.proto
ychebotarev: >I just want to confirm everything in this PR we are sure isn't better served live-obtaining from the worker.

One of the requirement - be able to handle 1M workers. And then you call ListWorker... Fan out to 1M workers? What about paging? Have internal list of active workers, select XXX, fun out only those? But we need to keep this list, and keep it up to date, and thus we come to heartbeats again. This is in short.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: changed to `worker_instance_key `

temporal/api/workflowservice/v1/service.proto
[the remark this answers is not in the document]
ychebotarev: I don't mind - not much work for me. But for the first iteration it is not needed. So up to SDK team. @Sushisource ?

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: updated the comments.

temporal/api/worker/v1/message.proto
ychebotarev: we will be providing what we have in SDK.

temporal/api/worker/v1/message.proto
ychebotarev: we will be providing what we have in SDK

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,106 @@
...
+  // Total number of failed tasks processed by the worker so far.
+  int32 failed_tasks = 4;
+
+  // Number of tasks processed in the last minute.
+  int32 processed_tasks_last_minute = 5;
+  // Number of failed tasks processed in the last minute.
+  int32 failure_tasks_last_minute = 6;
```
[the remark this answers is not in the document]
cretz: The problem with this kind of information is you can't really tell what the arbitrary "last minute" means which means this can never be used for trending. If you got explicit numbers for explicit windows instead of arbitrarily relative "last minute" it could at least be used for trending.

Having said that, if others agree with the last minute approach, ok. But I think it may be worth calling it `recently_processed_tasks` or something and just documenting that it represents a minute.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
cretz: The concern is that there is no value for the 3 fields, one of which is not host info (worker identity). Same reason you don't put sdk name and sdk version in "SDK info". It's unnecessary grouping IMO

temporal/api/workflowservice/v1/service.proto
cretz: That's the question here, is whether all the information here is worth being on list calls vs describe call. It's hard to know without even seeing a describe call to compare data with.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
cretz: I think the comments are missing whether it's successfully processed or just processed.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: For me this implies 'last minute, ending with the moment this request was send', not just some arbitrary point in time. Mostly because this is the only explanation that makes sense - you must understand that interval is not arbitrary.

I don't see how I can add more clarity with the name itself without making the name too big, Im open to suggestions.  
I also think "recently_processed_task" basically forces you to read documentation. (what means recently? what is the interval?)

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: I expect to add something eventually. I will be easier to add into separate message, then to have those fields be in different places.

temporal/api/worker/v1/message.proto
ychebotarev: ok I will add 'total/last/current' qualifies

temporal/api/workflowservice/v1/service.proto
ychebotarev: For now all information here will be in ListWorkers call. At least I don't see much reason to trim it.
We don't 100% know what will be in DescribeWorker call (besides this information), this is in discovery phase. We will not know until we implement worker commands.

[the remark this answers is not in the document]
ychebotarev: added to ShutdownWorker

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: that is because I didn't push changes, and technically didn't click "re-review" request.
Updated.

temporal/api/worker/v1/message.proto
ychebotarev: keeping this open

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
Sushisource: I would call it:

`xxx_last_interval`

IE: They are everything processed since the last reporting interval. This divorces the API from the interval, so that it can be configurable. It also acknowledges the reality that the reporting interval will have some small deviation from exactly one minute. Combine this with the fact that we record when the last heartbeat was received, and you've got all the info you need.

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
cretz: :+1: Any ideas on what it may be? Can we remove worker identity out? Do you want to group SDK name/version into its own message?

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: `xxx_last_interval` - I though about this. Mmmm... not sure I like it.
It also make this "processing rate" depend on heartbeat rate. 
So if users take a dependency they should also explicitly provide heartbeat interval. 
Like lets say they want to alert is task processing rate is low. with "last_minute" they can just use raw values.
Or 'xxx_last_minute'/60 if they want "per sec".

With `xxx_last_interval` they should use `xxx_last_interval`/'heartbeat_interval_in_min/sec/whatever'.
Though I do understand it is easier to implement (drop counter when we send heartbeats).

Hm. Should we add "previous_heartbeat_time"? 
Then is will become `xxx_last_interval`/(last_heartbeat_time-previous_heartbeat_time)

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
ychebotarev: yeah, some specific host info that depends on infra, like "K8s container", some generic environment info, etc. 
For SDK - it is just 2 fields, and what else we may add?

temporal/api/worker/v1/message.proto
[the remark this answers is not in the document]
Sushisource: > Hm. Should we add "previous_heartbeat_time"?

Yeah, I thought that was implied. SDK doesn't need to track, it, we just know server side when the last one was, and when this one was, and can therefore easily provide the elapsed time between the two.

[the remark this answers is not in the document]
ychebotarev: talked offline.
Will change this to 
`processed_tasks_last_interval`
` failed_tasks_last_interval`
and add
`elapsed_since_last_heartbeat`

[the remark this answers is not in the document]
ychebotarev: changed

temporal/api/enums/v1/common.proto
```diff
@@ -96,4 +96,13 @@ enum ApplicationErrorCategory {
...
     APPLICATION_ERROR_CATEGORY_BENIGN = 1;
-}
\ No newline at end of file
+}
+
+// (-- api-linter: core::0216::synonyms=disabled
+//     aip.dev/not-precedent: It seems we have both state and status, and status is a better fit for workers. --)
+enum WorkerStatus {
+    WORKER_STATUS_UNSPECIFIED = 0;
+    WORKER_STATUS_STATUS_RUNNING = 1;
```
Sushisource: Status status

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,111 @@
...
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "temporal/api/deployment/v1/message.proto";
+import "temporal/api/enums/v1/common.proto";
+
+message WorkerPollerInfo {
+
+  int32 active_pollers = 1;
+
+  int32 available_pollers = 2;
```
Sushisource: Honestly, we can probably just leave this off. Autoscaling will become the default eventually and this value doesn't apply there.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,111 @@
...
+
+  google.protobuf.Timestamp last_successful_poll_time = 3;
+}
+
+message WorkerSlotsInfo {
+
+  // Number of slots available for the worker to specific tasks.
+  int32 current_slots_available = 1;
```
```suggestion
  // Number of slots available for the worker to specific tasks. May be -1 if the upper bound is not known.
  int32 current_slots_available = 1;
```

temporal/api/worker/v1/message.proto
Sushisource: Right but how does that apply to host_name? To David's point, there will absolutely be multiple workers with the same host_name in the same namespace.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,111 @@
...
+  string worker_identity = 3;
+}
+
+// Worker info message, contains information about the worker and its current state.
+// All information is provided by the worker itself.
+// (-- api-linter: core::0140::prepositions=disabled
+//     aip.dev/not-precedent: Removing those words make names less clear. --)
+message WorkerInfo {
+  // Worker identifier, should be unique for the namespace.
+  // It is different from worker identity, which is usually a combination of host_name and process_id.
```
```suggestion
  // It is distinct from worker identity, which is not necessarily namespace-unique.
```

No need to repeat the info about the composition of identity here.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,111 @@
...
+
+  // Task queue this worker is polling for tasks.
+  string task_queue = 3;
+
+  temporal.api.deployment.v1.WorkerDeploymentVersion deployment_version = 4;
+
+  string sdk_name = 5;
+  string sdk_version = 6;
+
+  // Worker status. Possible values - "running", "shutting_down", "shutdown".
```
```suggestion
  // Worker status.
```
Wouldn't want this to get out-of-sync with actual enum values, which are easy enough to look at.

temporal/api/worker/v1/message.proto
ychebotarev: oh I just realized that 'host_name' has "unique" comment.... copy/paste error.
I though it was for "worker_instance_key".
Removed.

temporal/api/enums/v1/common.proto
ychebotarev: fixed

temporal/api/worker/v1/message.proto
ychebotarev: removed

temporal/api/worker/v1/message.proto
ychebotarev: discussed. removed

PR #601, opened 2025-06-04
Add worker visibility API - RecordWorkerHeartbeat and ListWorkers

ychebotarev:
**What changed?**
What changed?
* 2 new APIs added:
   * WorkerHeartbeat
   * ListWorkerStatus
* WorkerInfo information added to:
   * ShutDownWorker request
   * Poll(Activity|Nexus|Workflow)TaskQueue requests
   
This is "for merge".
Here is a previous PR, where we discuss this at length: https://github.com/temporalio/api/pull/599

**Why?**
Part of the worker visibility work. Workers should be able to report their status. 
Users should be able to get information about workers.

**Breaking changes**
Yes - new API is introduce.

**Server PR**
[Server PR](https://github.com/temporalio/temporal/pull/7870)

### DIFF
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -28,9 +28,9 @@ message WorkerSlotsInfo {
 
   // Number of slots available for the worker to specific tasks.
   // May be -1 if the upper bound is not known.
-  int32 current_slots_available = 1;
+  int32 current_available_slots = 1;
   // Number of slots used by the worker for specific tasks.
-  int32 current_slots_used = 2;
+  int32 current_used_slots = 2;
 
   // Kind of the slot supplier, which is used to determine how the slots are allocated.
   // Possible values: "Fixed | ResourceBased | Custom String"
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -1,0 +1,110 @@
+syntax = "proto3";
+
+package temporal.api.worker.v1;
+
+option go_package = "go.temporal.io/api/worker/v1;worker";
+option java_package = "io.temporal.api.worker.v1";
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Worker::V1";
+option csharp_namespace = "Temporalio.Api.Worker.V1";
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "temporal/api/deployment/v1/message.proto";
+import "temporal/api/enums/v1/common.proto";
+
+message WorkerPollerInfo {
+
+  int32 active_pollers = 1;
+
+  google.protobuf.Timestamp last_successful_poll_time = 2;
+}
+
+message WorkerSlotsInfo {
+
+  // Number of slots available for the worker to specific tasks.
+  // May be -1 if the upper bound is not known.
+  int32 current_slots_available = 1;
+  // Number of slots used by the worker for specific tasks.
+  int32 current_slots_used = 2;
+
+  // Total number of tasks processed (completed both successfully and unsuccesfully, or any other way)
+  // by the worker since the worker started. This is a cumulative counter.
+  int32 total_processed_tasks = 3;
+  // Total number of failed tasks processed by the worker so far.
+  int32 total_failed_tasks = 4;
+
+  // Number of tasks processed in since the last heartbeat from the worker.
+  // This is a cumulative counter, and it is reset to 0 each time the worker sends a heartbeat.
+  // Contains both successful and failed tasks.
+  int32 processed_tasks_last_interval = 5;
+  // Number of failed tasks processed since the last heartbeat from the worker.
+  int32 failure_tasks_last_interval = 6;
+}
+
+// Holds everything needed to identify the worker host/process context
+message WorkerHostInfo {
+  // Worker host identifier.
+  string host_name = 1;
+
+  // Worker process identifier, should be unique for the host.
+  string process_id = 2;
+
+  // Worker identity, set by the client, may not be unique.
+  // Usually host_name+(user group name)+process_id, but can be overwritten by the user.
+  string worker_identity = 3;
+}
+
+// Worker info message, contains information about the worker and its current state.
+// All information is provided by the worker itself.
+// (-- api-linter: core::0140::prepositions=disabled
+//     aip.dev/not-precedent: Removing those words make names less clear. --)
+message WorkerInfo {
+  // Worker identifier, should be unique for the namespace.
+  // It is distinct from worker identity, which is not necessarily namespace-unique.
+  string worker_instance_key = 1;
+
+  // Worker host information.
+  WorkerHostInfo host_info = 2;
+
+  // Task queue this worker is polling for tasks.
+  string task_queue = 3;
+
+  temporal.api.deployment.v1.WorkerDeploymentVersion deployment_version = 4;
+
+  string sdk_name = 5;
+  string sdk_version = 6;
+
+  // Worker status. Defined by SDK.
+  temporal.api.enums.v1.WorkerStatus status = 7;
+
+  // Worker start time.
+  // It can be used to determine worker uptime. (current time - start time)
+  google.protobuf.Timestamp start_time = 8;
+
+  // Last heartbeat time, coming from the worker. Worker should set it to "now".
+  google.protobuf.Timestamp last_heartbeat_time = 9;
+  // Elapsed time since the last heartbeat from the worker.
+  google.protobuf.Duration elapsed_since_last_heartbeat = 10;
+
+  WorkerSlotsInfo workflow_task_slots_info = 11;
+  WorkerSlotsInfo activity_task_slots_info = 12;
+  WorkerSlotsInfo nexus_task_slots_info = 13;
+  WorkerSlotsInfo local_activity_slots_info = 14;
+
+  WorkerPollerInfo workflow_poller_info = 15;
+  WorkerPollerInfo workflow_sticky_poller_info = 16;
+  WorkerPollerInfo activity_poller_info = 17;
+  WorkerPollerInfo nexus_poller_info = 18;
+
+  float current_host_cpu_usage = 19;
+  int64 current_host_mem_usage = 20;
+
+  // A Workflow Task found a cached Workflow Execution to run against.
+  int32 total_sticky_cache_hit = 21;
+  // A Workflow Task did not find a cached Workflow execution to run against.
+  int32 total_sticky_cache_miss = 22;
+  // Current cache size, expressed in number of Workflow Executions.
+  int32 current_sticky_cache_size = 23;
+}
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -93,6 +93,7 @@ message WorkerHeartbeat {
   google.protobuf.Timestamp start_time = 9;
 
   // Timestamp of this heartbeat, coming from the worker. Worker should set it to "now".
+  // Note that this timestamp comes directly from the worker and is subject to workers' clock skew.
   google.protobuf.Timestamp heartbeat_time = 10;
   // Elapsed time since the last heartbeat from the worker.
   google.protobuf.Duration elapsed_since_last_heartbeat = 11;
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -15,17 +15,16 @@ import "temporal/api/deployment/v1/message.proto";
 import "temporal/api/enums/v1/common.proto";
 
 message WorkerPollerInfo {
-
   // Number of polling RPCs that are currently in flight.
   int32 current_pollers = 1;
 
   google.protobuf.Timestamp last_successful_poll_time = 2;
 
+  // Set true if the number of concurrent pollers is auto-scaled
   bool is_autoscaling = 3;
 }
 
 message WorkerSlotsInfo {
-
   // Number of slots available for the worker to specific tasks.
   // May be -1 if the upper bound is not known.
   int32 current_available_slots = 1;
@@ -58,6 +57,12 @@ message WorkerHostInfo {
   // Worker process identifier, should be unique for the host.
   string process_id = 2;
 
+  // System used CPU as a float in the range [0.0, 1.0] where 1.0 is defined as all
+  // cores on the host pegged.
+  float current_host_cpu_usage = 3;
+  // System used memory as a float in the range [0.0, 1.0] where 1.0 is defined as
+  // all available memory on the host is used.
+  float current_host_mem_usage = 4;
 }
 
 // Worker info message, contains information about the worker and its current state.
@@ -108,19 +113,12 @@ message WorkerHeartbeat {
   WorkerPollerInfo activity_poller_info = 18;
   WorkerPollerInfo nexus_poller_info = 19;
 
-  // System used CPU as a float in the range [0.0, 1.0] where 1.0 is defined as all
-  // cores on the host pegged.
-  float current_host_cpu_usage = 20;
-  // System used memory as a float in the range [0.0, 1.0] where 1.0 is defined as
-  // all available memory on the host is used.
-  float current_host_mem_usage = 21;
-
   // A Workflow Task found a cached Workflow Execution to run against.
-  int32 total_sticky_cache_hit = 22;
+  int32 total_sticky_cache_hit = 20;
   // A Workflow Task did not find a cached Workflow execution to run against.
-  int32 total_sticky_cache_miss = 23;
+  int32 total_sticky_cache_miss = 21;
   // Current cache size, expressed in number of Workflow Executions.
-  int32 current_sticky_cache_size = 24;
+  int32 current_sticky_cache_size = 22;
 }
 
 message WorkerInfo {
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -92,8 +92,8 @@ message WorkerHeartbeat {
   // It can be used to determine worker uptime. (current time - start time)
   google.protobuf.Timestamp start_time = 9;
 
-  // Last heartbeat time, coming from the worker. Worker should set it to "now".
-  google.protobuf.Timestamp last_heartbeat_time = 10;
+  // Timestamp of this heartbeat, coming from the worker. Worker should set it to "now".
+  google.protobuf.Timestamp heartbeat_time = 10;
   // Elapsed time since the last heartbeat from the worker.
   google.protobuf.Duration elapsed_since_last_heartbeat = 11;
 
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -16,9 +16,12 @@ import "temporal/api/enums/v1/common.proto";
 
 message WorkerPollerInfo {
 
-  int32 active_pollers = 1;
+  // Number of polling RPCs that are currently in flight.
+  int32 current_pollers = 1;
 
   google.protobuf.Timestamp last_successful_poll_time = 2;
+
+  bool is_autoscaling = 3;
 }
 
 message WorkerSlotsInfo {
@@ -29,18 +32,22 @@ message WorkerSlotsInfo {
   // Number of slots used by the worker for specific tasks.
   int32 current_slots_used = 2;
 
+  // Kind of the slot supplier, which is used to determine how the slots are allocated.
+  // Possible values: "Fixed | ResourceBased | Custom String"
+  string slot_supplier_kind = 3;
+
   // Total number of tasks processed (completed both successfully and unsuccesfully, or any other way)
   // by the worker since the worker started. This is a cumulative counter.
-  int32 total_processed_tasks = 3;
+  int32 total_processed_tasks = 4;
   // Total number of failed tasks processed by the worker so far.
-  int32 total_failed_tasks = 4;
+  int32 total_failed_tasks = 5;
 
   // Number of tasks processed in since the last heartbeat from the worker.
   // This is a cumulative counter, and it is reset to 0 each time the worker sends a heartbeat.
   // Contains both successful and failed tasks.
-  int32 processed_tasks_last_interval = 5;
+  int32 last_interval_processed_tasks = 6;
   // Number of failed tasks processed since the last heartbeat from the worker.
-  int32 failure_tasks_last_interval = 6;
+  int32 last_interval_failure_tasks = 7;
 }
 
 // Holds everything needed to identify the worker host/process context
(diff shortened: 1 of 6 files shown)


temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+message WorkerHostInfo {
+  // Worker host identifier.
+  string host_name = 1;
+
+  // Worker process identifier, should be unique for the host.
+  string process_id = 2;
+
+  // Worker identity, set by the client, may not be unique.
+  // Usually host_name+(user group name)+process_id, but can be overwritten by the user.
+  string worker_identity = 3;
```
cretz: Mentioned on previous PR, this is not specific to host just because it has a default that is, and therefore this should be moved to top-level worker info IMO. I also think the other two fields don't deserve their own message any more than SDK name/version do, but not a strong opinion there.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+  // Total number of tasks processed (completed both successfully and unsuccesfully, or any other way)
+  // by the worker since the worker started. This is a cumulative counter.
+  int32 total_processed_tasks = 3;
+  // Total number of failed tasks processed by the worker so far.
+  int32 total_failed_tasks = 4;
+
+  // Number of tasks processed in since the last heartbeat from the worker.
+  // This is a cumulative counter, and it is reset to 0 each time the worker sends a heartbeat.
+  // Contains both successful and failed tasks.
+  int32 processed_tasks_last_interval = 5;
```
cretz: To confirm, this is completed tasks not just started tasks? So if since the last interval I started 30 activities, but they are all still running, this will be 0?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+  // May be -1 if the upper bound is not known.
+  int32 current_slots_available = 1;
+  // Number of slots used by the worker for specific tasks.
+  int32 current_slots_used = 2;
+
+  // Total number of tasks processed (completed both successfully and unsuccesfully, or any other way)
+  // by the worker since the worker started. This is a cumulative counter.
+  int32 total_processed_tasks = 3;
+  // Total number of failed tasks processed by the worker so far.
+  int32 total_failed_tasks = 4;
```
cretz: Inconsistent to put qualifier on the beginning in this case (i.e. `X_failed_tasks`) and end in the other case (i.e. `failed_tasks_X`). I recommend either `failed_tasks_total` + `failed_tasks_last_interval` or `total_failed_tasks` + `last_interval_failed_tasks`.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+option csharp_namespace = "Temporalio.Api.Worker.V1";
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "temporal/api/deployment/v1/message.proto";
+import "temporal/api/enums/v1/common.proto";
+
+message WorkerPollerInfo {
+
+  int32 active_pollers = 1;
```
cretz: Sometimes you use the term "active" and sometimes "current". I think this should be `current_pollers`.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+  WorkerSlotsInfo nexus_task_slots_info = 13;
+  WorkerSlotsInfo local_activity_slots_info = 14;
+
+  WorkerPollerInfo workflow_poller_info = 15;
+  WorkerPollerInfo workflow_sticky_poller_info = 16;
+  WorkerPollerInfo activity_poller_info = 17;
+  WorkerPollerInfo nexus_poller_info = 18;
+
+  float current_host_cpu_usage = 19;
+  int64 current_host_mem_usage = 20;
```
cretz: Can you document specifically what this represents for each value? There are a few things that could be taken to mean CPU and memory. I assume CPU is the equivalent of https://psutil.readthedocs.io/en/latest/#psutil.cpu_percent (i.e. across all cores)? I assume memory is something like what https://psutil.readthedocs.io/en/latest/#psutil.virtual_memory considers `used`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1170,6 +1180,9 @@ message GetSystemInfoResponse {
         // True if the server supports Nexus operations.
         // This flag is dependent both on server version and for Nexus to be enabled via server configuration.
         bool nexus = 11;
+
+        // True if the server support collecting workers information.
+        bool accept_worker_info = 12;
```
cretz: We are trying not to mutate this anymore. Cloud serves a fixed form of this. Worker should just try to send this info on poll regardless and as for the non-poll sending, worker should recognize an "unavailable" on first send and stop trying.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2346,3 +2362,45 @@ message TriggerWorkflowRuleResponse {
...
+message ListWorkersRequest {
+    string namespace = 1;
+    int32 page_size = 2;
+    bytes next_page_token = 3;
+
+    // `query` in ListWorkers is used to filter workers based on worker status info.
+    // The following worker status attributes are expected are supported as part of the query:
+    //* WorkerId
+    //* WorkerIdentity
+    //* HostId
```
cretz: Should this be HostName?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2346,3 +2362,45 @@ message TriggerWorkflowRuleResponse {
...
+    //* TaskQueue
+    //* DeploymentName
+    //* BuildId
+    //* SdkName
+    //* SdkVersion
+    //* StartTime
+    //* LastHeartbeatTime
```
cretz, line 2393: Are there concerns on updating this value so frequently in visibility store?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2346,3 +2362,45 @@ message TriggerWorkflowRuleResponse {
...
+    //* Status
+    // Currently metrics are not supported as a part of ListWorkers query.
+    string query = 4;
+}
+
+message ListWorkersResponse {
+    repeated temporal.api.worker.v1.WorkerInfo worker_info = 1;
```
cretz: This should be plural to match our other list responses

cretz: Consider making a wrapper message for this instead of returning the same worker info. There may be a time where we want server-set info and we'll regret we had nowhere to put it. Consider something like:

```proto
message WorkerDescription {
  temporal.api.worker.v1.WorkerInfo info = 1;
}
```

Or `WorkerListEntry` or anything. Basically some way for server to have room to add server-only things and not assume that the only thing we will ever return from list is the exact thing worker gives us. I expect some server-set info in here one day.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1173,4 +1173,25 @@ service WorkflowService {
         };
     }
 
+    // WorkerHeartbeat receive heartbeat request from the worker.
+    rpc RecordWorkerHeartbeat(RecordWorkerHeartbeatRequest) returns (RecordWorkerHeartbeatResponse) {
```
cretz: Pedantic, but inconsistent with yourself on whether you have a space between RPC name and open paren for request

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1173,4 +1173,25 @@ service WorkflowService {
...
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/worker-heartbeat"
```
```suggestion
            post: "/namespaces/{namespace}/workers/heartbeat"
```

temporal/api/worker/v1/message.proto
dnr: I know there was a long thread on these on the previous PR, sorry if I'm bringing up old stuff:

- If a heartbeat gets dropped, the data for that period is just lost?
- How is this intended to be used? It seems like a "rate" is maybe more generally useful? Should we calculate a rate from the # processed and interval and send that instead?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+
+  // Worker status. Defined by SDK.
+  temporal.api.enums.v1.WorkerStatus status = 7;
+
+  // Worker start time.
+  // It can be used to determine worker uptime. (current time - start time)
+  google.protobuf.Timestamp start_time = 8;
+
+  // Last heartbeat time, coming from the worker. Worker should set it to "now".
+  google.protobuf.Timestamp last_heartbeat_time = 9;
```
dnr: The comment/name is confusing.. the worker sets this to now() each time, so the server always sees this as a few ms in the past? So it's just measuring latency + clock skew? What is this useful for?

temporal/api/worker/v1/message.proto
dnr: Is this counting the number of polling loops running no matter what they're doing (i.e. maximum concurrent pollers) or the number of poll rpcs that are currently in flight?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+
+  // Worker identity, set by the client, may not be unique.
+  // Usually host_name+(user group name)+process_id, but can be overwritten by the user.
+  string worker_identity = 3;
+}
+
+// Worker info message, contains information about the worker and its current state.
+// All information is provided by the worker itself.
+// (-- api-linter: core::0140::prepositions=disabled
+//     aip.dev/not-precedent: Removing those words make names less clear. --)
```
dnr, line 71: there's no preposition in WorkerInfo?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+message WorkerInfo {
```
dnr: Sorry to be bike-shed-y, but could we try to improve this name? Maybe WorkerHeartbeat? WorkerStats? WorkerInfoSnapshot?

When I saw WorkerInfo in the poll requests below, I didn't realize it was including the full heartbeat.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2346,3 +2362,45 @@ message TriggerWorkflowRuleResponse {
     // True is the rule was applied, based on the rule conditions (predicate/visibility_query).
     bool applied = 1;
 }
+message RecordWorkerHeartbeatRequest {
+    // Namespace of the workflow which scheduled this activity.
+    string namespace = 1;
+
+    // The identity of the client who initiated this request.
+    string identity = 2;
```
dnr, line 2368: I don't think we need this, do we? There's every sort of identity you need inside of worker_info already.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2346,3 +2362,45 @@ message TriggerWorkflowRuleResponse {
...
+}
+
+message ListWorkersRequest {
+    string namespace = 1;
+    int32 page_size = 2;
+    bytes next_page_token = 3;
+
+    // `query` in ListWorkers is used to filter workers based on worker status info.
+    // The following worker status attributes are expected are supported as part of the query:
+    //* WorkerId
```
dnr: Didn't we rename this?

temporal/api/workflowservice/v1/request_response.proto
dnr: Can't we just add server-only fields to WorkerInfo and mark them as such?

There's no shortage of tag numbers...

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2346,3 +2362,45 @@ message TriggerWorkflowRuleResponse {
     // True is the rule was applied, based on the rule conditions (predicate/visibility_query).
     bool applied = 1;
 }
+message RecordWorkerHeartbeatRequest {
+    // Namespace of the workflow which scheduled this activity.
```
dnr: comment doesn't make sense

temporal/api/worker/v1/message.proto
Sushisource: Polling loops is kind of a pointless metric, and doesn't even apply to Core. This should be actual in-flight RPCs

temporal/api/worker/v1/message.proto
Sushisource: Yeah, I'm also not totally sure why we need it

temporal/api/workflowservice/v1/request_response.proto
cretz: I think it gets a little confusing to only accept some fields from worker on the same message. If we're willing to make separate messages for trivial things like host info, we should be willing to create separate messages for worker-provided data vs non-worker-provided data. I think of it like spec vs status in k8s. Or like in schedules case, `ScheduleListEntry`/`ScheduleListInfo` that wraps `ScheduleSpec`. I don't think it's very future proof to reuse the spec type as the list type with out some kind of wrapping. A single-field message just for list entries is mostly harmless but lets us grow as needed with the list response items.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+import "temporal/api/enums/v1/common.proto";
+
+message WorkerPollerInfo {
+
+  int32 active_pollers = 1;
+
+  google.protobuf.Timestamp last_successful_poll_time = 2;
+}
+
+message WorkerSlotsInfo {
```
Sushisource, line 27: Let's add a string here `slot_supplier_kind`. So we can say if fixed size/resource/based/user implemented.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Worker::V1";
+option csharp_namespace = "Temporalio.Api.Worker.V1";
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "temporal/api/deployment/v1/message.proto";
+import "temporal/api/enums/v1/common.proto";
+
+message WorkerPollerInfo {
```
Sushisource, line 17: Let's add `bool is_autoscaling` to indicate if the user has enabled auto-scaling pollers or not

temporal/api/worker/v1/message.proto
cretz: I think the idea is to get the exact window from a worker POV for the "current" metrics. If a worker is giving some time-based ephemeral data, it should give the exact time representation of that window as it sees it IMO. Also this is the exact same object we return back on list call with no expected server modification, so this gives callers information about the last time this was recorded according to worker clock.

If we do want the time window to be recorded by server, we may need to consider providing data since last successful heartbeat instead of a specific worker-provided window. But overall, yes, this is basically latency + clock skew (and client retries and such).

temporal/api/workflowservice/v1/request_response.proto
cretz: There is technically a difference between client identity and worker identity though we default the worker one to the client one. For most (all?) mutable RPCs I think it makes sense to accept client identity.

temporal/api/worker/v1/message.proto
ychebotarev: yes, we discuss it, and decide to go with raw numbers. 
* It is intended to be used to detect stuck workers (will be even better with historical data)
* we have "elapsed time", it can be used to calculate rate
* if heartbeat is dropped, yes data for that period will be lost. It is ok.
* even if heartbeat is dropped we still can calculate the rate, since "elasped time since the last heartbeat" is a part of heartbeat

Keeping open

temporal/api/worker/v1/message.proto
ychebotarev: qualifiers should be first. it is more common approach, and I would have more readable. I will change those "last_interval" names

temporal/api/worker/v1/message.proto
Sushisource: +1 to keeping the underlying data rather than sending the rate directly. Calculating rate server-side is easy, but if we send it directly we lose the ability to do better once we have historical data as Yuri mentions.

temporal/api/worker/v1/message.proto
ychebotarev: slot_supplier_kind added

temporal/api/worker/v1/message.proto
ychebotarev: done

temporal/api/worker/v1/message.proto
ychebotarev: changed name.

ychebotarev: also add comment. Keeping open since I'm not sure comment wording is correct.

temporal/api/worker/v1/message.proto
ychebotarev: i think last time I either missread this or the context was different. I agree, worker_identity shouldn't be here. I will move it out. But I will keep the structure - unlike SDK I feel it will make sense.

temporal/api/worker/v1/message.proto
ychebotarev: If comment is not clear - please propose a better comment. As for this field itself - we need it, at least to calculate uptime, but more like to be able to find "stuck" or "abandoned" workers.
I don't want to populate this way on a server side, for few reasons:
* for uptime (last_heartbeat_time - start_time), both times will come from the different sources
* indirectly it is a good way to spot workers that have completely craze time (like it is far in the future)
* I want all raw data come from workers.

temporal/api/worker/v1/message.proto
ychebotarev: there is `since` in the elapsed_since_last_heartbeat, and linter is not happy about it. But removing it will make is less clear.

temporal/api/worker/v1/message.proto
ychebotarev: I will skip long chain of thoughts...ok lets do this (hope there will be no much objections)

temporal/api/worker/v1/message.proto
cretz: > unlike SDK I feel it will make sense

Can you clarify how host name and process ID in its own object makes sense but SDK name and version does not?

temporal/api/worker/v1/message.proto
cretz: (also there's still my original question concerning started vs completed tasks)

temporal/api/worker/v1/message.proto
cretz: I do not see the comment

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: yes, changed

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I do expect to have something else (coming from server) in this answer, besides just heartbeat info. Creating separate structure for further extension make sense.

ychebotarev: also will allow us to separate receiving heartbeats from providing this info to the users.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: we are going to update them (in batch, but still) for every heartbeat. I don't think we can avoid updating this info/skip updates. Question is - will it be indexable? This is something to decide in P2 (I think it shouldn't, but may be users will lead us in different direction).

temporal/api/worker/v1/message.proto
ychebotarev: wait for me to ask for "re-review" :)

temporal/api/worker/v1/message.proto
ychebotarev: How can succesfull or failed task be not completed?

ychebotarev: in this case it will be running.

temporal/api/worker/v1/message.proto
ychebotarev: I already did in previous PR - this structure can be extended.

temporal/api/worker/v1/message.proto
ychebotarev: those are the metrics implemented in SDK. I got those comments from Spencer.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: I thought we still used this so I recommended it. Sending it every time on poll seems wasteful if it's just getting blackholed. I suppose we could have some logic where we always try to send it on the dedicated RPC on startup and then turn it off if it doesn't work. Slightly annoying, but, not horrible.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: keeping open

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ok I assume thing means we are removing it? removed (but will keep open)

temporal/api/worker/v1/message.proto
cretz: The SDK structure could be extended too. I am not sure I understand why host uniquely deserves its own message just because it may be extended like any other. It's obviously not a _big_ deal, just a strange inconsistency.

temporal/api/worker/v1/message.proto
cretz: Right. So to confirm, if since the last interval I started 30 activities, but they are all still running, this value will be 0? Do you users not want to know about running tasks instead of only completed ones? Are we ok intentionally hiding running tasks that may appear as if the worker is not busy? The answer may be "yes" because they can use used slots instead. Would like others' opinion here too.

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,110 @@
...
+  google.protobuf.Timestamp last_successful_poll_time = 2;
+}
+
+message WorkerSlotsInfo {
+
+  // Number of slots available for the worker to specific tasks.
+  // May be -1 if the upper bound is not known.
+  int32 current_slots_available = 1;
+  // Number of slots used by the worker for specific tasks.
+  int32 current_slots_used = 2;
```
cretz: > slots used by the worker for specific tasks.

@Sushisource - can you confirm that this is intentionally slots marked "used" and not meant to be slots reserved? Meaning you can't really get total slots on available + used (unless you also add active pollers, but even then there are races e.g. with eager activities/workflows).

temporal/api/worker/v1/message.proto
cretz: Hrmm, @Sushisource I wonder if we need to clarify somewhere what CPU and mem numbers we are getting (though doesn't have to be here if we're sharing this info with resource tuning). Or maybe it's obvious?

temporal/api/workflowservice/v1/request_response.proto
cretz: > I thought we still used this so I recommended it

Yeah, unfortunately due to this also being used as a health check higher than namespace level and the (publicly-discussed) multi-cell architecture of cloud, we can no longer rely on this. It is a fixed/static worldwide value in cloud and we've resolved to not update it anymore if at all possible. Can discuss internally how we can better make that clear for future implementers. This came up with the last one `nexus` before too, but despite disagreement, it was decided just decided to globally set it to true in cloud thereby limiting its value.

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1: My only concern here is whether indexable. Do we want to remove this from the indexable item list here?

ychebotarev: we for sure want to be able to query by "last heartbeat time".

temporal/api/worker/v1/message.proto
Sushisource: Yeah, I think having it be used makes good sense. It's more useful than reserved. I don't think being able to add them together to get the total is really useful in any meaningful way. Remaining available is the useful thing (if known).

temporal/api/worker/v1/message.proto
ychebotarev: The probability is different.  I think I've shared what I can on this topic, not much to add. Let's call it here.

temporal/api/worker/v1/message.proto
ychebotarev: I'm a bit lost in this question. lets me explain how I understand it.
If there are 10 tasks running, and nothing finished since the last heartbeat - `processed_tasks_last_interval` will be  0. 
But in this case the number of occupied slots will be 10, right?  
That is the whole idea (will be much clear if we have historical data) - if `processed_tasks_last_interval` is 0 (with historical data - for some time), and `current_slots_used` is high (with historical data - for some time) it means  that this worker is stuck (may be...only users know)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: removed, and closing this.

temporal/api/workflowservice/v1/request_response.proto
cretz: Do we? Or do we need to query by a certain state that may relate to last heartbeat time (e.g. staleness)?

> will it be indexable? [...] I think it shouldn't [...] we for sure want to

I am a bit confused here. Doesn't it have to be indexable to query on it? So if I have 5 million workers, as part of this first version, do we expect users to be able to query for all last heartbeat times before a certain timestamp? How can we defer indexability to a later decision but list it as an indexable attribute here?

temporal/api/worker/v1/message.proto
cretz: > But in this case the number of occupied slots will be 10, right?

Yes, after clarification from @Sushisource below that this doesn't include slots occupied for polling/eager reasons. So technically, yes, `current_slots_used` is effectively `current_processing_tasks` using this verbiage.

temporal/api/worker/v1/message.proto
cretz: I don't think the probability is different. I have noticed you have chosen not to put CPU and memory host info into host info, so I am suspecting you may not put other host info into host info? Also, how sure are you there is no more SDK info we may want in SDK info?

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,127 @@
...
+  google.protobuf.Timestamp last_successful_poll_time = 2;
+
+  bool is_autoscaling = 3;
+}
+
+message WorkerSlotsInfo {
+
+  // Number of slots available for the worker to specific tasks.
+  // May be -1 if the upper bound is not known.
+  int32 current_slots_available = 1;
```
cretz: Inconsistent word qualifier order with fields like `last_interval_processed_tasks`. It should be `current_available_slots` + `last_interval_processed_tasks` or `current_slots_available` + `last_interval_tasks_processed`.

ychebotarev: changed to `current_available_slots` (and next one)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: >How can we defer indelibility to a later decision

Easily - this is a two-way door decision, and right now it is a comment.
We are going to use "in-memory" approach for MVP. 
Id we decide not to index it later - we will remove it from the comment (and from filtering list).

temporal/api/worker/v1/message.proto
ychebotarev: I hear your concerns, but I'm not convinced the removal is necessary based on the arguments presented. Keeping has value. Let's keep this one for now.

temporal/api/worker/v1/message.proto
dnr: Oh I see, it applies to all the field names. I thought this was just for the message name and you had to comment them individually.

(Overall I think api-linter is a net negative and we should stop using it, but I won't bring that debate here :slightly_smiling_face:)

temporal/api/workflowservice/v1/request_response.proto
dnr: Fair enough. Do you have an example of when they would differ? Not a proxy, right, that would pass this through?

temporal/api/worker/v1/message.proto
dnr: I see the semantics, that makes sense.

On the name though: "last" in this message suggests it's the time of the previous one, not this one.

I would call it something more like "worker_timestamp", "heartbeat_timestamp", "worker_heartbeat_timestamp".

ychebotarev: Ah right. I was thinking about that in the beginning, decide agains it (with "uptime"), but then we change it to "elapsed since last time", and I didn't revisit.
Changed to just "heartbeat_time"

dnr (approved):

temporal/api/worker/v1/message.proto
```diff
@@ -0,0 +1,128 @@
...
+import "temporal/api/enums/v1/common.proto";
+
+message WorkerPollerInfo {
+
+  // Number of polling RPCs that are currently in flight.
+  int32 current_pollers = 1;
+
+  google.protobuf.Timestamp last_successful_poll_time = 2;
+
+  bool is_autoscaling = 3;
```
```suggestion
  // Set true if the number of concurrent pollers is auto-scaled
  bool is_autoscaling = 3;
```

temporal/api/worker/v1/message.proto
Sushisource: I don't think this is hugely important either way, but if you're convinced this is worth keeping, I'd agree with Chad that putting the CPU/Mem in there makes sense.

ychebotarev: Mmm.
Pro: it is host related, not worker related. So should be in host struct since we have it.
Con: not all metrics in one place.

Well. Yes, Pro > Con for me in this case, I will move them.

temporal/api/worker/v1/message.proto
ychebotarev: added.

cretz (approved):

PR #603, opened 2025-06-19
Add Stats to DescribeWorkerDeploymentVersion

stephanos:

**What changed?**

(1) Deprecated `task_queue_infos` in `deployment.WorkerDeploymentVersionInfo`.
(2) Added `version_task_queues` to `DescribeWorkerDeploymentVersionResponse`.

**Why?**

We want to report task queue stats for each task queue that is part of a worker deployment version.

The challenge is that the `taskqueue` package depends on the `deployment` package. So adding `TaskQueueStats` to `deployment.WorkerDeploymentVersionInfo` causes a cycle import error.

Weighing our options, we decided to effectively _move_ the task queue-related data from within the deployment package into the response message.

**Breaking changes**

Not yet; but in subsequent releases the deprecated field `task_queue_infos` will be removed.

**Server PR**

https://github.com/temporalio/temporal/pull/7959 (draft)

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -2030,6 +2030,15 @@ message DescribeWorkerDeploymentVersionRequest {
 
 message DescribeWorkerDeploymentVersionResponse {
     temporal.api.deployment.v1.WorkerDeploymentVersionInfo worker_deployment_version_info = 1;
+
+    // All the Task Queues that have ever polled from this Deployment version.
+    repeated VersionTaskQueue version_task_queues = 2;
+    // (-- api-linter: core::0123::resource-annotation=disabled --)
+    message VersionTaskQueue {
+        string name = 1;
+        temporal.api.enums.v1.TaskQueueType type = 2;
+        temporal.api.taskqueue.v1.TaskQueueStats stats = 3;
+    }
 }
 
 message DescribeWorkerDeploymentRequest {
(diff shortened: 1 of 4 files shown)


cretz (commented):
LGTM from a syntax POV, but will defer to other SDK team members more familiar w/ versioning. I admit I don't understand `VersionTaskQueueInfo` vs `VersionTaskQueue` and such.

stephanos:
Update: I just added `report_stats` since I realized it was missing.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2026,10 +2026,22 @@ message DescribeWorkerDeploymentVersionRequest {
     string version = 2 [deprecated = true];
     // Required.
     temporal.api.deployment.v1.WorkerDeploymentVersion deployment_version = 3;
+    // Report stats for version's task queues.
```
```suggestion
    // Report stats for task queues which have been polled by this version.
```

Is this accurate? If so, more understandable.

Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
stephanos: I _think_ polling is the only way to tie a task queue to a version? But I'm not 100% sure now.

ShahabT: > polling is the only way to tie a task queue to a version
Correct.

I agree that @Sushisource 's suggestion is more clear. The edge case is that user adds wfs to a TQ giving a versioning override forcing tasks to be backlogged against a version that never polled this TQ. In that case, the backlogged tasks would not be reflected here.

PR #604, opened 2025-06-20
UpdateTaskQueueConfig api changes

sivagirish81:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -315,3 +315,39 @@ message PollerScalingDecision {
   // all slots are already occupied.
   int32 poll_request_delta_suggestion = 1;
 }
+
+message RateLimit { 
+
+    // Zero is a valid rate limit.
+    float requests_per_second = 1;
+
+}
+
+message ConfigMetadata {
+
+    // Reason for why the rate limit was set.
+    string reason = 1;
+    
+    // Can be "system", "worker" or "api".
+    string source = 2;
+    
+    // Identity of the last updater.
+    // Set by the request's identity field.
+    string identity = 3;
+    
+    // Time of the last update.
+    google.protobuf.Timestamp update_time = 4;
+}
+
+message RateLimitConfig { 
+    RateLimit rate_limit = 1;    
+    ConfigMetadata metadata = 2;
+}
+
+message TaskQueueConfig {
+    TaskQueue task_queue = 1;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 2;
+    // Unless modified, this is the system-defined rate limit.
+    RateLimitConfig queue_rate_limit = 3;
+    RateLimitConfig fairness_keys_rate_limit_default = 4;
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -985,6 +985,9 @@ message RespondQueryTaskCompletedRequest {
     // traces.
     // Mutually exclusive with `query_result`. Set when the query fails.
     temporal.api.failure.v1.Failure failure = 7;
+    // Why did the task fail? It's important to note that many of the variants in this enum cannot
+    // apply to worker responses. See the type's doc for more.
+    temporal.api.enums.v1.WorkflowTaskFailedCause cause = 8;
 }
 
 message RespondQueryTaskCompletedResponse {
@@ -1083,6 +1086,9 @@ message DescribeTaskQueueRequest {
     // Report task reachability for the requested versions and all task types (task reachability is not reported
     // per task type).
     bool report_task_reachability = 10 [deprecated = true];
+
+    // Show Task Queue Config
+    bool show_task_queue_config = 11;
 }
 
 message DescribeTaskQueueResponse {
@@ -1110,6 +1116,8 @@ message DescribeTaskQueueResponse {
     // Only returned in ENHANCED mode.
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3 [deprecated = true];
+
+    temporal.api.taskqueue.v1.TaskQueueConfig configs  = 6;
 }
 
 message GetClusterInfoRequest {
@@ -2403,4 +2411,49 @@ message ListWorkersResponse {
 
     // Next page token
     bytes next_page_token = 2;
+}
+
+message UpdateTaskQueueConfigRequest {
+	message RateLimitUpdate { 
+
+	  float requests_per_second = 1;
+
+	  // Reason for why the rate limit was set.
+	  string reason = 2;
+
+	}
+
+  message UpdateQueueRateLimit {
+    // Rate limit to apply to task queue.
+    // It must be less than or equal to the system-level setting at the time
+    // of the request. If null, removes the existing rate limit.
+    RateLimitUpdate rate_limit = 1;
+  }
+
+  message UpdateFairnessKeyRateLimitDefault {
+    // Default rate limit to apply for all fairness keys.
+    // If null, removes the existing rate limit.
+    RateLimitUpdate rate_limit = 1;
+  }
+
+  string namespace = 1;
+  string identity = 2;
+   
+  // Selects the task queue to update.
+  string task_queue_name = 3;
+
+  temporal.api.enums.v1.TaskQueueType task_queue_type = 4;
+ 
+  // Update to queue-wide rate limit.
+  // If not set, this configuration is unchanged.
+  // NOTE: A limit set by the worker is overriden; and restored again when reset.
+  UpdateQueueRateLimit update_queue_rate_limit = 5;
+  
+  // Update to the default fairness key rate limit.
+  // If not set, this configuration is unchanged.
+  UpdateFairnessKeyRateLimitDefault update_fairness_key_rate_limit_default = 6;
+}
+
+message UpdateTaskQueueConfigResponse {
+    temporal.api.taskqueue.v1.TaskQueueConfig configs = 1;
 }
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -1194,4 +1194,17 @@ service WorkflowService {
             }
         };
     }
+
+    // Update taskqueue config and uncouple taskqueue settings from the worker lifecycle
+    rpc UpdateTaskQueueConfig (UpdateTaskQueueConfigRequest) returns (UpdateTaskQueueConfigResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/task-queues/update-config"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/task-queues/update-config"
+                body: "*"
+            }
+        };
+    }
+    
 }
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -1457,6 +1457,45 @@
         ]
       }
     },
+    "/api/v1/namespaces/{namespace}/task-queues/update-config": {
+      "post": {
+        "summary": "Update taskqueue config and uncouple taskqueue settings from the worker lifecycle",
+        "operationId": "UpdateTaskQueueConfig2",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1UpdateTaskQueueConfigResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpcStatus"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "namespace",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/WorkflowServiceUpdateTaskQueueConfigBody"
+            }
+          }
+        ],
+        "tags": [
+          "WorkflowService"
+        ]
+      }
+    },
     "/api/v1/namespaces/{namespace}/task-queues/{taskQueue.name}": {
       "get": {
         "summary": "DescribeTaskQueue returns the following information about the target task queue, broken down by Build ID:\n  - List of pollers\n  - Workflow Reachability status\n  - Backlog info for Workflow and/or Activity tasks",
@@ -1600,6 +1639,13 @@
           {
             "name": "reportTaskReachability",
             "description": "Deprecated (as part of the ENHANCED mode deprecation).\nReport task reachability for the requested versions and all task types (task reachability is not reported\nper task type).",
+            "in": "query",
+            "required": false,
+            "type": "boolean"
+          },
+          {
+            "name": "showTaskQueueConfig",
+            "description": "Show Task Queue Config",
             "in": "query",
             "required": false,
             "type": "boolean"
@@ -5097,6 +5143,45 @@
         ]
       }
     },
+    "/namespaces/{namespace}/task-queues/update-config": {
+      "post": {
+        "summary": "Update taskqueue config and uncouple taskqueue settings from the worker lifecycle",
+        "operationId": "UpdateTaskQueueConfig",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1UpdateTaskQueueConfigResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpcStatus"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "namespace",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/WorkflowServiceUpdateTaskQueueConfigBody"
+            }
+          }
+        ],
+        "tags": [
+          "WorkflowService"
+        ]
+      }
+    },
     "/namespaces/{namespace}/task-queues/{taskQueue.name}": {
       "get": {
         "summary": "DescribeTaskQueue returns the following information about the target task queue, broken down by Build ID:\n  - List of pollers\n  - Workflow Reachability status\n  - Backlog info for Workflow and/or Activity tasks",
@@ -5240,6 +5325,13 @@
           {
             "name": "reportTaskReachability",
             "description": "Deprecated (as part of the ENHANCED mode deprecation).\nReport task reachability for the requested versions and all task types (task reachability is not reported\nper task type).",
+            "in": "query",
+            "required": false,
+            "type": "boolean"
+          },
+          {
+            "name": "showTaskQueueConfig",
+            "description": "Show Task Queue Config",
             "in": "query",
             "required": false,
             "type": "boolean"
@@ -7149,6 +7241,37 @@
       },
       "description": "An operation completed successfully."
     },
+    "UpdateTaskQueueConfigRequestRateLimitUpdate": {
+      "type": "object",
+      "properties": {
+        "requestsPerSecond": {
+          "type": "number",
+          "format": "float"
+        },
+        "reason": {
+          "type": "string",
+          "description": "Reason for why the rate limit was set."
+        }
+      }
+    },
+    "UpdateTaskQueueConfigRequestUpdateFairnessKeyRateLimitDefault": {
+      "type": "object",
+      "properties": {
+        "rateLimit": {
+          "$ref": "#/definitions/UpdateTaskQueueConfigRequestRateLimitUpdate",
+          "description": "Default rate limit to apply for all fairness keys.\nIf null, removes the existing rate limit."
+        }
+      }
+    },
+    "UpdateTaskQueueConfigRequestUpdateQueueRateLimit": {
+      "type": "object",
+      "properties": {
+        "rateLimit": {
+          "$ref": "#/definitions/UpdateTaskQueueConfigRequestRateLimitUpdate",
+          "description": "Rate limit to apply to task queue.\nIt must be less than or equal to the system-level setting at the time\nof the request. If null, removes the existing rate limit."
+        }
+      }
+    },
     "UpdateWorkerBuildIdCompatibilityRequestAddNewCompatibleVersion": {
       "type": "object",
       "properties": {
@@ -8448,6 +8571,29 @@
         "searchAttributes": {
           "$ref": "#/definitions/v1SearchAttributes",
           "description": "Schedule search attributes to be updated.\nDo not set this field if you do not want to update the search attributes.\nA non-null empty object will set the search attributes to an empty map.\nNote: you cannot only update the search attributes with `UpdateScheduleRequest`,\nyou must also set the `schedule` field; otherwise, it will unset the schedule."
+        }
+      }
+    },
+    "WorkflowServiceUpdateTaskQueueConfigBody": {
+      "type": "object",
+      "properties": {
+        "identity": {
+          "type": "string"
+        },
+        "taskQueueName": {
+          "type": "string",
+          "description": "Selects the task queue to update."
+        },
+        "taskQueueType": {
+          "$ref": "#/definitions/v1TaskQueueType"
+        },
+        "updateQueueRateLimit": {
+          "$ref": "#/definitions/UpdateTaskQueueConfigRequestUpdateQueueRateLimit",
+          "description": "Update to queue-wide rate limit.\nIf not set, this configuration is unchanged.\nNOTE: A limit set by the worker is overriden; and restored again when reset."
+        },
+        "updateFairnessKeyRateLimitDefault": {
+          "$ref": "#/definitions/UpdateTaskQueueConfigRequestUpdateFairnessKeyRateLimitDefault",
+          "description": "Update to the default fairness key rate limit.\nIf not set, this configuration is unchanged."
         }
       }
     },
@@ -9888,6 +10034,28 @@
         }
       }
     },
+    "v1ConfigMetadata": {
+      "type": "object",
+      "properties": {
+        "reason": {
+          "type": "string",
+          "description": "Reason for why the rate limit was set."
+        },
+        "source": {
+          "type": "string",
+          "description": "Can be \"system\", \"worker\" or \"api\"."
+        },
+        "identity": {
+          "type": "string",
+          "description": "Identity of the last updater.\nSet by the request's identity field."
+        },
+        "updateTime": {
+          "type": "string",
+          "format": "date-time",
+          "description": "Time of the last update."
+        }
+      }
+    },
     "v1ContinueAsNewInitiator": {
       "type": "string",
       "enum": [
@@ -10305,6 +10473,9 @@
             "$ref": "#/definitions/v1TaskQueueVersionInfo"
           },
           "description": "Deprecated.\nOnly returned in ENHANCED mode.\nThis map contains Task Queue information for each Build ID. Empty string as key value means unversioned."
+        },
+        "configs": {
+          "$ref": "#/definitions/v1TaskQueueConfig"
         }
       }
     },
@@ -12670,6 +12841,27 @@
       },
       "description": "Range represents a set of integer values, used to match fields of a calendar\ntime in StructuredCalendarSpec. If end < start, then end is interpreted as\nequal to start. This means you can use a Range with start set to a value, and\nend and step unset (defaulting to 0) to represent a single value."
     },
+    "v1RateLimit": {
+      "type": "object",
+      "properties": {
+        "requestsPerSecond": {
+          "type": "number",
+          "format": "float",
+          "description": "Zero is a valid rate limit."
+        }
+      }
+    },
+    "v1RateLimitConfig": {
+      "type": "object",
+      "properties": {
+        "rateLimit": {
+          "$ref": "#/definitions/v1RateLimit"
+        },
+        "metadata": {
+          "$ref": "#/definitions/v1ConfigMetadata"
+        }
+      }
+    },
     "v1RecordActivityTaskHeartbeatByIdResponse": {
       "type": "object",
       "properties": {
@@ -14428,6 +14620,24 @@
       },
       "title": "See https://docs.temporal.io/docs/concepts/task-queues/"
     },
+    "v1TaskQueueConfig": {
+      "type": "object",
+      "properties": {
+        "taskQueue": {
+          "$ref": "#/definitions/v1TaskQueue"
+        },
+        "taskQueueType": {
+          "$ref": "#/definitions/v1TaskQueueType"
+        },
+        "queueRateLimit": {
+          "$ref": "#/definitions/v1RateLimitConfig",
+          "description": "Unless modified, this is the system-defined rate limit."
+        },
+        "fairnessKeysRateLimitDefault": {
+          "$ref": "#/definitions/v1RateLimitConfig"
+        }
+      }
+    },
     "v1TaskQueueKind": {
       "type": "string",
       "enum": [
@@ -14878,6 +15088,14 @@
     },
     "v1UpdateScheduleResponse": {
       "type": "object"
+    },
+    "v1UpdateTaskQueueConfigResponse": {
+      "type": "object",
+      "properties": {
+        "configs": {
+          "$ref": "#/definitions/v1TaskQueueConfig"
+        }
+      }
     },
     "v1UpdateWorkerBuildIdCompatibilityResponse": {
       "type": "object",
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -1300,6 +1300,37 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/namespaces/{namespace}/task-queues/update-config:
+    post:
+      tags:
+        - WorkflowService
+      description: Update taskqueue config and uncouple taskqueue settings from the worker lifecycle
+      operationId: UpdateTaskQueueConfig
+      parameters:
+        - name: namespace
+          in: path
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UpdateTaskQueueConfigRequest'
+        required: true
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/UpdateTaskQueueConfigResponse'
+        default:
+          description: Default error response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
   /api/v1/namespaces/{namespace}/task-queues/{taskQueue}/worker-build-id-compatibility:
     get:
       tags:
@@ -1499,6 +1530,11 @@ paths:
             Deprecated (as part of the ENHANCED mode deprecation).
              Report task reachability for the requested versions and all task types (task reachability is not reported
              per task type).
+          schema:
+            type: boolean
+        - name: showTaskQueueConfig
+          in: query
+          description: Show Task Queue Config
           schema:
             type: boolean
       responses:
@@ -4566,6 +4602,37 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /namespaces/{namespace}/task-queues/update-config:
+    post:
+      tags:
+        - WorkflowService
+      description: Update taskqueue config and uncouple taskqueue settings from the worker lifecycle
+      operationId: UpdateTaskQueueConfig
+      parameters:
+        - name: namespace
+          in: path
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UpdateTaskQueueConfigRequest'
+        required: true
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/UpdateTaskQueueConfigResponse'
+        default:
+          description: Default error response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
   /namespaces/{namespace}/task-queues/{taskQueue}/worker-build-id-compatibility:
     get:
       tags:
@@ -4765,6 +4832,11 @@ paths:
             Deprecated (as part of the ENHANCED mode deprecation).
              Report task reachability for the requested versions and all task types (task reachability is not reported
              per task type).
+          schema:
+            type: boolean
+        - name: showTaskQueueConfig
+          in: query
+          description: Show Task Queue Config
           schema:
             type: boolean
       responses:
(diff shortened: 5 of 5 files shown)


temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,39 @@ message PollerScalingDecision {
...
 }
+
+message RateLimit { 
+
+    // Zero is a valid rate limit.
+    float requests_per_second = 1;
+
+}
+
+message ConfigMetadata {
```
Sushisource: Seems like both of these should just be inlined into `RateLimitConfig`, esp. since they relate to each other.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,39 @@ message PollerScalingDecision {
...
+
+    // Reason for why the rate limit was set.
+    string reason = 1;
+    
+    // Can be "system", "worker" or "api".
+    string source = 2;
```
Sushisource: Should be an enum if it has known fixed values

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1113,6 +1116,8 @@ message DescribeTaskQueueResponse {
     // Only returned in ENHANCED mode.
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3 [deprecated = true];
+
+    temporal.api.taskqueue.v1.TaskQueueConfig configs  = 6;
```
Sushisource: Plural would imply `repeated` here, and it seems like it should be? Otherwise how are we getting different configs for different `task_queue_type`s?
```

File: temporal/api/workflowservice/v1/service.proto [hunk 4]
```diff
@@ -1194,4 +1194,17 @@ service WorkflowService {
             }
         };
     }
+
+    // Update taskqueue config and uncouple taskqueue settings from the worker lifecycle
```
Sushisource: This comment reads oddly. "...and uncouple" implies that the uncoupling happens as part of the update. Is that true?

IE: Does calling this mean that if, previously, workers had been setting a ratelimit, any future attempts by workers to do so will no longer be respected? If so, that should be spelled out in more detail here.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2412,48 @@ message ListWorkersResponse {
     // Next page token
     bytes next_page_token = 2;
 }
+
+message UpdateTaskQueueConfigRequest {
+	message RateLimitUpdate {
```
cretz: Indentation is off

temporal/api/workflowservice/v1/request_response.proto
cretz:
```suggestion
    temporal.api.taskqueue.v1.TaskQueueConfig config = 6;
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1086,6 +1086,9 @@ message DescribeTaskQueueRequest {
     // Report task reachability for the requested versions and all task types (task reachability is not reported
     // per task type).
     bool report_task_reachability = 10 [deprecated = true];
+
+    // Show Task Queue Config
+    bool show_task_queue_config = 11;
```
cretz: Can you confirm that task queue config is expensive to fetch therefore justifying it having to be opt-in?

temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,39 @@ message PollerScalingDecision {
   // all slots are already occupied.
   int32 poll_request_delta_suggestion = 1;
 }
+
+message RateLimit {
```
cretz, line 319: For `TaskQueueStats` we chose to make these associated messages as children of the common parent, should we move these under `TaskQueueConfig` similarly?

temporal/api/taskqueue/v1/message.proto
cretz, line 324: Can this be named to be clear it's for rate limit metadata and not general config metadata?

temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,39 @@ message PollerScalingDecision {
...
+    RateLimit rate_limit = 1;    
+    ConfigMetadata metadata = 2;
+}
+
+message TaskQueueConfig {
+    TaskQueue task_queue = 1;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 2;
+    // Unless modified, this is the system-defined rate limit.
+    RateLimitConfig queue_rate_limit = 3;
+    RateLimitConfig fairness_keys_rate_limit_default = 4;
```
cretz: Can some documentation be added to this field? I can't completely understand it at first glance

temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,39 @@ message PollerScalingDecision {
...
+    float requests_per_second = 1;
+
+}
+
+message ConfigMetadata {
+
+    // Reason for why the rate limit was set.
+    string reason = 1;
+    
+    // Can be "system", "worker" or "api".
```
cretz: If this is a limited enumerate, can we make it an `enum`?

temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,39 @@ message PollerScalingDecision {
...
+    // Set by the request's identity field.
+    string identity = 3;
+    
+    // Time of the last update.
+    google.protobuf.Timestamp update_time = 4;
+}
+
+message RateLimitConfig { 
+    RateLimit rate_limit = 1;    
+    ConfigMetadata metadata = 2;
```
cretz, line 338: Should we just inline the fields from these two messages?

temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,39 @@ message PollerScalingDecision {
...
+
+    // Reason for why the rate limit was set.
+    string reason = 1;
+    
+    // Can be "system", "worker" or "api".
+    string source = 2;
+    
+    // Identity of the last updater.
+    // Set by the request's identity field.
+    string identity = 3;
```
cretz: May want to rename to `update_identity` or similar to be clear config doesn't have an identity

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2412,48 @@ message ListWorkersResponse {
...
+  // NOTE: A limit set by the worker is overriden; and restored again when reset.
+  UpdateQueueRateLimit update_queue_rate_limit = 5;
+  
+  // Update to the default fairness key rate limit.
+  // If not set, this configuration is unchanged.
+  UpdateFairnessKeyRateLimitDefault update_fairness_key_rate_limit_default = 6;
+}
+
+message UpdateTaskQueueConfigResponse {
+    temporal.api.taskqueue.v1.TaskQueueConfig configs = 1;
```
cretz:
```suggestion
    temporal.api.taskqueue.v1.TaskQueueConfig config = 1;
```

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,17 @@ service WorkflowService {
...
+    rpc UpdateTaskQueueConfig (UpdateTaskQueueConfigRequest) returns (UpdateTaskQueueConfigResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/task-queues/update-config"
```
```suggestion
            post: "/namespaces/{namespace}/task-queues/{task_queue_name}/update-config"
```

I think this makes more sense (same for the additional binding)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2412,48 @@ message ListWorkersResponse {
...
+    // Default rate limit to apply for all fairness keys.
+    // If null, removes the existing rate limit.
+    RateLimitUpdate rate_limit = 1;
+  }
+
+  string namespace = 1;
+  string identity = 2;
+   
+  // Selects the task queue to update.
+  string task_queue_name = 3;
```
cretz: In `DescribeTaskQueue` we accept a full `TaskQueue` type, would there be value in doing so here? Or maybe it was a mistake to do so there?

temporal/api/taskqueue/v1/message.proto
cretz: To confirm, these two fields are never null?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2412,48 @@ message ListWorkersResponse {
...
+    // It must be less than or equal to the system-level setting at the time
+    // of the request. If null, removes the existing rate limit.
+    RateLimitUpdate rate_limit = 1;
+  }
+
+  message UpdateFairnessKeyRateLimitDefault {
+    // Default rate limit to apply for all fairness keys.
+    // If null, removes the existing rate limit.
+    RateLimitUpdate rate_limit = 1;
+  }
```
Sushisource: These also seem like they could be inlined unless we expect more things to go in here besides the rate limit

temporal/api/taskqueue/v1/message.proto
dnr: But it is for general config metadata: suppose we add another field (e.g. for partition behavior) and we also want to track a reason, update time, etc.? Let's be able to reuse the same thing.

Yeah, the comment on `reason` should be changed

temporal/api/taskqueue/v1/message.proto
dnr: I'd describe it like
```
// If set, each individual fairness key will be limited to this rate, scaled by the weight of the fairness key.
```

dnr: I think we'll have `queue_rate_limit` returned with some value no matter what, but the fairness key limit can be missing.

temporal/api/workflowservice/v1/request_response.proto
dnr: It's not expensive to fetch (it should be kept in memory always), but it could be a little large.. do we want selective returning in that case? I'm okay with always returning it to simplify things.

temporal/api/workflowservice/v1/request_response.proto
dnr: The catch is that there needs to be a way to unset the rate limit (either one or both). With this schema you can do that by providing an empty UpdateQueueRateLimit or UpdateFairnessKeyRateLimitDefault. If they're inlined you need an extra `bool unset` in RateLimitUpdate. Which is a valid way to do it if you prefer that

temporal/api/workflowservice/v1/request_response.proto
dnr: There were a lot of historical mistakes... the main one was not to put type in TaskQueue. Kind is a mostly useless value and all these config/versioning/deployment apis don't work on sticky queues so there's not really any reason to use TaskQueue there (except on polls).

There is kind of a reason to take TaskQueue on Describe though: Describe will cause the tq to be loaded if it's not loaded, and when we load it we should have kind available to switch some behavior. And you can Describe a sticky queue. But you can't configure it, so the plain string here makes sense, even if it feels inconsistent.

(Arguably we shouldn't have had a kind enum either and gotten the sticky bit from somewhere higher level)

temporal/api/workflowservice/v1/service.proto
dnr: Yes, a rate limit set by this api overrides the worker-set rate limit. If it's unset, then the worker-set limit takes effect again.

temporal/api/workflowservice/v1/request_response.proto
cretz: Same, I have no strong opinion. I wish describe were infrequent, but with stats on it I suspect it may be a bit frequent. I am less concerned with data size than performance. Up to y'all, just wanted to toss this out there.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2412,48 @@ message ListWorkersResponse {
     // Next page token
     bytes next_page_token = 2;
 }
+
+message UpdateTaskQueueConfigRequest {
```
cretz, line 2442: Do we need some kind of conflict-token approach to ensure updates are not clashing?

temporal/api/taskqueue/v1/message.proto
cretz: Being in the task queue package means it has limited use. Regardless, I actually think this information can just be inlined where needed, not sure a separate message has value.

temporal/api/workflowservice/v1/request_response.proto
dnr: Hmm, good question. My initial inclination is no/add it later: for versioning/deployment stuff, the state is more complex and you might make changes that depend (implicitly) on other parts of the state. Here, we have a bag of LWW fields. It doesn't feel necessary. But if we want to add it later we can.

temporal/api/workflowservice/v1/request_response.proto
dnr: How are we translating these opt-in bools in the sdk/cli so far? Is it annoying to add more? (for us or the user)

Size/performance: in some sense size affects performance since we have to serialize and send it. But we're not going to do an extra database read or rpc to get this data, fwiw

temporal/api/taskqueue/v1/message.proto
dnr: I like the idea of sharing it among multiple settings. Yeah, it's only ~~4~~ 3 fields so it could be repeated. I don't feel that strongly about it

temporal/api/workflowservice/v1/service.proto
Sushisource: OK, yeah, comment needs way more detail about that

temporal/api/taskqueue/v1/message.proto
sivagirish81: There’s a use case where omitting the rate_limit field signals that we want to reset to the worker defaults. By not inlining metadata within RateLimit, we retain the flexibility to provide context while intentionally unsetting the limit. This avoids ambiguity—since 0 is a valid value for requests_per_second.

sivagirish81: Updated the comment against the config reason to make the configMetaData generic for any taskQueueRelated configs.

temporal/api/taskqueue/v1/message.proto
sivagirish81: Answered previously.
https://github.com/temporalio/api/pull/604/files#r2167166804

temporal/api/workflowservice/v1/service.proto
sivagirish81: Updated the comment.

temporal/api/taskqueue/v1/message.proto
sivagirish81: We could include the RateLimitConfigs as child messages inside the TaskQueueConfig. I think the configMetaData will still need to be separate for it to be reusable. Open to suggestions.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Fixed the indentation. Tab spaces were off due to vscode.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -325,15 +325,12 @@ message RateLimit {
...
     string reason = 1;
-    
-    // Can be "system", "worker" or "api".
-    string source = 2;
-    
+
     // Identity of the last updater.
     // Set by the request's identity field.
-    string identity = 3;
+    string update_identity = 3;
```
dnr: use tags 2 and 3 now

temporal/api/taskqueue/v1/message.proto
```diff
@@ -345,9 +342,8 @@ message RateLimitConfig {
 }
 
 message TaskQueueConfig {
-    TaskQueue task_queue = 1;
-    temporal.api.enums.v1.TaskQueueType task_queue_type = 2;
     // Unless modified, this is the system-defined rate limit.
     RateLimitConfig queue_rate_limit = 3;
```
dnr: use tags 1 and 2 now

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2414,46 +2414,46 @@ message ListWorkersResponse {
...
-  // If not set, this configuration is unchanged.
-  // NOTE: A limit set by the worker is overriden; and restored again when reset.
-  UpdateQueueRateLimit update_queue_rate_limit = 5;
-  
-  // Update to the default fairness key rate limit.
-  // If not set, this configuration is unchanged.
-  UpdateFairnessKeyRateLimitDefault update_fairness_key_rate_limit_default = 6;
+    message RateLimitUpdate { 
+
+        float requests_per_second = 1;
```
dnr: since we have taskqueue.RateLimit, let's use that here so that we can unset the rate limit while supplying a reason for that change

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2414,46 +2414,46 @@ message ListWorkersResponse {
...
+
+        // Reason for why the rate limit was set.
+        string reason = 2;
+
+    }
+
+    message UpdateQueueRateLimit {
```
dnr: if we use taskqueue.RateLimit in RateLimitUpdate then we don't need these two wrapper messages, we can just use RateLimitUpdate for `update_queue_rate_limit` and `update_fairness_key_rate_limit_default`

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1196,12 +1196,14 @@ service WorkflowService {
     }
 
     // Update taskqueue config and uncouple taskqueue settings from the worker lifecycle
+    // rate limit set by this api overrides the worker-set rate limit
+    // If unset worker rate limits continue to take effect
```
dnr: Use full sentences and punctuation and everything:
```suggestion
    // Updates task queue configuration.
    // For the overall queue rate limit: the rate limit set by this api overrides the worker-set rate limit,
    // which uncouples the rate limit from the worker lifecycle.
    // If the overall queue rate limit is unset, the worker-set rate limit takes effect.
```

temporal/api/taskqueue/v1/message.proto
dnr: I'd rather keep it top-level and then reuse this RateLimit in the update request

temporal/api/taskqueue/v1/message.proto
```diff
@@ -315,3 +315,35 @@ message PollerScalingDecision {
   // all slots are already occupied.
   int32 poll_request_delta_suggestion = 1;
 }
+
+message RateLimit { 
+
```
dnr: remove these extra blank lines here and elsewhere

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1113,6 +1116,8 @@ message DescribeTaskQueueResponse {
     // Only returned in ENHANCED mode.
     // This map contains Task Queue information for each Build ID. Empty string as key value means unversioned.
     map<string, temporal.api.taskqueue.v1.TaskQueueVersionInfo> versions_info = 3 [deprecated = true];
+
+    temporal.api.taskqueue.v1.TaskQueueConfig config  = 6;
```
dnr: We agreed that this field would have just the config that was set. So now we could add an "effective rate limit" message here (with rate limit + "source") if we want. I'm okay with doing that either in this PR or in a followup.

dnr: Similar, move this so the current fields are on top and deprecated are below to make it easier to read. And add a comment about what it is and that it's only populated if show_task_queue_config is set.

temporal/api/workflowservice/v1/request_response.proto
dnr: Let's organize this with all current fields first, then deprecated fields at the bottom. I.e. move this after report_stats.

temporal/api/taskqueue/v1/message.proto
sivagirish81: Implemented.

temporal/api/taskqueue/v1/message.proto
sivagirish81: Implemented

temporal/api/taskqueue/v1/message.proto
sivagirish81: Renamed to update_identity.

temporal/api/workflowservice/v1/service.proto
sivagirish81: Implemented.

temporal/api/workflowservice/v1/service.proto
sivagirish81: Updated the comment.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Added an enum for source in the DescribeTaskQueueResponse.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Implemented

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Implemented.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Implemented.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Implemented.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: In the DescribeTaskQueue we do accept a TaskQueueType and if nothing is provided we default to the TASK_QUEUE_TYPE_WORKFLOW. So at any point of time we will only be returning a config corresponding to a TaskQueueType so it will be singular itself. I have fixed this.  We have depricated the enhanced version of the api so these config changes will not be affecting that flow.

temporal/api/workflowservice/v1/request_response.proto
cretz: > How are we translating these opt-in bools in the sdk/cli so far? Is it annoying to add more?

We haven't really for any of this newer stuff. Unfortunately these APIs are being designed without cross-SDK design, but it should be just trivial booleans there too. So Go does have an old form at https://pkg.go.dev/go.temporal.io/sdk/internal#DescribeTaskQueueEnhancedOptions but I think I heard "enhanced" is being deprecated before it ever stabilized.

> Is it annoying to add more?

Not really. Just opt-in bools that affect whether some data is in the response I think.

temporal/api/taskqueue/v1/message.proto
cretz: This link isn't going to a discussion for me, can you link the anchor from the conversation tab instead of files tab?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1058,6 +1058,9 @@ message DescribeTaskQueueRequest {
     // Report stats for the requested task queue type(s).
     bool report_stats = 8;
 
+    // Show Task Queue Config
+    bool show_task_queue_config = 11;
```
cretz: A bit of inconsistency between the bool verbs "report", "show" and "include" in this same message. Not a big deal, we can just make sure we're consistent in SDKs/CLIs.

cretz (approved):
Nothing blocking from my POV

temporal/api/enums/v1/common.proto
```diff
@@ -106,3 +106,11 @@ enum WorkerStatus {
     WORKER_STATUS_SHUTTING_DOWN = 2;
     WORKER_STATUS_SHUTDOWN = 3;
 }
+
+// Source for the update
+enum Source {
```
dnr: This is pretty specific to the task queue rate limit, I don't want to grab the very generic name `enumspb.Source`... can we put this in the taskqueue package and call it `RateLimitSource`?

temporal/api/workflowservice/v1/request_response.proto
dnr: Yeah, let's be consistent with `report` for all new flags

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2416,31 @@ message ListWorkersResponse {
...
+    string namespace = 1;
+    string identity = 2;
+    // Selects the task queue to update.
+    string task_queue_name = 3;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 4;
+    // Update to queue-wide rate limit.
+    // If not set, this configuration is unchanged.
+    // NOTE: A limit set by the worker is overriden; and restored again when reset.
+    // It must be less than or equal to the system-level setting at the time
+    // of the request. If null, removes the existing rate limit.
```
dnr: I don't think we should enforce any relation to the system setting.

And let's be clear about how to remove the existing limit:

Fields of proto messages aren't "null", pointers in some programming language can be null. Fields in proto messages can be "not present" or "not set" or "missing".

So I'd write it like
```suggestion
    // If the `rate_limit` field in the `RateLimitUpdate` is missing, remove the existing rate limit.
```
and put the same comment on update_fairness_key_rate_limit_default

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1105,6 +1105,15 @@ message DescribeTaskQueueResponse {
     // a Pinned VersioningOverride or are Child Workflows of a Pinned parent).
     temporal.api.taskqueue.v1.TaskQueueVersioningInfo versioning_info = 4;
 
+    // Statistics for the task queue. Only populated when `report_stats` is set to true in the request.
+    temporal.api.taskqueue.v1.TaskQueueStats stats = 5;
+
+    // Only populated if show_task_queue_config is set to true.
+    temporal.api.taskqueue.v1.TaskQueueConfig config  = 6;
+
+    // Source of the update
```
dnr: What is "the update"?

The source is a property of the effective rate limit, and there's no field here that has the effective rate limit. If you want to add one now, maybe we should add a submessage of DescribeTaskQueueResponse that has a taskqueue.RateLimit and a source?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,19 @@ service WorkflowService {
...
         };
     }
+
+    // Updates task queue configuration.
+    // For the overall queue rate limit: the rate limit set by this api overrides the worker-set rate limit,
+    // which uncouples the rate limit from the worker lifecycle.
+    // If the overall queue rate limit is unset, the worker-set rate limit takes effect.
+    rpc UpdateTaskQueueConfig (UpdateTaskQueueConfigRequest) returns (UpdateTaskQueueConfigResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/task-queues/{task_queue_name}/update-config"
```
dnr: Hmm, just looking at this now: how do we get the type in this url? it has just the name. Same question for DescribeTaskQueue, actually.

temporal/api/taskqueue/v1/message.proto
Shivs11: I kinda like having two fields as separate messages as it's much easier to read + `ConfigMetadata` seems to encapsulate all the fields that are more or less related

cretz: I actually think it makes it more difficult to read (you have to hop around to see basic info) and increases the API surface area unnecessarily. Lots of messages have fields in them that could be classified as "metadata", including rate limit config.

I think this is clearest:

```proto
message RateLimitConfig { 
    RateLimit rate_limit = 1;

    // Reason for why the config was set.
    string reason = 2;

    // Identity of the last updater.
    // Set by the request's identity field.
    string update_identity = 3;

    // Time of the last update.
    google.protobuf.Timestamp update_time = 4;
}
```

Also stuff like `ConfigMetadata` shows up in people's intellisense when they develop with our SDKs. (arguably this rate limit config is specific to task queue config and could be a child of that message, but meh)

Shivs11: I mean sure, I think it's kinda personal in this sense; In my head, it seemed cleaner to compartmentalize the last three fields into a separate message - I don't wanna block on this though ok with what you are suggesting

cretz: Can get other opinions, but yes would like not to group these into a top-level message if we don't have to. Same type of deal where in `Request` objects we accept things like `namespace` and `identity` but don't want a `RequestMetadata` message. Not a big deal/blocker though.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Ah, yeah, that makes sense

temporal/api/workflowservice/v1/service.proto
sivagirish81: We don't get the type in this url. In case of describeTaskQueue we pass the task queue type as a query parameter and in case if the UpdateTaskQueueConfig api we will get the TaskQueueType as part of the request body. Is there any reason why we would want it as part of the url?

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Updated the comment.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Added a new EffectiveRateLimit message and introduce a float to store the effective rate limit and the source of the update.

temporal/api/workflowservice/v1/service.proto
sivagirish81: Updated the url and the additional binding to add the {task_queue_type} as well in the url to be indicative of the exact resource that is being updated as part of the POST request.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Changed to report to be consistent.

temporal/api/enums/v1/common.proto
sivagirish81: Moved the Source enum under TaskQueue and changed the name of the enum to RateLimitSource.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1105,6 +1105,25 @@ message DescribeTaskQueueResponse {
...
+        float requests_per_second = 1;
+
+        // Source of the RateLimit Configuration,which can be one of the following values:
+        // - SOURCE_API: The rate limit that is set via the TaskQueueConfig api.
+        // - SOURCE_WORKER: The rate limit is the value set using the workerOptions in TaskQueueActivitiesPerSecond.
+        // - SOURCE_SYSTEM: The rate limit is the default value set by the system
+        temporal.api.enums.v1.RateLimitSource rate_limit_source = 2;
+    }
+
+    EffectiveRateLimit effective_rate_limit = 7;
```
cretz, line 1125: Any reason why this wouldn't be considered stats or config?

temporal/api/taskqueue/v1/message.proto
dnr: It is all subjective and can be case-by-case. Personally I'm much more in favor of message reuse than we do. I think a RequestMetadata message would be great! There's lots of benefits from that.

And here, I do see some benefits from a common ConfigMetadata. But of course the SDK API can inline everything if that's how you want to do it in SDKs. Why does ConfigMetadata show up in intellisense when working with SDKs? It's a proto message in a different package/module.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1058,6 +1058,9 @@ message DescribeTaskQueueRequest {
     // Report stats for the requested task queue type(s).
     bool report_stats = 8;
 
+    // Report Task Queue Config
+    bool report_task_queue_config = 11;
```
dnr: we can be a little more concise, "task queue" is implied here
```suggestion
    bool report_config = 11;
```

temporal/api/workflowservice/v1/request_response.proto
dnr: For historical reasons it is kind of a different thing:

I want "config" to be a more resource-oriented value: you set/unset it, you get back what you set/unset.

But because we previously had a rate limit passed from the workers, there can be a rate limit that is set, but not part of the "config". We want to report that for informative purposes and to avoid confusion, but not in the config.

I also don't think it's "stats", it has nothing to do with the actual flow of tasks, historical or present. It's just a sort of runtime piece of data that comes from one of those three places and one of them is in effect at the moment.

Even if it was sorta "stats", it shouldn't go in `temporal.api.taskqueue.v1.TaskQueueStats` because that's a per-physical-queue thing (which can be aggregated) and we are/will be using that message to report per-priority-level stats, but the rate limit applies to the whole queue, above priority levels.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,19 @@ service WorkflowService {
...
         };
     }
+
+    // Updates task queue configuration.
+    // For the overall queue rate limit: the rate limit set by this api overrides the worker-set rate limit,
+    // which uncouples the rate limit from the worker lifecycle.
+    // If the overall queue rate limit is unset, the worker-set rate limit takes effect.
+    rpc UpdateTaskQueueConfig (UpdateTaskQueueConfigRequest) returns (UpdateTaskQueueConfigResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/task-queues/{task_queue_name}/update-config/{task_queue_type}"
```
dnr: Btw @cretz did you have an opinion on task queue type in URLs? In my mind it should clearly be in the URL since it's identifying the resource being modified. But why didn't we do it for DescribeTaskQueue?

dnr (approved):
I'm good with this now, maybe just that one rename

temporal/api/workflowservice/v1/request_response.proto
cretz: It just came off a bit inconsistent because we have these other two messages that you have to opt in to viewing, are effectively-single-use messages in the `taskqueue` package, etc. We just a bit strange. Also, I assume that unlike the config which you opt in to, this we have decided we will always return no matter what because it won't be large and config will?

temporal/api/workflowservice/v1/service.proto
cretz: I think we should remove this from the URL (sorry I didn't see this before).

```suggestion
            post: "/namespaces/{namespace}/task-queues/{task_queue_name}/update-config"
```

Traditionally the URLs have just been named qualifiers and if there are other qualifiers they are via body/query, even required ones (though in describe's case it defaults to workflow IIUC). Basically the URL alone isn't all things needed to qualify what resource your affecting, it's just a name for the API call on the resource (same for run ID on workflow calls), recognizing the URL may not represent a single resource like it would if we were more RESTful. If we had a single string that uniquely identified a task queue and type, that'd be worth using instead, but it's awkward to pack tuples into URLs.

`GET https://temporal-api/namespaces/my-namespace/task-queues/my-task-queue?task_queue_type=TASK_QUEUE_TYPE_ACTIVITY` works for describe, and here you'll have to use `POST https://temporal-api/namespaces/my-namespace/task-queues/my-task-queue/update-config` with the full body including type and other things. This is why it makes sense for this to be `POST /namespaces/{namespace}/task-queues/{task_queue_name}/update-config` instead of `POST /namespaces/{namespace}/task-queues/{task_queue_name}`, because we are not truly RESTful.

temporal/api/workflowservice/v1/request_response.proto
dnr: Yeah, there's no way to do it that's not awkward somewhere: either config has something that's not a config (or at least has a very different data flow from the rest of config), or stats has something that's not a stat, or we have this unusual "effective rate limit" thing. And yeah, config may be large, stats needs to be aggregated across partitions, but effective rate limit is a single value we can just pull from memory.

temporal/api/workflowservice/v1/service.proto
dnr: It doesn't really make sense to me to have the namespace name and task queue name in one place and the type in another. You need all three to identify the task queue. But I also don't care, so I'm fine with removing the type here.

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1: We will probably be able to make it look prettier in SDKs if/when we make high level forms of this. Can you confirm these new fields are set regardless of "mode"?

temporal/api/workflowservice/v1/service.proto
cretz: Our URLs are not about uniquely identifying something because they can't do so well with our non-single-key resource design (same for workflows, activities, workers, etc). This is especially true with operations that sometimes have optional items to make up the key (such as some task queue and workflow calls). We just need a common base for operations on the general thing from an HTTP API simplicity POV.

cretz (commented):
Only slight request to remove the enumerate from the task queue URL for consistency, otherwise LGTM

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Implemented.

temporal/api/workflowservice/v1/service.proto
sivagirish81: removed the {task_queue_type} from the url as suggested.

cretz (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2438,31 @@ message ListWorkersResponse {
...
+        // Rate Limit to be updated
+        temporal.api.taskqueue.v1.RateLimit rate_limit = 1;
+        // Reason for why the rate limit was set.
+        string reason = 2;
+    }
+
+    string namespace = 1;
+    string identity = 2;
+    // Selects the task queue to update.
+    string task_queue = 3;
```
sivagirish81, line 2453: Updated the task_queue_name to task_queue as the routing magic login in temporal server needs this field to be named task_queue to get picked up automatically.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,19 @@ service WorkflowService {
...
         };
     }
+
+    // Updates task queue configuration.
+    // For the overall queue rate limit: the rate limit set by this api overrides the worker-set rate limit,
+    // which uncouples the rate limit from the worker lifecycle.
+    // If the overall queue rate limit is unset, the worker-set rate limit takes effect.
+    rpc UpdateTaskQueueConfig (UpdateTaskQueueConfigRequest) returns (UpdateTaskQueueConfigResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/task-queues/{task_queue}/update-config"
```
sivagirish81, line 1204: @cretz had to update the task_queue_name mapping to task_queue as some routing logic in the server repo relies on this field being named task_queue. Hence updating this api gateway bindings as well.

temporal/api/workflowservice/v1/request_response.proto
cretz: Great catch! I am so happy we wrote (some of?) the impl server-side before merging this so we didn't have an API release with this other field name present.

PR #605, opened 2025-06-20
Add new workflow service APIs for Nexus operations

pdoerner:
**What changed?**
New Workflow Service APIs to support non-workflow callers invoking/completing Nexus operations over gRPC.

**Server PR**
https://github.com/temporalio/temporal/pull/8105


temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,42 @@ service WorkflowService {
...
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: Nexus has a separately defined HTTP API. --)
+    rpc CancelNexusOperation (CancelNexusOperationRequest) returns (CancelNexusOperationResponse) {
+    }
+
+    // Request to complete an operation.
+    //
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: Nexus has a separately defined HTTP API. --)
+    rpc CompleteNexusOperation (CompleteNexusOperationRequest) returns (CompleteNexusOperationResponse) {
```
Quinn-With-Two-Ns: Don't we need other APIs to Fail an operation? I see we also have `CancelNexusOperation`, but the API docs say it is for "Requesting" so do we need another API to actually cancel?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,42 @@ service WorkflowService {
...
+    rpc StartNexusOperation (StartNexusOperationRequest) returns (StartNexusOperationResponse) {
+    }
+
+    // Request to cancel an operation. The operation may later complete as canceled or any other outcome. Handlers
+    // should ignore multiple cancelations of the same operation and return successfully if cancelation was already
+    // requested.
+    //
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: Nexus has a separately defined HTTP API. --)
+    rpc CancelNexusOperation (CancelNexusOperationRequest) returns (CancelNexusOperationResponse) {
```
Quinn-With-Two-Ns: nit: Maybe we should call this `RequestCancelNexusOperation` since we also do need an API to cancel the operation

temporal/api/workflowservice/v1/service.proto
Quinn-With-Two-Ns: Or complete needs to accept a variant of different ways to complete the operation

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+    map<string, string> header = 4;
+
+    // Name of service to start the operation in.
+    string service = 5;
+    // Operation name.
+    string operation = 6;
+    // A request ID that can be used as an idempotentency key.
+    string request_id = 7;
+    // Callback URL to call upon completion if the started operation is async.
+    string callback = 8;
```
Quinn-With-Two-Ns, line 2575: How does callback interact with the `CompleteNexusOperation` RPC since there is now sort of two ways to complete an operation?

temporal/api/nexus/v1/message.proto
```diff
@@ -121,12 +161,54 @@ message StartOperationResponse {
...
+    oneof variant {
+        // The operation completed successfully.
+        Complete complete_success = 1;
+        // The operation completed unsuccessfully (failed or canceled).
+        Error operation_error = 2;
+        // The operation is still running, When waiting for completion, the caller may re-issue this request to start a
+        // new long poll.
+        StillRunning still_running = 3;
+        // The server gave up waiting for operation completion. The request may be retried by the caller.
+        Timeout timeout = 4;
```
Quinn-With-Two-Ns: What is the value of this over a generic `gRPC` deadline exceeded?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+
+    // Request headers to forward to the Handler.
+    map<string, string> header = 4;
+
+    // Service name.
+    string service = 5;
+    // Operation name.
+    string operation = 6;
+    // Operation token as originally generated by a Handler via an async StartOperation response.
+    string operation_token = 7;
```
Quinn-With-Two-Ns: What happens if multiple operations have the same `operation_token`? Do they all get completed?

temporal/api/workflowservice/v1/service.proto
pdoerner: Complete accepts an OperationError variant. To complete as canceled, you would just call complete with an OperationError indicating the cancellation.

temporal/api/workflowservice/v1/service.proto
pdoerner: Yeah good point, `RequestCancelNexusOperation` is a better name since the handler can still choose to ignore the cancel request.

temporal/api/nexus/v1/message.proto
pdoerner: It is more clear that the server gave up waiting for the handler to respond. Deadline exceeded might indicate a failure to contact the server at all.

temporal/api/workflowservice/v1/request_response.proto
pdoerner: `operation_token` should be unique for each operation

temporal/api/workflowservice/v1/request_response.proto
pdoerner: Good point, perhaps the callback should be more generic so that we can put either HTTP or gRPC information there, not just a URL. The other option is to say this is optional and the handler may choose to call the CompleteOperation gRPC API using a Temporal client instead.

pdoerner:
> (waiting on PR to be marked ready-for-review before reviewing...)

Only marked as draft since the Server PR is not quite ready, but I don't think that should stop discussion about the API signature.

temporal/api/nexus/v1/message.proto
```diff
@@ -42,6 +43,13 @@ message Link {
     string type = 2;
 }
 
+message OperationInfo {
+    // See https://github.com/nexus-rpc/api/blob/main/SPEC.md#operationinfo.
+    // Operation token as originally generated by a Handler.
+    string token = 1;
+    string state = 2;
```
cretz, line 50: I think this should be an enum too (and is a clear representation of what the API handles, not to mention what a good type-safe API should have)

temporal/api/nexus/v1/message.proto
```diff
@@ -42,6 +43,13 @@ message Link {
     string type = 2;
 }
 
+message OperationInfo {
+    // See https://github.com/nexus-rpc/api/blob/main/SPEC.md#operationinfo.
+    // Operation token as originally generated by a Handler.
+    string token = 1;
```
cretz, line 49: If we're willing to include the token, why not the service and operation names too? And will those field names match their request counterparts?

temporal/api/nexus/v1/message.proto
```diff
@@ -121,12 +161,54 @@ message StartOperationResponse {
 message CancelOperationResponse {
 }
 
+// Response variant for GetOperationInfoRequest.
+message GetOperationInfoResponse {
+    OperationInfo info = 1;
+    repeated Link links = 2;
```
cretz: Does get operation result have links? It's not showing in spec.

temporal/api/nexus/v1/message.proto
```diff
@@ -121,12 +161,54 @@ message StartOperationResponse {
...
+}
+
+// Response variant for GetOperationResultRequest.
+message GetOperationResultResponse {
+    // The operation completed successfully.
+    message Complete {
+        temporal.api.common.v1.Payload payload = 1;
+        repeated Link links = 2;
```
cretz: Not seeing links for operation result in spec either

temporal/api/nexus/v1/message.proto
cretz: Isn't there a difference between Nexus handler impl returning after its `wait` and server failing after gRPC timeout? When given a gRPC timeout lower than the `wait` given in the get result call, what is the value of `wait` is sent to the Nexus handler? Is it some kind of minimum of both? Is there a leeway/buffer period between gRPC deadline and wait that we want to accept to make sure the server isn't returning a still-running without receiving that literally from the handler?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,42 @@ service WorkflowService {
...
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: Nexus has a separately defined HTTP API. --)
+    rpc GetNexusOperationInfo (GetNexusOperationInfoRequest) returns (GetNexusOperationInfoResponse) {
+    }
+
+    // Retrieve operation details.
+    //
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: Nexus has a separately defined HTTP API. --)
+    rpc GetNexusOperationResult (GetNexusOperationResultRequest) returns (GetNexusOperationResultResponse) {
```
cretz, line 1277: Can you confirm these two calls go to the handler implementations of these and that the server does nothing with them and does not try to intercept them? For instance, if I had a custom "get result" handler implementation, the server is not going to try to short-circuit it with its own known result right?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+
+}
+message GetNexusOperationInfoResponse {
+    temporal.api.nexus.v1.OperationInfo info = 1;
+    repeated temporal.api.nexus.v1.Link links = 2;
```
cretz: Not seeing info response links in spec

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
     // Next page token
     bytes next_page_token = 2;
 }
+
+message StartNexusOperationRequest {
```
cretz: Can you accept an `identity` on these calls even if it's not passed down to the handler (and we do nothing with it yet)? I think we need to generally accept `identity` on all side-effecting Temporal RPC calls. Note, if get info and get result calls are technically durable which I suspect they might be, they may need identity too.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+    // The caller namespace. The endpoint below may point to a different namespace.
+    string namespace = 1;
+    // A registered endpoint. Mutually exclusive with task queue.
+    string endpoint = 2;
+    // Dispatch directly to a task queue on this namespace.
+    string task_queue = 3;
```
cretz: Should these be in a oneof?

temporal/api/workflowservice/v1/request_response.proto
cretz: I think callback should be a whole object with callback headers in there too. I guess we didn't with the Nexus form, we flattened it, so we can keep flatted here too.

> The other option is to say this is optional

Is this not optional anyways? I don't have to have a callback right I can just make get-result calls?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+    // Service name.
+    string service = 5;
+    // Operation name.
+    string operation = 6;
+    // Operation token as originally generated by a Handler via an async StartOperation response.
+    string operation_token = 7;
+
+    // If set and non-zero, reflects the duration the caller has indicated that it wants to wait for operation
+    // completion, turning the request into a long poll.
+    google.protobuf.Duration wait = 8;
```
cretz: I think we should consider just making this gRPC deadline (or just before) for the Temporal side of this call and maybe just exposing `bool wait`. That way there's no ambiguity on which is in use.

temporal/api/nexus/v1/message.proto
cretz: I am a bit confused in spec between the difference between 408 and 412. How is a timeout for this call any different than a timeout for any other call?

temporal/api/workflowservice/v1/request_response.proto
cretz: The question is reasonable though, what does happen if it's reused for this complete call?

temporal/api/workflowservice/v1/service.proto
cretz: > Complete accepts an OperationError variant

I didn't see it, `CompleteNexusOperationRequest` only accepts a `payload`

temporal/api/workflowservice/v1/request_response.proto
Quinn-With-Two-Ns: Right now `operation_token` is not unique for each operation, even built in workflow run operation can easily generate multiple operations with the same `operation_token`

temporal/api/workflowservice/v1/service.proto
pdoerner: Right, sorry I meant that payload should be a serialized OperationError with a canceled reason in order to complete the operation as canceled.

temporal/api/workflowservice/v1/request_response.proto
pdoerner: Yes, it is always optional. The handler does not necessarily have to invoke the callback.

temporal/api/nexus/v1/message.proto
pdoerner: If the handler returned after its wait, we should not be returning timeout, we should be returning `still_running` or a result. Yes, we will use the minimum of gRPC timeout and `wait`. If we do not receive a response from the handler before the `wait` / gRPC deadline, we would return `timeout` not `still_running`.

temporal/api/workflowservice/v1/request_response.proto
pdoerner: In the temporal implementation, we use a `completion_token` which contains a reference to a specific state machine. The `operation_token` is only used to fabricate the started event if the completion is received before the response to the StartOperation call.

temporal/api/nexus/v1/message.proto
pdoerner: Links are a general purpose header that may be sent with any request/response: https://github.com/nexus-rpc/api/blob/main/SPEC.md#nexus-link

temporal/api/nexus/v1/message.proto
pdoerner: Links are a general purpose header that may be sent with any request/response: https://github.com/nexus-rpc/api/blob/main/SPEC.md#nexus-link

temporal/api/workflowservice/v1/service.proto
pdoerner: That's right. The server should not store results.

temporal/api/workflowservice/v1/request_response.proto
pdoerner: Links are a general purpose header that may be sent with any request/response: https://github.com/nexus-rpc/api/blob/main/SPEC.md#nexus-link

temporal/api/workflowservice/v1/request_response.proto
pdoerner: What is identity used for? If it will not be passed down to the handler maybe we can just accept it as a metadata header and not as an explicit request param?

temporal/api/workflowservice/v1/request_response.proto
pdoerner: I like that idea. It doesn't quite match the spec, but it does match the behavior of other Temporal APIs (e.g. GetWorkflowExecutionHistory)

temporal/api/nexus/v1/message.proto
pdoerner: 408 means the server didn't receive a response from the handler. 412 means that the handler explicitly returned `still_running`.

temporal/api/nexus/v1/message.proto
pdoerner: Service and operation names are part of the Temporal implementation of Nexus, not a part of the general spec. We could still return them, but I think they should be separate fields in the response object, not part of the `OperationInfo` struct.

temporal/api/workflowservice/v1/request_response.proto
pdoerner: But it is a good point that we should support passing callback headers too. I will add a separate field for that.

temporal/api/workflowservice/v1/request_response.proto
Quinn-With-Two-Ns: Right, but this API takes a `operation_token ` so how will you know which state machine the user means?

temporal/api/workflowservice/v1/service.proto
cretz: Not following this exactly. How is a payload serialized with operation error? What does that payload look like from an encoding and data POV? I would expect, like workflow tasks and activity tasks, that we provide a way to explicitly set all of the things that you can set with `GetOperationResultResponse`.

temporal/api/nexus/v1/message.proto
cretz: I see. I guess I am a bit confused on why only this can timeout and not other Nexus calls. But I guess that is an issue with the spec and not this PR.

temporal/api/nexus/v1/message.proto
cretz: I see, makes sense, no need to change anything here

temporal/api/nexus/v1/message.proto
cretz: > may be sent with any request/response

If that were truly the case, they could be on the request/response object. But they are absent on some requests and responses here. Do we need to add them to every request/response? Also if they are general purpose, not understanding why the spec specifically calls them out on the start operation and not elsewhere?

Can you get confirmation of whether links are supported on _every_ request/response or only some?

temporal/api/nexus/v1/message.proto
cretz: > 408 means the server didn't receive a response from the handler

Doesn't 408 represent the handler returning a timeout? So if handler returns a timeout for this call, does server relay that through? Am a bit confused from a spec POV why request timeout response is represented one way and wait timeout response is another (not to be confused with "still running").

temporal/api/workflowservice/v1/request_response.proto
cretz: It is for our own server-side purposes that could be used by logs or other things in the future, even if we don't use it today. In general I think all mutable RPC calls should accept an explicit `identity` parameter (even though I'm sure there are some we forgot or didn't do). We should go ahead and accept it so when we decide to use it, older SDKs will already be sending it.

In general though, how can I as a user using Temporal UI or Temporal API see my outstanding Nexus operations? Do we have plans to show users Nexus operations they have started (even if they aren't the namespaces running them)?

temporal/api/workflowservice/v1/request_response.proto
pdoerner: There is a state machine reference within the `completion_token`. We will ignore the `operation_token` unless we receive the completion before the start response in which case it is used to fabricate the started event, not for any lookup or validation. There should not be any ambiguity in the state machine reference from the completion_token

temporal/api/nexus/v1/message.proto
pdoerner: They could be on the request/response object, that's true. I only added them where I thought it made sense because they would provide value there. Are there any request/responses that do not have links that you would expect to? I omitted it from some places because it doesn't make sense to return a link for a timeout, for example.

The spec specifically calls them out for start operation I think just for discoverability. We could certainly adjust the spec to make it more clear if you think it warrants it. As Temporal is currently the only implementation of Nexus, links are supported on any request/response where we decide to support them. And in general, this would be a decision of any implementer of Nexus how to handle links.

temporal/api/workflowservice/v1/request_response.proto
pdoerner: At the moment we have UI for operations that are started from a workflow because those can be looked up from the workflow's state. AFAIK there are not currently plans to support a ListOperations type API for operations that were started by external callers. You would have to use GetOperationInfo for each operation that is started.

temporal/api/workflowservice/v1/service.proto
pdoerner: So the user sends a payload which represents the result of the operation. That payload could contain a result value or an OperationError. In order to complete the operation as canceled, users should send an OperationError with a canceled cause. I guess I'm not opposed to providing a more explicit structure than an opaque payload, this is just more similar to how we handle completions in the HTTP API.

temporal/api/nexus/v1/message.proto
bergundy: Service and operation names are not part of the Temporal implementation, they are generic Nexus concepts.
TBH I don't think we should have put the token in `OperationInfo` since there's no need to echo this back to the caller of `GetInfo` (BTW @cretz you're the one who suggested this a while back).

temporal/api/nexus/v1/message.proto
bergundy: We made this a string with a fixed set of options in Nexus over HTTP, let's be consistent. If we ever decide to add more states we will stay compatible between the two transports.

temporal/api/nexus/v1/message.proto
bergundy: I don't think we need links here for now, they're not part of the GetInfo response in Nexus. Only StartOperation has links for the time being.

temporal/api/workflowservice/v1/request_response.proto
bergundy: It can be but I would be okay with putting this directly in the message and switching in the code.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+    string callback = 8;
+    // Full request body from the incoming HTTP request.
+    temporal.api.common.v1.Payload payload = 9;
+    // Header that is expected to be attached to the callback request when the operation completes.
+    map<string, string> callback_header = 10;
+    // Links contain caller information and can be attached to the operations started by the handler.
+    repeated temporal.api.nexus.v1.Link links = 11;
+}
+
+message StartNexusOperationResponse {
```
bergundy, line 2584: We need to decide how to represent `HandlerError`s too here. I think the best option is to embed them in the response variants. Just note that `HandlerError`s are more like grpc status errors and we may need to figure out the observability for these errors if we return them in an "ok" response.

bergundy: An alternative would be to create a service error that's a generic `HandlerError` that can be associated with any grpc status code and always map grpc statuses to nexus handler error types and back on the server + SDK.

Not 100% clear to me which option is better ATM.

bergundy: This comment applies to all of the responses.

temporal/api/workflowservice/v1/request_response.proto
bergundy: No need for links here, we only support them on `StartOperation` ATM.

temporal/api/workflowservice/v1/request_response.proto
bergundy: @cretz when we designed the Go client we made `wait` and context deadline separate at your request. The idea was to allow the client to differentiate between an unresponsive server and an operation that hasn't completed yet.
We can still revert this decision IMHO since the Go client's GetResult method is considered experimental.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+    // If set and non-zero, reflects the duration the caller has indicated that it wants to wait for operation
+    // completion, turning the request into a long poll.
+    google.protobuf.Duration wait = 8;
+}
+
+message GetNexusOperationResultResponse {
+    // The operation completed successfully.
+    message Complete {
+        temporal.api.common.v1.Payload payload = 1;
+        repeated temporal.api.nexus.v1.Link links = 2;
```
bergundy: Remove links here and in the other variants, we don't support them yet (although we may eventually).

temporal/api/workflowservice/v1/request_response.proto
bergundy: You may want to add a comment as we've done in the SPEC and SDK to mention that this is only used in the complete-before-start race condition.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+    // Dispatch directly to a task queue on this namespace.
+    string task_queue = 3;
+
+    // Request headers to forward to the Handler.
+    map<string, string> header = 4;
+
+    // Service name.
+    string service = 5;
+    // Operation name.
+    string operation = 6;
```
bergundy: This isn't part of the callback spec.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+
+message CompleteNexusOperationRequest {
+    // The caller namespace. The endpoint below may point to a different namespace.
+    string namespace = 1;
+    // A registered endpoint. Mutually exclusive with task queue.
+    string endpoint = 2;
+    // Dispatch directly to a task queue on this namespace.
+    string task_queue = 3;
+
+    // Request headers to forward to the Handler.
```
bergundy: Unclear what the "Handler" is here (it's the original caller).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+        // The operation is still running, When waiting for completion, the caller may re-issue this request to start a
+        // new long poll.
+        StillRunning still_running = 3;
+        // The server gave up waiting for operation completion. The request may be retried by the caller.
+        Timeout timeout = 4;
+    }
+}
+
+message CompleteNexusOperationRequest {
+    // The caller namespace. The endpoint below may point to a different namespace.
```
bergundy: This is the "handler" namespace, not the caller one. The handler would not have permissions to connect directly to the caller namespace.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+    string namespace = 1;
+    // A registered endpoint. Mutually exclusive with task queue.
+    string endpoint = 2;
+    // Dispatch directly to a task queue on this namespace.
+    string task_queue = 3;
```
bergundy: This isn't dispatched by endpoint or task queue, it's dispatched by a callback URL. Essentially you're asking the server to invoke the callback on your behalf.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2407,3 +2407,182 @@ message ListWorkersResponse {
...
+
+    // Service name.
+    string service = 5;
+    // Operation name.
+    string operation = 6;
+    // Operation token as originally generated by a Handler via an async StartOperation response.
+    string operation_token = 7;
+
+    // Operation result.
+    temporal.api.common.v1.Payload payload = 8;
```
bergundy: This should take the `state` of the operation (`successful`, `failed`, `canceled`) and an optional error. Can potentially make it a oneof since the result and the error would be mutually exclusive.

temporal/api/nexus/v1/message.proto
bergundy: We don't support links on anything other than `StartOperation` yet. We intend to but let's add that capability separately.

temporal/api/nexus/v1/message.proto
bergundy: > Is there a leeway/buffer period between gRPC deadline and wait that we want to accept to make sure the server isn't returning a still-running without receiving that literally from the handler?

That's the idea (your proposal BTW @cretz)

temporal/api/nexus/v1/message.proto
cretz: Is @pdoerner correct that that section of the spec implies these are available on every call? If so, can we update the spec if they are not? My reading of the spec is that they are not available on every call, but I can see the confusion/ambiguity.

temporal/api/nexus/v1/message.proto
cretz: > The spec specifically calls them out for start operation I think just for discoverability

and

> We don't support links on anything other than StartOperation yet

Are in conflict. Can we update the spec to unambiguously clarify what links are accepted on?

temporal/api/workflowservice/v1/request_response.proto
cretz: Can we update the spec to remove this ambiguity so others don't have a similar reading?

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1: Regardless, I think we should accept an identity for all calls like these, even if we don't use it (yet)

temporal/api/workflowservice/v1/request_response.proto
cretz: > we designed the Go client we made wait and context deadline separate at your request

Because the context deadline is not sent over HTTP like it is gRPC. However, Temporal APIs have used the gRPC deadline to define how long they are willing to wait for things. I guess since we're going to have to have a common client interface for HTTP and gRPC, we are going to have to ignore the gRPC/Temporal approach here. Can we document on this field how Temporal server will handle the situation where deadline is less than wait and deadline is reached? Same deal for if the wait is beyond the max long poll of the server.

temporal/api/nexus/v1/message.proto
pdoerner: It is true that the spec implies that it is a general purpose header. Currently we only implemented support for it on Start and Callback requests. In the future we think it is likely that we will want links on GetResult and GetInfo and maybe on Cancel.

Since we do not have links support for those yet, I will remove them from this PR and we can add them once they are fully supported.

At the moment I don't think we need to update the spec, implementers of Nexus could still make their own choices about when and where to support links.

temporal/api/nexus/v1/message.proto
pdoerner: So you're saying you would rather not have this timeout variant? I think it is nice to be explicit that the request made it to the server and the server did not hear back from the handler before the wait deadline. A simple request timeout might indicate a failure to connect to the server at all.

This timeout is different from other calls because this is the only API that supports `wait`.

temporal/api/nexus/v1/message.proto
cretz: > implementers of Nexus could still make their own choices about when and where to support links.

I think we don't want this ambiguous in the spec (but we can clearly call out that it is optional in spec if needed). My interpretation was that the spec didn't imply it is available on all (it's just documented in a general place), so I think we should clarify if it is available on all or not in the spec.

bergundy: It is a general purpose header, that is it should be available on more than one method. We currently only support it on start though. I think we want to document links once and mention which methods they're available in. Can definitely add that clarification.

bergundy: In the meantime, lets remove links until we decide to support them in `GetInfo`

temporal/api/nexus/v1/message.proto
bergundy: We should.

bergundy: I think we will have links in `GetResult` eventually, it's just not supported quite yet, please remove them from here.

temporal/api/nexus/v1/message.proto
bergundy: > Am a bit confused from a spec POV why request timeout response is represented one way and wait timeout response is another

I think we can use `408` for any request timeout errors, I can fix that for Go. It hasn't really come up because in the Go SDK the client sets a context deadline and will not get anything back from the server. In the Temporal implementation we give ourselves some buffer to avoid timing out requests and respond with a 520 UPSTREAM_ERROR handler error. This helps us distinguish between timeouts that we think are our fault vs. the user's fault.

So I tend to agree that we probably don't need a special timeout response here.

bergundy: Worth also updating the Go SDK and the spec to return 408 for any request timeout (e.g. context deadline errors).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,203 @@ message ListWorkersResponse {
...
+    // Callback URL to call upon completion if the started operation is async.
+    string callback = 8;
+    // Full request body from the incoming HTTP request.
+    temporal.api.common.v1.Payload payload = 9;
+    // Header that is expected to be attached to the callback request when the operation completes.
+    map<string, string> callback_header = 10;
+    // Links contain caller information and can be attached to the operations started by the handler.
+    repeated temporal.api.nexus.v1.Link links = 11;
+
+    // The identity of the client who initiated this request.
```
bergundy: nit: I'd move this generic metadata up so it doesn't get sandwiched when we add more fields later.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,203 @@ message ListWorkersResponse {
...
+    message Sync {
+        temporal.api.common.v1.Payload payload = 1;
+        repeated temporal.api.nexus.v1.Link links = 2;
+    }
+
+    // The operation will complete asynchronously.
+    // The returned token can be used to reference this operation.
+    message Async {
+        string operation_token = 1;
+        repeated temporal.api.nexus.v1.Link links = 2;
```
bergundy: Put the link at the top level here since we said that we want it to be available with `operation_error` too.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,203 @@ message ListWorkersResponse {
...
+    }
+
+    oneof variant {
+        Sync sync_success = 1;
+        Async async_success = 2;
+        // The operation completed unsuccessfully (failed or canceled).
+        temporal.api.nexus.v1.UnsuccessfulOperationError operation_error = 3;
+        temporal.api.nexus.v1.HandlerError handler_error = 4;
```
bergundy, line 2607: On the fence whether we should return this as part of a successful response. I'm leaning towards translating this to a grpc status error and serializing it into the grpc response headers. This would map better semantically.

bergundy (commented):
I had to stop mid-review. I'll get back to it, wanted to post what I have so far.

temporal/api/nexus/v1/message.proto
cretz: Sorry, I may have confused myself...

So I understand now how "server didn't get response from client in time" timeout error works, and why `temporal.api.workflowservice.v1.GetNexusOperationResultResponse.timeout` exists on the Temporal gRPC. I am less understanding why it appears here in the Nexus protos. Can you clarify for me the situation where `temporal.api.nexus.v1.GetOperationResultResponse.timeout` is set by any code? Are handlers expecting to set this value when wait expires on _their_ side? And so then there's two ways `temporal.api.workflowservice.v1.GetNexusOperationResultResponse.timeout` can be set, 1) handler returns it, 2) server actually times out?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,203 @@ message ListWorkersResponse {
...
+}
+
+message RequestCancelNexusOperationRequest {
+    // The caller namespace. The endpoint below may point to a different namespace.
+    string namespace = 1;
+    // A registered endpoint. Mutually exclusive with task queue.
+    string endpoint = 2;
+    // Dispatch directly to a task queue on this namespace.
+    string task_queue = 3;
+
```
bergundy: If we're using a oneof in StartNexusOperationRequest, we should be consistent in the other request types.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,203 @@ message ListWorkersResponse {
...
+    map<string, string> header = 4;
+
+    // Service name.
+    string service = 5;
+    // Operation name.
+    string operation = 6;
+    // Operation token as originally generated by a Handler via an async StartOperation response.
+    string operation_token = 7;
+
+    // The identity of the client who initiated this request.
```
bergundy: Same as my comment above, move this up to avoid getting it sandwiched between more specific fields later.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,203 @@ message ListWorkersResponse {
...
+    // The operation completed successfully.
+    message Complete {
+        temporal.api.common.v1.Payload payload = 1;
+    }
+
+    message StillRunning {
+    }
+
+    message Timeout {
+    }
```
bergundy: Get rid of this variant per the comments above.

bergundy: A request timeout in the gRPC world would be context deadline exceeded.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,203 @@ message ListWorkersResponse {
...
+
+    // Operation token as originally generated by a Handler via an async StartOperation response. Used only when
+    // the operation completion is received before the response to the StartNexusOperation call.
+    string operation_token = 5;
+
+    message Success {
+        // Operation result.
+        temporal.api.common.v1.Payload payload = 1;
+        // Links containing handler information which can be attached to operations scheduled by the caller.
+        repeated temporal.api.nexus.v1.Link links = 2;
```
bergundy: I would put links on `CompleteNexusOperationRequest` since they may be relevant in other variants (such as operation_error).

bergundy (commented):
Just inserting the obvious note that this PR should not be merged until we've implemented the server and one client and have confidence that this API is stable enough to merge.

temporal/api/nexus/v1/message.proto
cretz: Not following this here. Just because HTTP doesn't have enums doesn't mean gRPC doesn't. If we update the spec to say this value can technically be any string but these are the suggested, that could make more sense.

temporal/api/nexus/v1/message.proto
cretz: @bergundy - can we clarify in spec whether links can happen on these calls or not?

temporal/api/nexus/v1/message.proto
cretz: Can I get clarification on the situation described above

cretz (approved):
Nothing blocking, but would like us to 1) be clear about which calls can have links, and 2) confirm there is no ambiguity between user-provided get result timeout and server-set get result timeout based on deadline.

temporal/api/nexus/v1/message.proto
bergundy: We are considering adding another state to support queuing requests when a handler is unavailable. I know gRPC can support this but it is more difficult to support it with enums IME.

temporal/api/nexus/v1/message.proto
bergundy: Not today. Let's not add links in the API until we plan to support this in the SDKs.

temporal/api/nexus/v1/message.proto
bergundy: Not removed yet. Please remove from all of the variants.

temporal/api/nexus/v1/message.proto
bergundy: I don't think we need this variant.

We can use context deadline errors and handler errors with upstream timeout for when a request times out and the `StillRunning` variant for when the timeout expired before the request deadline.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,42 @@ service WorkflowService {
...
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: Nexus has a separately defined HTTP API. --)
+    rpc RequestCancelNexusOperation (RequestCancelNexusOperationRequest) returns (RequestCancelNexusOperationResponse) {
+    }
+
+    // Request to complete an operation.
+    //
+    // (-- api-linter: core::0127::http-annotation=disabled
+    //     aip.dev/not-precedent: Nexus has a separately defined HTTP API. --)
+    rpc CompleteNexusOperation (CompleteNexusOperationRequest) returns (CompleteNexusOperationResponse) {
```
bergundy: I'm debating whether we should add this to WorkflowService or not. I agree we 100% want this API but we may want to surface it as a system nexus service to be accessible from workflows.

bergundy (commented):
Please address all of the comments. There are still pending comments from a couple of weeks ago.

temporal/api/workflowservice/v1/service.proto
pdoerner: I thought workflow service APIs are available from workflows?

temporal/api/workflowservice/v1/request_response.proto
bergundy: We settled this offline. Keeping HandlerError in the responses but avoiding having the server return handler errors.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,191 @@ message ListWorkersResponse {
...
+        temporal.api.common.v1.Payload sync_success = 1;
+        // The operation will complete asynchronously.
+        // The returned string is an operation token which can be used to reference this operation.
+        string async_success = 2;
+        // The operation completed unsuccessfully (failed or canceled).
+        temporal.api.nexus.v1.UnsuccessfulOperationError operation_error = 3;
+        temporal.api.nexus.v1.HandlerError handler_error = 4;
+    }
+
+    repeated temporal.api.nexus.v1.Link links = 5;
```
bergundy, line 2610: nit: put this before the variant since it's considered metadata.

temporal/api/workflowservice/v1/service.proto
bergundy: They are not.

bergundy (approved):
I would like to think about the completion API a bit more but overall LGTM.

temporal/api/nexus/v1/message.proto
```diff
@@ -121,12 +161,37 @@ message StartOperationResponse {
...
+}
+
+// Response variant for GetOperationResultRequest.
+message GetOperationResultResponse {
+    message StillRunning {
+    }
+
+    oneof variant {
+        // The operation completed successfully.
+        temporal.api.common.v1.Payload complete_success = 1;
```
dandavison: I'm finding the name "complete_success" for a payload a bit surprising/awkward.

How about following Workflow Update and naming it `success`?

```protobuf
// The outcome of a Workflow Update: success or failure.
message Outcome {
    oneof value {
        temporal.api.common.v1.Payloads success = 1;
        temporal.api.failure.v1.Failure failure = 2;
    }
}
```

temporal/api/nexus/v1/message.proto
```diff
@@ -121,12 +161,37 @@ message StartOperationResponse {
...
+        // The operation completed unsuccessfully (failed or canceled).
+        UnsuccessfulOperationError operation_error = 2;
+        // The operation is still running, When waiting for completion, the caller may re-issue this request to start a
+        // new long poll.
+        StillRunning still_running = 3;
```
dandavison, line 189: `StillRunning` implies that we know that it is still running. Are we sure that we won't want to use this in situations where we want to suggest to the client to resubmit their long-poll while remaining agnostic about what state the operation is in? If we do, we could call it something vaguer such as `unresolved`.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,191 @@ message ListWorkersResponse {
...
+    // Request headers to forward to the Handler.
+    map<string, string> header = 5;
+
+    // Name of service to start the operation in.
+    string service = 6;
+    // Operation name.
+    string operation = 7;
+    // A request ID that can be used as an idempotentency key.
+    string request_id = 8;
+    // Callback URL to call upon completion if the started operation is async.
```
dandavison, line 2574: nit: since nexus operations in general have the ability to decide at handling time whether to respond synchronously or asynchronously.

```suggestion
    // Callback URL to call upon completion if the operation responds asynchronously.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2419,191 @@ message ListWorkersResponse {
...
+    repeated temporal.api.nexus.v1.Link links = 12;
+}
+
+message StartNexusOperationResponse {
+    oneof variant {
+        // An operation completed successfully.
+        temporal.api.common.v1.Payload sync_success = 1;
+        // The operation will complete asynchronously.
+        // The returned string is an operation token which can be used to reference this operation.
+        string async_success = 2;
```
dandavison: Shouldn't this be

```protobuf
        OperationInfo async_success = 2
```


https://github.com/nexus-rpc/api/blob/main/SPEC.md#response-codes

temporal/api/nexus/v1/message.proto
dandavison: Also, the Go SDK is proposing to use an API like

```go
if result.Pending != nil {
```

but "Pending" and "StillRunning" would be two different terms for precisely the same underlying concept; so let's consider sharing the terminology between the proto and the SDK.

temporal/api/workflowservice/v1/request_response.proto
dandavison: I'm thinking that, seeing as the spec says `OperationInfo`, we should return that, because we may add fields / new operation state values to `OperationInfo` in the future.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2550,4 +2550,171 @@ message DescribeWorkerRequest {
...
+    }
+
+    oneof variant {
+        // The handler has indicated the operation was completed synchronously.
+        Sync sync_success = 1;
+        // The handler has indicated that the operation was started and will complete asynchronously.
+        Async async_success = 2;
+        // The operation completed unsuccessfully (failed or canceled).
+        Unsuccessful unsuccessful = 3;
+        // There was an error connecting to the handler.
```
bergundy:
```suggestion
        // The user handler has responded with an error or failed to respond in time.
```

bergundy (approved):
Looks great, added a small comment, nothing blocking.

PR #606, opened 2025-06-25
bug: add timestamp to TriggerImmediatelyRequest

spkane31:

**What changed?**
adding a timestamp to `TriggerImmediatelyRequest` message

We can make the wft “more idempotent”: put the desired trigger time in the trigger immediately request itself, so it goes in the signal payload, so all wfts attempt to start the target with the same id.

**Breaking changes**

[**Server PR**](https://github.com/temporalio/temporal/pull/7968)

### DIFF
--- a/temporal/api/schedule/v1/message.proto
+++ b/temporal/api/schedule/v1/message.proto
@@ -278,6 +278,9 @@ message ScheduleState {
 message TriggerImmediatelyRequest {
     // If set, override overlap policy for this one request.
     temporal.api.enums.v1.ScheduleOverlapPolicy overlap_policy = 1;
+    
+    // Timestamp for when this request should be triggered.
+    google.protobuf.Timestamp trigger_time = 2;
 }
 
 message BackfillRequest {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -14748,6 +14748,11 @@
         "overlapPolicy": {
           "$ref": "#/definitions/v1ScheduleOverlapPolicy",
           "description": "If set, override overlap policy for this one request."
+        },
+        "triggerTime": {
+          "type": "string",
+          "format": "date-time",
+          "description": "Timestamp for when this request should be triggered."
         }
       }
     },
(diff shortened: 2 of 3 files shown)


temporal/api/schedule/v1/message.proto
```diff
@@ -278,6 +278,9 @@ message ScheduleState {
 message TriggerImmediatelyRequest {
     // If set, override overlap policy for this one request.
     temporal.api.enums.v1.ScheduleOverlapPolicy overlap_policy = 1;
+    
+    // Timestamp for when this request should be triggered.
```
dnr: This time is not "when" anything will happen, it will always happen immediately. I'd say it more like:
```suggestion
    // Timestamp used for the identity of the target workflow.
```

cretz: Even with that explanation I don't understand it exactly. Is this the value that is in `TemporalScheduledStartTime` SA (and what helps make up the workflow ID) or is this in the workflow start event or something?

dnr: Yeah, it'll be the value in that SA. But more importantly it'll be the value in the workflow id, so the "same" trigger always makes the same workflow id.

Ideally we could fix this issue without touching api, but it does make some sense to have an explicit timestamp here (backfill has explicit timestamps, for example). And putting it here simplifies the server. I don't think it has to be exposed in SDKs but it could be if anyone wants it.

cretz: :+1: I suppose my only suggestion would be to call this `scheduled_time` to clarify it is used as the deterministic/idempotent time the workflow was considered scheduled at, but no strong opinion. I assume we will default this to "now" (also worth clarifying).

dnr: Oh yeah, we should add that it defaults to now in the comment. I'm fine with either name.

spkane31: Renamed to `scheduled_time` and updated the comment with @dnr and @cretz suggestions

dnr (approved):

PR #607, opened 2025-06-27
Add http routes for add/remove search attributes

Alex-Tideman:

**What changed?**
Add HTTP route for adding and removing search attributes

**Why?**

**Breaking changes**

**Server PR**


cretz (approved):
LGTM. But you probably need to re-run the OpenAI generator stuff. Would be eager to remove the `/api/v1` endpoints at some point.

PR #608, opened 2025-07-07
Add process key, make heartbeats repeatable field

ychebotarev:
**What changed?**

Add "process_key" to the worker host info.
Make "worker_heartbeat" repeatable in "PollNexus" and "RecordHeartbeat" APIs.

**Why?**

We decide to have a single Q per process. Thus we need a way to agree on task queue name, and for that we need this information to be:
* available as a part of heartbeat
* unique across at least namespace
Process ID is not unique.

Since we send all heartbeats at once - we make them repeated.

**Breaking changes**
Not yet.

### DIFF
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -54,13 +54,16 @@ message WorkerHostInfo {
   // Worker host identifier.
   string host_name = 1;
 
-  // Worker process identifier, should be unique across the fleet.
-  // It is to disambiguate processed with the same id.
-  // Used to build worker command task queue name:
-  // f"temporal-sys/worker-commands/{process_key}"
+
+  // Worker process identifier. This id should be unique for all _processes_
+  // running workers in the namespace, and should be shared by all workers
+  // in the same process.
+  // This will be used to build the worker command nexus task queue name:
+  // "temporal-sys/worker-commands/{process_key}"
   string process_key = 5;
 
-  // Worker process identifier, should be unique for the host.
+  // Worker process identifier. Unlike process_key, this id only needs to be unique
+  // within one host (so using e.g. a unix pid would be appropriate).
   string process_id = 2;
 
   // System used CPU as a float in the range [0.0, 1.0] where 1.0 is defined as all
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -57,7 +57,7 @@ message WorkerHostInfo {
   // Worker process identifier, should be unique across the fleet.
   // It is to disambiguate processed with the same id.
   // Used to build worker command task queue name:
-  // f"sys/worker/command/{process_key}/"
+  // f"temporal-sys/worker-commands/{process_key}"
   string process_key = 5;
 
   // Worker process identifier, should be unique for the host.
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -54,6 +54,12 @@ message WorkerHostInfo {
   // Worker host identifier.
   string host_name = 1;
 
+  // Worker process identifier, should be unique across the fleet.
+  // It is to disambiguate processed with the same id.
+  // Used to build worker command task queue name:
+  // f"sys/worker/command/{process_key}/"
+  string process_key = 5;
+
   // Worker process identifier, should be unique for the host.
   string process_id = 2;
 
@@ -77,7 +83,6 @@ message WorkerHeartbeat {
   // Worker identity, set by the client, may not be unique.
   // Usually host_name+(user group name)+process_id, but can be overwritten by the user.
   string worker_identity = 2;
-
 
   // Worker host information.
   WorkerHostInfo host_info = 3;
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1772,9 +1772,7 @@ message PollNexusTaskQueueRequest {
     temporal.api.deployment.v1.WorkerDeploymentOptions deployment_options = 6;
 
     // Worker info to be sent to the server.
-    reserved 7;
-    reserved "worker_heartbeat";
-    repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeats = 8;
+    repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 7;
 }
 
 message PollNexusTaskQueueResponse {
@@ -2386,9 +2384,7 @@ message RecordWorkerHeartbeatRequest {
     // The identity of the client who initiated this request.
     string identity = 2;
 
-    reserved 3;
-    reserved "worker_heartbeat";
-    repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeats = 4;
+    repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 3;
 }
 
 message RecordWorkerHeartbeatResponse {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1772,7 +1772,9 @@ message PollNexusTaskQueueRequest {
     temporal.api.deployment.v1.WorkerDeploymentOptions deployment_options = 6;
 
     // Worker info to be sent to the server.
-    temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 7;
+    reserved 7;
+    reserved "worker_heartbeat";
+    repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeats = 8;
 }
 
 message PollNexusTaskQueueResponse {
@@ -2384,7 +2386,9 @@ message RecordWorkerHeartbeatRequest {
     // The identity of the client who initiated this request.
     string identity = 2;
 
-    temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 3;
+    reserved 3;
+    reserved "worker_heartbeat";
+    repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeats = 4;
 }
 
 message RecordWorkerHeartbeatResponse {
--- a/buf.yaml
+++ b/buf.yaml
@@ -13,8 +13,8 @@ breaking:
     - WIRE_JSON
   ignore:
     - google
-    # TODO (yuri) remove this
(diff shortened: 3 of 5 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1772,7 +1772,9 @@ message PollNexusTaskQueueRequest {
     temporal.api.deployment.v1.WorkerDeploymentOptions deployment_options = 6;
 
     // Worker info to be sent to the server.
-    temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 7;
+    reserved 7;
+    reserved "worker_heartbeat";
+    repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeats = 8;
```
dnr: You don't have to do this, you can just change the single to repeated

temporal/api/worker/v1/message.proto
```diff
@@ -54,6 +54,12 @@ message WorkerHostInfo {
   // Worker host identifier.
   string host_name = 1;
 
+  // Worker process identifier, should be unique across the fleet.
+  // It is to disambiguate processed with the same id.
+  // Used to build worker command task queue name:
+  // f"sys/worker/command/{process_key}/"
```
dnr: Is that trailing slash supposed to be there? why?

Do we really need to document the command task queue name here? Why `sys/worker/command/...`? Are there going to be other `sys/...` queue names? What if those conflict with user-set names?

If we're going to reserve a prefix of task queue names for internal purposes, it should probably have `temporal` in there somewhere (similar to reserved search attributes, etc.) so `temporal-sys/...`

temporal/api/worker/v1/message.proto
```diff
@@ -54,6 +54,12 @@ message WorkerHostInfo {
...
+  string process_key = 5;
```
dnr, line 63: We now have a worker_instance_key, worker_identity, process_key, and process_id. This is pretty confusing. First, why can't we use worker_instance_key here? Second, can we maybe have some examples of how these keys should be chosen by a reasonable client to achieve the uniqueness that we want?

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: Yes, in theory going from "optional" to "repeated" is not a breaking change, and that was the first thing I did. 
but linter complains. I add it to the description.

temporal/api/worker/v1/message.proto
ychebotarev: it is here mostly to agree with SDK team on format. I like "temporal-sys". I don't know if there will be other system queues, may be we may want to reuse it for something else. I'm just not sure what that something else can be (process level commands? kind of weird).

But we could be less specific and have something like:  f"temporal-sys/{process_key}"

temporal/api/workflowservice/v1/request_response.proto
dnr: It's not a breaking change at the wire level. We're not using these in history events so I don't care about json-wire compatibility. If the linter complains, disable the linter. We should not make code worse to appease linters.

Sushisource: Just squelch the linter. No reason for this noise for something that hadn't been used yet anyway.

temporal/api/worker/v1/message.proto
dnr: I would want to add one level so we can use the prefix for other things. Like `temporal-sys/worker-command/{process_key}`

temporal/api/worker/v1/message.proto
ychebotarev: We need to have a way to identify the worker. Worker identity is not guarantied to be unique.
We need to have a way to identify the worker process. Process id is not guarantied to be unique.
But we still want to have worker identity and process id. Thus - 4 fields. I just don't see a way to avoid it.

> why can't we use worker_instance_key here?

In the description. We decided that we are going to have a single task queue per ->process<- So we can't use just worker ket.

>can we maybe have some examples.

We define the criteria - unique. I can add "per namespace". But I don't think I want to add the concrete algorithm (like "generate guild and add a hash of host name if pid")

Sushisource: We can't use worker instance key because this queue needs to be unique per-process (IE: multiple workers running in the same process use the same key here). The queue is addressed to a whole process (and may contain a worker instance key when needed to address a specific worker).

All the fields do serve some useful purpose

temporal/api/worker/v1/message.proto
ychebotarev: sounds good to me. @Sushisource ?

temporal/api/worker/v1/message.proto
dnr: Ok, I see. Maybe if I just saw that sentence "multiple workers running in the same process use the same key here" it would jog my mind (and future readers) in the right way so these make sense. Let's add that as a comment?

temporal/api/worker/v1/message.proto
Sushisource: Works for me

cretz: I'm not following a bit myself. Will the server be enforcing a specific stringly-typed format here? Why can't it be an opaque string and if server needs more structured format we can discuss more structure with structured types instead of string formats?

temporal/api/worker/v1/message.proto
cretz: > We need to have a way to identify the worker process

Can I get some background on why this is needed? It makes sense, I am just curious on how it will be used server side.

Sushisource: The nexus task channel will be per-worker process (really, runtime, so there's no global static) to avoid unnecessary connection overhead for use cases with many workers per-process.

ychebotarev: @cretz 
1. What Spencer said
2. Technically we don't need specifically "process key". But we need some way to understand from heartbeat what task queue worker will use to communicate with server. And that task queue will be one per process (so many workers can use it). Thus worker should send some piece of information about that task queue. We decide that semantically "process key" is the best fit. 
3. Other option was ot directly send "temporal system process queue name". Pro: - no need to agree on format. Con - exposing too much internal details, clumsy name, plus we may have more sys queues, so we decide against it.

temporal/api/worker/v1/message.proto
ychebotarev: See my other comment.
We did look into "just send 'process_task_queue' instead of "process_key".
Benefit - no need to agree on format, SDK can choose whatever they want.
Con: too much exposure, and we may later use it for multiple queues instead of sending multiple queue names.

Thus we decide to go with "process_key" and agree on how we build that sys task queue name.

I'm honestly ok with either of those approaches.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ok was not as easy as it should be. I disable breaking check for that file. We should enable it in the follow up PR.

temporal/api/worker/v1/message.proto
cretz: Hrmm, I see. I was only thinking about data pushed and the server shouldn't care where the push came from so long as the pushes contain the globally unique worker identifiers. But I see we're talking about the other direction where the server calls the worker process? I assume we'll define the shape of the implicit Nexus service each worker implements somewhere? Am curious what features it might support, but no need to mess with that now.

temporal/api/worker/v1/message.proto
cretz: Ah, I see. So process key is opaque (can just be a process-level uuidv4) and the comment is saying how the task queue name is built from the key. Works for me.

cretz (approved):

temporal/api/worker/v1/message.proto
ychebotarev: well yes, we are going to receive data via that channel (nexus, specific task queue), and we can just store that task queue name when we receive the heartbeat. And then we are going to use the same channel to send commands to the workers.

But in some cases we may use dedicated "RecordHeartbeat" call (and "ShutdownWorker" ), and in that case we don't know that tasks queue name. So to be safe that information should come with every heartbeat.

temporal/api/worker/v1/message.proto
```diff
@@ -54,6 +54,12 @@ message WorkerHostInfo {
   // Worker host identifier.
   string host_name = 1;
 
+  // Worker process identifier, should be unique across the fleet.
+  // It is to disambiguate processed with the same id.
```
dnr:
```suggestion
  // It is to disambiguate processes with the same id.
```
I don't get what that sentence is supposed to mean though

temporal/api/worker/v1/message.proto
```diff
@@ -54,6 +54,12 @@ message WorkerHostInfo {
...
+  // Used to build worker command task queue name:
+  // f"temporal-sys/worker-commands/{process_key}"
```
dnr:
```suggestion
  // Worker process identifier. This id should be unique for all _processes_
  // running workers in the namespace, and should be shared by all workers
  // in the same process.
  // This will be used to build the worker command nexus task queue name:
  // "temporal-sys/worker-commands/{process_key}"
```

temporal/api/worker/v1/message.proto
```diff
@@ -54,6 +54,12 @@ message WorkerHostInfo {
...
+  string process_key = 5;
+
   // Worker process identifier, should be unique for the host.
```
dnr:
```suggestion
  // Worker process identifier. Unlike process_key, this id only needs to be unique
  // within one host (so using e.g. a unix pid would be appropriate).
```

ychebotarev: changed.

temporal/api/worker/v1/message.proto
ychebotarev: changed. 
mmm...may be my understanding of that word is wrong... But I tested with ChatGPT :)

There is no guarantee that processes on different hosts will have unique process id. So we may happen with two processed on a different hosts with the same process_id.

temporal/api/worker/v1/message.proto
ychebotarev: ok, sounds cleaner.

temporal/api/worker/v1/message.proto
dnr: I see what you meant now, it's just confusing because we're using "id" for like four things now, it wasn't clear at first which one

PR #609, opened 2025-07-07
Remove buf breaking exception

ychebotarev:
**What changed?**
remove exception for "workflowservice/v1/request_response.proto"

**Why?**

This file is important, I don't want to hide it from this check for a long time.

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -13,8 +13,6 @@ breaking:
     - WIRE_JSON
   ignore:
     - google
-    # TODO (yuri) remove this. Added for "heartbeat" change (add
-    - temporal/api/workflowservice/v1/request_response.proto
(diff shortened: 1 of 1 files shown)


buf.yaml
```diff
@@ -13,8 +13,6 @@ breaking:
     - WIRE_JSON
   ignore:
     - google
-    # TODO (yuri) remove this. Added for "heartbeat" change (add
-    - temporal/api/workflowservice/v1/request_response.proto
```
dnr, line 17: we could leave one commented out so it's clear to the next person that this should contain a filename and what the filename should be relative to

dnr (approved):

Sushisource (approved):

PR #610, opened 2025-07-08
Update Workflow and Activity Priority

stephanos:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -42,4 +42,5 @@ message ActivityOptions {
     google.protobuf.Duration heartbeat_timeout = 5;
 
     temporal.api.common.v1.RetryPolicy retry_policy = 6;
+    temporal.api.common.v1.Priority priority = 7;
 }
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -300,7 +300,6 @@ message PendingActivityInfo {
     // If nil, the activity has not yet been dispatched or was last dispatched to an unversioned worker.
     temporal.api.deployment.v1.WorkerDeploymentVersion last_deployment_version = 25;
 
-
     // Priority metadata
     temporal.api.common.v1.Priority priority = 22;
 
@@ -529,6 +528,9 @@ message NexusOperationCancellationInfo {
 message WorkflowExecutionOptions {
     // If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
     VersioningOverride versioning_override = 1;
+
+    // If set, overrides the workflow's priority sent by the SDK.
+    temporal.api.common.v1.Priority priority = 2;
 }
 
 // Used to override the versioning behavior (and pinned deployment version, if applicable) of a
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -8809,6 +8809,9 @@
         },
         "retryPolicy": {
           "$ref": "#/definitions/v1RetryPolicy"
+        },
+        "priority": {
+          "$ref": "#/definitions/v1Priority"
         }
       }
     },
@@ -15868,6 +15871,10 @@
         "versioningOverride": {
           "$ref": "#/definitions/v1VersioningOverride",
           "description": "If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion."
+        },
+        "priority": {
+          "$ref": "#/definitions/v1Priority",
+          "description": "If set, overrides the workflow's priority sent by the SDK."
         }
       }
     },
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -6218,6 +6218,8 @@ components:
           description: Maximum permitted time between successful worker heartbeats.
         retryPolicy:
           $ref: '#/components/schemas/RetryPolicy'
+        priority:
+          $ref: '#/components/schemas/Priority'
     ActivityPropertiesModifiedExternallyEventAttributes:
       type: object
       properties:
@@ -13370,6 +13372,10 @@ components:
           allOf:
             - $ref: '#/components/schemas/VersioningOverride'
           description: If set, takes precedence over the Versioning Behavior sent by the SDK on Workflow Task completion.
+        priority:
+          allOf:
+            - $ref: '#/components/schemas/Priority'
+          description: If set, overrides the workflow's priority sent by the SDK.
     WorkflowExecutionOptionsUpdatedEventAttributes:
       type: object
       properties:


temporal/api/activity/v1/message.proto
```diff
@@ -42,4 +42,5 @@ message ActivityOptions {
     google.protobuf.Duration heartbeat_timeout = 5;
 
     temporal.api.common.v1.RetryPolicy retry_policy = 6;
+    temporal.api.common.v1.Priority priority = 7;
```
stephanos, line 45: This is a bit awkward since I now realized that `PendingActivityInfo` contains `ActivityOptions` already; and we've already added `temporal.api.common.v1.Priority priority = 22;` there ... 😞

temporal/api/workflow/v1/message.proto
```diff
@@ -300,8 +300,8 @@ message PendingActivityInfo {
     // If nil, the activity has not yet been dispatched or was last dispatched to an unversioned worker.
     temporal.api.deployment.v1.WorkerDeploymentVersion last_deployment_version = 25;
 
-
     // Priority metadata
+    // Deprecated. Use `activity_options.priority`.
```
stephanos: This is a unfortunate since I only now realized that `PendingActivityInfo` contains `ActivityOptions` but we've already added `temporal.api.common.v1.Priority priority = 22;` there ... 😞

temporal/api/activity/v1/message.proto
```diff
@@ -42,4 +42,7 @@ message ActivityOptions {
     google.protobuf.Duration heartbeat_timeout = 5;
 
     temporal.api.common.v1.RetryPolicy retry_policy = 6;
+
+    // Overrides the activity's priority sent by the SDK.
+    temporal.api.common.v1.Priority priority = 7;
```
Sushisource: This field has two meanings, one when writing, one when reading. This comment only describes the write behavior. 

The read behavior might be an override or the original.

temporal/api/workflow/v1/message.proto
Sushisource: We need to make sure this value keeps getting written by server for a while.

Should also use `[deprecated = true]`.

Alternatively, related to my other comment, maybe we clarify that this field is _always_ the original priority, and `activity_options` has the override, if any?

temporal/api/activity/v1/message.proto
stephanos: I like that! I'll change it to that.

stephanos: @dnr we could lean even more into this and make the field in `PendingActivityInfo` always represent the _user set_ value and `ActivityOptions` to contain the latest value. Meaning, when it inherits the workflow's priority, the data would show that the user set no priority but the actual priority was the one from the workflow (or a manual override).

Semantically, this would be a breaking change. Maybe fine in pre-release, though.

dnr: I sorta like the idea of differentiating what was explicitly set in on the activity from what it inherited from the workflow... and from what it was later overridden to? There are really four values: 1. inherited from workflow, 2. explicitly set on activity by workflow, 3. overridden by update call, and 4. "effective" priority which is the merged value of the first three in that order.

But it's not obvious to me why PendingActivityInfo would have one of those and ActivityOptions would have another. I mean, if you asked me I probably would have said ActivityOptions should have the explicit set value and PendingActivityInfo would have the "effective" one (backwards from what you said I think).

So if we really want to split them out, maybe we should include multiple Priorities in each of those locations? But before we get into that complexity, let's consider if it's really worth it. Why do users want to read this value and which one are they more interested in?

The most intuitive to me is: this field is 2+3 in my initial list: the explicitly-set Priority, which can be changed by update call. The "effective" value is not reported explicitly anywhere. But I could easily be convinced otherwise.

temporal/api/history/v1/message.proto
```diff
@@ -797,6 +797,7 @@ message WorkflowExecutionOptionsUpdatedEventAttributes {
     string attached_request_id = 3;
     // Completion callbacks attached to the running workflow execution.
     repeated temporal.api.common.v1.Callback attached_completion_callbacks = 4;
+    temporal.api.common.v1.Priority priority = 5;
```
dnr: Maybe a comment here? Does this contain the full new Priority or only the changed fields?

stephanos: 👍 I've added a comment now

Since this is for the workflow options, it's valid to allow zero values in the update (e.g. for fairness key). Therefore we must store the full priority message even if the user only updated one field to disambiguate between default/not changed and a change.

temporal/api/activity/v1/message.proto
stephanos: So I'm inclined to do the same here to KISS; ie make both fields return the same priority message/struct.

stephanos: Question that remains is, deprecate the one in `PendingActivityInfo` or not? So far the other duplicated fields - such as `schedule_to_close_timeout` etc - have not been marked deprecated.

stephanos: Coming back here, I think we have two options:

(a) Deprecate the field in `PendingActivityInfo`; keep the field in `ActivityOptions`. Until we can remove it, hydrate both from the server with the explicitly set data (ie 2+3); ie status quo. 

(b) As proposed above, make the `ActivityOptions` reflect the explicitly set data (ie 2+3); and then I'd argue that realistically leaves only (4) for `PendingActivityInfo` since (1) seems not helpful.

I like (b) only because it would allow us to easily show what the effective priority options are in the CLI/UI. That in itself seems useful to communicate. But that would technically be a breaking behavior change; although, I don't think anyone relies on this (not sure it's exposed?). And it's pre-release.

If we don't find that valuable (or possibly confusing; the biggest risk here AFAICT), we still have option (a) since the feature is in pre-release.

temporal/api/history/v1/message.proto
dnr: Do we have to change SDKs to pick this out and update the priority in workflow info?

temporal/api/activity/v1/message.proto
dnr: Looking again with a fresh perspective: it seems like anything other than (a) (duplicate in both places, maybe deprecate the old one) is inevitably going to confuse someone badly, and I would just go with that. PendingActivityInfo is always understood in the context of a workflow, right? So anyone who sees a PendingActivityInfo can compute the effective Priority. (Standalone activities have no workflow to inherit from so that's fine too.) _If_ we want to expose effective priority to avoid duplicating that logic, let's do it in a new field.

(Re-reading comments, that basically matches what I said on Aug 4, which is good)

temporal/api/history/v1/message.proto
cretz: Yes, most likely.

Hrmm. Workflow info has traditionally been considered immutable for the life of the workflow in some SDKs. This is because people store it, clone it, mock it in tests, etc. People often have expectations of workflow start info as being immutable in some SDKs (but not all).

We can work around this as needed, but it is a bit rough to turn immutable values into mutable in backwards-compatible ways.

temporal/api/activity/v1/message.proto
stephanos: Agreed!

temporal/api/history/v1/message.proto
stephanos: Huh; is that different from versioning overrides? They already update the WorkflowExecutionInfo: https://github.com/temporalio/temporal/blob/44017dcc4949143cb65701694938deb898bcef72/service/history/api/updateworkflowoptions/api.go#L143 which updates https://github.com/temporalio/api/blob/master/temporal/api/workflow/v1/message.proto#L93 AFAICT

dnr: Search attributes and memo are in workflow info too and they are very mutable (and that's been true forever, at least for SAs)

Sushisource: Yeah, I don't think it's a big problem. For those things we have provided them as methods rather than field access, but unfortunately this is already a field IIRC is really the only issue.

Either way, fine.

cretz: > Huh; is that different from versioning overrides?

I believe workflow deployment information in SDKs is provided via a getter for this exact reason and _not_ put on workflow info. But I would have to check, as I don't even see worker deployment info available to workflows in all SDKs at this time.

> Search attributes and memo are in workflow info too and they are very mutable (and that's been true forever, at least for SAs)

Right, they are mutable collections on the info in many SDKs, we don't replace anything on the info itself.

> but unfortunately this is already a field IIRC is really the only issue

Right, but we can keep the field representation as "original" and add getters if we want, or we can just hack this in. But it is a hack for some languages. Because when we say, e.g. in Python, `Info` is a _frozen_ dataclass, the cheating it requires is a bit gross (as it should be, we have set an expectation on how deterministically mutable one can expect these values when they see we mark the info as immutable).

In Java and .NET and Ruby it will also a be a bad hack. If the Rust SDK existed, we just wouldn't be able to do this at all, but we call it "initial information" there maybe to clarify.

Doesn't affect anything in this API/PR though, just bringing up since SDK discussion started here.

PR #612, opened 2025-07-10
Add worker comamnds initial API

ychebotarev:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/sdk/v1/worker_commands.proto
+++ b/temporal/api/sdk/v1/worker_commands.proto
@@ -1,0 +1,71 @@
+syntax = "proto3";
+
+package temporal.api.sdk.v1;
+
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "WorkerCommandsProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+import "google/protobuf/field_mask.proto";
+
+message WorkerConfig {
+    message SimplePollerBehavior {
+      int32 max_pollers = 1;
+    }
+    message AutoscalingPollerBehavior {
+      // At least this many poll calls will always be attempted (assuming slots are available).
+      // Cannot be zero.
+      int32 min_pollers = 1;
+      // At most this many poll calls will ever be open at once. Must be >= `minimum`.
+      int32 max_pollers = 2;
+      // This many polls will be attempted initially before scaling kicks in. Must be between
+      // `minimum` and `maximum`.
+      int32 initial_pollers = 3;
+    }
+
+  int32 max_cached_workflows = 1;
+  oneof poller {
+      SimplePollerBehavior simple_poller = 2;
+      AutoscalingPollerBehavior autoscaling_poller = 3;
+  }
+}
+
+message GetWorkerConfigCommand {
+
+}
+
+// TO DISCUSS: not sure I like the word "response" here, it has a different meaning in the context of commands.
+// ?
+message GetWorkerConfigResponse {
+  // The worker configuration.
+  temporal.api.sdk.v1.WorkerConfig worker_config = 1;
+}
+
+
+message SetWorkerConfigCommand {
+  // The worker configuration to set.
+  message PartialWorkerConfig {
+    WorkerConfig worker_config = 1;
+
+    // DISCUSSION POINT: should we use a field mask here? Or too much complexity for the user?
+    // if not - we can remove this message and just use `WorkerConfig` directly.
+    // Controls which fields from `worker_config` will be applied
+    google.protobuf.FieldMask update_mask = 2;
+  }
+
+  // DISCUSSION POINT: should we use only structured data? I feel there will be a value to just send json string as config
+  // especially for multiple workers in batch.
+  oneof source {
+    // JSON string representation of the worker config
+    PartialWorkerConfig partial_config = 1;
+    string json_config = 2;
+  }
+}
+
+message SetWorkerConfigResponse {
+  // The updated worker configuration.
+  temporal.api.sdk.v1.WorkerConfig worker_config = 1;
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -32,6 +32,7 @@ import "temporal/api/namespace/v1/message.proto";
 import "temporal/api/query/v1/message.proto";
 import "temporal/api/replication/v1/message.proto";
 import "temporal/api/rules/v1/message.proto";
+import "temporal/api/sdk/v1/worker_commands.proto";
 import "temporal/api/schedule/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
 import "temporal/api/update/v1/message.proto";
@@ -2418,4 +2419,42 @@ message ListWorkersResponse {
 
     // Next page token
     bytes next_page_token = 2;
+}
+
+message WorkerCommandRequest {
+    // Namespace this worker belongs to.
+    string namespace = 1;
+
+    // The identity of the client who initiated this request.
+    string identity = 2;
+
+    // for audit purposes
+    string reason = 3;
+
+    // ->THIS IS TO DISCUSS<-
+    // I add this as oneof because we will need a way to send command to multiple workers.
+    // Options are:
+    // - query (will be used as query to ListWorkers, same format as in ListWorkersRequest.query)
+    // - task queue (just a shortcut. Same as query=' "TaskQueue"="my-task-queue" ')
+    // - may be something else? Like SDK, deployment, etc.
+    //   All of these can be replaced with a query, but it is not convenient.
+    // string filter = 5;
+    // string task_queue = 6;
+    // ...
+    oneof selector {
+        // Worker instance key to which the command should be sent.
+        string worker_instance_key = 4;
+    }
+
+    oneof command {
+        temporal.api.sdk.v1.GetWorkerConfigCommand get_worker_config = 10;
+        temporal.api.sdk.v1.SetWorkerConfigCommand set_worker_config = 11;
+    }
+}
+
+message WorkerCommandResponse {
+    oneof response {
+        temporal.api.sdk.v1.GetWorkerConfigResponse get_worker_config = 1;
+        temporal.api.sdk.v1.SetWorkerConfigResponse set_worker_config = 2;
+    }
 }
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -1194,4 +1194,13 @@ service WorkflowService {
             }
         };
     }
+    // ListWorkers is a visibility API to list worker status information in a specific namespace.
+    rpc WorkerCommand (WorkerCommandRequest) returns (WorkerCommandResponse) {
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/workers/command"
+            additional_bindings {
+                get: "/api/v1/namespaces/{namespace}/workers/command"
+            }
+        };
+    }
 }
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -2243,6 +2243,110 @@
         ]
       }
     },
+    "/api/v1/namespaces/{namespace}/workers/command": {
+      "get": {
+        "summary": "ListWorkers is a visibility API to list worker status information in a specific namespace.",
+        "operationId": "WorkerCommand2",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1WorkerCommandResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpcStatus"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "namespace",
+            "description": "Namespace this worker belongs to.",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "identity",
+            "description": "The identity of the client who initiated this request.",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "reason",
+            "description": "for audit purposes",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "workerInstanceKey",
+            "description": "Worker instance key to which the command should be sent.",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.maxCachedWorkflows",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.simplePoller.maxPollers",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.autoscalingPoller.minPollers",
+            "description": "At least this many poll calls will always be attempted (assuming slots are available).\nCannot be zero.",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.autoscalingPoller.maxPollers",
+            "description": "At most this many poll calls will ever be open at once. Must be >= `minimum`.",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.autoscalingPoller.initialPollers",
+            "description": "This many polls will be attempted initially before scaling kicks in. Must be between\n`minimum` and `maximum`.",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.updateMask",
+            "description": "DISCUSSION POINT: should we use a field mask here? Or too much complexity for the user?\nif not - we can remove this message and just use `WorkerConfig` directly.\nControls which fields from `worker_config` will be applied",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "setWorkerConfig.jsonConfig",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          }
+        ],
+        "tags": [
+          "WorkflowService"
+        ]
+      }
+    },
     "/api/v1/namespaces/{namespace}/workers/heartbeat": {
       "post": {
         "summary": "WorkerHeartbeat receive heartbeat request from the worker.",
@@ -5851,6 +5955,110 @@
         ]
       }
     },
+    "/namespaces/{namespace}/workers/command": {
+      "get": {
+        "summary": "ListWorkers is a visibility API to list worker status information in a specific namespace.",
+        "operationId": "WorkerCommand",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1WorkerCommandResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpcStatus"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "namespace",
+            "description": "Namespace this worker belongs to.",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "identity",
+            "description": "The identity of the client who initiated this request.",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "reason",
+            "description": "for audit purposes",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "workerInstanceKey",
+            "description": "Worker instance key to which the command should be sent.",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.maxCachedWorkflows",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.simplePoller.maxPollers",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.autoscalingPoller.minPollers",
+            "description": "At least this many poll calls will always be attempted (assuming slots are available).\nCannot be zero.",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.autoscalingPoller.maxPollers",
+            "description": "At most this many poll calls will ever be open at once. Must be >= `minimum`.",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.workerConfig.autoscalingPoller.initialPollers",
+            "description": "This many polls will be attempted initially before scaling kicks in. Must be between\n`minimum` and `maximum`.",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "setWorkerConfig.partialConfig.updateMask",
+            "description": "DISCUSSION POINT: should we use a field mask here? Or too much complexity for the user?\nif not - we can remove this message and just use `WorkerConfig` directly.\nControls which fields from `worker_config` will be applied",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "setWorkerConfig.jsonConfig",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          }
+        ],
+        "tags": [
+          "WorkflowService"
+        ]
+      }
+    },
     "/namespaces/{namespace}/workers/heartbeat": {
       "post": {
         "summary": "WorkerHeartbeat receive heartbeat request from the worker.",
@@ -7142,6 +7350,19 @@
       },
       "description": "UpdateWorkflowOptions represents updating workflow execution options after a workflow reset.\nKeep the parameters in sync with temporal.api.workflowservice.v1.UpdateWorkflowExecutionOptionsRequest."
     },
+    "SetWorkerConfigCommandPartialWorkerConfig": {
+      "type": "object",
+      "properties": {
+        "workerConfig": {
+          "$ref": "#/definitions/v1WorkerConfig"
+        },
+        "updateMask": {
+          "type": "string",
+          "title": "DISCUSSION POINT: should we use a field mask here? Or too much complexity for the user?\nif not - we can remove this message and just use `WorkerConfig` directly.\nControls which fields from `worker_config` will be applied"
+        }
+      },
+      "description": "The worker configuration to set."
+    },
     "StartOperationResponseAsync": {
       "type": "object",
       "properties": {
@@ -7312,6 +7533,35 @@
       ],
       "default": "PINNED_OVERRIDE_BEHAVIOR_UNSPECIFIED",
       "description": "Used to specify different sub-types of Pinned override that we plan to add in the future.\n\n - PINNED_OVERRIDE_BEHAVIOR_UNSPECIFIED: Unspecified.\n - PINNED_OVERRIDE_BEHAVIOR_PINNED: Override workflow behavior to be Pinned."
+    },
+    "WorkerConfigAutoscalingPollerBehavior": {
+      "type": "object",
+      "properties": {
+        "minPollers": {
+          "type": "integer",
+          "format": "int32",
+          "description": "At least this many poll calls will always be attempted (assuming slots are available).\nCannot be zero."
+        },
+        "maxPollers": {
+          "type": "integer",
+          "format": "int32",
+          "description": "At most this many poll calls will ever be open at once. Must be >= `minimum`."
+        },
+        "initialPollers": {
+          "type": "integer",
+          "format": "int32",
+          "description": "This many polls will be attempted initially before scaling kicks in. Must be between\n`minimum` and `maximum`."
+        }
+      }
+    },
+    "WorkerConfigSimplePollerBehavior": {
+      "type": "object",
+      "properties": {
+        "maxPollers": {
+          "type": "integer",
+          "format": "int32"
+        }
+      }
     },
     "WorkerDeploymentInfoWorkerDeploymentVersionSummary": {
       "type": "object",
@@ -10810,6 +11060,19 @@
       },
       "title": "[cleanup-wv-pre-release]"
     },
+    "v1GetWorkerConfigCommand": {
+      "type": "object"
+    },
+    "v1GetWorkerConfigResponse": {
+      "type": "object",
+      "properties": {
+        "workerConfig": {
+          "$ref": "#/definitions/v1WorkerConfig",
+          "description": "The worker configuration."
+        }
+      },
+      "title": "TO DISCUSS: not sure I like the word \"response\" here, it has a different meaning in the context of commands.\n?"
+    },
     "v1GetWorkerTaskReachabilityResponse": {
       "type": "object",
       "properties": {
@@ -13757,6 +14020,27 @@
       },
       "title": "[cleanup-wv-pre-release] Pre-release deployment APIs, clean up later"
     },
+    "v1SetWorkerConfigCommand": {
+      "type": "object",
+      "properties": {
+        "partialConfig": {
+          "$ref": "#/definitions/SetWorkerConfigCommandPartialWorkerConfig",
+          "title": "JSON string representation of the worker config"
+        },
+        "jsonConfig": {
+          "type": "string"
+        }
+      }
+    },
+    "v1SetWorkerConfigResponse": {
+      "type": "object",
+      "properties": {
+        "workerConfig": {
+          "$ref": "#/definitions/v1WorkerConfig",
+          "description": "The updated worker configuration."
+        }
+      }
+    },
     "v1SetWorkerDeploymentCurrentVersionResponse": {
       "type": "object",
       "properties": {
@@ -15173,6 +15457,32 @@
         }
       },
       "description": "Specifies client's intent to wait for Update results."
+    },
+    "v1WorkerCommandResponse": {
+      "type": "object",
+      "properties": {
+        "getWorkerConfig": {
+          "$ref": "#/definitions/v1GetWorkerConfigResponse"
+        },
+        "setWorkerConfig": {
+          "$ref": "#/definitions/v1SetWorkerConfigResponse"
+        }
+      }
+    },
+    "v1WorkerConfig": {
+      "type": "object",
+      "properties": {
+        "maxCachedWorkflows": {
+          "type": "integer",
+          "format": "int32"
+        },
+        "simplePoller": {
+          "$ref": "#/definitions/WorkerConfigSimplePollerBehavior"
+        },
+        "autoscalingPoller": {
+          "$ref": "#/definitions/WorkerConfigAutoscalingPollerBehavior"
+        }
+      }
     },
     "v1WorkerDeploymentInfo": {
       "type": "object",
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -2041,6 +2041,92 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/namespaces/{namespace}/workers/command:
+    get:
+      tags:
+        - WorkflowService
+      description: ListWorkers is a visibility API to list worker status information in a specific namespace.
+      operationId: WorkerCommand
+      parameters:
+        - name: namespace
+          in: path
+          description: Namespace this worker belongs to.
+          required: true
+          schema:
+            type: string
+        - name: identity
+          in: query
+          description: The identity of the client who initiated this request.
+          schema:
+            type: string
+        - name: reason
+          in: query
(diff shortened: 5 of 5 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,13 @@ service WorkflowService {
             }
         };
     }
+    // ListWorkers is a visibility API to list worker status information in a specific namespace.
```
cretz: This comment doesn't seem accurate. Also ideally there's a blank line after the previous RPC's closing brace.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,13 @@ service WorkflowService {
...
+    rpc WorkerCommand (WorkerCommandRequest) returns (WorkerCommandResponse) {
```
cretz: Would prefer our RPC calls to start with a verb. Can be "send" or something.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,41 @@ message ListWorkersResponse {
...
+    //   All of these can be replaced with a query, but it is not convenient.
+    // string filter = 5;
+    // string task_queue = 6;
+    // ...
+    oneof selector {
+        // Worker instance key to which the command should be sent.
+        string worker_instance_key = 4;
+    }
+
+    oneof command {
```
cretz: We can only send one worker command at a time?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,41 @@ message ListWorkersResponse {
...
 }
+
+message WorkerCommandRequest {
+    // Namespace this worker belongs to.
+    string namespace = 1;
+
+    // The identity of the client who initiated this request.
+    string identity = 2;
+
+    // for audit purposes
```
cretz: Inconsistent with whether you capitalize your comments (and inconsistent with whether you have trailing punctuation)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,41 @@ message ListWorkersResponse {
...
+    string reason = 3;
+
+    // ->THIS IS TO DISCUSS<-
+    // I add this as oneof because we will need a way to send command to multiple workers.
```
cretz: > we will need a way to send command to multiple workers

But the response only has a single response

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,71 @@
...
+      // Cannot be zero.
+      int32 min_pollers = 1;
+      // At most this many poll calls will ever be open at once. Must be >= `minimum`.
+      int32 max_pollers = 2;
+      // This many polls will be attempted initially before scaling kicks in. Must be between
+      // `minimum` and `maximum`.
+      int32 initial_pollers = 3;
+    }
+
+  int32 max_cached_workflows = 1;
```
cretz: Inconsistent with whether you use 2 space or 4 space indentation

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,71 @@
...
+    // Controls which fields from `worker_config` will be applied
+    google.protobuf.FieldMask update_mask = 2;
+  }
+
+  // DISCUSSION POINT: should we use only structured data? I feel there will be a value to just send json string as config
+  // especially for multiple workers in batch.
+  oneof source {
+    // JSON string representation of the worker config
+    PartialWorkerConfig partial_config = 1;
+    string json_config = 2;
```
cretz: I don't think our API needs to be dealing in JSON. The callers of the API can deal in JSON if they need to.

temporal/api/workflowservice/v1/service.proto
cretz: While technically many of our RPCs could support multiple commands, we have chosen to make them individual RPCs. For instance, we could have `WorkflowCommand` to start, say, 5 workflows at a time, but we instead just have start workflow. There's even value in 5 atomic start workflow calls, we just haven't done it yet. I think we should consider supporting explicit get and set worker config calls, and if we find ourselves needing to support multiple different worker calls at once, we should leverage `ExecuteMultiOperation` (even if that means special code has to be written for that to collect which calls go to which workers). I am struggling to think of situations where multiple different worker commands are needed any more than multiple any other commands are needed.

temporal/api/sdk/v1/worker_commands.proto
Sushisource: +1, let's only use the proto

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,71 @@
...
+
+message SetWorkerConfigCommand {
+  // The worker configuration to set.
+  message PartialWorkerConfig {
+    WorkerConfig worker_config = 1;
+
+    // DISCUSSION POINT: should we use a field mask here? Or too much complexity for the user?
+    // if not - we can remove this message and just use `WorkerConfig` directly.
+    // Controls which fields from `worker_config` will be applied
+    google.protobuf.FieldMask update_mask = 2;
```
Sushisource: I think it's worth it. People typically aren't going to be using gRPC directly and the field mask allows us to avoid weird sigil values.

temporal/api/workflowservice/v1/service.proto
Sushisource: There's a few things going on:

* Piping all command types down one rpc is maybe slightly annoying because it's possible for the response type to not match the request command. IMO, not really a big deal though.
* Sending multiple commands to happen "atomically" is possibly useful
* Fixing the first thing doesn't really work with the second thing at the same time.

So, we can:
* Have dedicated RPCs for specific commands (which can share a `WorkerCommandRequest` message containing the id/reason/selector)
* Have an `ExecuteMultiWorkerCommands`. I don't think re-using `ExecuteMultiOperation` makes sense as it's workflow specific. Putting worker commands in there mixes too many unrelated concerns.

My take is it's going to be easier for us to iterate if we do that. We can start with a specific `UpdateWorkerConfig` rpc (with a `WorkerCommandRequest` (maybe `WorkerCommandRequestData`?) inside it. A selector which selects multiple workers would then imply that rpcs like `UpdateWorkerConfig` have a repeated `responses` field, where each response contains `WorkerCommandResponseData` (or whatever name) with identifying information about which worker that response belongs to.

cretz: > Piping all command types down one rpc is maybe slightly annoying because it's possible for the response type to not match the request command. IMO, not really a big deal though.

Not following this exactly. Are you saying the response for get/set config request isn't a get/set config response? Or maybe I'm conflating this user-facing API with the Nexus API the worker will implement?

> Sending multiple commands to happen "atomically" is possibly useful

I can understand that, but it's a single oneof today from what I can see

> So, we can:

I like the first, but I am technically ok with the approach today too. But just to clarify on `ExecuteMultiOperation`, we didn't name it `ExecuteMultiWorkflowOperation` on purpose because it was meant to support all needs, w/ request types added lazily as needed and requests that didn't work together would be a runtime error. It's also very explicitly why we didn't add any shared workflow-like fields at the top level.

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: ok keeping

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: ok.
But.
My concern is automation. 
1. We will return configs as json or some other json-like format.
2. Users may want to create config, and then distribute it to the workers via CLI calls
3. But without it they will need to write a script to convert their configs into proper structure

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: fixed

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I added "job id". If we send command to multiple workers it will be "batch operation" in the background

temporal/api/sdk/v1/worker_commands.proto
cretz: Yes, IMO we either need field mask or we need conflict token. People should be able to comfortably update only one config item. If this field mask is passed down to workers though, it means each SDK will implement their own implementation of this and so I expect them to do it very naively, e.g. a string switch that has a case for `simple_poller.max_pollers`. And we need to discuss what it means if that is provided and they are currently using an autoscaling poller config (does it switch or error?).

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,71 @@
...
+
+message GetWorkerConfigCommand {
+
+}
+
+// TO DISCUSS: not sure I like the word "response" here, it has a different meaning in the context of commands.
+// ?
+message GetWorkerConfigResponse {
+  // The worker configuration.
+  temporal.api.sdk.v1.WorkerConfig worker_config = 1;
```
cretz: Don't have to fully qualify within the file

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,71 @@
...
+      // At most this many poll calls will ever be open at once. Must be >= `minimum`.
+      int32 max_pollers = 2;
+      // This many polls will be attempted initially before scaling kicks in. Must be between
+      // `minimum` and `maximum`.
+      int32 initial_pollers = 3;
+    }
+
+  int32 max_cached_workflows = 1;
+  oneof poller {
+      SimplePollerBehavior simple_poller = 2;
```
cretz: Would recommend we suffix this and the oneof with `_behavior`, it was a bit confusing seeing this as just poller in field masks.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: yes, if we need multiple atomic commands we will add "SendWorkerCommands" (look at the code for ExecuteMultiOperation... I don't think we will use it)

temporal/api/workflowservice/v1/service.proto
ychebotarev: I add blank line. But what is inaccurate?

ychebotarev: oh copy paste...fixed.

temporal/api/workflowservice/v1/service.proto
ychebotarev: Ok I will switch to individual commands. This will also make more sense for invoking commands for multiple workers (we can send config to multiple workers at once, but getting it from multiple workers at once makes less sense, especially given that it will be batch job...)

temporal/api/sdk/v1/worker_commands.proto
dnr: agreed we should not stuff json into proto. we can build tools to convert.

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: removed

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: done

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,71 @@
...
+      AutoscalingPollerBehavior autoscaling_poller = 3;
+  }
+}
+
+message GetWorkerConfigCommand {
+
+}
+
+// TO DISCUSS: not sure I like the word "response" here, it has a different meaning in the context of commands.
```
yuandrew: How does `GetWorkerCommandResult` sound?

temporal/api/sdk/v1/worker_commands.proto
yuandrew: > Users may want to create config, and then distribute it to the workers via CLI calls

I think Chad is suggesting in this case the CLI would be the translation layer that would convert config to JSON, so our API doesn't need to deal with JSON itself.

I agree, I think it's better to have structured proto here

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: better, but seems we decide to go with separate API per command, thus irrelevant.

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: >case the CLI would be the translation layer that would convert config to JSON

If we remove json we will need an opposite - convert JSON to the structured config. Probably not a bad idea - we are moving config validation from workers to the CLI.

ychebotarev: ok that make sense, we can do json validation in CLI, I will close this one

temporal/api/workflowservice/v1/request_response.proto
cretz: So if the RPC is only for sending a single command, is there value in combining all possible commands into a single RPC call?

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,36 @@
...
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "WorkerCommandsProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+message WorkerConfig {
+    message SimplePollerBehavior {
+      int32 max_pollers = 1;
```
cretz: Inconsistent 4-space vs 2-space indention

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,72 @@ message ListWorkersResponse {
...
+
+}
+
+message GetWorkerConfigResponse {
+    oneof response {
+        // The worker configuration.
+        temporal.api.sdk.v1.WorkerConfig worker_config = 1;
+
+        // If the command was sent to a batch job, this is the job ID.
+        string job_id = 3;
```
cretz: How does the system choose whether this will be an immediate response or a batch job? I don't see any "selector" on the get config request. Relatedly, we should try to implement most of this before we merge so we can catch issues like "missing worker instance key on API".

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I think we decided to go with a separate API for different commands?

temporal/api/sdk/v1/worker_commands.proto
cretz, line 1: The name of this file may need to change as there aren't commands in here anymore

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,36 @@
...
+package temporal.api.sdk.v1;
+
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "WorkerCommandsProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+message WorkerConfig {
```
cretz: Do we need some way to version this or have a conflict token to make sure I am not overriding a later call's config, especially in the case of batch jobs that are eventually consistent? We can table now and come back later if we want.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,72 @@ message ListWorkersResponse {
...
+    message PartialWorkerConfig {
+        // The worker configuration to set.
+        temporal.api.sdk.v1.WorkerConfig worker_config = 1;
+
+        // Controls which fields from `worker_config` will be applied
+        google.protobuf.FieldMask update_mask = 2;
+    }
+
+    // JSON string representation of the worker config
+    PartialWorkerConfig partial_config = 5;
```
cretz: Arguably doesn't deserve a sub-message, can just have two fields on the top-level here. This is actually helpful because the absence of a field mask is meant to assume all fields (can you confirm we'll be treating it this way?), and therefore it'd be natural to call set worker config with the full config, not having to set a field called `partial_config` when it actually isn't partial w/ no update mask (the default).

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1194,4 +1194,28 @@ service WorkflowService {
...
+
+    // GetWorkerConfig returns the worker configuration for a specific worker.
+    rpc GetWorkerConfig (GetWorkerConfigRequest) returns (GetWorkerConfigResponse) {
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/workers/config"
+            additional_bindings {
+                get: "/api/v1/namespaces/{namespace}/workers/config"
+            }
+        };
+    }
```
```suggestion
    rpc FetchWorkerConfig (FetchWorkerConfigRequest) returns (FetchWorkerConfigResponse) {
        option (google.api.http) = {
            post: "/namespaces/{namespace}/workers/fetch-config"
            additional_bindings {
                post: "/api/v1/namespaces/{namespace}/workers/fetch-config"
            }
        };
    }
```

This is a bit pedantic, but there are certain assumptions about getters, and especially about HTTP GET, and it is that they should not have side effects (like creating batch jobs). It can actually be a security issue in some situations if an HTTP GET is side effecting.

I'd also be ok updating the `Set` to be something like `Update`, but your call there.

cretz (commented):
Sorry to keep making comments, ideally I'd have them all up front, but some were only realized until later.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: >I don't see any "selector" on the get config request

You are looking at different messages. There is no "job id" or multiple worekrs for "Get", but there is a selector in "SetWorkerConfigRequest", and this ` string job_id = 3;` is from SetWorkerConfigResponse.

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: I don't know how IDE  is doing this...

ychebotarev: or may be I know. We have a mix of 2- and 4- space indentions. 
I will go with 4.

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: renamed

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: good point, but lets put this aside for now..

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: sub-message removed

temporal/api/workflowservice/v1/service.proto
ychebotarev: Set* may support batch, Get - I don't see much value in this.
For now Get will be "single worker" operation. 

I will change Set to be `Update` - more in line with our other APIs. (UpdateSchedule, UpdateActivityOptions, UpdateWorkflowOptions, etc.) and "update_mask".

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: or not... fixed

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,62 @@ message ListWorkersResponse {
...
+
+message UpdateWorkerConfigRequest {
+    // Namespace this worker belongs to.
+    string namespace = 1;
+
+    // The identity of the client who initiated this request.
+    string identity = 2;
+
+    // Reason for sending worker command, can be used for audit purpose.
+    string reason = 3;
```
dnr, line 2507: what other commands are we thinking of here?

in this case, "set" is a superset of "get": if you set without changing anything, it returns the current config, right? so we don't necessarily need a separate "get".

but then on the other hand: we define read-only vs read-write apis at the top-level method level, so we can do it in the authorizer. I think we want to allow a read-only user to get worker configs, but not set them, right? so maybe we actually need two top-level methods, one for read-only commands and one for read-write.

dnr: hmm, I started writing this yesterday and a bunch of stuff changed so it got attached to the wrong line.

I think the current version addresses this concern.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,62 @@ message ListWorkersResponse {
...
+    // Eventually we will extend this as way to send command to multiple workers.
+    // Options are:
+    // - query (will be used as query to ListWorkers, same format as in ListWorkersRequest.query)
+    // - task queue (just a shortcut. Same as query=' "TaskQueue"="my-task-queue" ')
+    // - etc.
+    //   All of these can be replaced with a query, but it is not convenient.
+    // string filter = 5;
+    // string task_queue = 6;
+    // ...
+    oneof selector {
```
cretz: Why are we ok updating (or at least laying foundation to update) multiple configs at once, but not getting multiple configs at once?

temporal/api/workflowservice/v1/service.proto
cretz: We really should not have HTTP GETs (and general getters) having side effects such as communicating with workers. It's the same reason `QueryWorkflow` is not an HTTP GET. There are expectations of getters, especially HTTP GET, which this does not satisfy. It actually can be a security issue with CSRF, though less so for our API than in other web-oriented contexts.

In addition to not altering server state, there are also a lot of other expectations around `GET` such as idempotency and the ability to cache. This is not really an HTTP GET (or really even a "getter") by RPC standards.

ychebotarev: ok, changed to POST

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1185,7 +1185,7 @@ service WorkflowService {
         };
     };
 
-    // ListWorkers is a visibility API to list worker status information in a specific namespace.
+    // ListWorkers is a visibility API to list workers status information in a specific namespace.
```
Sushisource:
```suggestion
    // ListWorkers is a visibility API to list worker status information in a specific namespace.
```

This was right before (or maybe "worker's", but, this is fine)

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: Lets say you have 200 workers.
1. It is natural to run "update all workers with this new config" command. They usually run in the same environment.  But less natural to "get configs for all my workers". 
2. This is batch command, and it will return "job_id". When "get configs for 200 workers" batch job will be complete - where user can go and see the results? We don't have a machinery for it. So if we want this we will need to design it. It is out of scope for this phase.
3. Same will be for other "get" commands, or in general for commands that return some results. We don't have a machinery for "store result of batch and show them to the users". (to my best knowledge).

To summarize - less needed, and we have no machinery to return results of batch jobs.(yes, we can make that workflow, and keep them in memory, etc. But all of this is "more work".)

Sushisource: That explanation makes sense to me, we can always add it later... but I think it should go in some shared message. I can't really imagine a reason certain selectors wouldn't work for certain commands, seems they should always work with all of them. 

IMO this we should have a shared message for namespace/identity/reason, too, but, I'm not super stuck on that. I think David mentioned on some recent review we don't have as many shared messages for stuff like that as it seems we should & I agree.

temporal/api/workflowservice/v1/service.proto
ychebotarev: changed.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: we have "namespace", "identity", "reason".
Beside those we have "key" for "get", and "selector" for set.

We may have shared message for  "namespace", "identity", "reason", but I don't think we should do that for "key" vs "selector". Or create separate "single worker commands" vs "multiple worker command" - too much abstractions. Though no strong opinion.

ychebotarev: also - what will be a good name?
```
message CommandContext {
  string namespace = 1;
  string identity = 2;
  string reason = 3;
}

```
?

temporal/api/sdk/v1/worker_config.proto
```diff
@@ -0,0 +1,36 @@
...
+    message AutoscalingPollerBehavior {
+        // At least this many poll calls will always be attempted (assuming slots are available).
+        // Cannot be zero.
+        int32 min_pollers = 1;
+
+        // At most this many poll calls will ever be open at once. Must be >= `minimum`.
+        int32 max_pollers = 2;
+
+        // This many polls will be attempted initially before scaling kicks in. Must be between
+        //  minimum` and `maximum`.
```
dnr:
```suggestion
        //  `minimum` and `maximum`.
```

temporal/api/sdk/v1/worker_config.proto
```diff
@@ -0,0 +1,36 @@
...
+        int32 initial_pollers = 3;
+    }
+
+    int32 max_cached_workflows = 1;
```
dnr: consider?

```suggestion
    int32 workflow_cache_size = 1;
```

unless that name comes from sdk directly

temporal/api/workflowservice/v1/request_response.proto
cretz: > To summarize - less needed, and we have no machinery to return results of batch jobs.(yes, we can make that workflow, and keep them in memory, etc. But all of this is "more work".)

It is a bit strange to tell people to make 200 calls for fetching worker config but 1 call for updating because of our internal batch representations. We have clearly decided the update call should return the updated config, correct? How are we returning the updated config when multiple are present? I don't mind the inconsistency here as I can understand our batch doesn't support multiple results, but it is confusing from a user POV using the application IMO. When we add updating multiple config, do we have an answer for users that want to fetch multiple configs?

> I think David mentioned on some recent review we don't have as many shared messages for stuff like that as it seems we should & I agree.

I don't agree as much, just because some commands share some fields doesn't necessarily necessitate a new message. Inlining the same fields is fine (I am thankful we don't have a `NamespaceAndIdentity` field for many of our other RPC calls for instance).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,64 @@ message ListWorkersResponse {
...
+    }
+}
+
+message UpdateWorkerConfigResponse {
+    oneof response {
+        // The worker configuration. Will be returned if the command was sent to a single worker.
+        temporal.api.sdk.v1.WorkerConfig worker_config = 1;
+
+        // If the command was sent to a multiple workers - it will be converted into a batch job, and this is the job ID.
+        string job_id = 2;
```
cretz: This will never be used with the current API, correct? Can we remove this for now until there is a selector that supports more than one worker?

temporal/api/workflowservice/v1/service.proto
cretz: I don't think we should call the RPC get either (it is strange for it not to match general HTTP/RPC get expectations but be called "get")

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,64 @@ message ListWorkersResponse {
...
+        string worker_instance_key = 6;
+    }
+}
+
+message UpdateWorkerConfigResponse {
+    oneof response {
+        // The worker configuration. Will be returned if the command was sent to a single worker.
+        temporal.api.sdk.v1.WorkerConfig worker_config = 1;
+
+        // If the command was sent to a multiple workers - it will be converted into a batch job, and this is the job ID.
```
dnr: We haven't written this "batch" thing yet, have we? I suspect it's not going to fit into the existing "batch" framework (which is very tied to visibility) and it's going to be a different internal workflow that does batch-y things, but maybe we should give it a different name since "batch" is used? Or I guess we could let it be ambiguous and determined by context.

The reason I ask this here is because: job_id is currently defined for the batch api, but this thing will probably need new apis, so it may be confusing to call it job_id. Or again, could leave it ambiguous.

temporal/api/workflowservice/v1/request_response.proto
Sushisource: I think the shared message for selector is _more_ important than the `CommandContext` message. If Chad disagrees we need that then I'm not super strong on it.

The reason I'm strong on the selector is that, as you mention in the comments above it, many more selector types seem possible/likely in the future, and it seems much easier to accidentally diverge them in different commands when I think it would be strongly desirable to make sure they are consistent.

temporal/api/workflowservice/v1/service.proto
ychebotarev: propose other name

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: > We have clearly decided the update call should return the updated config, correct? How are we returning the updated config when multiple are present? 

We are returning either config for a single worker, or job_id for multi-worker call.

>When we add updating multiple config, do we have an answer for users that want to fetch multiple configs?

Not at the moment. To have the answer to this (or any other worker command that will return data) question we will need the machinery to store intermediate results, and then return them by "job_id" once job is complete.

ychebotarev: > think the shared message for selector is more important than the CommandContext message. If Chad disagrees we need that then I'm not super strong on it.

How do you see it? We can't use selector with "Get" kind of operations. 
Or you mean "shared massege for all operations that CAN be called for multiple workers"? If so - ok, I will add it.

But I also will add WorkerCommandContext, so we never forget "reason".

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: it will still use the same machinery, just use ListWorkers instead of ListWorkflows

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: how about this?

cretz: > > When we add updating multiple config, do we have an answer for users that want to fetch multiple configs?

> Not at the moment. To have the answer to this (or any other worker command that will return data) [...]

This is my concern from a user POV. Think if I am using the CLI. I can see the result of my update if I update one, but then it's hidden from me if I update two. I can see the result of my config if I ask for it on one, but I am not allowed to ask for it on two.

Thinking from the user POV, it's just strange that I can do some things to multiple but not others, can get results to my config updates sometimes but not others.

Having said all of this, I understand why our server/batch limitations result in user-facing limitations, it's just rough from a user POV (the main POV I am concerned about). The only solution I can come up with is batch being able to store lots of worker configs for fetch/update, but if we can't reasonably build that here (or even to a reasonable max), then I have no other suggested solution if we must use batch.

> If Chad disagrees we need that then I'm not super strong on it.

No super strong opinion myself, but in general doesn't seem worth it to encapsulate into another context any more than any other API.

I will say this though, every non-get-system-info API that is supported on cloud _must_ have `namespace` string as a top-level field due to how we leverage gRPC interceptors to set headers, ref https://github.com/temporalio/sdk-go/blob/2ce51377ca5ca5de78b66c17ce628c5bdfdaccaf/internal/grpc_dialer.go#L162-L172 (and other SDKs).

temporal/api/workflowservice/v1/request_response.proto
cretz: Looks like this comment is referring to an absent field

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2419,3 +2420,38 @@ message ListWorkersResponse {
     // Next page token
     bytes next_page_token = 2;
 }
+
+
+message GetWorkerConfigRequest {
+    temporal.api.sdk.v1.WorkerCommandContext context = 1;
```
cretz: Mentioned at bottom of https://github.com/temporalio/api/pull/612#discussion_r2205735332 why `namespace` must be a field in the request

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,63 @@
...
+    // - query (will be used as query to ListWorkers, same format as in ListWorkersRequest.query)
+    // - task queue (just a shortcut. Same as query=' "TaskQueue"="my-task-queue" ')
+    // - etc.
+    //   All of these can be replaced with a query, but it is not convenient.
+    // string filter = 5;
+    // string task_queue = 6;
+    // ...
+    oneof selector {
+        // Worker instance key to which the command should be sent.
+        string worker_instance_key = 6;
```
Sushisource:
```suggestion
        string worker_instance_key = 1;
```

temporal/api/sdk/v1/worker_commands.proto
```diff
@@ -0,0 +1,63 @@
...
+package temporal.api.sdk.v1;
+
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "WorkerCommandsProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+message WorkerConfig {
```
cretz: Kinda a dumb question, but curious, if SDK team adds a field to this message, does server need to be made aware, or will it automatically just work when the new proto model is pulled in to server?

The reason I ask is because we had traditionally reserved the `temporal.api.sdk.v1` for models that SDK could update independent of server. You can see the `CODEOWNERS` file setup that way. (same issues for other messages in here)

Sushisource (commented):
Personally, I quite like this iteration. Per one of the other comments (I think it was David?) - I'd honestly be fine eliminating "GetWorkerConfig" at this point and just saying that an update to a single worker that doesn't actually update anything is how you "get". Lessens the surface area.

Other than resolving if we want to do that, I have no other blockers.

temporal/api/sdk/v1/worker_commands.proto
Sushisource: It should work that way, yes.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: ok. guess I will revert "WorkerContext".
As for the rest I can only reiterate - we don't have a machinery to collect responses from multiple "Get" kind of requests. Which means that if you want to be consistent then it will be no batches at all. If you want to be useful - no batched just for "Get".

Between those two I select "no batches for Get". Mostly because "oh @#$%@#%$ we need to update all configs" is super common scenario. (and the urgency is much higher.)

temporal/api/sdk/v1/worker_commands.proto
cretz: :+1: Then worker selector probably needs to move out of this package

temporal/api/workflowservice/v1/service.proto
cretz: I did above in the original suggestion - `FetchWorkerConfig`

temporal/api/workflowservice/v1/request_response.proto
cretz: :+1: Makes sense to me our user experience is affected by our system limitations, we just need to be able to explain the inconsistencies here when they manifest in the CLI and other places. Users don't care about our batch problems and our urgency and all of that, they just see something works one way but not the next.

dnr:
> Personally, I quite like this iteration. Per one of the other comments (I think it was David?) - I'd honestly be fine eliminating "GetWorkerConfig" at this point and just saying that an update to a single worker that doesn't actually update anything is how you "get". Lessens the surface area.

I half took back that comment, though, since we have to set apis read-only or r/w at the method level and there's value in having a read-only get config (e.g. the UI can call it and show it to users who can't necessarily edit the config).

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: moved out

temporal/api/sdk/v1/worker_commands.proto
ychebotarev: replaced

temporal/api/workflowservice/v1/service.proto
ychebotarev: ok. accepted.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: I think it is better to say users "you can't do batch get because we don't have the way to do that" then "you can do mass Update because we think you will be 'confused'".

>and our urgency

this is not our urgency. This is user urgency. Situation when users want to do batch config updates are usually urgent (problem that is happening right now).

ychebotarev: closing this one. We agree that we will add selector to FetchConfig, but will validate the request and only accept a single key.

temporal/api/common/v1/message.proto
```diff
@@ -270,3 +270,22 @@ message Priority {
...
+// Will be extended to be able to send command to multiple workers.
+message WorkerSelector {
+    // Options are:
+    // - query (will be used as query to ListWorkers, same format as in ListWorkersRequest.query)
+    // - task queue (just a shortcut. Same as query=' "TaskQueue"="my-task-queue" ')
+    // - etc.
+    //   All but 'query' are shortcuts, can be replaced with a query, but it is not convenient.
+    // string query = 5;
+    // string task_queue = 6;
+    // ...
```
Sushisource, line 285: Probably cut the options that don't exist out of this comment for now.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1209,4 +1209,30 @@ service WorkflowService {
...
+            post: "/namespaces/{namespace}/workers/fetch-config"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/workers/fetch-config"
+                body: "*"
+            }
+        };
+    }
+
+    // UpdateWorkerConfig updates the worker configuration for a specific worker.
```
Sushisource:
```suggestion
    // UpdateWorkerConfig updates the worker configuration of one or more workers.
```

As written, this conflicts with the 3rd comment line. We could keep this and remove that line (since it's technically accurate) or change it to this to be forward-looking.

Sushisource (approved):
A few comment changes but otherwise LGTM

cretz (approved):

cretz:
I would like to request we get pretty far along on the implementation before we merge this to `master`

PR #613, opened 2025-07-11
Proto changes to support activity pause by type options

gow:
We decided to support pausing of activities by type with the following options:
 - All: Pause all running and future activities of the given type.
 - Running: Pause just the currently running activities.
 - Future: Pause just the future activities if the given type.

This is to support different options while pausing by activity type.

### DIFF
--- a/temporal/api/enums/v1/workflow.proto
+++ b/temporal/api/enums/v1/workflow.proto
@@ -162,10 +162,10 @@ enum VersioningBehavior {
     VERSIONING_BEHAVIOR_AUTO_UPGRADE = 2;
 }
 
-// Pause option activity status is used to specify if we want to pause/unpause just the current activity, or just the future activities or both.
-enum PauseOptionActivityStatus {
-    PAUSE_OPTION_ACTIVITY_STATUS_UNSPECIFIED = 0;
-    PAUSE_OPTION_ACTIVITY_STATUS_RUNNING = 1;
-    PAUSE_OPTION_ACTIVITY_STATUS_FUTURE = 2;
-    PAUSE_OPTION_ACTIVITY_STATUS_BOTH = 3;
+// Pause option activity state is used to specify if we want to pause/unpause just the current activity, or just the future activities or both.
+enum PauseOptionActivityState {
+    PAUSE_OPTION_ACTIVITY_STATE_UNSPECIFIED = 0;
+    PAUSE_OPTION_ACTIVITY_STATE_RUNNING = 1;
+    PAUSE_OPTION_ACTIVITY_STATE_FUTURE = 2;
+    PAUSE_OPTION_ACTIVITY_STATE_BOTH = 3;
 }
--- a/temporal/api/enums/v1/workflow.proto
+++ b/temporal/api/enums/v1/workflow.proto
@@ -161,3 +161,11 @@ enum VersioningBehavior {
     // complete on the old Version.
     VERSIONING_BEHAVIOR_AUTO_UPGRADE = 2;
 }
+
+// Pause option activity status is used to specify if we want to pause/unpause just the current activity, or just the future activities or both.
+enum PauseOptionActivityStatus {
+    PAUSE_OPTION_ACTIVITY_STATUS_UNSPECIFIED = 0;
+    PAUSE_OPTION_ACTIVITY_STATUS_RUNNING = 1;
+    PAUSE_OPTION_ACTIVITY_STATUS_FUTURE = 2;
+    PAUSE_OPTION_ACTIVITY_STATUS_BOTH = 3;
+}
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1905,12 +1905,28 @@ message PauseActivityRequest {
     oneof activity {
         // Only the activity with this ID will be paused.
         string id = 4;
-        // Pause all running activities of this type.
-        string type = 5;
+        // Pause by type with option to pause all, running, or future activities of this type.
+        // (-- api-linter: core::0140::prepositions=disabled --)
+        PauseByType pause_by_type = 7;
     }
+    reserved 5;
+    reserved "type";
 
     // Reason to pause the activity.
     string reason = 6;
+
+    // (-- api-linter: core::0123::resource-annotation=disabled --)
+    message PauseByType{
+        string name = 1;
+        PauseByTypeOption option = 2;
+    }
+
+    enum PauseByTypeOption{
+        PAUSE_BY_TYPE_OPTION_UNSPECIFIED = 0;
+        PAUSE_BY_TYPE_OPTION_ALL = 1;
+        PAUSE_BY_TYPE_OPTION_RUNNING = 2;
+        PAUSE_BY_TYPE_OPTION_FUTURE = 3;
+    }
 }
 
 message PauseActivityResponse {
@@ -1980,6 +1996,7 @@ message ResetActivityRequest {
     // Default options are then options activity was created with.
     // They are part of the first SCHEDULE event.
     bool restore_original_options = 9;
+
 
 }
 
@@ -2208,8 +2225,8 @@ message ListWorkerDeploymentsResponse {
         google.protobuf.Timestamp create_time = 2;
         temporal.api.deployment.v1.RoutingConfig routing_config = 3;
         // Summary of the version that was added most recently in the Worker Deployment.
-        temporal.api.deployment.v1.WorkerDeploymentInfo.WorkerDeploymentVersionSummary latest_version_summary = 4; 
-        // Summary of the current version of the Worker Deployment. 
+        temporal.api.deployment.v1.WorkerDeploymentInfo.WorkerDeploymentVersionSummary latest_version_summary = 4;
+        // Summary of the current version of the Worker Deployment.
         temporal.api.deployment.v1.WorkerDeploymentInfo.WorkerDeploymentVersionSummary current_version_summary = 5;
         // Summary of the ramping version of the Worker Deployment.
         temporal.api.deployment.v1.WorkerDeploymentInfo.WorkerDeploymentVersionSummary ramping_version_summary = 6;
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1905,12 +1905,17 @@ message PauseActivityRequest {
     oneof activity {
         // Only the activity with this ID will be paused.
         string id = 4;
-        // Pause all running activities of this type.
+        // Pause activities of this type.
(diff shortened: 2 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1905,12 +1905,28 @@ message PauseActivityRequest {
...
+        PauseByType pause_by_type = 7;
     }
+    reserved 5;
+    reserved "type";
 
     // Reason to pause the activity.
     string reason = 6;
+
+    // (-- api-linter: core::0123::resource-annotation=disabled --)
+    message PauseByType{
```
cretz: Pedantic, but usually there is a space between identifier and open brace

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1905,12 +1905,28 @@ message PauseActivityRequest {
...
+        string name = 1;
+        PauseByTypeOption option = 2;
+    }
+
+    enum PauseByTypeOption{
```
cretz: I was under the impression we were considering visibility-like queries for deciding what activities certain actions would affect, is that the case?

cretz: Might I want to choose all, running, or future for more than just activity type? For instance, how do I pause all future activities regardless of type? Whether the predicate applies to all, running, or future doesn't seem to relate to the predicate itself.

cretz: Unpause has a concept of type also, does this same concern apply there? Also, unpause has a concept of "all" there too, does this same feature need to apply here?

gow: This is for activity pause/unpause APIs. The primary purpose is to allow someone to pause *all* activities of a certain type. The current implementation [explicitly pauses just the running instances](https://github.com/temporalio/temporal/blob/main/service/history/api/unpauseactivity/api.go#L69-L75) of an activity and doesn't pause any future instances.
Regarding rules & actions - nothing changes. Rules will still use queries and identify the activities (by id) to be paused.

gow: These are part of single activity operation APIs (pause/unpause/reset). Pausing all activities within a workflow is under the scope of workflow pause (which is being speced currently).

gow: No. I'm not planning to support selectively unpausing activity by type. We can add it in the future if needed?

cretz: > This is for activity pause/unpause APIs

Mentioned in other comment, but this is not applying to unpause that I can see

> The primary purpose is to allow someone to pause all activities of a certain type

What if I want to pause all activities regardless of type? In unpause that is a boolean, not here?

> The current implementation [explicitly pauses just the running instances](https://github.com/temporalio/temporal/blob/main/service/history/api/unpauseactivity/api.go#L69-L75) of an activity and doesn't pause any future instances.

So how can I pause future instances without being forced to provide a type? Or pause all?

We should be consistent here. Why does whether I want my predicate to apply to all or running or future have to do with what that predicate is?

cretz: What about only pausing future activities regardless of type? It is confusing to only be able to do some activity pausing via workflow pause. I don't want to pause my workflow, I want to pause activities. I think we need to think about this from a devx/user POV. I want to pause all activities, I want to pause running activities, I want to pause future activities, I want to pause running activities of a certain type, etc etc need to be a cohesive, easy to understand interface. To tell someone half of that is done elsewhere is not good devx.

cretz: We already support unpausing activity by type today, see the API. So today we support pausing and unpausing activities with the same predicate capabilities. This now makes pause selection different than unpause selection and therefore confusing from a user understanding POV.

gow: Are we conflating rules and individual activity operations here? My understanding of these individual activity operations is that they operate on specified activity/s. Activities are identified by `id` or by `type` in the API input. When identified by `id` it's implicit that they are currently running activities (so no additional options apply here). When identified by `type` they could be currently running or the ones started in future. So we need additional options (running, future, all). So actions differ based on how you identify the activity.
> So how can I pause future instances without being forced to provide a type?

If we want to do this via activity pause/unpause APIs we can support wildcard `*` to identify all activities? The question is how do we identify/represent "all activities" in these APIs.

Workflow rules have predicates that are evaluated and automated action is taken. This is were typically actions are independent of predicates.
Let me know if my understanding of how these APIs and rules would be employed is incorrect.

gow: Agree. All cases you mentioned are covered except being able to specify "all" activity types. Like mentioned above, we can add a wildcard `*` to allow users to indicate that they want to pause all activities. Thoughts?

gow: I agree about the inconsistency between pause and unpause API. My initial reasoning was that there is no use case to selectively unpause just running, future, all (when unpausing by `type`). So I left it out. I'll add similar unpause options to unpause API as well.

cretz: > My understanding of these individual activity operations is that they operate on specified activity/s.

Why can't I apply activity operations on all activities for a workflow? Why are we limiting to only certain IDs or types?

> When identified by id it's implicit that they are currently running activities (so no additional options apply here). When identified by type they could be currently running or the ones started in future.

Regardless of how identified (id, type, all, selector in the future, etc), I should be able to specify an activity state they should apply to.

Can I pause all running activities? Can I pause all future activities? Can I pause all activities regardless of whether future or running?

> If we want to do this via activity pause/unpause APIs we can support wildcard * to identify all activities?

No, unpause already has this, a boolean to apply to all. We're strangely inconsistent with ourselves on how to apply something to all activities (or all running or all future). We can become consistent by adding a boolean here if we want for "all" or we can change that one.

cretz: Mentioned in other reply, we should be consistent with unpause. Why have a boolean for all there and a special string for all here? And why should I even have to set a type at all if I don't care about type?

cretz: > I'll add similar unpause options to unpause API as well.

:+1: Will be interested in what this looks like when done. I think you'll find that the enumerate state of the activity is wholly unrelated to its type and therefore `PauseByType` doesn't make sense because it doesn't sense to combine the two since they are distinct filters.

gow: I see your point. Added `PauseOptionActivityState` (open to other name suggestions). Now one can pause/unpause any combination of
- activity: {id, type, all}
- activity_state: {running, future, both}

There are some invalid combinations like `{activity: id, activity_state: future}`. But these will be handled in the API implementation.

cretz: This looks great to me!

cretz (approved):
LGTM, but can we make sure to at least get most of the impl done server side before we merge this? I want to confirm the server can work with these changes and we won't learn something during implementation.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1905,12 +1905,17 @@ message PauseActivityRequest {
...
+        // Pause activities of this type.
         string type = 5;
+        // Pause all activities.
+        bool pause_all = 7;
     }
 
     // Reason to pause the activity.
     string reason = 6;
+
+    // which activities to pause - current, future or both.
```
ychebotarev, line 1917: comment says "which activities to pause" - but previous oneof selector selects which activities to pause.
also what is the interaction between id and activity_state? What does it mean 'future activities by id'?

temporal/api/enums/v1/workflow.proto
```diff
@@ -161,3 +161,11 @@ enum VersioningBehavior {
     // complete on the old Version.
     VERSIONING_BEHAVIOR_AUTO_UPGRADE = 2;
 }
+
+// Pause option activity state is used to specify if we want to pause/unpause just the current activity, or just the future activities or both.
+enum PauseOptionActivityState {
```
bergundy, line 166: Why is this in `workflow.proto` and not `activity.proto`?

bergundy: Or maybe this could be generalized for workflows too? Would we want this for workflow pause?

temporal/api/enums/v1/workflow.proto
```diff
@@ -161,3 +161,11 @@ enum VersioningBehavior {
...
+    PAUSE_OPTION_ACTIVITY_STATE_UNSPECIFIED = 0;
+    PAUSE_OPTION_ACTIVITY_STATE_RUNNING = 1;
+    PAUSE_OPTION_ACTIVITY_STATE_FUTURE = 2;
+    PAUSE_OPTION_ACTIVITY_STATE_BOTH = 3;
```
bergundy, line 170:
```suggestion
    PAUSE_OPTION_ACTIVITY_STATE_ANY = 3;
```

bergundy (commented):
I thought we said that we want pause for future activities to be done globally in namespace rules. Surprised to see us adding per-workflow future activity pause.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: same for "match all". I don't think "match all" and "for all, now and in the future" should coexist...
hm. I think you separated those qualifiers into activity_state after discussion with Chad, but I don't think this is the right idea.

it should be part of "type".
(not going to block on this)

temporal/api/enums/v1/workflow.proto
ychebotarev, line 166: either create acitvity.proto, or add to "common". 
Or remove "activity" and change it to "ENTITY" to generalize?

ychebotarev: why this comment is here...

temporal/api/workflowservice/v1/request_response.proto
cretz: IMO they are independent filters and if both are present, both must be satisfied. The other is "which activities", this is "also limit by current state" IMO.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1943,6 +1948,9 @@ message UnpauseActivityRequest {
 
     // If set, the activity will start at a random time within the specified jitter duration.
     google.protobuf.Duration jitter = 9;
+
+    // which activities to unpause - current, future or both.
```
cretz, line 1952: Inconsistent with capitalization of comments

cretz, line 1952: Safe to assume "any" is the default when unspecified? Can you clarify that in comment?

PR #617, opened 2025-07-21
Fairness

stephanos:

**What changed?**

Already [discussed and approved previously here](https://github.com/temporalio/api/pull/513).

**Breaking changes**

N/A

**Server PR**

https://github.com/temporalio/temporal/tree/feature/fairness


temporal/api/common/v1/message.proto
```diff
@@ -266,7 +279,48 @@ message Priority {
...
+    // 1. Weights for a small set of keys can be overridden in task queue
+    //    configuration with an API.
+    // 2. It can be attached to the workflow/activity in this field.
+    // 3. The default weight of 1.0 will be used.
+    // Note that if the weight for a key is attached in this field, for best
+    // results, the same weight should be used for the same key for a reasonable
+    // amount of time (minutes). It may change, but it may take some time for
+    // the change to be reflected.
+    //
+    // The recommended range of usable weights is [0.001, 1000].
```
Sushisource: Why this recommendation? Might be good to elaborate.

stephanos: Let me see ... so the code won't allow anything [smaller than 0.001](https://github.com/temporalio/temporal/blob/feature/fairness/service/matching/fairness_util.go#L23) or [larger than 1000](https://github.com/temporalio/temporal/blob/feature/fairness/service/matching/fair_task_writer.go#L123).

So they are _allowed_ to pass in something smaller or larger; it just won't have an effect. I don't know if we have plans to change that range in the future and don't want to hard-code that as a hard-requirement.

Sushisource: I'd be fine with saying values are currently clamped to that range (more than fine - it seems key that that will happen vs. just "recommended")

temporal/api/common/v1/message.proto
```diff
@@ -266,7 +279,48 @@ message Priority {
...
+    // balancing mechanism. It may correspond to a tenant id, or to a fixed
+    // string like "high" or "low". The default is the empty string.
+    //
+    // The fairness mechanism attempts to dispatch tasks for a given key in
+    // proportion to its weight. For example, using a thousand distinct tenant
+    // ids, each with a weight of 1.0 (the default) will result in each tenant
+    // getting a roughly equal share of task dispatch throughput.
+    //
+    // (Note: this does not imply equal share of worker capacity! Fairness
+    // decisions are made only at dispatch time based on queue statistics, not
```
dnr: This is maybe a little too specific, and currently incorrect:

```suggestion
    // decisions are made only at enqueue time based on queue statistics, not
```

or maybe

```suggestion
    // decisions are made based on queue statistics, not
```

temporal/api/common/v1/message.proto
```diff
@@ -266,7 +279,48 @@ message Priority {
...
+    // Fairness weight for a task can come from multiple sources for
+    // flexibility. From highest to lowest precedence:
+    // 1. Weights for a small set of keys can be overridden in task queue
+    //    configuration with an API.
+    // 2. It can be attached to the workflow/activity in this field.
+    // 3. The default weight of 1.0 will be used.
+    // Note that if the weight for a key is attached in this field, for best
+    // results, the same weight should be used for the same key for a reasonable
+    // amount of time (minutes). It may change, but it may take some time for
+    // the change to be reflected.
```
dnr: let's remove this paragraph since it's not really true anymore

```suggestion
```

Sushisource (approved):

PR #618, opened 2025-07-22
Update descriptions of WorkflowService and OperatorService

dandavison:
Small improvements to descriptions of the services. E.g.
- OperatorService docstrings were falsely claiming that it included namespace registration
- WorkflowService was failing to document the diversity of APIs that it now supports

### DIFF
--- a/temporal/api/operatorservice/v1/service.proto
+++ b/temporal/api/operatorservice/v1/service.proto
@@ -13,10 +13,10 @@ option csharp_namespace = "Temporalio.Api.OperatorService.V1";
 import "temporal/api/operatorservice/v1/request_response.proto";
 import "google/api/annotations.proto";
 
-// OperatorService API defines how Temporal SDKs and other clients interact with the Temporal server
-// to perform administrative functions like registering a search attribute or a namespace.
-// APIs in this file could be not compatible with Temporal Cloud, hence it's usage in SDKs should be limited by
-// designated APIs that clearly state that they shouldn't be used by the main Application (Workflows & Activities) framework.
+// The OperatorService API defines how Temporal SDKs and other clients interact with the Temporal server to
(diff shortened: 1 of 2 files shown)


temporal/api/operatorservice/v1/service.proto
```diff
@@ -13,10 +13,10 @@ option csharp_namespace = "Temporalio.Api.OperatorService.V1";
 import "temporal/api/operatorservice/v1/request_response.proto";
 import "google/api/annotations.proto";
 
-// OperatorService API defines how Temporal SDKs and other clients interact with the Temporal server
-// to perform administrative functions like registering a search attribute or a namespace.
-// APIs in this file could be not compatible with Temporal Cloud, hence it's usage in SDKs should be limited by
-// designated APIs that clearly state that they shouldn't be used by the main Application (Workflows & Activities) framework.
+// The OperatorService API defines how Temporal SDKs and other clients interact with the Temporal server to
+// perform administrative functions. APIs in this file may be incompatible with Temporal Cloud, hence its
```
mjameswh, line 17: > APIs in this file **may** be incompatible with Temporal Cloud

Not something you changed, but wondering… Is there even a single API in `OperatorService` that can be used on Cloud? I don't think so. If none, we should consider making this description more specific.

temporal/api/operatorservice/v1/service.proto
```diff
@@ -13,10 +13,10 @@ option csharp_namespace = "Temporalio.Api.OperatorService.V1";
...
+// usage in SDKs should be limited to APIs that clearly state that they shouldn't be used by the main
```
cretz, line 18: > usage in SDKs should be limited to APIs that clearly state that they shouldn't be used

Not following what you are meaning here. Is this documentation/notes for the SDK team? It was a bit confusing to read at first putting myself in a user's shoes.

PR #619, opened 2025-07-22
Add plugin info to worker heartbeats

Sushisource:

**What changed?**
Record which plugins are in use on WFT completions

**Why?**
Visibility into plugin usage and helpful for debugging

**Breaking changes**
No

**Server PR**

### DIFF
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -127,8 +127,18 @@ message WorkerHeartbeat {
   int32 total_sticky_cache_miss = 21;
   // Current cache size, expressed in number of Workflow Executions.
   int32 current_sticky_cache_size = 22;
+
+  // Plugins currently in use by this SDK.
+  repeated PluginInfo plugins = 23;
 }
 
 message WorkerInfo {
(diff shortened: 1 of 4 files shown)


temporal/api/worker/v1/message.proto
```diff
@@ -127,8 +127,18 @@ message WorkerHeartbeat {
...
+  repeated PluginInfo plugins = 23;
 }
 
 message WorkerInfo {
   WorkerHeartbeat worker_heartbeat = 1;
-}
\ No newline at end of file
+}
+
+message PluginInfo {
```
bergundy, line 139: Should this be `SDKPluginInfo` or be a part of the `sdk` package?

bergundy (approved):
Blind approval on my end, I have zero context here.

temporal/api/worker/v1/message.proto
Sushisource: Probably belongs here for the same reason the other heartbeat stuff is here - it'll be consumed by server for reporting purposes so it's not purely transparent to the server like other things in SDK are.

yuandrew (approved):

PR #620, opened 2025-07-22
Add a BatchOperations proto

spkane31:
Add a `BatchOperation` message 

This proto allows the server code to remove a struct used to safely serialize proto to json and back

No, additions only

https://github.com/temporalio/temporal/pull/8081


temporal/api/batch/v1/message.proto
```diff
@@ -19,6 +19,40 @@ import "temporal/api/enums/v1/reset.proto";
 import "temporal/api/rules/v1/message.proto";
 import "temporal/api/workflow/v1/message.proto";
 
+message BatchOperation {
```
cretz, line 22: This message doesn't seem to be used anywhere in the API. Is there an API use for this? If it is server only, I think it should be in server protos only (or just a server struct or something).

spkane31: ahhh this might be the part I was missing here, I need a request and response proto somewhere

PR #621, opened 2025-07-23
Added-failoverandinitial-version‑fields

vaibhavyadav-dev:

## Describe what has changed in this PR

- Added two new fields to the `GetClusterInfoResponse` proto:
```protobuf
  int64 failover_version_increment = 9;
  int64 initial_failover_version   = 10;

```
## Tell your future self why have you made these changes
These fields were already returned by the server, but never exposed in the API layer.

## Are there any breaking changes on binary or code level?
NO

## If this breaks the Server, please provide the Server PR to merge right after this PR was merged. 
*[*Server PR**
](https://github.com/temporalio/temporal/pull/8090)


bergundy:
> st add a tdbg command for calling the admin api. But I am ok with just exposing them in the existing workflow api as well, don't really have a strong opinion here.
> 
> Please help update the original issue saying we are going with this approach

I don't see a reason not to expose this here.

yycptt:
> I don't see a reason not to expose this here.

Me neither. It just a race condition that we were also discussing this issue during standup yesterday and assigned the task to @simvlad.

PR #622, opened 2025-07-24
Add worker commands payloads

ychebotarev:
**What changed?**
Add payloads for worker commands:
FetchWorkerConfig
UpdateWorkerConfig

Also add DescribeWorker API.

**Why?**
payloads - we need to align with SDK on command payload schemas.

DescribeWorker - will be needed to get all the heartbeat information from the worker.
Partially this can be covered by "ListWorkers" with query "WorkerInstanceKey = 'xxx'", but this is a dedicated API that doesn't require query processing.

**Breaking changes**
Yes (new API)

[Server PR](https://github.com/temporalio/temporal/pull/8094)

### DIFF
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -11,7 +11,6 @@ option csharp_namespace = "Temporalio.Api.Worker.V1";
 
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
-
 import "temporal/api/deployment/v1/message.proto";
 import "temporal/api/enums/v1/common.proto";
 
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -11,11 +11,9 @@ option csharp_namespace = "Temporalio.Api.Worker.V1";
 
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
-import "google/protobuf/field_mask.proto";
 
 import "temporal/api/deployment/v1/message.proto";
 import "temporal/api/enums/v1/common.proto";
-import "temporal/api/sdk/v1/worker_config.proto";
 
 message WorkerPollerInfo {
   // Number of polling RPCs that are currently in flight.
@@ -144,22 +142,4 @@ message PluginInfo {
     string name = 1;
     // The version of the plugin, may be empty.
     string version = 2;
-}
-
-// Will be send to the worker as a payload of the FetchWorkerConfig command.
-message FetchWorkerConfigCommandPayload {
-  // Worker identifier, should be unique for the namespace.
-  string worker_instance_key = 1;
-}
-
-// Will be send to the worker as a payload of the UpdateWorkerConfig command.
-message UpdateWorkerConfigCommandPayload {
-  // Worker identifier, should be unique for the namespace.
-  repeated string worker_instance_key = 1;
-
-  // The new worker config to be applied.
-  temporal.api.sdk.v1.WorkerConfig worker_config = 2;
-
-  // Controls which fields from `worker_config` will be applied
-  google.protobuf.FieldMask update_mask = 3;
 }
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -155,7 +155,7 @@ message FetchWorkerConfigCommandPayload {
 // Will be send to the worker as a payload of the UpdateWorkerConfig command.
 message UpdateWorkerConfigCommandPayload {
   // Worker identifier, should be unique for the namespace.
-  string worker_instance_key = 1;
+  repeated string worker_instance_key = 1;
 
   // The new worker config to be applied.
   temporal.api.sdk.v1.WorkerConfig worker_config = 2;
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -11,8 +11,11 @@ option csharp_namespace = "Temporalio.Api.Worker.V1";
 
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
+import "google/protobuf/field_mask.proto";
+
 import "temporal/api/deployment/v1/message.proto";
 import "temporal/api/enums/v1/common.proto";
+import "temporal/api/sdk/v1/worker_config.proto";
 
 message WorkerPollerInfo {
   // Number of polling RPCs that are currently in flight.
@@ -141,4 +144,22 @@ message PluginInfo {
     string name = 1;
     // The version of the plugin, may be empty.
     string version = 2;
+}
+
+// Will be send to the worker as a payload of the FetchWorkerConfig command.
+message FetchWorkerConfigCommandPayload {
+  // Worker identifier, should be unique for the namespace.
+  string worker_instance_key = 1;
+}
+
+// Will be send to the worker as a payload of the UpdateWorkerConfig command.
+message UpdateWorkerConfigCommandPayload {
+  // Worker identifier, should be unique for the namespace.
+  string worker_instance_key = 1;
+
+  // The new worker config to be applied.
+  temporal.api.sdk.v1.WorkerConfig worker_config = 2;
+
+  // Controls which fields from `worker_config` will be applied
+  google.protobuf.FieldMask update_mask = 3;
 }
--- a/temporal/api/worker/v1/worker_nexus_service_commands.proto
+++ b/temporal/api/worker/v1/worker_nexus_service_commands.proto
@@ -38,23 +38,24 @@ import "temporal/api/sdk/v1/worker_config.proto";
 // --)
 
 // Will be sent to the worker as a payload of the FetchWorkerConfig command.
-message FetchWorkerConfigRequest {
+message FetchWorkerConfigRequestPayload {
   // List of worker identifiers. For now only a single worker instance key is supported.
   repeated string worker_instance_key = 1;
 }
 
-message FetchWorkerConfigResponse {
-  message WorkerConfigEntry {
-    // Worker instance key the config is for.
-    string worker_instance_key = 1;
+message WorkerConfigEntry {
+  // Worker instance key the config is for.
+  string worker_instance_key = 1;
 
-    temporal.api.sdk.v1.WorkerConfig worker_config = 2;
-  }
+  temporal.api.sdk.v1.WorkerConfig worker_config = 2;
+}
+
+message FetchWorkerConfigResponsePayload {
   repeated WorkerConfigEntry worker_configs = 1;
 }
 
 // Will be sent to the worker as a payload of the UpdateWorkerConfig command.
-message UpdateWorkerConfigRequest {
+message UpdateWorkerConfigRequestPayload {
   // List of worker identifiers.
   repeated string worker_instance_key = 1;
 
@@ -65,5 +66,6 @@ message UpdateWorkerConfigRequest {
   google.protobuf.FieldMask update_mask = 3;
 }
 
-message UpdateWorkerConfigResponse {
+message UpdateWorkerConfigResponsePayload {
+  repeated WorkerConfigEntry worker_configs = 1;
 }
--- a/temporal/api/worker/v1/worker_nexus_service_commands.proto
+++ b/temporal/api/worker/v1/worker_nexus_service_commands.proto
@@ -1,0 +1,73 @@
+syntax = "proto3";
+
+package temporal.api.worker.v1;
+
+option go_package = "go.temporal.io/api/worker/v1;worker";
+option java_package = "io.temporal.api.worker.v1";
+option java_multiple_files = true;
+option java_outer_classname = "WorkerNexusServiceCommandsProto";
+option ruby_package = "Temporalio::Api::Worker::V1";
+option csharp_namespace = "Temporalio.Api.Worker.V1";
+
+import "google/protobuf/field_mask.proto";
+import "temporal/api/sdk/v1/worker_config.proto";
+
+// (--
+/////////////////////////////////////////////////////////////////////
+// This file contains:
+// - Conventions between server and worker.
+// - Definitions for commands and payloads for server-worker communication via Nexus
+//
+// WORKER COMMANDS AND CONVENTIONS:
+//
+// Worker commands are used to manage worker configurations, operations, etc.
+// Command names should match names defined in the server API.
+// Currently supported commands:
+// - FetchWorkerConfig
+// - UpdateWorkerConfig.
+//
+// Command names are provided in StartOperationRequest.Operation field.
+//
+// PAYLOAD CONVENTIONS:
+//
+// In/out payloads namings follow the same convention as the regular API:
+// - CommandNameRequest (input payload)
+// - CommandNameResponse (output payload).
+// - Empty payload if response is not needed/not expected
+//
+// COMMUNICATION PROTOCOL:
+// - Transport: Nexus messages on dedicated channel
+// - Server identifier: "sys-worker-service"
+// - Task queue: "temporal-sys/worker-tq/{namespace_name}/{process_key}"
+// --)
+
+// Will be send to the worker as a payload of the FetchWorkerConfig command.
+message FetchWorkerConfigRequest {
+  // List of worker identifiers. For now only a single worker instance key is supported.
+  repeated string worker_instance_key = 1;
+}
+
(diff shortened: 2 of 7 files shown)


temporal/api/worker/v1/message.proto
```diff
@@ -142,3 +145,21 @@ message PluginInfo {
     // The version of the plugin, may be empty.
     string version = 2;
 }
+
+// Will be send to the worker as a payload of the FetchWorkerConfig command.
```
cretz: Not familiar with "FetchWorkerConfig command" or what is meant by "payload" here (we have things already called "commands" and "payloads" in Temporal). Where is this protobuf used? Is it in some Nexus service? If so, can that Nexus service be publicly defined/documented somewhere? Specifically, it is hard to tell where this is used, what "command" means, what the response to the "command" should be (and how it's relayed), etc.

temporal/api/worker/v1/message.proto
```diff
@@ -142,3 +145,21 @@ message PluginInfo {
...
+message FetchWorkerConfigCommandPayload {
+  // Worker identifier, should be unique for the namespace.
+  string worker_instance_key = 1;
```
cretz: Shouldn't this be a worker selector in case they want many? Or is server expected to expand future "many selector" to several individual instance-key commands thereby being limited to only ones it may know about?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1235,4 +1235,15 @@ service WorkflowService {
             }
         };
     }
+
+    // DescribeWorker returns information about the specified worker execution.
+    rpc DescribeWorker (DescribeWorkerRequest) returns (DescribeWorkerResponse) {
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/workers/{worker_instance_key}"
```
cretz: I do not think worker instance key should be in the path here any more than any other `namespaces/{namespace}/workers` based HTTP API. Path pieces that have unique identifiers shouldn't _also_ have API names. So `/namespaces/{namespace}/workers/{worker_instance_key}` and `/namespaces/{namespace}/workers/update-config` don't make sense in the same API

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2537,3 +2537,20 @@ message UpdateWorkerConfigResponse {
...
+
+message DescribeWorkerRequest {
+    // Namespace this worker belongs to.
+    string namespace = 1;
+
+    // The identity of the client who initiated this request.
+    string identity = 2;
+
+    // Worker instance key to describe.
+    string worker_instance_key = 3;
```
cretz: To confirm, we allow fetching many worker configs at once but not many workers at once? Do we need a `ListWorkers`?

temporal/api/worker/v1/message.proto
yuandrew: [FetchWorkerConfig](https://github.com/temporalio/api/blob/c149ac698d5355b582d82187e94f2114d3779859/temporal/api/workflowservice/v1/request_response.proto#L2491C9-L2491C33)

Yes, this is the nexus portion of worker commands, where the server sends commands over to each worker. I believe `Payload` referenced here is for the [payload field in StartOperationRequest](https://github.com/temporalio/api/blob/c149ac698d5355b582d82187e94f2114d3779859/temporal/api/nexus/v1/message.proto#L56). To our understanding, that is how we would send these requests over nexus operations and where we'd need a shared location to store these service definitions

cretz: Can we put a public spec somewhere for this Nexus service? I know we don't really have an IDL yet, but if some of it is based on models inside our API repo, I may need _something_ to properly review how these models are leveraged. It can be just as simple as service name, operation names, and operation inputs/outputs.

temporal/api/worker/v1/message.proto
ychebotarev: when server receives "many" it will fanout to many workers.

cretz: So if I have 30 workers in the same process using the same client, you're going to start 30 Nexus operations instead of just 1?

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: we don't allow fetching many worker configs at once.

ychebotarev: We also have ListWorkers, though I'm not sure what is the question here.

temporal/api/workflowservice/v1/service.proto
ychebotarev: ok, add "describe" as a part of the path

temporal/api/workflowservice/v1/request_response.proto
cretz: > we don't allow fetching many worker configs at once.

https://github.com/temporalio/api/pull/622#discussion_r2228967667 talks about receiving "many"

> We also have ListWorkers, though I'm not sure what is the question here.

I see it now, can ignore this comment thread

temporal/api/worker/v1/message.proto
ychebotarev: we don't support 'fetch many configs'

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2537,3 +2537,20 @@ message UpdateWorkerConfigResponse {
...
+}
+
+message DescribeWorkerResponse {
+    temporal.api.worker.v1.WorkerInfo worker_info = 1;
```
cretz, line 2552: At https://github.com/temporalio/api/pull/601#discussion_r2127359787 it was mentioned we should wrap this for here and for list because more info may come over besides just `WorkerInfo`. But it looks like we may not have wrapped it there (or here).

ychebotarev: if we will need to add more info we will add more fields, either to this message or (if it belongs to the worker_info) to the WorkerInfo. 
What exactly is this comment about?

temporal/api/worker/v1/message.proto
ychebotarev: but this make sense for "update".

ychebotarev: and it shouldn't be "selector". From client perspective selector is either key, or query (or "pseudo-query" like task-queue), and those queries are "executed" in server, during batch processing. 

I added "repeated" to the key.

temporal/api/worker/v1/message.proto
ychebotarev: is this for me or for Andrew? Because I don't know where I can put it, seems should be somewhere in SDK?

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: one of the possible things we may want to add is something like "status" - what is running on the worker, and how things are going. But for that we will need to perform worker command.  I'm not sure this call should be anything more then just query... But we will cross this bridge when we start thinking more about that command.

(also Spenser just added "plugins" to the worker info - is that what you mean?)

temporal/api/worker/v1/message.proto
yuandrew: If it's shared, it would have to go somewhere in this repo, right? This seems like the first of potentially more shared nexus IDLs between server and SDK, seems like we need a new section in this repo to put these in?

Sushisource: We can just define it in a comment here or something. I don't really know where else it would go. The corresponding workflowservice RPCs can mention where to find the comment containing the service definition.

For now I would suggest moving these commands into another proto file inside `worker/v1` like `worker_nexus_service_commands.proto` and a big block comment at the top can say the service name, and the operation names, and how they correspond to workflowservice rpcs

Whether or not it uses some future IDL is something we can decide later. For now, using it looks just like calling any other workflowservice RPC.

temporal/api/worker/v1/message.proto
cretz: > we don't support 'fetch many configs'

But we chose to use worker selector on the RPC for fetch config and just commented that we only support one at a time for now there too. Can we just make this similarly repeated for the same future proofing reasons?

temporal/api/workflowservice/v1/request_response.proto
cretz: >  What exactly is this comment about?

I thought there was discussion/consensus before that we wanted to wrap `WorkerInfo` in something else before returning from list (see the linked comment). I am ok not wrapping it though, I just noticed the comment from before where we were concerned about being able to add a field to each list result without altering the shared-with-describe info (note this only applies to list, describe here can just add fields to the response).

If it's the case that describe response will always be a superset of the same info that comes back in list, probably all good here.

temporal/api/worker/v1/message.proto
```diff
@@ -142,3 +145,21 @@ message PluginInfo {
...
+}
+
+// Will be send to the worker as a payload of the UpdateWorkerConfig command.
+message UpdateWorkerConfigCommandPayload {
+  // Worker identifier, should be unique for the namespace.
+  repeated string worker_instance_key = 1;
```
cretz: To confirm, you can only update worker config if the server knows about the worker. So if you say "update worker config for every worker on X process", if a specific worker on that process hasn't heartbeat, it doesn't get the update? I think this is probably ok, just want to confirm that is the expected behavior.

temporal/api/worker/v1/message.proto
dnr: Maybe even a new package, `worker_nexus_service`  (parallel to workflowservice, operatorservice, etc. except with nexus in the name to make it clear it's not grpc)?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2537,3 +2537,20 @@ message UpdateWorkerConfigResponse {
         // Once we support sending update to a multiple workers - it will be converted into a batch job, and job id will be returned.
     }
 }
+
+message DescribeWorkerRequest {
+    // Namespace this worker belongs to.
+    string namespace = 1;
+
+    // The identity of the client who initiated this request.
+    string identity = 2;
```
dnr: do we need this for a simple get request? no other Describe calls take an identity, afaict only calls that modify things or are long-polls do

temporal/api/worker/v1/message.proto
ychebotarev: I added worker_nexus_service_commands.proto inside "worker".
I don't want it to be separate package - all other "XXXservice" are gRPC services, will be confusing. Also we don't define "operations", only payloads (request/response)

I added big comment describing all the conventions.

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: > describe response will always be a superset of the same info that comes back in list

This. Or at least this is my intention, and I will try to keep it this way.

temporal/api/worker/v1/message.proto
ychebotarev: >But we chose to use worker selector on the RPC

well that is why I was against it...I think it is "fake consistency". 
This will also force me to add multiple "return configs", which introduce extra level of indirection.

But ok, added "repeated", should be ok since this is not user facing, we control each step.
@yuandrew

temporal/api/worker/v1/message.proto
ychebotarev: yes, it doesn't get an update.
But that will be very weird situation - heartbeats are coming via the same nexus poll request...

temporal/api/workflowservice/v1/request_response.proto
ychebotarev: removed

cretz (commented):
All LGTM. Only wanting to wait for approval pending the server and ideally SDK use of this (basically can we have both just referencing this branch while they develop this new Nexus service?)

temporal/api/worker/v1/worker_nexus_service_commands.proto
```diff
@@ -0,0 +1,73 @@
...
+// - Conventions between server and worker.
+// - Definitions for commands and payloads for server-worker communication via Nexus
+//
+// WORKER COMMANDS AND CONVENTIONS:
+//
+// Worker commands are used to manage worker configurations, operations, etc.
+// Command names should match names defined in the server API.
+// Currently supported commands:
+// - FetchWorkerConfig
+// - UpdateWorkerConfig.
```
Sushisource: This list might not get updated, and is kinda redundant with the message definitions anyway. I'd just remove it.

temporal/api/worker/v1/worker_nexus_service_commands.proto
```diff
@@ -0,0 +1,73 @@
...
+//
+// PAYLOAD CONVENTIONS:
+//
+// In/out payloads namings follow the same convention as the regular API:
+// - CommandNameRequest (input payload)
+// - CommandNameResponse (output payload).
+// - Empty payload if response is not needed/not expected
+//
+// COMMUNICATION PROTOCOL:
+// - Transport: Nexus messages on dedicated channel
```
Sushisource:
```suggestion
// - Transport: Nexus tasks on dedicated task queue
```

temporal/api/worker/v1/worker_nexus_service_commands.proto
```diff
@@ -0,0 +1,73 @@
...
+// - Server identifier: "sys-worker-service"
+// - Task queue: "temporal-sys/worker-tq/{namespace_name}/{process_key}"
+// --)
+
+// Will be send to the worker as a payload of the FetchWorkerConfig command.
```
Sushisource:
```suggestion
// Will be sent to the worker as a payload of the FetchWorkerConfig command.
```

temporal/api/worker/v1/worker_nexus_service_commands.proto
```diff
@@ -0,0 +1,73 @@
...
+  message WorkerConfigEntry {
+    // Worker instance key the config is for.
+    string worker_instance_key = 1;
+
+    temporal.api.sdk.v1.WorkerConfig worker_config = 2;
+  }
+  repeated WorkerConfigEntry worker_configs = 1;
+}
+
+// Will be send to the worker as a payload of the UpdateWorkerConfig command.
```
Sushisource:
```suggestion
// Will be sent to the worker as a payload of the UpdateWorkerConfig command.
```

temporal/api/worker/v1/worker_nexus_service_commands.proto
ychebotarev: ok

temporal/api/worker/v1/worker_nexus_service_commands.proto
ychebotarev: done

temporal/api/worker/v1/worker_nexus_service_commands.proto
ychebotarev: done

temporal/api/worker/v1/worker_nexus_service_commands.proto
ychebotarev: done

Sushisource (commented):
Agree w/ Chad that let's not merge this until we have something working in branches w/ both your and Andrew's code

temporal/api/worker/v1/worker_nexus_service_commands.proto
```diff
@@ -0,0 +1,69 @@
...
+// - Empty payload if response is not needed/not expected
+//
+// COMMUNICATION PROTOCOL:
+// - Transport: Nexus tasks on task queue
+// - Server identifier: "sys-worker-service"
+// - Task queue: "temporal-sys/worker-tq/{namespace_name}/{process_key}"
+// --)
+
+// Will be sent to the worker as a payload of the FetchWorkerConfig command.
+message FetchWorkerConfigRequest {
```
yuandrew: I feel like it's a little confusing that these are named the exact same name as the request they're used for, could we differentiate them slightly? Maybe add in `Command`, `Payload`, or `Nexus` into the name?

ychebotarev: we need "Request/Response" pair. "Command" is fine for "Request", but what about "Response"?
Or you mean we should name them

`FetchWorkerConfigCommand[Request/Resonse] `? If so I'm fine with that.

ychebotarev: Or `FetchWorkerConfig[Request|Response]Payload`? Which is exactly what it is

PR #626, opened 2025-08-05
Allow fairness weight overrides via UpdateTaskQueueConfigApi

sivagirish81:

**What changed?**
+ Added a update fairness weights map to the UpdateTaskQueueConfig api request.
+ Added a fairness_weights map in the taskQueueConfig message to enable persistence and return response.

**Why?**
+ Allow override of fairness weights for individual fairness keys via UpdateTaskQueueConfig api.

**Breaking changes**
+ No.

**Server PR**

### DIFF
--- a/temporal/api/taskqueue/v1/message.proto
+++ b/temporal/api/taskqueue/v1/message.proto
@@ -343,4 +343,6 @@ message TaskQueueConfig {
     RateLimitConfig queue_rate_limit = 1;
     // If set, each individual fairness key will be limited to this rate, scaled by the weight of the fairness key.
     RateLimitConfig fairness_keys_rate_limit_default = 2;
+    // If set, overrides the fairness weights for the corresponding fairness keys.
+    map<string, float> fairness_weights = 3;
 }
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -2484,6 +2484,9 @@ message UpdateTaskQueueConfigRequest {
     // If not set, this configuration is unchanged.
     // If the `rate_limit` field in the `RateLimitUpdate` is missing, remove the existing rate limit.
     RateLimitUpdate update_fairness_key_rate_limit_default = 6;
+    // If set, overrides the fairness weights for the fairness keys.
+    // Set as -1 to unset the fairness key.
+    map<string, float> update_fairness_weights = 7;
 }
 
 message UpdateTaskQueueConfigResponse {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -8940,6 +8940,14 @@
         "updateFairnessKeyRateLimitDefault": {
           "$ref": "#/definitions/UpdateTaskQueueConfigRequestRateLimitUpdate",
           "description": "Update to the default fairness key rate limit.\nIf not set, this configuration is unchanged.\nIf the `rate_limit` field in the `RateLimitUpdate` is missing, remove the existing rate limit."
+        },
+        "updateFairnessWeights": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "number",
+            "format": "float"
+          },
+          "description": "If set, overrides the fairness weights for the fairness keys.\nSet as -1 to unset the fairness key."
         }
       }
     },
@@ -15145,6 +15153,14 @@
         "fairnessKeysRateLimitDefault": {
           "$ref": "#/definitions/v1RateLimitConfig",
           "description": "If set, each individual fairness key will be limited to this rate, scaled by the weight of the fairness key."
+        },
+        "fairnessWeights": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "number",
+            "format": "float"
+          },
+          "description": "If set, overrides the fairness weights for the corresponding fairness keys."
         }
       }
     },
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -12196,6 +12196,12 @@ components:
           allOf:
             - $ref: '#/components/schemas/RateLimitConfig'
           description: If set, each individual fairness key will be limited to this rate, scaled by the weight of the fairness key.
+        fairnessWeights:
+          type: object
+          additionalProperties:
+            type: number
+            format: float
+          description: If set, overrides the fairness weights for the corresponding fairness keys.
     TaskQueueReachability:
       type: object
       properties:
@@ -12776,6 +12782,14 @@ components:
             Update to the default fairness key rate limit.
              If not set, this configuration is unchanged.
              If the `rate_limit` field in the `RateLimitUpdate` is missing, remove the existing rate limit.
+        updateFairnessWeights:
+          type: object
+          additionalProperties:
+            type: number
+            format: float
+          description: |-
+            If set, overrides the fairness weights for the fairness keys.
+             Set as -1 to unset the fairness key.
     UpdateTaskQueueConfigRequest_RateLimitUpdate:
       type: object
       properties:


temporal/api/taskqueue/v1/message.proto
```diff
@@ -343,4 +343,6 @@ message TaskQueueConfig {
     RateLimitConfig queue_rate_limit = 1;
     // If set, each individual fairness key will be limited to this rate, scaled by the weight of the fairness key.
     RateLimitConfig fairness_keys_rate_limit_default = 2;
+    // If set, overrides the fairness weights for the corresponding fairness keys.
+    map<string, float> fairness_weights = 3;
```
dnr:
```suggestion
    map<string, float> fairness_weight_overrides = 3;
```
more verbose but I want to make it clear these are overrides for the weights attached to Priorities

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2484,6 +2484,11 @@ message UpdateTaskQueueConfigRequest {
     // If not set, this configuration is unchanged.
     // If the `rate_limit` field in the `RateLimitUpdate` is missing, remove the existing rate limit.
     RateLimitUpdate update_fairness_key_rate_limit_default = 6;
+    // If set, overrides the fairness weights for the fairness keys.
+    // Sentinel value -1 is used to unset a fairness weight override.
+    // If unset, the fairness weights fall back to the values set during task creation (if any),
+    // or to the default weight of 1.0.
+    map<string, float> update_fairness_weights = 7;
```
dnr:
```suggestion
    map<string, float> update_fairness_weight_overrides = 7;
```

dnr (approved):
I think this is simple and sufficient, I'd just tweak the names slightly.

Does anyone have objections to the "-1 to unset"?

cretz (commented):
(from a structure POV all LGTM, but am deferring SDK review to someone on SDK team more familiar w/ the priority/fairness project, maybe @Sushisource?)

temporal/api/taskqueue/v1/message.proto
sivagirish81: Updated.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Updated.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2525,6 +2525,11 @@ message UpdateTaskQueueConfigRequest {
     // If not set, this configuration is unchanged.
     // If the `rate_limit` field in the `RateLimitUpdate` is missing, remove the existing rate limit.
     RateLimitUpdate update_fairness_key_rate_limit_default = 6;
+    // If set, overrides the fairness weights for the fairness keys.
+    // Sentinel value -1 is used to unset a fairness weight override.
```
cretz: This is a bit rough. How do we expect unsetting an override to look in CLI, will we ask users to use `-1`? Arguably a separate list of keys to remove could be clearer, but I do not have strong opinion

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2525,6 +2525,11 @@ message UpdateTaskQueueConfigRequest {
...
+    // If unset, the fairness weights fall back to the values set during task creation (if any),
+    // or to the default weight of 1.0.
```
cretz: If other values in this request are unset, things remain as is. Are we inconsistently saying for this one field if you don't set it, we are going to revert to fall back values instead of leaving the overrides alone? Or do you mean "If unset" as in if a certain key override is not present at runtime, not whether this field is unset? It is a bit confusing looking at other fields in this request talk about "If not set" which does refer to the field, not the runtime behavior of the value.

sivagirish81: >If other values in this request are unset, things remain as is. Are we inconsistently saying for this one field if you don't set it, we are going to revert to fall back values instead of leaving the overrides alone? 

No. The same behavior applies for this field as well. If the values are not set in the api calls we will leave the overrides untouched. 
```suggestion
    // If unset via sentinel value, the fairness weights fall back to the values set during task creation (if any),
    // or to the default weight of 1.0.
```
If the value of a particular fairness key is unset in the map by using the sentinel value then the statement applies. I can see how it might be confusing.

temporal/api/workflowservice/v1/request_response.proto
sivagirish81: Addressing this here : https://github.com/temporalio/cli/pull/840/files#r2379720580

stephanos: Options I'm considering here now:

(1)
```
map<string, float> set_fairness_weight_overrides = 7;  // float must be zero or greater
repeated string unset_fairness_weight_overrides = 8;
```

(2)
```
message FairnessWeightUpdate {
  string key = 1;
  oneof op {
    float set = 2;
    bool unset = 3;      
  }
}

repeated FairnessWeightUpdate update_fairness_weight_overrides = 7;
```

(3) AI also suggested this to me. I didn't know proto3 actually had `optional`. [Looks legit](https://protobuf.dev/programming-guides/proto3/#option-retention).

```
message FairnessWeightUpdate {
    optional float weight = 1;
}

map<string, FairnessWeightUpdate> update_fairness_weight_overrides = 7;
```

---

(1) seems clearest and easiest.

cretz: Agreed, 1 is clearest IMO. Would like to avoid optional and empty messages if we can. It is less clear to users to differentiate scalars between `0` and "none" (if optional even makes sense that way). People expect `0` and unset to be the same thing.

sivagirish81: Approach 1 looks good to me. A couple of points to consider for the server implementation:
- What should the expected behavior be if a user tries to set a fairness key as well as unset a fairness key in the same request. Do we do a sanity check and drop the request as a whole or do we prioritize the update.
- We should enforce a limit on the number of keys that can be deleted in a single request, similar to how we enforce a limit on the number of fairness weight overrides allowed at a time.
- Do we also want to support a mechanism for unsetting all overrides at once? This could be useful for reset scenarios.

stephanos: Good points to consider! I'd say

- "reject if a request contains conflicting requests"
- +1 to combined update limit
- could add a `bool` field (now validation becomes trickier) or declare out of scope for now

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2525,6 +2525,12 @@ message UpdateTaskQueueConfigRequest {
     // If not set, this configuration is unchanged.
     // If the `rate_limit` field in the `RateLimitUpdate` is missing, remove the existing rate limit.
     RateLimitUpdate update_fairness_key_rate_limit_default = 6;
+    // If set, overrides the fairness weight for each specified fairness key.
+    // Fairness keys not listed in this map will keep their existing overrides (if any).
+    map<string, float> set_fairness_weight_overrides = 7;
+    // If set, removes any existing fairness weight overrides for each specified fairness key.
+    // Fairness keys not listed will keep their existing overrides (if any).
+    repeated string unset_fairness_weight_overrides = 8;
```
sivagirish81, line 2537:
```suggestion
    // If set, removes any existing fairness weight overrides for each specified fairness key.
    // Fairness weights for corresponding keys fall back to the values set during task creation (if any),
    // or to the default weight of 1.0.
    repeated string unset_fairness_weight_overrides = 8;
```
When fairness keys are unset we can mention what the values will revert back to. Considering this is an unset field we don't need to mention 
>// Fairness keys not listed will keep their existing overrides (if any).

ShahabT (approved):

PR #627, opened 2025-08-07
[DO NOT MERGE] Use Temporal Failures for Nexus APIs

bergundy:
Also add a failure representation of the Nexus SDK's OperationError.
The main motivation for this change is for consistency with the rest of the Temporal APIs and to make payload visiting in proxies simpler.

I do not intend to merge this PR until I have the corresponding Go SDK and server PRs.
I do want to get early feedback on this direction before I polish the work on the two remaining PRs.


temporal/api/failure/v1/message.proto
```diff
@@ -125,6 +131,7 @@ message Failure {
         ChildWorkflowExecutionFailureInfo child_workflow_execution_failure_info = 12;
         NexusOperationFailureInfo nexus_operation_execution_failure_info = 13;
         NexusHandlerFailureInfo nexus_handler_failure_info = 14;
+        NexusSDKOperationFailureInfo nexus_sdk_operation_failure_info = 15;
```
Quinn-With-Two-Ns, line 140: I am worried about adding this, can we confirm this error won't end up as the cause of a nexus operation failure in history?

temporal/api/failure/v1/message.proto
```diff
@@ -91,6 +92,11 @@ message NexusHandlerFailureInfo {
     temporal.api.enums.v1.NexusHandlerErrorRetryBehavior retry_behavior = 2;
 }
 
+// Representation of the Nexus SDK OperationError object.
+message NexusSDKOperationFailureInfo {
+    string state = 1;
```
cretz, line 97: What is "state" here? May need just a quick comment about what's accepted or what it refers to.

temporal/api/failure/v1/message.proto
bergundy: Yeah, the idea is that it would only be used in non-workflow callers and in handlers. But it may still show up in the cause chain in workflow, just not the top level failure.

temporal/api/nexus/v1/message.proto
```diff
@@ -86,6 +93,8 @@ message Request {
     //     aip.dev/not-precedent: Not following linter rules. --)
     google.protobuf.Timestamp scheduled_time = 2;
 
+    Capabilities capabilities = 100;
```
cretz, line 96: Who is responsible for setting this? I assume the server correct? So is this Nexus `Request` message only expected to ever be constructed by the server? If this is a Temporal-server-set-only thing, should it instead be on `PollNexusTaskQueueResponse`?

temporal/api/failure/v1/message.proto
bergundy: It's `failed | canceled` per the Nexus spec. We leave it as a string to allow adding statuses on the Nexus side without requiring a Temporal upgrade.

temporal/api/nexus/v1/message.proto
bergundy: Either of these are constructed only by the server. No strong opinion on where this should go.

temporal/api/failure/v1/message.proto
Quinn-With-Two-Ns: When would this be part of the cause chain in a workflow?

bergundy: Since any failure can appear as a cause, there's nothing preventing users from setting `OperationError` as the cause of an `ApplicationError` or `HandlerError`. I don't expect this to typically happen though.
Maybe when we have non-workflow callers and your operation handler uses the client to forward the operation and the handler wraps that error. Here's an example:

```
func (*myHandler) StartOperation(...) (...) {
  sc := temporalnexus.GetClient(...).NexusServiceClient(...)
  _, err := sc.StartOperation(...)
  return fmt.Errorf("failed to forward call: %w", err)
}
```

temporal/api/failure/v1/message.proto
cretz: :+1: Worth adding a comment here saying this is an enum defined in the Nexus spec

temporal/api/nexus/v1/message.proto
cretz: I like the idea of being on the poll response myself, but I similarly do not have a strong opinion

temporal/api/nexus/v1/message.proto
```diff
@@ -114,6 +123,9 @@ message StartOperationResponse {
         Async async_success = 2;
         // The operation completed unsuccessfully (failed or canceled).
         UnsuccessfulOperationError operation_error = 3;
+        // The operation completed unsuccessfully (failed or canceled).
+        // Failure object must contain a NexusSDKOperationFailureInfo object.
+        temporal.api.failure.v1.Failure failure = 4;
```
cretz, line 128: May want to say if this is set, the `operation_error` is ignored, but not that important

cretz (approved):

temporal/api/failure/v1/message.proto
Quinn-With-Two-Ns: Yes that's true, my main concern is we don't change the current error chains type users see in workflows. Can we confirm that will remain the same?

temporal/api/nexus/v1/message.proto
bergundy: It's already a oneof so that's not possible.

temporal/api/failure/v1/message.proto
bergundy: Yup, those will stay the same except for that we won't construct an implicit `ApplicationError` as the cause when users call construct a handler error with a message and both caller and handler have been upgraded.

PR #629, opened 2025-08-12
pause: add pause information to workflow execution info

spkane31:
Add workflow and activity pause info to the `WorkflowExecutionInfo` protobuf. This can be used by the cli (in `temporal workflow describe`) to display information about workflows paused.

To enable discovery of paused workflows.

No

Server PR: https://github.com/temporalio/temporal/pull/8188

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -9296,29 +9296,6 @@
         }
       }
     },
-    "v1ActivityPauseInfo": {
-      "type": "object",
-      "properties": {
-        "updateTime": {
-          "type": "string",
-          "format": "date-time",
-          "description": "The time when the activity pause info was updated."
-        },
-        "activityType": {
-          "type": "string",
-          "description": "The type of the activity to be paused."
-        },
-        "identity": {
-          "type": "string",
-          "description": "The identity of the actor that paused the activity."
-        },
-        "reason": {
-          "type": "string",
-          "description": "The reason for pausing the activity."
-        }
-      },
-      "description": "ActivityPauseInfo contains the details of the request to pause an activity."
-    },
     "v1ActivityPropertiesModifiedExternallyEventAttributes": {
       "type": "object",
       "properties": {
@@ -12742,6 +12719,29 @@
     },
     "v1PauseActivityResponse": {
       "type": "object"
+    },
+    "v1PausedActivities": {
+      "type": "object",
+      "properties": {
+        "updateTime": {
+          "type": "string",
+          "format": "date-time",
+          "description": "The time when the activity pause info was updated."
+        },
+        "activityType": {
+          "type": "string",
+          "description": "The type of the activity to be paused."
+        },
+        "identity": {
+          "type": "string",
+          "description": "The identity of the actor that paused the activity."
+        },
+        "reason": {
+          "type": "string",
+          "description": "The reason for pausing the activity."
+        }
+      },
+      "description": "PausedActivities contains information about a paused activity type.\nAny pending activity of this type should be either in PAUSE_REQUESTED or PAUSED states.\nFuture activities of this type will automatically be put into PAUSED state."
     },
     "v1Payload": {
       "description": "Arbitrary payload data in an unconstrained format.\nThis may be activity input parameters, a workflow result, a memo, etc.\n"
@@ -17013,13 +17013,13 @@
     "v1WorkflowPauseInfo": {
       "type": "object",
       "properties": {
-        "activityPauseInfos": {
+        "pausedActivities": {
           "type": "array",
           "items": {
             "type": "object",
-            "$ref": "#/definitions/v1ActivityPauseInfo"
+            "$ref": "#/definitions/v1PausedActivities"
           },
-          "description": "List of activities that are to be paused."
+          "description": "List of activity types that are paused, any future activities of this type will automatically be put into PAUSED state."
         }
       },
       "description": "WorkflowPauseInfo contains the details of the request to pause the workflow."
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -6508,23 +6508,6 @@ components:
           description: Maximum permitted time between successful worker heartbeats.
         retryPolicy:
           $ref: '#/components/schemas/RetryPolicy'
-    ActivityPauseInfo:
-      type: object
-      properties:
-        updateTime:
-          type: string
-          description: The time when the activity pause info was updated.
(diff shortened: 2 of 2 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -100,6 +100,27 @@ message WorkflowExecutionInfo {
     temporal.api.common.v1.Priority priority = 24;
 }
 
+// WorkflowPauseInfo contains the details of the request to pause the workflow.
+message WorkflowPauseInfo {
+    // List of activities that are to be paused.
+    repeated ActivityPauseInfo activity_pause_infos = 1;
+}
+
+// ActivityPauseInfo contains the details of the request to pause an activity.
```
bergundy:
```suggestion
// ActivityPauseInfo contains information about a paused activity type.
// Any pending activity of this type should be either in PAUSE_REQUESTED or PAUSED states.
// Future activities of this type will automatically be put into PAUSED state.
```

temporal/api/workflow/v1/message.proto
```diff
@@ -100,6 +100,27 @@ message WorkflowExecutionInfo {
     temporal.api.common.v1.Priority priority = 24;
 }
 
+// WorkflowPauseInfo contains the details of the request to pause the workflow.
+message WorkflowPauseInfo {
+    // List of activities that are to be paused.
+    repeated ActivityPauseInfo activity_pause_infos = 1;
```
cretz: Can we come up with a better term for "thing that pauses future activities"? Because I would expect "activity pause info" to be information about a paused activity, not a "rule" or thing to apply to future activities. Here are some suggestions:

* "activity pause rules"
* "activity pause conditions"
* "activity pause policies"

etc. Basically the comment "List of activities that are to be paused" is not really accurate, it's more of "List of _things_ that will pause future matching activities".

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1042,6 +1042,7 @@ message DescribeWorkflowExecutionResponse {
     repeated temporal.api.workflow.v1.CallbackInfo callbacks = 6;
     repeated temporal.api.workflow.v1.PendingNexusOperationInfo pending_nexus_operations = 7;
     temporal.api.workflow.v1.WorkflowExecutionExtendedInfo workflow_extended_info = 8;
+    temporal.api.workflow.v1.WorkflowPauseInfo workflow_pause_info = 9;
```
cretz, line 1045: How do I know if workflow tasks are paused (what I would consider a traditional concept of "workflow pause")? Does the simple presence/absence of this determine that? Or should we add a field? Does it get represented as a status in another enum somewhere?

temporal/api/workflow/v1/message.proto
spkane31: This is a list of activity types that are both currently paused and will be paused if new ones are scheduled. I don't think calling these rules/conditions/policies is totally accurate and could conflict with future namespace rules that could have pausing information. What are your thoughts of changing this to `repeated PausedActivities paused_activities = 1;` This would encompass both current and future paused activities and I'd update the comment to indicate that

temporal/api/workflowservice/v1/request_response.proto
spkane31: This work is only concerned with pausing activities, pausing workflow is a separate stream of work.

spkane31: > How do I know if workflow tasks are paused 

The `workflow_pause_info` is a list of all activity types that are paused.

> Does the simple presence/absence of this determine that? Or should we add a field? Does it get represented as a status in another enum somewhere?

I don't quite understand these questions, this is a new field that gives the paused activities. There is no other enum

temporal/api/workflow/v1/message.proto
cretz: Hrmm, they are not paused activities, they are "policies" or "rules" or whatever to pause future activities. I would expect a collection of paused activities to be a collection of activities that are paused, same as I would expect a collection pending activities to be a collection of activities that are pending. Can we think of any term that represents a setting to pause future activities since they are not actually activities?

temporal/api/workflowservice/v1/request_response.proto
cretz: >  pausing workflow is a separate stream of work

But this is called "workflow pause info" so I assume it will reuse this message for other pause things? I think we should consider "workflow pause" (and therefore "workflow pause info") the _same_ stream of work if we want a cohesive view of pausing workflows and things in it.

> I don't quite understand these questions, this is a new field that gives the paused activities. There is no other enum

The question is about the rest of workflow pause and to make sure we aren't myopically designing for only one form. If we expect task pause to just be a bool in pause info, that makes sense to me. I am a bit surprised we're doing in-workflow activity pausing before workflow task pausing. I just want to confirm what we're thinking task pause may look like within this new general-purpose structure that is not specific to activity pause (even if its only field is).

temporal/api/workflow/v1/message.proto
```diff
@@ -100,6 +100,29 @@ message WorkflowExecutionInfo {
     temporal.api.common.v1.Priority priority = 24;
 }
 
+// WorkflowPauseInfo contains the details of the request to pause the workflow.
+message WorkflowPauseInfo {
+    // List of activity types that are paused, any future activities of this type will automatically be put into PAUSED state.
+    repeated PausedActivities paused_activities = 1;
```
cretz, line 106: When trying to see how I would set this as a user, I can't find the workflow pause API, is this yet to be developed or is it as part of another API? To clarify, I am trying to figure out how to remove a "pause policy" I have previously set.

spkane31: https://github.com/temporalio/temporal/pull/8161/files I think you are looking for this PR. @gow should be able to provide more clarity here

cretz: I mean from a user/API POV how do I issue a current-and-future pause for activities and how to I remove that current-and-future pause? The reason I ask is because how undo occurs can help drive what this API looks like (e.g. if you need a unique identifier for this "setting"/"policy"). I think this API PR should go hand-in-hand with the API PR that is used to mutate this value.

cretz: Ok, I think I missed that this is implemented via `PauseActivity`? I still think the pause/unpause is a bit confusing on whether it applies to the future, but that's a separate topic. I think we're good on this thread specifically.

PR #632, opened 2025-08-26
add Manager field to WorkerDeployment, and SetWorkerDeploymentManager API

carlydf:

add Manager field to WorkerDeployment, and SetWorkerDeploymentManager API

I want the Worker Controller to stop making changes to a Worker Deployment if it detects that another client has written to the same Worker Deployment. This is so that, if a user calls `SetCurrentVersion` or `SetRampingVersion` from the CLI during an incident, the Controller does not overwrite that change.

`LastModifiedBy` is not enough for this feature, because we need a way for the "other client" (likely user CLI in most cases, maybe UI if we allow such changes through the UI some day) to indicate to the controller that manual changes are no longer needed and the controller can safely take ownership over writing to this Worker Deployment again.

I thought about doing this via a `Metadata` on the Deployment, similar to WorkerDeploymentVersion Metadata, but I can't think of any other use cases for Deployment-level metadata, and I think this is load-bearing enough to deserve it's own field, where we can clearly document semantics.

No

Not implemented in server yet because this is just a proposal

### DIFF
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -923,6 +923,19 @@ service WorkflowService {
         };
     }
 
+    // Set/unset the Manager of a Worker Deployment.
+    // Experimental. This API might significantly change or be removed in a future release.
+    rpc SetWorkerDeploymentManager (SetWorkerDeploymentManagerRequest) returns (SetWorkerDeploymentManagerResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/worker-deployments/{deployment_name}/set-manager"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/worker-deployments/{deployment_name}/set-manager"
+                body: "*"
+            }
+        };
+    }
+
     // Invokes the specified Update function on user Workflow code.
     rpc UpdateWorkflowExecution(UpdateWorkflowExecutionRequest) returns (UpdateWorkflowExecutionResponse) {
         option (google.api.http) = {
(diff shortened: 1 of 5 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -923,6 +923,19 @@ service WorkflowService {
         };
     }
 
+    // Set/unset the Owner of a Worker Deployment.
+    // Experimental. This API might significantly change or be removed in a future release.
+    rpc SetWorkerDeploymentOwner (SetWorkerDeploymentOwnerRequest) returns (SetWorkerDeploymentOwnerResponse) {
```
cretz: Just for the record, if we had it to do over again, I would have preferred a single `UpdateWorkerDeployment` so this could be atomically updated alongside everything else, and so we don't have to add a new RPC call for every future mutable thing we add to worker deployment. But can't change now.

carlydf:
@cretz I simplified the semantics a bit so that the user doesn't have to do an awkward `set-owner --unset` and so that the user does not think they have to find and copy-paste the exact identity of themselves or another client. Replaced with `ownership --take` and `ownership --release` see Notion doc for the full scenario, but I think this is much simpler for the user and streamlines them to use the recommended path. 

IMO we can always add `ownership --assign "specific-owner-id"` later if there is a use case for it.

I also removed the `--ignore-owner` flag from SetCurrent and SetRamping since I wouldn't recommend using it; if two users are in contention over an object, `--ignore-owner` will just lead to their changes being clobbered by the other. Users should explicitly take or release ownership and then make their change IMO.

pdoerner (approved):

Shivs11 (approved):

PR #633, opened 2025-08-26
Worker heartbeat: Remove heartbeat from non-nexus polling

yuandrew:

**What changed?**
Removed the `worker_heartbeat` field from non-nexus calls

**Why?**
We only want `worker_heartbeat` info sent on nexus polls and dedicated worker heartbeat requests.

**Breaking changes**
Breaking change for a WIP feature, should have no effect. Also tried grepping for this field in server and didn't find it used anywhere.

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -267,9 +267,6 @@ message PollWorkflowTaskQueueRequest {
     // Worker deployment options that user has set in the worker.
     // Experimental. Worker Deployments are experimental and might significantly change in the future.
     temporal.api.deployment.v1.WorkerDeploymentOptions deployment_options = 6;
-
-    // Worker info to be sent to the server.
-    temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 7;
 }
 
 message PollWorkflowTaskQueueResponse {
@@ -441,10 +438,6 @@ message PollActivityTaskQueueRequest {
     temporal.api.common.v1.WorkerVersionCapabilities worker_version_capabilities = 5 [deprecated = true];
     // Worker deployment options that user has set in the worker.
     temporal.api.deployment.v1.WorkerDeploymentOptions deployment_options = 6;
-
-    // Worker info to be sent to the server.
-    temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 7;
-
 }
 
 message PollActivityTaskQueueResponse {
@@ -1007,8 +1000,6 @@ message ShutdownWorkerRequest {
     string sticky_task_queue = 2;
     string identity = 3;
     string reason = 4;
-
-    temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 5;
 }
 
 message ShutdownWorkerResponse {


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1007,8 +1004,8 @@ message ShutdownWorkerRequest {
     string sticky_task_queue = 2;
     string identity = 3;
     string reason = 4;
-
-    temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 5;
+    reserved "worker_heartbeat";
+    reserved 5;
```
Sushisource: Hmm... this one's more interesting and we may need to keep. Since we do (will?) have this worker status field, we're going to stop polling on the nexus queue while shutting down... so we need to be able to send the final (or, I guess, semi-final "shutting down") status on this rpc I think.

Sushisource (commented):
You don't even need the "reserved" fields since this was never actually done in server.

temporal/api/workflowservice/v1/request_response.proto
yuandrew: > so we need to be able to send the final status on this rpc

Is having the worker's state on shutdown that important? I'm not against keeping this, especially since this doesn't really cost anything, but I wanna better understand the benefit. Having heartbeat info included on shutdown seems like it's slightly better than just having a worker's info disappear on the next nexus poll, but only if server uses that shutdown heartbeat info?

> we're going to stop polling on the nexus queue while shutting down

Yes, but just to be clear, only when we shutdown the last worker on the namespace

yuandrew:
Without the `reserved` fields I got `make`/CI errors like 
```
temporal/api/workflowservice/v1/request_response.proto:254:1:Previously present field "7" with name "worker_heartbeat" on message "PollWorkflowTaskQueueRequest" was deleted without reserving the name "worker_heartbeat".
temporal/api/workflowservice/v1/request_response.proto:254:1:Previously present field "7" with name "worker_heartbeat" on message "PollWorkflowTaskQueueRequest" was deleted without reserving the number "7".
```

Is there a way to get around that?

Sushisource:
> Without the `reserved` fields I got `make`/CI errors like
> 
> ```
> temporal/api/workflowservice/v1/request_response.proto:254:1:Previously present field "7" with name "worker_heartbeat" on message "PollWorkflowTaskQueueRequest" was deleted without reserving the name "worker_heartbeat".
> temporal/api/workflowservice/v1/request_response.proto:254:1:Previously present field "7" with name "worker_heartbeat" on message "PollWorkflowTaskQueueRequest" was deleted without reserving the number "7".
> ```
> 
> Is there a way to get around that?

Yeah, you need to temporarily disable the breaking change check, then reinstate it in a subsequent PR. Ex: https://github.com/temporalio/api/pull/608/files#diff-1a5ba9cba93e971f532139f694d7da802776bfe578e3f753b9c3f25968dbf42dL16

temporal/api/workflowservice/v1/request_response.proto
Sushisource: Yeah, I think it's important in that we can differentiate "went missing" vs "shut down properly". The former we can alert on and the latter is safe.

However, to your point, this really only matters for the "last" worker to turn off, as the others can send it with the overall poll. That said, it's still important, since I think in 90%+ of cases, there's only going to be one worker total.

Sushisource (approved):
LGTM

You'll probably want to save a note somewhere that we'll want to test this shutdown case.

PR #635, opened 2025-08-27
Add RESOURCE_EXHAUSTED_CAUSE_GRPC_MESSAGE_TOO_LARGE

maciejdudko:
**What changed?**
Added `RESOURCE_EXHAUSTED_CAUSE_GRPC_MESSAGE_TOO_LARGE` enum value to `ResourceExhaustedCause`.

**Why?**
Go SDK recently merged PR for not retrying gRPC-message-too-large errors: https://github.com/temporalio/sdk-go/pull/2026

The implementation detects this error situation and sets the details of the gRPC status object to an instance of `ResourceExhaustedFailure` with `cause` set to a specific value so that the information is preserved by the error converter. However, no existing `ResourceExhaustedCause` value was appropriate here, so a new one was added, currently as a non-standard extension.

This PR upstreams the new enum value so that the non-standard extension in Go SDK can be removed.

**Breaking changes**
No

### DIFF
--- a/temporal/api/enums/v1/failed_cause.proto
+++ b/temporal/api/enums/v1/failed_cause.proto
@@ -123,6 +123,8 @@ enum ResourceExhaustedCause {
     RESOURCE_EXHAUSTED_CAUSE_CIRCUIT_BREAKER_OPEN = 8;
     // Namespace exceeds operations rate limit.
     RESOURCE_EXHAUSTED_CAUSE_OPS_LIMIT = 9;
+    // GRPC message was too large.
+    RESOURCE_EXHAUSTED_CAUSE_GRPC_MESSAGE_TOO_LARGE = 10;
 }
 
 enum ResourceExhaustedScope {


Quinn-With-Two-Ns:
In the Go SDK can we just wrap the exception rather then add a new cause that will only be used by the SDK?

Quinn-With-Two-Ns:
>Meaning we always return a service error for all client calls.

That isn't strictly true, the SDK may wrap error between the grpc call and the the high level Temporal client although it isn't commonly done. In Go wrapping errors is generally a pretty safe change since Go's standard library will deal with unwrapping them for the users. 

You could also just check the message contents no? I understand that is a bit of effort for every error. 

Not a big issue, If you and @yuandrew are okay with the current approach then I will disagree and commit

temporal/api/enums/v1/failed_cause.proto
```diff
@@ -123,6 +123,8 @@ enum ResourceExhaustedCause {
     RESOURCE_EXHAUSTED_CAUSE_CIRCUIT_BREAKER_OPEN = 8;
     // Namespace exceeds operations rate limit.
     RESOURCE_EXHAUSTED_CAUSE_OPS_LIMIT = 9;
+    // GRPC message was too large.
+    RESOURCE_EXHAUSTED_CAUSE_GRPC_MESSAGE_TOO_LARGE = 10;
```
cretz, line 127: This is meant to be a server-sent enumerate for use in `temporal.api.errordetails.v1.ResourceExhaustedFailure` (and converted via https://pkg.go.dev/go.temporal.io/api/serviceerror#FromStatus on client side). But servers can't intercept gRPC message too large to ever set this enum, so I don't think it should be defined in the server/client API because it's not part of the server/client API. If client-side Go needs specific things, we can do that in client-side Go only.

Sushisource: Yeah, when we talked about this item in planning I was thinking it was adding it here: https://github.com/temporalio/api/blob/2e5751c82c13013ec7c486efb36969e56a116fe5/temporal/api/enums/v1/failed_cause.proto#L83

But, in fact, that already exists.

So not sure we really need to do anything here?

maciejdudko:
We decided to implement another approach which doesn't require API changes.

PR #637, opened 2025-09-10
Worker Heartbeat: Use Worker Grouping instead of Runtime for Key in nexus TQ

yuandrew:

**What changed?**
use `worker_grouping_key` instead of `runtime_key` for worker command nexus task queue

**Why?**
There can be multiple clients (i.e. with different auths) within the same runtime/namespace, so we want a more specific category for these nexus worker command requests

**Breaking changes**
No, WIP feature

**Server PR**

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -16309,7 +16309,7 @@
         },
         "workerGroupingKey": {
           "type": "string",
-          "title": "Worker grouping identifier. This id should be unique for all worker sets (grouped\nby clients they share) running workers in the namespace, and should be shared by all\nworkers in the same client+namespace+process.\nThis will be used to build the worker command nexus task queue name:\n\"temporal-sys/worker-commands/{worker_grouping_key}\""
+          "title": "Worker grouping identifier. A key to group workers that share the same client+namespace+process.\nThis will be used to build the worker command nexus task queue name:\n\"temporal-sys/worker-commands/{worker_grouping_key}\""
         },
         "processId": {
           "type": "string",
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -54,9 +54,7 @@ message WorkerHostInfo {
   // Worker host identifier.
   string host_name = 1;
 
-  // Worker grouping identifier. This id should be unique for all worker sets (grouped
-  // by clients they share) running workers in the namespace, and should be shared by all
-  // workers in the same client+namespace+process.
+  // Worker grouping identifier. A key to group workers that share the same client+namespace+process.
   // This will be used to build the worker command nexus task queue name:
   // "temporal-sys/worker-commands/{worker_grouping_key}"
   string worker_grouping_key = 5;
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -13651,9 +13651,7 @@ components:
         workerGroupingKey:
           type: string
           description: |-
-            Worker grouping identifier. This id should be unique for all worker sets (grouped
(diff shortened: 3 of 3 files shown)


temporal/api/worker/v1/message.proto
```diff
@@ -54,15 +54,14 @@ message WorkerHostInfo {
   // Worker host identifier.
   string host_name = 1;
 
-
-  // Worker process identifier. This id should be unique for all _processes_
-  // running workers in the namespace, and should be shared by all workers
-  // in the same process.
+  // Worker process identifier. This id should be unique for all worker sets, or
```
```suggestion
  // Worker set identifier. This id should be unique for all worker sets (grouped
  // by clients they share) running workers in the namespace, and should be shared by all
  // workers in the same worker set.
```

openapi/openapiv2.json
```diff
@@ -16272,13 +16272,13 @@
           "type": "string",
           "description": "Worker host identifier."
         },
-        "processKey": {
+        "workerSetKey": {
           "type": "string",
-          "title": "Worker process identifier. This id should be unique for all _processes_\nrunning workers in the namespace, and should be shared by all workers\nin the same process.\nThis will be used to build the worker command nexus task queue name:\n\"temporal-sys/worker-commands/{process_key}\""
+          "title": "Worker process identifier. This id should be unique for all worker sets, or\nclients, running workers in the namespace, and should be shared by all\nworkers in the same worker set.\nThis will be used to build the worker command nexus task queue name:\n\"temporal-sys/worker-commands/{worker_set_key}\""
```
rkannan82: Why does this say worker process identifier?

Also, please help me understand the following. I am not able to correlate the PR description with this comment. For example, the description talks about runtime key and how clients with different auth need to be categorized. But here, it sounds like all workers running on the same host will have the id?

yuandrew: Good callout, I should update this as well to `Worker Set Identifier`. We previously used a worker process identifier each separate process. But I later discovered we can have multiple clients with different auth settings (i.e. self host & cloud) within the same process, so we need a more granular key. 

In the scenario where we have 2 different clients being used in the same process, we would need 2 separate nexus task queues to handle heartbeating

yuandrew: In terms of "runtime", we were previously using that synonymous to process, but that's also technically not true SDK side, not sure why anyone would want to have multiple runtimes in the same process, but they technically could, I think

rkannan82: Is this an already agreed upon name, or are we open to suggestions ? 😄 
If so, I would suggest calling it as a workerGroupingKey. I feel the intent is more direct by calling it a grouping-key than a set-key.

- A host can have multiple processes.
- Each process can have 1 or more workers.
- Workers within a process can be grouped together using a grouping key.
IIUC, the SDK would construct this automatically based on the worker attributes that the user has defined (example, auth settings).

Question: How is this key generated?

yuandrew: Discussed offline, we agreed on `workerGroupingKey` here

temporal/api/worker/v1/message.proto
```diff
@@ -54,15 +54,14 @@ message WorkerHostInfo {
   // Worker host identifier.
   string host_name = 1;
 
-
-  // Worker process identifier. This id should be unique for all _processes_
-  // running workers in the namespace, and should be shared by all workers
-  // in the same process.
+  // Worker grouping identifier. This id should be unique for all worker sets (grouped
```
rkannan82: The comment is still using worker sets. Also
For clarity, suggest revising this comment along these lines ... Please apply this to the yaml and json as well.

A key to group workers that share the same client within a process. This is used to build the nexus task queue name: ...

PR #638, opened 2025-09-15
Update description of Visibility Scan API

rodrigozhou:

**What changed?**
Update description of Visibility Scan API

**Why?**
Visibility Scan API has been marked as deprecated since Temporal v1.27.0.
Temporal Server is changing to simply call the `List` API, so updating the description so it is clear.

**Breaking changes**

**Server PR**
https://github.com/temporalio/temporal/pull/8328

### DIFF
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -458,7 +458,8 @@ service WorkflowService {
         };
     }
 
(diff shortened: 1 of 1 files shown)


bergundy:
I don't think we have ever removed a server API, this will in turn break the Go SDK. You should just mark it as deprecated forever IMHO.

cretz (approved):

PR #639, opened 2025-09-16
Initial Proposal: Make Worker-Deployment API's async in nature.

Shivs11:

- The main aim of this PR is to add an enum in the RoutingConfig.

- To allow users to have a peek at the exact stage where an update, scheduled by a worker-versioning write API, has reached.

- None

**Server PR**
- WIP right now.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -14111,9 +14111,10 @@
           "format": "date-time",
           "description": "Last time ramping version percentage was changed.\nIf ramping version is changed, this is also updated, even if the percentage stays the same."
         },
-        "routingConfigUpdateState": {
-          "$ref": "#/definitions/v1RoutingConfigUpdateState",
-          "description": "Indicates whether the routing_config has been fully propagated to all\nrelevant task queues and their partitions."
+        "revisionNumber": {
+          "type": "string",
+          "format": "int64",
+          "description": "Monotonically increasing value which is incremented on every mutation \nto any field of this message to achieve eventual consistency between task queues and their partitions."
         }
       }
     },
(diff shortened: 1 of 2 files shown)


temporal/api/enums/v1/task_queue.proto
```diff
@@ -100,3 +100,15 @@ enum RateLimitSource {
...
+// Indicates whether a change to the Routing Config has been
+// propagated to all relevant Task Queues and their partitions.
+enum RoutingConfigUpdateState {
+    ROUTING_CONFIG_UPDATE_STATE_UNSPECIFIED = 0;
+    // Update to the RoutingConfig is currently in progress.
+    ROUTING_CONFIG_UPDATE_STATE_IN_PROGRESS = 1;
+    // Update to the RoutingConfig has completed successfully.
+    ROUTING_CONFIG_UPDATE_STATE_COMPLETED = 2;
+    // Update to the RoutingConfig has failed.
+    ROUTING_CONFIG_UPDATE_STATE_FAILED = 3;
```
carlydf, line 113: if we do keep an internal timeout (such as 10m or 5m) and the update times out, would that cause it to be in failed state?

Shivs11: Yes

temporal/api/deployment/v1/message.proto
```diff
@@ -285,4 +285,7 @@ message RoutingConfig {
     // Last time ramping version percentage was changed.
     // If ramping version is changed, this is also updated, even if the percentage stays the same.
     google.protobuf.Timestamp ramping_version_percentage_changed_time = 6;
+    // Indicates whether the routing_config has been fully propagated to all
+    // relevant task queues and their partitions.
+    temporal.api.enums.v1.RoutingConfigUpdateState routing_config_update_state = 10;
```
ShahabT: thinking if this should be in the `WorkerDeploymentInfo` itself because we want to store `RoutingConfig` in TQs and it doesn't make sense for it to have this status there.

temporal/api/deployment/v1/message.proto
ShahabT, line 261: do you mind adding the revision number in this too? something like

```
    // Incremented every time time other fields of this struct change.
    int64 revision_number = 10;
```

PR #640, opened 2025-09-17
[DO NOT MERGE] Add standalone activity APIs

bergundy:
**What changed?**

Added first revision of APIs for standalone activities.

Missing API support for:

- [ ] PauseActivity
- [ ] UnpauseActivity
- [ ] ResetActivity
- [ ] UpdateActivityOptions

**Notes for reviewers**

- See the open TODOs, some decisions need to be finalized.
- Do we want to have an activity specific ID reuse policy and conflict policy or can we make this generic?
  - There are already other cases planned where we would want to use the same definitions.
- Do we need a generic form of `ActivityExecution`? What would it be called? `EntityExecution`?
- Does pause info need to be generic?
- How does this integrate with versioning? Can the client specify any versioning constraints on the `StartActivityExecutionRequest`?
- Do we want to use `ActivityOptions` in `StartActivityExecutionRequest`? Would more of the fields in the request be considered options in the future?

**Server PR**

Not done yet, this will be the next step.


temporal/api/enums/v1/activity.proto
```diff
@@ -0,0 +1,40 @@
...
+// The status is updated once, when the activity is originally scheduled, and again when the activity reaches a terminal
+// status.
+// TODO: Should this be a common execution status? Seems like the other archetypes will share this status.
+// (-- api-linter: core::0216::synonyms=disabled
+//     aip.dev/not-precedent: Named consistently with WorkflowExecutionStatus. --)
+enum ActivityExecutionStatus {
+    ACTIVITY_EXECUTION_STATUS_UNSPECIFIED = 0;
+    // The activity is not in a terminal status. This does not necessarily mean that there is a currently running
+    // attempt. The activity may be backing off between attempts or waiting for a worker to pick it up.
+    ACTIVITY_EXECUTION_STATUS_RUNNING = 1;
```
bergundy, line 22: We will likely add a `PAUSED` status here since this is the direction workflow pause is going.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+message GetActivityExecutionResultResponse {
+    // The run ID of the completed activity, may be used in case a run ID was not specified in the request.
+    string run_id = 1;
+
+    oneof outcome {
+        // The result if the activity completed successfully.
+        temporal.api.common.v1.Payloads result = 2;
+        // The failure if the activity completed unsuccessfully.
+        temporal.api.failure.v1.Failure failure = 3;
+    }
```
dandavison: Maybe worth comparing the comments in [`PollWorkflowExecutionUpdateResponse`](https://github.com/temporalio/api/blob/standalone-activity/temporal/api/workflowservice/v1/request_response.proto#L1777-L1797). When we introduced that there was some discussion, leading to a decision to use a non-error response to signify that a long-polling client should retry.

temporal/api/activity/v1/message.proto
```diff
@@ -9,10 +9,27 @@ option java_outer_classname = "MessageProto";
...
+import "temporal/api/enums/v1/activity.proto";
+import "temporal/api/enums/v1/workflow.proto";
+import "temporal/api/failure/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
+import "temporal/api/sdk/v1/user_metadata.proto";
 
-import "google/protobuf/duration.proto";
+// When StartActivityExecution uses the ID_CONFLICT_POLICY_USE_EXISTING and there is already an existing running
+// activity, OnConflictOptions defines actions to be taken on the existing running activity, updating its state.
+message OnConflictOptions {
```
cretz: Would be ok not doing this for the first version. If we are to keep this, the comments above the fields need to be updated to not refer to workflows.

temporal/api/activity/v1/message.proto
```diff
@@ -40,6 +57,143 @@ message ActivityOptions {
...
+            // The activity was paused by direct API invocation.
+            Manual manual = 2;
+        }
+    }
+
+    PauseInfo pause_info = 30;
+}
+
+// Limited activity information returned in the list response.
+message ActivityListInfo {
```
cretz: To be consistent with workflow, _this_ one should be called `ActivityExecutionInfo` and we can add `ActivityExecutionExtendedInfo` or something if you need things that are not in visibility. I cover this in a later comment.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+    // Token returned in ListActivityExecutionsResponse.
+    bytes next_page_token = 3;
+    // Visibility query, see https://docs.temporal.io/list-filter for the syntax.
+    string query = 4;
+}
+
+message ListActivityExecutionsResponse {
+    repeated temporal.api.activity.v1.ActivityListInfo executions = 1;
+    // Token to use to fetch the next page. If empty, there is no next page.
+    bytes next_page_token = 2;
```
cretz, line 2798: To match workflow, describe should be a _superset_ of list response, not a completely different structure. This makes sense with many of our SDKs where describe returns a class that extends what list returns. IMO this should be (ignoring long poll stuff):

```proto
message DescribeActivityExecutionResponse {
    temporal.api.activity.v1.ActivityExecutionInfo info = 1;
    temporal.api.activity.v1.ActivityExecutionExtendedInfo extended_info = 2;

    // ...
}

// ...

message ListActivityExecutionsResponse {
    repeated temporal.api.activity.v1.ActivityExecutionInfo executions = 1;
    // ...
```

If there are some things that don't feel like they overlap, e.g. task queue or other activity options on the shared info only represent what it was started with, then those fields can be named/documented appropriately. It makes no sense to get more information from list. Describe should always have at least what visibility does.

temporal/api/enums/v1/activity.proto
cretz: I don't think either are a wise direction. What is happening between "running" and "done" are not traditionally execution statuses. Are you going to have a "backing off" status too? Execution status has traditionally been "running or done", and this changes those expectations unnecessarily. Can you use a separate enum to provide more detailed information about the state the running activity/workflow is in and leave the execution status alone?

temporal/api/enums/v1/id.proto
```diff
@@ -0,0 +1,40 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "IdProto";
+option ruby_package = "Temporalio::Api::Enums::V1";
+option csharp_namespace = "Temporalio.Api.Enums.V1";
+
+// Defines whether to allow re-using an ID from a previously *closed* execution.
+// If the request is denied, the server returns an `ExecutionAlreadyStarted` error.
+//
+// See `IdConflictPolicy` for handling ID duplication with a *running* execution.
+enum IdReusePolicy {
```
cretz: This doesn't seem general purpose, it seems each top-level Temporal component has its own ID reuse/conflict policy enumerates. Shouldn't this be `ActivityIdReusePolicy` and `ActivityIdConflictPolicy`? I think this actually may have value if/when we need something different for one component or the other. From an SDK POV, I would expect these to be separately named enums too because we like consistency.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+    // Header for context propagation and tracing purposes.
+    temporal.api.common.v1.Header header = 12;
+    // Request to get the first activity task inline in the response bypassing matching service and worker polling.
+    // If set to `true` the caller is expected to have a worker available and capable of processing the task.
+    // The returned task will be marked as started and is expected to be completed by the specified
+    // `schedule_to_close_timeout`.
+    bool request_eager_execution = 13;
+    // Callbacks to be called by the server when this activity reaches a terminal status.
+    // Callback addresses must be whitelisted in the server's dynamic configuration.
+    repeated temporal.api.common.v1.Callback completion_callbacks = 14;
```
cretz: I would be ok not putting this in the first version if it adds extra server effort and can be done later

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+message StartActivityExecutionResponse {
+    // The run ID of the activity that was started - or used (via ID_CONFLICT_POLICY_USE_EXISTING).
+    string run_id = 1;
+    // If true, a new activity was started.
+    bool started = 2;
+    // When `request_eager_execution` is set on the `StartActivityExecutionRequest`, the server will return the first
+    // activity task to be eagerly executed.
+    // The caller is expected to have a worker available to process the task.
+    PollActivityTaskQueueResponse eager_task = 3;
+    // Link to the workflow event.
```
cretz: Is this copy/paste docs issue? Seen a few in the PR, may be worth revisiting the comments.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+}
+
+message DescribeActivityExecutionRequest {
+    string namespace = 1;
+    string activity_id = 2;
+    // Activity run ID, targets the latest run if run_id is empty.
+    string run_id = 3;
+
+    // If true, the activity input is returned in the response.
+    bool include_input = 4;
```
cretz: Why add this new feature to activities before we add it to workflows? It seems like a new Temporal feature to request input on describe.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+
+    // If not empty, turns this request into a long poll that is unblocked when the activity state changes from the time
+    // the token was returned.
+    // This token is returned as part of the `DescribeActivityExecutionResponse`.
+    bytes long_poll_token = 5;
```
cretz: One wonders if there is value in "activity event" ledger. People are going to long poll and wonder why they can't get updated for each attempt that has an issue.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+message ListActivityExecutionsResponse {
+    repeated temporal.api.activity.v1.ActivityListInfo executions = 1;
+    // Token to use to fetch the next page. If empty, there is no next page.
+    bytes next_page_token = 2;
+}
+
+message CountActivityExecutionsRequest {
+    string namespace = 1;
+    // Visibility query, see https://docs.temporal.io/list-filter for the syntax.
+    string query = 2;
```
cretz, line 2804: Are the built-in search attributes for activities still TBD?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+
+message GetActivityExecutionResultRequest {
+    string namespace = 1;
+    string activity_id = 2;
+    // Activity run ID, targets the latest run if run_id is empty.
+    string run_id = 3;
+    // If set, turns this request into a long poll that is unblocked when the activity reaches a terminal status.
+    // The wait duration is capped by the request's context deadline or by the maximum enforced long poll interval
+    // allowed by the server.
+    bool wait = 4;
```
cretz: This is a bit confusing from a UI/CLI POV. So if I want to wait on an update for _anything but result_ I have to use one call, and to wait on result, I have to use this call separately? For workflows, interfaces can use the same call to wait on any kind of update, whether its result or events or whatever. I think we should consider doing the same for activities. I can give some suggestions on ideas if we don't want a full activity ledger, but many boil down to having one RPC that can opt-in to whether it cares about only result or any update, and that doesn't have to have been called first before it can be long polled. This can be mashed into describe if you must, but I like describe to be a fast synchronous call for all components.

I recommend something like:

```proto
message GetActivityExecutionStatusRequest {
    string namespace = 1;
    string activity_id = 2;
    string run_id = 3;

    bool include_info = 4;
    bool wait_info = 5; // Cannot be true if include_info is false
    bytes continue_info_wait_token = 6; // Cannot have a value if either of the above two are false

    bool include_outcome = 7;
    bool wait_outcome = 8; // Cannot be true if include_outcome is false
}

message GetActivityExecutionStatusRequest {
    // These three only set if include_info is true (and if it's true, they are _always_ all set)
    temporal.api.activity.v1.ActivityExecutionInfo current_info = 1;
    temporal.api.activity.v1.ActivityExecutionExtendedInfo current_extended_info = 2;
    bytes continue_info_wait_token = 3;

    // This is only set if is activity completed and include_outcome is true
    oneof result {
        temporal.api.common.v1.Payloads result = 4;
        temporal.api.failure.v1.Failure failure = 5;
    }
}
```

This is important to reduce interface polling calls that need to know _both_ of these things. I could see an SDK however splitting these into separate calls. It also allows us to keep describe as a simple synchronous getter like it is everywhere else.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2593,3 +2607,364 @@ message DescribeWorkerRequest {
...
+    string request_id = 5;
+    // Reason for requesting the cancellation, recorded and available via the DescribeActivityExecution API.
+    // Not propagated to a worker if an activity attempt is currently running.
+    string reason = 6;
+}
+
+message RequestCancelActivityExecutionResponse {
+}
+
+message TerminateActivityExecutionRequest {
```
cretz, line 2842: We're a bit inconsistent on whether our side-effecting calls need a request ID for dedupe. I understand that a terminate is final and therefore a successive call is "safe", but, a user could see an error for the second terminate call I assume. We do not need to solve in this PR since it's a general problem, but IMO we should see if we can make a policy of "request ID on all side-effecting calls" though I understand it causes server side effort/persistence to dedupe.

cretz, line 2842: Just want to note that it is a bit confusing seeing `PauseActivityExecution` work for multiple forms of activity, but `TerminateActivityExecution` (and cancel and get result and describe and others) only work for one kind of activity. In the SDK we're going to have to be careful to structure it in a way that is less inconsistent than our API has chosen to structure it here. I do wish we could be clearer in the naming and/or have a pathway to really be generic in these concepts.

temporal/api/enums/v1/activity.proto
bergundy: That's what we are doing for workflow pause, we should be consistent for any type of execution.

temporal/api/activity/v1/message.proto
bergundy: Yeah, this will only be added when we do the Nexus integration. I put it here for completeness.

temporal/api/activity/v1/message.proto
bergundy: I think this was not done correctly for workflows.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I don't think the consistency here is important TBH. Using the same info types for workflow list and describe was an oversight that we recognize today. I would be open to changing this if message reuse simplifies anything significantly on the SDK side.

temporal/api/enums/v1/id.proto
bergundy: I already know that I want this for "standalone" Nexus operations, and generally these reuse and conflict policies are framework capabilities. I am already at the point of wanting to generalize this and not having to redefine for every new primitive.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Yeah, I put it here for completeness. It won't make it into the first release.

bergundy: I will remove these before merging.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Yup, I will review the comments. Thanks for catching it.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Just because there is no other way of getting updates when the activity is running, unlike workflow where you can wait for history events. But I 100% agree that we want this for workflows too, we're just focusing on standalone activities right now.

temporal/api/workflowservice/v1/request_response.proto
bergundy: We discussed this and decided not to add it for now.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Not fully fleshed out yet. I believe this will be the rough set of attributes:

```
ActivityId
RunId
ActivityType
TaskQueue
StartTime
ExecutionTime
CloseTime
ExecutionStatus
ExecutionDuration
PauseInfo -- when we add pause support
```

temporal/api/workflowservice/v1/request_response.proto
bergundy: I like this idea. Let me take a stab at it. This just means that we won't have `Describe` at all but I'm fine with that.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I copied this from workflow terminate, I wonder if it's worth having consistency there or not. It's not much trouble to store the terminate request ID and return an error if the activity is already terminated if the request ID does not match.

bergundy: This is intentional. We don't want to add these APIs for workflow activities yet but they are structured in a way that it would be possible to add `workflow_id` later to operate on a workflow activity.

temporal/api/enums/v1/activity.proto
cretz: Right, I am saying we shouldn't do it for workflow pause either (and have mentioned it in those situations too). It's the same reason that "backing off" doesn't need to be a status. Sure it is effectively paused during backoff, but that doesn't affect that Temporal defines activities and workflows as running if they are not completed.

There are many many cases where the expectations would be violated. For instance, people expect ID reuse policy to only apply to not-running workflows, and ID conflict policy to only apply to running workflows. This violates both of those. People expect `ExecutionStatus != Running` query to give them all completed workflows. I could come up with many more.

temporal/api/activity/v1/message.proto
cretz: I disagree, and even if I did agree, there's no harm in being consistent here and making list a subset of describe

temporal/api/workflowservice/v1/request_response.proto
cretz: I don't think it was an oversight. Do we believe that describe data is a superset of the list data? If so, we need to model it that way. The SDKs (that have types for list and describe) are modeled that way _very_ intentionally (via inheritance or similar) because it makes perfect sense that visibility is a subset of the data.

temporal/api/enums/v1/id.proto
cretz: I think 3 sets of enums here for their distinct uses is better than being inconsistent (even if they happen to have the same values so far). How do you expect the SDKs to model this? To be inconsistent with themselves, or are SDKs held to a higher standard of consistency?

temporal/api/workflowservice/v1/request_response.proto
cretz: Not following. Are you saying activity input can be updated? This comment about including the original starting input in describe response right?

temporal/api/workflowservice/v1/request_response.proto
cretz: We should absolutely have describe at least from a user POV. I understand the API is not held to a very high standard of consistency/clarity, but our SDKs are, so I guess we can reuse this API call and hide our unfortunate lack of API consistency.

temporal/api/workflowservice/v1/request_response.proto
cretz: I think we can tackle the "idempotency keys for _all_ side-effecting requests" requirement/project separately, but would definitely be nice to have.

cretz: > yet

If there are plans to, that makes sense then at least

temporal/api/activity/v1/message.proto
bergundy: I don't think the consistency argument is that strong here. We have seen multiple times that we assume something needs to only be in describe or list and realize we actually want it in both. The two separate messages don't allow for repurposing the fields without deprecation or duplication.

temporal/api/enums/v1/activity.proto
bergundy: I think it'd really nice to be able to see paused executions and their counts in the main list view. I don't think "backing off" is the same as pause, we don't update visibility and don't have enough of an incentive to do that for intermediate transient states.

In any case, we shouldn't be having this debate here. Whatever we decide to do for workflows we will copy to activities. For now, we are not even implementing standalone activity pause. That will likely come in the second phase of this project.

temporal/api/enums/v1/id.proto
bergundy: I think we're going to get tired of redefining these enums everywhere, and for the most part, SDKs can alias types for the workflow flavors of these enums.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Ah, I thought you were referring to the ability to long poll.
The UI and CLI would include the input only on the first request, and from then on, when long polling, assume it cannot be updated and turn this flag off.
There's no other way of getting the input back and it's required for UIs.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Commented on this above.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Also adding `StateTransitionCount`. I think that may be the list we end up going with.

temporal/api/workflowservice/v1/request_response.proto
bergundy: I just don't see a good enough reason to duplicate the APIs apart from the consistency argument, which I don't think is super important. Workflows have both history and mutable state, activities only have mutable state. You don't need two APIs here.

temporal/api/workflowservice/v1/request_response.proto
bergundy: It's inconclusive, but I didn't think we'd need to support describe for workflow activities initially either. It turns out that if we have list, we need describe. Just don't want to get into a corner where it's harder to add it later.

bergundy: Can you think of other requests that we would want to do this with? Probably pause, unpause, update options...

temporal/api/activity/v1/message.proto
dandavison: I'm thinking that `ActivityExecutionListInfo` is a more consistent name for this. The underlying entity in both cases is an activity execution; it's the amount of info that differs. Thus

- `ActivityExecutionInfo`: full info for an activity execution, returned by `describe`
- `ActivityExecutionListInfo`: reduced info for an activity execution, returned by `list`

temporal/api/enums/v1/activity.proto
cretz: > I think it'd really nice to be able to see paused executions and their counts in the main list view

Sure, but that doesn't mean affecting existing fields/concepts like status that have a clear defined expectation and always have

> In any case, we shouldn't be having this debate here

👍

temporal/api/activity/v1/message.proto
cretz: > I don't think the consistency argument is that strong here

It becomes one when the SDK gets involved. Do you agree we expect consistency there with regards to activity listing and workflow listing? If not, then we need to discuss what the SDK results look like for activity. And if so, we should really hold our API to similar standards and not let the API just come inconsistent because we don't feel 

> We have seen multiple times that we assume something needs to only be in describe or list and realize we actually want it in both.

Duplicating fields is not the best way to achieve this. It's actually the opposite because you'll forget things in describe that you have in list.

Do you agree if it is list it should be in describe? If so, how are you keeping those in sync? How are you going to answer a user that asks "why do I have to use the list API to get the task queue it started with"? Why can't we just make sure that describe includes what list has? This is easiest and most consistent by having the same message for both and describe having the extra data.

> I'm thinking that ActivityExecutionListInfo is a more consistent name for this

It's actually less consistent, what I suggested for both name and how it is composed is more consistent with workflow listing.

> ActivityExecutionListInfo: reduced info for an activity execution, returned by list

This is _not_ the case in this PR right now, this is what I am advocating for in this PR. Today there are things in list that aren't in describe. You should want describe to have everything in list, which is why the describe result needs to return _at least_ what the list result returns. Using the same message as list as part of the describe result is a good way to achieve that (not to mention it is consistent).

I am now fairly convinced we need to design the SDK APIs to understand what the server APIs need to be. I was hoping this wasn't going to be needed.

temporal/api/workflowservice/v1/request_response.proto
cretz: Can we at least agree that describe data is a superset of the list data?

temporal/api/enums/v1/id.proto
cretz: We are already redefining them in other user-facing places. SDKs can not always alias "workflow flavors", not every language works like that. It is important to see this on the whole and understand all user interfaces, not assume what user-facing aspects can do. We're not going to get tired of these two enums on a situational basis any more than we are with cancellation types or other things where we have done this. The SDKs will at least remain as consistent as possible, even if the API sadly continues to choose not to.

I am now growing convinced we need to see user-facing SDK designs before we can understand the APIs needed for them.

temporal/api/workflowservice/v1/request_response.proto
cretz: > There's no other way of getting the input back and it's required for UIs.

If we are allowing users to get input via describe, do you believe we should consistently do it for workflows?

temporal/api/workflowservice/v1/request_response.proto
cretz: > I just don't see a good enough reason to duplicate the APIs apart from the consistency argument

If you discount the good enough reason, then yes it will be hard to see a good enough reason. I was referring to the SDK at least being consisten.

I am now convinced this API cannot be designed well without designing the SDK APIs that will use it. SDK will have a synchronous describe and it sounds like API will not, so we need to see how it fits.

temporal/api/workflowservice/v1/request_response.proto
cretz: Anything side-effecting tbh (which includes things that have no user-facing side effects, like queries, because they technically do things), but can be opened as a separate issue

temporal/api/activity/v1/message.proto
dandavison: > This is easiest and most consistent by having the same message for both and describe having the extra data.

@cretz one of the points @bergundy is making is the following: suppose we start off with a field in describe output only, and then later need to expose it in visibility. With the struct composition design that field is now going to be duplicated _within_ the describe output. That is bad. But with independent/decoupled structs, this problem does not exist.

> let the API just come inconsistent

We're explicitly considering sacrificing consistency with workflow in order to achieve a design that avoids this that can be used consistently for Standalone Activity, Nexus, and future CHASM archetypes.

> Do you agree if it is list it should be in describe? ... "why do I have to use the list API to get the task queue it started with"?

Yes, list should be a subset of describe. I assume the omission of task queue is an oversight.

> It's actually the opposite because you'll forget things in describe that you have in list.

But forgetting a field and then adding it is backwards-compatible and not a problem.


> It's actually less consistent

My comment was referring to internal consistency regarding Activity, not consistency with Workflow. The comment is accurate interpreted with its intended scope: it's more consistent to name them both with the "ActivityExecution" prefix, than to have one use just "Activity" as the prefix.

temporal/api/workflowservice/v1/request_response.proto
dandavison: >  when the activity state changes

Let's define "activity state" in the comment here. (What is the intended definition here?)

temporal/api/workflowservice/v1/request_response.proto
dandavison: > Can we at least agree that describe data is a superset of the list data?

I agree with that.

temporal/api/activity/v1/message.proto
cretz: > one of the points @bergundy is making is the following: suppose we start off with a field in describe output only, and then later need to expose it in visibility. With the struct composition design that field is now going to be duplicated within the describe output. That is bad. But with independent/decoupled structs, this problem does not exist.

This is too small of a concern to completely upend model reuse for everyone. This can be used as justification to never use a model in more than one place. We could just inline/copy all fields everywhere.

> We're explicitly considering sacrificing consistency with workflow in order to achieve a design that avoids this that can be used consistently for Standalone Activity, Nexus, and future CHASM archetypes.

But those other things don't exist, so it's just guessing. Right now as of this PR, it's just inconsistent. And maybe the next developer doing a future CHASM archetype decides to be inconsistent in their way that they think represents the future and has plenty of justification at that time. Or we could just be consistent in this case everywhere, regardless of whether it's past, present, or future and not subject to the whims of each individual project.

> Yes, list should be a subset of describe. I assume the omission of task queue is an oversight.

Then we should model it as such

> But forgetting a field and then adding it is backwards-compatible and not a problem.

Not being able to forget is a better solution

> My comment was referring to internal consistency regarding Activity, not consistency with Workflow.

There is no concept of activity listing today, maybe I'm misunderstanding what past activity list consistence this refers to

temporal/api/activity/v1/message.proto
```diff
@@ -40,6 +57,143 @@ message ActivityOptions {
...
+
+    // Search attributes from the start request.
+    temporal.api.common.v1.SearchAttributes search_attributes = 8;
+
+    // The task queue this activity was scheduled on when it was originally started, updated on activity options update.
+    string task_queue = 9;
+    // Updated on terminal status.
+    int64 state_transition_count = 10;
+    // Updated once on scheduled and once on terminal status.
+    int64 state_size_bytes = 11;
```
cretz, line 188: Why inconsistently make these as fields instead of search attributes like workflow? Are they _also_ in search attributes, or are we just forbidding the use of this information in queries for activities even though we allow it for workflows? By making them fields, we are ensuring that list and describe field behaviors diverge for the same field which is confusing to have to explain for the same field name.

temporal/api/activity/v1/message.proto
bergundy: > Yes, list should be a subset of describe. I assume the omission of task queue is an oversight.

Task queue is in activity options.

> And maybe the next developer doing a future CHASM archetype decides to be inconsistent in their way that they think represents the future and has plenty of justification at that time.

That's why we are having these reviews to agree on a direction. Being consistent with a single API, for the sake of consistency, is not a strong argument IMHO. We have a chance to make things better, which Dan and I believe this is achieving.

> This is too small of a concern to completely upend model reuse for everyone. This can be used as justification to never use a model in more than one place. We could just inline/copy all fields everywhere.

Why is it too small of a concern? There are definitely places where this doesn't apply, nobody says we need to take this guideline as a strict rule, it should be applied as necessary.

temporal/api/activity/v1/message.proto
bergundy: This is available for workflows though...

https://github.com/temporalio/api/blob/1ae5b673d66b0a94f6131c3eb06bc7173ae2c326/temporal/api/workflow/v1/message.proto#L43-L45

Search attributes returned here are custom ones provided by the user in the start request.

temporal/api/enums/v1/id.proto
bergundy: I'm now leaning towards having a specific one for activities just to avoid mis-predicting the future where we would want to allow different archetypes to have different ID reuse and conflict policies. It's worth the duplication IMHO.

temporal/api/workflowservice/v1/request_response.proto
dandavison: I agree with the proposal here -- that the API should have a single endpoint with the option for long-poll for various stages including completed, thus allowing result to be retrieved and / or other status info ("describe"). The SDKs can present separate APIs if they choose to.

temporal/api/enums/v1/id.proto
dandavison: Opened https://github.com/temporalio/api/pull/650 targeting this PR switching to activity-specific reuse and conflict policy.

temporal/api/workflowservice/v1/request_response.proto
dandavison: Removed these copy/paste errors in 1e99f6f5f19aa317c86390d2171aee7ed96a655a

temporal/api/workflowservice/v1/request_response.proto
dandavison: Resolution here (i.e. matching latest PR state, which is pending SDK approval) is that server is only going to return latest state; it will not persist a ledger/history. Added documentation:

```
// returning the new state. Note that activity state may change multiple times between
// requests, therefore it is not guaranteed that a client making a sequence of long-poll
// requests will see a complete sequence of state changes.
```

temporal/api/workflowservice/v1/request_response.proto
dandavison: Resolution here (i.e. matching latest PR state, which is pending SDK approval) is that the API will offer a method named `PollActivityExecutionRequest` to poll for state and/or result, optionally including input, with no-wait and long-poll variants. The name and functionality aims for clarity and increased consistency (with `WorkflowExecutionUpdate` specifically):

- "Poll" is accurate: the method offers two forms of poll and long-poll
- The functionality is similar to `PollWorkflowExecutionUpdate` and the name thus emphasizes this consistency.
- It feels better IMO to use "poll" to wait for a result than to use "describe" for that.
- It makes sense IMO that an SDK would implement "describe" and "get result" in terms of a lower-level "poll" API

temporal/api/activity/v1/message.proto
```diff
@@ -9,10 +9,27 @@ option java_outer_classname = "MessageProto";
 option ruby_package = "Temporalio::Api::Activity::V1";
 option csharp_namespace = "Temporalio.Api.Activity.V1";
```
dandavison: <dummy comment to start review>

temporal/api/activity/v1/message.proto
dandavison: Resolution here (i.e. matching latest PR state, which is pending SDK approval) is that the API will use independent response structs:

```proto
// Information about a standalone activity.
message ActivityExecutionInfo {

// Limited activity information returned in the list response.
message ActivityExecutionListInfo {
```

temporal/api/workflowservice/v1/request_response.proto
dandavison: Resolution here (i.e. matching latest PR state, which is pending SDK approval) is that the API will use independent response structs:

```proto
// Information about a standalone activity.
message ActivityExecutionInfo {

// Limited activity information returned in the list response.
message ActivityExecutionListInfo {
```

temporal/api/enums/v1/id.proto
dandavison: Resolution here (i.e. matching latest PR state, which is pending SDK approval) is that the API will use activity-specific reuse and conflict policies, as opposed to policies intended to be shared between different execution types.

temporal/api/workflowservice/v1/request_response.proto
dandavison: Resolution here (i.e. matching latest PR state, which is pending SDK approval) is that an `include_input` option will be included:

```proto
message PollActivityExecutionRequest {
    ...
    // Include the input field in the response.
    bool include_input = 5;
```

dandavison:
@cretz the PR has changed substantially, would you like to review again?

temporal/api/activity/v1/message.proto
cretz: Can you clarify whether you resolve to make sure that everything that is in list is also in non-list? Anything beyond relying on close PR scrutiny to ensure that?

temporal/api/activity/v1/message.proto
cretz: Ah I was confused by the name list info here and not there (I probably won't be the last to be confused by this inconsistency). There are other mismatched fields though, but can bring those up separately.

temporal/api/workflowservice/v1/request_response.proto
cretz: > I agree with that.

The model already violates this. For instance, we have decided for list response we are giving a field for `execution_duration`. Same for `close_time` too unless I'm missing where these two fields are defined.

How are you planning to uphold this assurance in the absence of normal model reuse? My concern is the only way it will be upheld is if I personally review every single field very closely. I think we can expect for this to happen plenty if it's happening already before the first PR lands (again, unsure, maybe it's in a model I missed somewhere, hard to know without proper model reuse).

temporal/api/workflowservice/v1/request_response.proto
cretz: I do wish there was still a describe to match our other APIs

temporal/api/activity/v1/message.proto
dandavison: My view is that all the information in list should be in non-list. However, it may not be a simple subset, as the example of `task_queue` showed (its present in non-list as part of `ActivityOptions`). I've added the following comment:

```proto
// When adding fields here, ensure that it is also present in ActivityExecutionInfo (note that it
// may already be present in ActivityExecutionInfo but not at the top-level).
```

So there's that that, and PR scrutiny.

temporal/api/workflowservice/v1/request_response.proto
dandavison: On the plus side, it would be a backwards-compatible change to add it in the future, for example if we find ourselves in possession of additional evidence that it would be a good addition despite the redundancy.

temporal/api/activity/v1/message.proto
dandavison: The server could have a test that verifies that every field in list-info can be computed from info. It wouldn't stop PRs here merging but it would ensure the server is not released violating this.

temporal/api/workflowservice/v1/request_response.proto
dandavison: > How are you planning to uphold this assurance in the absence of normal model reuse? My concern is the only way it will be upheld is if I personally review every single field very closely.

The server could have a test that verifies that every field in list-info can be computed from info.

> I think we can expect for this to happen plenty if it's happening already before the first PR lands

(Actually, I'd say this PR will be subject to _more_ stringent validation when we start implementing with it than now when we're just eyeballing it)

cretz: A test server side would make me much more comfortable, thanks

temporal/api/workflowservice/v1/request_response.proto
maciejdudko, line 446: Do we want a standalone activity to have access to its own run ID? If so then `PollActivityTaskQueueResponse` should be extended to include it.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2595,3 +2609,380 @@ message DescribeWorkerRequest {
...
+    temporal.api.common.v1.Link link = 4;
+}
+
+message PollActivityExecutionRequest {
+    string namespace = 1;
+    string activity_id = 2;
+    // Activity run ID. If empty the request targets the latest run.
+    string run_id = 3;
+    // Exclude the info field from the response.
+    bool exclude_info = 4;
```
maciejdudko: Why is this field `exclude` when others are `include`? If it's to make returning info the default for this operation, I don't think it's the right choice as I imagine the most common use case for polling execution would be waiting for activity completion and getting only the result.

cretz: > I imagine the most common use case for polling execution would be waiting for activity completion and getting only the result.

We make long poll opt-in on long-poll-optional calls. So the default here actually represents what a "describe activity" would do, but they aren't making a describe activity, so they are just making this behave like that by default. The "poll for result" which they've also put inside of this same call is the opt-in alternative call use. It makes sense from a caller/HTTP API POV to have the non-poll "describe" behavior be the default behavior despite the RPC name.

dandavison: > Why is this field exclude when others are include? If it's to make returning info the default for this operation, I don't think it's the right choice as I imagine the most common use case for polling execution would be waiting for activity completion and getting only the result.

Good point @maciejdudko. We made it `exclude_info` because in an earlier version of this PR the method was named `DescribeActivityExecution`. However, I subsequently renamed it `PollActivityExecution` in line with our aim for consistency and clarity in the gRPC API. The new name reflects the fact that this is a low-level API, which SDKs will build functionality such as `DescribeActivity` and `FetchResult(longPoll: bool)` on top of. As such, I think you're right, that the fields should now all have `include` semantics and that it is appropriate for the low-level "poll" gRPC API to default to not including any of the bulky data. Clients (such as SDK authors) making use of the API will not have difficulty understanding that they have to set one of the `include` fields.

cretz: > Clients (such as SDK authors) making use of the API will not have difficulty understanding that they have to set one of the include fields.

Disagree. I think `curl https://my-temporal-host/namespaces/my-ns/activities/my-activity` should return info like `curl https://my-temporal-host/namespaces/my-ns/workflows/my-workflow`, `curl https://my-temporal-host/namespaces/my-ns/schedules/my-schedule`, etc. I think we're regressing and making our APIs harder to use than we did in the past. I don't think we have to put the onus on users and clients to simplify complex APIs, we can develop non-complex APIs from the get-go.

maciejdudko: @cretz makes a good point with curl. I'm on the side of leaving it as `exclude_info` now.

PR #641, opened 2025-09-22
add dispatch ID parameter to create-release.yaml GH action

lilydoar:
**What changed?**

Added the input param dispatch_id to the create-release.yaml GH action.

**Why?**

This parameter is required by the launchpad release tooling to track GH actions that are run as part of Temporal WFs.

<https://temporal.io/blog/running-github-actions-temporal-guide#the-dispatch-id-challenge>

**Breaking changes**

None. A default value of "none" is provided. The new parameter can be ignored when the action is run manually.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -8842,6 +8842,10 @@
         "priority": {
           "$ref": "#/definitions/v1Priority",
           "title": "Priority metadata"
+        },
+        "eagerWorkerDeploymentOptions": {
+          "$ref": "#/definitions/v1WorkerDeploymentOptions",
+          "description": "Deployment Options of the worker who will process the eager task. Passed when `request_eager_execution=true`."
         }
       }
     },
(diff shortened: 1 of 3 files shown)


.github/workflows/create-release.yml
```diff
@@ -16,8 +16,16 @@ on:
       skip_sdk_check:
         description: "Skip sdk-go compatibility check"
         type: boolean
-
+      dispatch_id:
+        description: An ID used by the launchpad worker to identify workflow runs(can be left empty when running manually)
```
cretz: Pedantic, but can be more generic in this open source repo (e.g. "An ID used by external tools ..."), but no big deal

.github/workflows/create-release.yml
```diff
@@ -16,8 +16,16 @@ on:
...
+        default: "none"
+        type: string
 jobs:
+  dispatch:
```
cretz, line 24: I'm guessing this is intentionally a separate job instead of in another job so it's easily programmatically obtainable by external tooling? Just curious

lilydoar: The **important** part is that a step with the ID in its name exists, i.e.

```
steps:
      - name: Dispatch ${{ inputs.dispatch_id }}
...
```

I've just placed it on a different job so that it doesn't interrupt/influence the existing functionality of the action

lilydoar: <https://temporal.io/blog/running-github-actions-temporal-guide#the-dispatch-id-challenge>

PR #642, opened 2025-09-25
Mark pause activity by type API as experimental

gow:
What: Just annotating the pause activity by type API as experimental.

Why: We might revisit this API and change its behavior. So marking it as experimental.

### DIFF
--- a/temporal/api/workflowservice/v1/service.proto
+++ b/temporal/api/workflowservice/v1/service.proto
@@ -133,9 +133,9 @@ service WorkflowService {
             }
         };
     }
-    
-    // GetWorkflowExecutionHistoryReverse returns the history of specified workflow execution in reverse 
-    // order (starting from last event). Fails with`NotFound` if the specified workflow execution is 
+
+    // GetWorkflowExecutionHistoryReverse returns the history of specified workflow execution in reverse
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/service.proto
```diff
@@ -1062,6 +1062,7 @@ service WorkflowService {
     // - The activity should respond to the cancellation accordingly.
     //
     // Returns a `NotFound` error if there is no pending activity with the provided ID or type
+    // Note: Experimental API the parameters and behavior of pause by activity type might change in a future release.
```
bergundy: I would just mark that pause by type is experimental on the `oneof`, I think we are happy with the rest and would like to move forward with it.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -133,9 +133,9 @@ service WorkflowService {
             }
```
gow, line 133: Just some white space changes that the editor automatically did in this file.

temporal/api/workflowservice/v1/service.proto
gow: Done.

PR #643, opened 2025-09-30
add degraded workflow visibility into NamespaceInfo.Capabilities

spkane31:

**What changed?**
Adding `degraded_workflow_visibility` to `NamespaceInfo.Capabilities`.

This will allow us to programmatically turn on the saved query for degraded workflow visibility.

**Breaking changes**
No

### DIFF
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -36,6 +36,8 @@ message NamespaceInfo {
         bool async_update = 3;
         // True if the namespace supports worker heartbeats
         bool worker_heartbeats = 4;
+        // True if the namespace supports degraded workflow visibility
+        bool degraded_workflow_visibility = 5;
     }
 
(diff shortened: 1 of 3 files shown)


bergundy (approved):

temporal/api/namespace/v1/message.proto
```diff
@@ -36,6 +36,8 @@ message NamespaceInfo {
         bool async_update = 3;
         // True if the namespace supports worker heartbeats
         bool worker_heartbeats = 4;
+        // True if the namespace supports degraded workflow visibility
+        bool degraded_workflow_visibility = 5;
```
cretz: Would suggest a rename, e.g. `reported_problems_search_attribute` or something heh. At first glance I read this as there was a form of workflow visibility that was degraded from the normal form if this was true. This is not blocking, though I am not sure the word "degraded" is the best product name.

spkane31: Renamed

PR #661, opened 2025-11-11
Add lease duration to worker heartbeat request

rkannan82:
Added a lease duration field to worker heartbeat request. If not set, or 0, the server will use a default value (1 minute).

This gives SDK the flexibility to determine the heartbeat timeout interval. The server will compute the lease expiry based on the time at which it receives the heartbeat.

No breaking changes. This field is currently not being used on the server.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2467,6 +2467,11 @@ message RecordWorkerHeartbeatRequest {
     string identity = 2;
 
     repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 3;
+
+    // Duration for which the worker lease should be valid.
+    // Server will calculate the actual expiration time based on when it receives this request.
+    // If not specified or zero, the server will use a default lease duration (typically 60 seconds).
+    google.protobuf.Duration lease_duration = 4;
```
Sushisource: Not immediately clear to me what "lease" means in this context. I assume it's how long until the server considers the worker no-longer-live if there isn't another heartbeat in that window, but we should just state that explicitly in the comment if accurate.

rkannan82: Added more details.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2467,6 +2467,21 @@ message RecordWorkerHeartbeatRequest {
...
+    // Duration for which the worker lease should be valid. During this time, the server considers the worker to be active.
+    // The worker is expected to send periodic heartbeats to renew its lease before it expires.
+    //
+    // Server will calculate the actual expiration time based on when it receives this request.
+    // If not specified or zero, the server will use a default lease duration (typically 60 seconds).
+    //
+    // There are 3 states for a worker: Active, Inactive, and CleanedUp.
+    // Lifecycle transitions:
+    // - Active->Active: Each time the server receives a heartbeat from the worker, it will renew the lease and keep the worker in the active state.
+    // - Active->Inactive: When the lease expires, the server will consider the worker to be inactive, and reschedule activities that were known to be running as of that time.
```
Sushisource, line 2481: Is this already true? If not we may want to go back and update the comment after it is, or, alternatively, let's hold off merging this until the implementation behind it is also ready.

Love the detail in the comment now though!

rkannan82: No, none of this is implemented yet.
- Do we typically merge only after the server changes are merged/close to being merged? If so, what is the best practice for new feature development when we have cross repo dependencies like this?
- Can we just add a comment saying under development and merge?
- For context, later I will be evolving this proto to pass activities. Server side requires these protos for implementation.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2467,6 +2467,21 @@ message RecordWorkerHeartbeatRequest {
     string identity = 2;
 
     repeated temporal.api.worker.v1.WorkerHeartbeat worker_heartbeat = 3;
+
+    // Duration for which the worker lease should be valid. During this time, the server considers the worker to be active.
+    // The worker is expected to send periodic heartbeats to renew its lease before it expires.
+    //
+    // Server will calculate the actual expiration time based on when it receives this request.
+    // If not specified or zero, the server will use a default lease duration (typically 60 seconds).
```
yuandrew: > typically 60 seconds

might be pedantic, but if we're talking about a default, shouldn't that be more set in stone? Use of the word "typically" here makes it seem configurable/uncertain

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2467,6 +2467,21 @@ message RecordWorkerHeartbeatRequest {
...
+    //
+    // There are 3 states for a worker: Active, Inactive, and CleanedUp.
+    // Lifecycle transitions:
+    // - Active->Active: Each time the server receives a heartbeat from the worker, it will renew the lease and keep the worker in the active state.
+    // - Active->Inactive: When the lease expires, the server will consider the worker to be inactive, and reschedule activities that were known to be running as of that time.
+    // - Inactive->Active: If the server receives subsequent heartbeat from this worker, then it will transition it back to the active state.
+    // - Inactive->CleanedUp: If the worker remains inactive for a prolonged period, the server will cleanup the worker state. This is a terminal state.
+    //   If the server receives subsequent heartbeat from this worker, then it will return an error.
```
yuandrew: > it will return an error

As in the heartbeat response back to worker will have an error message in it? If an SDK worker receives this, then they should just shutdown/restart, right? Since server no longer sees it as active

temporal/api/workflowservice/v1/request_response.proto
cretz: For all but the most obvious things, we usually would like an accompanying server implementation to ensure that the API matches what we want. We learn a lot during implementation and it can cause problems providing "under development" fields. We find that we so often cut releases of API that people will employ these models into their proxies and in other ways that makes backwards incompatible alterations problematic.

Would strongly suggest _at least_ having server implementation built before merging this. For larger efforts, even with the server implementation built, we may want it in a separate branch so SDKs can design against it and we can apply any API learnings from that, but not sure this qualifies for such a need.

temporal/api/workflowservice/v1/request_response.proto
rkannan82: Corrected. Ideally, I didn't want the client to make any assumptions on the exact value.

temporal/api/workflowservice/v1/request_response.proto
rkannan82: That's right. Updated the comment as well.

temporal/api/workflowservice/v1/request_response.proto
rkannan82: Sounds good. I have converted this into a draft PR, so you can still review it. I will not be merging this.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2467,6 +2467,25 @@ message RecordWorkerHeartbeatRequest {
...
+    // Server will calculate the actual expiration time based on when it receives this request.
+    // If not specified or zero, the server will use a default lease duration of 1 minute.
+    //
+    // There are 3 states for a worker: Active, Inactive, and CleanedUp.
+    // Lifecycle transitions:
+    // - Active->Active: Each time the server receives a heartbeat from the worker, it will renew the lease and keep the worker in the active state.
+    //
+    // - Active->Inactive: When the lease expires, the server will consider the worker to be inactive, and reschedule activities that were known to be running as of that time.
+    //
+    // - Inactive->Active: If the server receives subsequent heartbeat from this worker, then it will transition it back to the active state.
```
bergundy, line 2483: I would call out that once the state machine transitions to inactive, any activity task that was assigned to this worker is considered failed and should be canceled by the worker.

The actual protocol for syncing the activity task inventory should be defined sooner rather than later IMHO.

bergundy (approved):
Approved with a comment.

PR #685, opened 2025-12-09
Standalone Nexus Operations

bergundy:
**What changed?**

- Added the full API scope for standalone nexus operations.

**Server PR**

https://github.com/temporalio/temporal/pull/9869


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2651,3 +2652,210 @@ message UnpauseWorkflowExecutionRequest {
...
 message UnpauseWorkflowExecutionResponse { }
+
+message StartNexusOperationRequest {
+    string namespace = 1;
+    // The identity of the client who initiated this request.
+    string identity = 2;
+    // A unique identifier for this start request. Typically UUIDv4.
+    string request_id = 3;
+
+    string operation_id = 4;
```
pdoerner, line 3094: Probably deserves a comment that this is only a caller-side ID, distinct from the operation ID that the handler will generate

temporal/api/nexus/v1/message.proto
```diff
@@ -198,3 +202,136 @@ message EndpointTarget {
...
+
+    // Schedule-to-close timeout for this operation.
+    // This is the only timeout settable for a Nexus operation.
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 8;
+
+    // The number of attempts made to start/deliver the operation request.
+    // This number represents a minimum bound since the attempt is incremented after the request completes.
+    int32 attempt = 9;
```
pdoerner: This is attempt to deliver the start request. Will we support overall operation retry in the future? Will this name be confusing if we do? Maybe we should call it `start_attempt` so that people will not confuse it with activity attempt which has a different meaning.

temporal/api/nexus/v1/message.proto
```diff
@@ -198,3 +202,136 @@ message EndpointTarget {
...
+
+    // If the state is BLOCKED, blocked reason provides additional information.
+    string blocked_reason = 18;
+
+    // Request ID allocated by the server when handling the StartNexusOperation request.
+    // Used as an idempotency token for submitting start requests to the handler.
+    string request_id = 19;
+
+    // Operation token. Only set for asynchronous operations after a successful StartOperation call.
+    string operation_token = 20;
```
pdoerner: I thought we said we didn't want to expose this to callers? They should only have one way of referencing their operations: their caller-side operation ID.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2651,3 +2652,210 @@ message UnpauseWorkflowExecutionRequest {
 
 // Response to a successful UnpauseWorkflowExecution request.
 message UnpauseWorkflowExecutionResponse { }
+
+message StartNexusOperationRequest {
+    string namespace = 1;
+    // The identity of the client who initiated this request.
+    string identity = 2;
+    // A unique identifier for this start request. Typically UUIDv4.
+    string request_id = 3;
```
pdoerner, line 3090: Deserves a comment that this is the request ID for this caller-side request. StartOperation requests sent to the handler will use a different request ID, generated by the caller server.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2651,3 +2652,210 @@ message UnpauseWorkflowExecutionRequest {
...
+
+    // Stage to wait for. The operation may be in a more advanced stage when the poll is unblocked.
+    temporal.api.enums.v1.NexusOperationWaitStage wait_stage = 4;
+}
+
+message PollNexusOperationResponse {
+    // The run ID of the operation, useful when run_id was not specified in the request.
+    string run_id = 1;
+
+    // Stage to wait for. The operation may be in a more advanced stage when the poll is unblocked.
```
pdoerner: I'm confused, is this the stage the original request sent? Or does it represent the current stage of the operation?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -1287,4 +1287,95 @@ service WorkflowService {
...
+    rpc TerminateNexusOperation (TerminateNexusOperationRequest) returns (TerminateNexusOperationResponse) {
+        option (google.api.http) = {
+            post: "/namespaces/{namespace}/nexus-operations/{operation_id}/terminate"
+            body: "*"
+            additional_bindings {
+                post: "/api/v1/namespaces/{namespace}/nexus-operations/{operation_id}/terminate"
+                body: "*"
+            }
+        };
+    }
```
pdoerner, line 1878: Won't we need `DeleteNexusOperation` as well? Otherwise we won't be able to clean up operations after namespace deletion.

bergundy: Thanks, we need this.

temporal/api/nexus/v1/message.proto
```diff
@@ -198,3 +202,136 @@ message EndpointTarget {
...
+    // available in the list response.
+    temporal.api.enums.v1.NexusOperationStatus status = 8;
+
+    // Search attributes from the start request.
+    temporal.api.common.v1.SearchAttributes search_attributes = 9;
+
+    // Updated on terminal status.
+    int64 state_transition_count = 10;
+    // Updated once on scheduled and once on terminal status.
+    int64 state_size_bytes = 11;
```
cretz: Is this intentionally a field only present in list? It was mentioned for standalone activities that everything in list was expected to be in describe.

Also, for standalone activities it was mentioned there would be a tool that would make sure everything in list was also in describe result. Can we prioritize that? It's a lot of effort for me to have to continually confirm our assertion on every PR and find these issues since we chose not to reuse types.

bergundy: Just want to call this out that we don't have this guarantee for schedules or batch which are much older archetypes: https://github.com/temporalio/api/blob/master/temporal/api/schedule/v1/message.proto https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L1715-L1751.

I don't think this guarantee needs to be high priority but we should keep track of it because I do think that it is nice to have. Ideally the SDKs would allow the types to have completely different fields, there's no need to reuse the models here.

cretz: Right, but this guarantee/promise was made as part of not reusing models knowing the SDK will need this guarantee. Was not expecting a "nice to have" guarantee when the promise/guarantee was made.

bergundy: Let's take this offline.

bergundy: Discussed offline, we will write a tool soon.

temporal/api/workflowservice/v1/request_response.proto
bergundy: The handler doesn't generate an operation ID but I agree it's worth calling out.

temporal/api/nexus/v1/message.proto
bergundy: I want to keep this for consistency with `PendingNexusOperationInfo`.

temporal/api/nexus/v1/message.proto
bergundy: It's still worth exposing this information as we do for workflow callers.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Will do.

temporal/api/workflowservice/v1/request_response.proto
bergundy: The current stage. Let me fix the docstring.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2651,3 +2652,210 @@ message UnpauseWorkflowExecutionRequest {
 
 // Response to a successful UnpauseWorkflowExecution request.
 message UnpauseWorkflowExecutionResponse { }
+
+message StartNexusOperationRequest {
```
Quinn-With-Two-Ns: Is it intentional we don't have `Priority` support?

bergundy: Yes, priorities only apply to durable matching queues, those are not used for nexus tasks.

temporal/api/nexus/v1/message.proto
stephanos: I have no horse in this race, but I'm curious, why is it useful to have?

bergundy: It's useful for debugging and can be used in the direct Nexus APIs to reattach to the same operation (future capability).

temporal/api/nexus/v1/message.proto
```diff
@@ -213,3 +216,136 @@ message EndpointTarget {
...
+    // The last attempt's failure, if any.
+    temporal.api.failure.v1.Failure last_attempt_failure = 5;
+    // The time when the next attempt is scheduled.
+    google.protobuf.Timestamp next_attempt_schedule_time = 6;
+
+    // If the state is BLOCKED, blocked reason provides additional information.
+    string blocked_reason = 7;
+
+    // A reason that may be specified in the CancelNexusOpertionRequest.
+    string reason = 8;
```
stephanos, line 242: This was set to `24` and I changed it; I don't see why it should be `24`.

temporal/api/nexus/v1/message.proto
```diff
@@ -284,7 +284,8 @@ message NexusOperationInfo {
     // The time when the next attempt is scheduled.
     google.protobuf.Timestamp next_attempt_schedule_time = 15;
 
-    // How long this operation has been running for, including all attempts and backoff between attempts.
+    // Elapsed time from schedule_time to now for running operations or to close_time for closed
+    // operations, including all attempts and backoff between attempts.
```
stephanos, line 300: Clarifies it works for running operations (as opposed to `NexusOperationListInfo`); correct me if that's wrong.

temporal/api/enums/v1/nexus.proto
```diff
@@ -21,8 +21,6 @@ enum NexusHandlerErrorRetryBehavior {
 }
 
 // Status of a standalone Nexus operation.
-// The status is updated once, when the operation is originally scheduled, and again when the operation reaches a
-// terminal status.
```
stephanos: Moved this to NexusOperationListInfo where it seems to be applicable; and only there?

temporal/api/nexus/v1/message.proto
stephanos: (I wonder why we don't have a linter for this?)

temporal/api/enums/v1/nexus.proto
bergundy: It's applicable here too.

stephanos: Oh, I see so `UNSPECIFIED` -> `RUNNING` -> [terminal state].

temporal/api/common/v1/message.proto
```diff
@@ -240,9 +240,17 @@ message Link {
         string job_id = 1;
     }
 
+    // A link to a standalone Nexus operation execution.
+    message NexusOperationExecution {
```
bergundy: Looks like the naming is inconsistent with `BatchJob` and `WorkflowEvent`. Neither of those use the term `Execution`. I'm on the fence whether this is acceptable or not.

temporal/api/enums/v1/nexus.proto
```diff
@@ -20,3 +20,63 @@ enum NexusHandlerErrorRetryBehavior {
...
+    NEXUS_OPERATION_EXECUTION_STATUS_COMPLETED = 2;
+    // The operation completed with failure.
+    NEXUS_OPERATION_EXECUTION_STATUS_FAILED = 3;
+    // The operation completed as canceled.
+    // Requesting to cancel an operation does not automatically transition the operation to canceled status, depending
+    // on the current operation status and the cancelation type used.
+    NEXUS_OPERATION_EXECUTION_STATUS_CANCELED = 4;
+    // The operation was terminated. Termination happens immediately without notifying the handler.
+    NEXUS_OPERATION_EXECUTION_STATUS_TERMINATED = 5;
+    // The operation has timed out by reaching the specified schedule-to-close timeout.
```
bergundy: This could be any timeout now that we've added start-to-close and schedule-to-close.

temporal/api/enums/v1/nexus.proto
```diff
@@ -20,3 +20,63 @@ enum NexusHandlerErrorRetryBehavior {
...
+    NEXUS_OPERATION_EXECUTION_STATUS_RUNNING = 1;
+    // The operation completed successfully.
+    NEXUS_OPERATION_EXECUTION_STATUS_COMPLETED = 2;
+    // The operation completed with failure.
+    NEXUS_OPERATION_EXECUTION_STATUS_FAILED = 3;
+    // The operation completed as canceled.
+    // Requesting to cancel an operation does not automatically transition the operation to canceled status, depending
+    // on the current operation status and the cancelation type used.
+    NEXUS_OPERATION_EXECUTION_STATUS_CANCELED = 4;
+    // The operation was terminated. Termination happens immediately without notifying the handler.
```
bergundy, line 41: Note to future us: This comment may not be true if we add a Nexus equivalent of a "parent close policy".

temporal/api/nexus/v1/message.proto
bergundy: Copy pasta probably. Thanks for catching this.

temporal/api/nexus/v1/message.proto
```diff
@@ -213,3 +216,138 @@ message EndpointTarget {
...
+    // Updated once when the operation is originally scheduled, and again when it reaches a terminal status.
+    temporal.api.enums.v1.NexusOperationExecutionStatus status = 6;
+    // More detailed breakdown of NEXUS_OPERATION_EXECUTION_STATUS_RUNNING.
+    temporal.api.enums.v1.PendingNexusOperationState state = 7;
+
+    // Schedule-to-close timeout for this operation.
+    // This is the only timeout settable for a Nexus operation.
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 8;
```
bergundy, line 267: Might as well add `start_to_close_timeout` and `schedule_to_start_timeout` already. Those are already supported by the server for workflow ops.

temporal/api/nexus/v1/message.proto
bergundy: LGTM

bergundy (commented):
@stephanos added a few more comments but this is almost good to merge AFAIC.

temporal/api/common/v1/message.proto
stephanos: Removed `Execution`

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3017,6 +3018,159 @@ message ListActivityExecutionsResponse {
...
+    // Service name.
+    string service = 6;
+    // Operation name.
+    string operation = 7;
+
+    // Schedule-to-close timeout for this operation.
+    // This is the only timeout settable for a Nexus operation.
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 8;
```
Quinn-With-Two-Ns, line 3107: What the other timeouts like schedule to start and start to close

bergundy: Those should be added.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3017,6 +3018,159 @@ message ListActivityExecutionsResponse {
...
+    // - Endpoint
+    // - Service
+    // - Operation
+    // - RequestId
+    // - StartTime
+    // - ExecutionTime
+    // - CloseTime
+    // - ExecutionStatus
+    // - ExecutionDuration
+    // - StateTransitionCount
```
bergundy, line 3244: @stephanos can you double check this against the implementation please?

temporal/api/workflowservice/v1/request_response.proto
Quinn-With-Two-Ns: Will they be in this PR or later?

Quinn-With-Two-Ns (approved):

temporal/api/workflowservice/v1/request_response.proto
stephanos: I'll add it now: https://github.com/temporalio/api/pull/685/commits/1d12d58f4838e8470c594c75a2ad68c680245aa6

temporal/api/enums/v1/nexus.proto
```diff
@@ -20,3 +20,63 @@ enum NexusHandlerErrorRetryBehavior {
...
+// Note that it is *never* valid to have two running instances of the same operation ID.
+//
+// See `NexusOperationIdReusePolicy` for handling operation ID duplication with a *closed* operation.
+enum NexusOperationIdConflictPolicy {
+    NEXUS_OPERATION_ID_CONFLICT_POLICY_UNSPECIFIED = 0;
+    // Don't start a new operation; instead return `NexusOperationAlreadyStarted` error.
+    NEXUS_OPERATION_ID_CONFLICT_POLICY_FAIL = 1;
+    // Don't start a new operation; instead return a handle for the running operation.
+    NEXUS_OPERATION_ID_CONFLICT_POLICY_USE_EXISTING = 2;
+}
```
Sushisource, line 82: We don't support terminate like w/ workflows?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3027,6 +3028,176 @@ message ListActivityExecutionsResponse {
...
+message DescribeNexusOperationExecutionRequest {
+    string namespace = 1;
+    string operation_id = 2;
+    // Operation run ID. If empty the request targets the latest run.
+    string run_id = 3;
+    // Include the input field in the response.
+    bool include_input = 4;
+    // Include the outcome (result/failure) in the response if the operation has completed.
+    bool include_outcome = 5;
+    // Token from a previous DescribeNexusOperationExecutionResponse. If present, long-poll until operation
```
```suggestion
    // Token from a previous DescribeNexusOperationExecutionResponse. If present, this RPC will long-poll until operation
```

Took me a minute to understand what this was trying to say w/o this clarification.

Sushisource: Also does this mean in order to long poll on complete you must make a minimum of two RPCs? One to get the token and one to wait on it? If so what's the use case for the long poll token here instead of just using `PollNexusOperationExecutionRequest`?

Quinn-With-Two-Ns: Polling on `DescribeNexusOperationExecution` is for the UI. So the UI will describe once to render the page and then long poll so any update can be reflected in the UI.

temporal/api/enums/v1/nexus.proto
stephanos: cc @bergundy

temporal/api/workflowservice/v1/request_response.proto
stephanos: fwiw, [DescribeActivityExecutionRequest](https://github.com/temporalio/api/pull/688/changes#diff-2c30b34f996ae016a7532d5c70a798c8387a14519c1846fd2e4149cd10d244e0R2739) follows the same approach.

Also a noteable difference is, like Quinn alludes to, the `Describe` returns `NexusOperationExecutionInfo` which Poll doesn't.

temporal/api/enums/v1/nexus.proto
stephanos: I'll add that I don't have context for why it's not there; but it seems like something we can always add later if we want to.

bergundy: We don't support TERMINATE_IF_RUNNING for anything that isn't workflow.

temporal/api/nexus/v1/message.proto
```diff
@@ -213,3 +216,148 @@ message EndpointTarget {
...
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_start_timeout = 9;
+
+    // Start-to-close timeout for this operation.
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration start_to_close_timeout = 10;
+
+    // The number of attempts made to start/deliver the operation request.
+    // This number represents a minimum bound since the attempt is incremented after the request completes.
```
bergundy:
```suggestion
  // The number of attempts made to deliver the start operation request.
  // This number is approximate, it is incremented when a task is added to the history queue.
  // In practice, there could be more attempts if a task is executed but fails to commit, or less attempts if a task was
  // never executed.
```

bergundy: @stephanos please also update this comment for PendingNexusOperationInfo in `workflow/v1/message.go`.

stephanos: Updated in https://github.com/temporalio/api/pull/685/commits/734ce9d40245f2925bd435b0a2eda6f0acda7d2c

temporal/api/workflowservice/v1/request_response.proto
stephanos: All but `RequestId` are supported; fixing that up now: https://github.com/temporalio/temporal/pull/10032

Sushisource (approved):

PR #703, opened 2026-01-26
initial server metadata proto fields for history event

chene40:

**What changed?**

**Why?**

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -24,6 +24,7 @@ import "temporal/api/update/v1/message.proto";
 import "temporal/api/workflow/v1/message.proto";
 import "temporal/api/sdk/v1/task_complete_metadata.proto";
 import "temporal/api/sdk/v1/user_metadata.proto";
+import "temporal/api/server/v1/message.proto";
 
 // Always the first event in workflow history
 message WorkflowExecutionStartedEventAttributes {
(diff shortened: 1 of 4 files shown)


temporal/api/server/v1/message.proto
cretz, line 1: I do not think we need an entirely new proto package for this

temporal/api/history/v1/message.proto
```diff
@@ -1115,6 +1116,9 @@ message HistoryEvent {
     // Links associated with the event.
     repeated temporal.api.common.v1.Link links = 302;
     // The event details. The type must match that in `event_type`.
+    // Actions metering metadata for the event.
+    temporal.api.server.v1.ServerMetadata server_metadata = 303;
```
bergundy, line 1120: I think most of the fields here could be considered server metadata, you could just inline the ActionMetadata here.

bergundy: Having a separate server metadata field would make sense if it is supposed to be applied in other places though.

cretz, line 1120: Unfortunately the PR where the discussion was occurring was closed. But bringing back the discussion from https://github.com/temporalio/api/pull/701#discussion_r2722764611, I do not believe we should inundate history with redundant information. If people can iterate history and count actions, they can iterate history and count (certain) events.

We should not use the burden of having to read history, which already has many cryptic elements and we provide UIs and CLIs for understanding, as justification for adding redundant data.

Also, this doesn't even solve the problem of missing data (e.g. query and heartbeat counts). History is not the place for this information IMO. We should have a more general discussion about how one can derive action count from history, but this not only doesn't solve that, it inundates _everyone_'s history format with redundant information.

temporal/api/server/v1/message.proto
dnr: If we do need a new package, please use a different name: we have `temporal.server.api...` for internal protos which is already somewhat confusing, let's not also have `temporal.api.server`. Maybe `temporal.api.metering` or `temporal.api.cloudmetering` to be clear this is a cloud concept?

PR #708, opened 2026-02-03
Add Nexus worker service for server-to-worker commands

rkannan82:
## Summary

Defines a Nexus service for server-to-worker communication, starting with activity cancellation support.

## Design Decision

We chose a **generic command API** (`ExecuteCommandsRequest` with `oneof` command types) instead of a cancel-specific API. This allows a future optimization to batch multiple commands (cancel, pause, etc) in a single request and deliver to a worker in one RPC.

## Files

- `temporal/api/nexusservices/workerservice/v1/request_response.proto` - request response definitions
- `nexus-rpc/temporal-proto-models-nexusrpc.yaml` - Nexus service definition

## Related

- [Server PR](https://github.com/temporalio/temporal/pull/9233)


temporal/api/worker/v1/message.proto
```diff
@@ -139,3 +140,29 @@ message PluginInfo {
...
+// Request to cancel running activities on this worker.
+message CancelActivityTask {
+    // The workflow execution that owns the activities.
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 1;
+    // The scheduled event IDs of activities to cancel.
+    // If empty, cancel all activities for this workflow running on this worker.
+    repeated int64 scheduled_event_ids = 2;
+    // Human-readable reason for cancellation.
+    string reason = 3;
+}
```
cretz: Arguably we shouldn't have to use proto for this, but even if we do, can these be defined at the same time as the exact Nexus service?

temporal/api/worker/v1/message.proto
```diff
@@ -139,3 +140,29 @@ message PluginInfo {
     // The version of the plugin, may be empty.
     string version = 2;
 }
+
+// Container for batching multiple control tasks delivered to a worker in one Nexus operation.
+message WorkerControlTasks {
+    repeated WorkerControlTask tasks = 1;
```
cretz: Why put all of these in one call instead of having independent Nexus operations?

temporal/api/worker/v1/message.proto
```diff
@@ -139,3 +140,29 @@ message PluginInfo {
     // The version of the plugin, may be empty.
     string version = 2;
 }
+
+// Container for batching multiple control tasks delivered to a worker in one Nexus operation.
+message WorkerControlTasks {
```
cretz: If this is meant to be the Nexus input/request, it should be named as such and you need a output/response of an operation (even if it is empty at first). You should also define even what the service name and operation are expected to be. Having an input model only with no output model, no service name, and no operation name makes me concerned this isn't fully designed yet.

temporal/api/worker/v1/message.proto
```diff
@@ -139,3 +140,29 @@ message PluginInfo {
...
+    // Timestamp when this task was created.
+    google.protobuf.Timestamp create_time = 1;
+
+    oneof task {
+        CancelActivityTask cancel_activity = 2;
+    }
+}
+
+// Request to cancel running activities on this worker.
+message CancelActivityTask {
```
cretz: This should contain the task token to reference the activity by IMO, not workflow + event ID which doesn't help standalone activities.

temporal/api/worker/v1/message.proto
```diff
@@ -139,3 +140,29 @@ message PluginInfo {
...
+    // The workflow execution that owns the activities.
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 1;
+    // The scheduled event IDs of activities to cancel.
+    // If empty, cancel all activities for this workflow running on this worker.
+    repeated int64 scheduled_event_ids = 2;
+    // Human-readable reason for cancellation.
+    string reason = 3;
```
cretz: I don't think this is accurate. I assume this is for any form of activity interruption, and therefore this is missing information. This should be modeled off of `RecordActivityTaskHeartbeatResponse`. Also, there is no such thing as a `reason` to deliver to an activity for its cancellation (if that is a new feature, we should discuss it in general), and this doesn't account for activity pause/reset.

temporal/api/worker/v1/message.proto
cretz: IMO input to the worker Nexus service should provide the worker instance key the control call is targeting. This is a shared Nexus worker.

On a side note, I think we will want to implement a form of this end-to-end before we merge what we think the models needed will be.

temporal/api/worker/v1/message.proto
```diff
@@ -139,3 +140,29 @@ message PluginInfo {
...
+
+// Container for batching multiple control tasks delivered to a worker in one Nexus operation.
+message WorkerControlTasks {
+    repeated WorkerControlTask tasks = 1;
+}
+
+// A single control task for a worker.
+message WorkerControlTask {
+    // Timestamp when this task was created.
+    google.protobuf.Timestamp create_time = 1;
```
cretz: Why is this needed? Not saying there aren't "nice to have" fields we can continually add to these things, but it doesn't seem necessary any more than many other things we can add as needed.

temporal/api/worker/v1/message.proto
yuandrew: agree that we should have something in server and SDK working (in branches) with this definition before we merge

yuandrew: I think this should be a new file, as it's defining a new nexus service, not the typical grpc service, see the previous discussion here, https://github.com/temporalio/api/pull/622#discussion_r2228479768, 
specifically what spencer said 

> For now I would suggest moving these commands into another proto file inside worker/v1 like worker_nexus_service_commands.proto and a big block comment at the top can say the service name, and the operation names, and how they correspond to workflowservice rpcs

I think we were all on board with Yuri's design of `worker_nexus_service_commands.proto`, and we should do something similar here where the new file contains the block comment giving more information about the new nexus service

```
// (--
/////////////////////////////////////////////////////////////////////
// This file contains:
// - Conventions between server and worker.
// - Definitions for commands and payloads for server-worker communication via Nexus
... and so on
```

temporal/api/worker/v1/message.proto
rkannan82: Changed this to be independent operations. I had an incorrect understanding of how the Nexus operation would be modeled; thought we could batch multiple commands together.

temporal/api/worker/v1/message.proto
rkannan82: Please take another look. I have incorporated Yuris' PR style in here.

temporal/api/worker/v1/message.proto
rkannan82: Done.

temporal/api/worker/v1/message.proto
rkannan82: Done.

temporal/api/worker/v1/message.proto
rkannan82: Removed. That was an oversight; thought this is something that worker would use for debug logs.

temporal/api/worker/v1/message.proto
rkannan82: Sorry, what did you mean by "define at the same time as Nexus service"? I have now documented the nexus service and command name.

temporal/api/worker/v1/message.proto
rkannan82: FYI: The WIP server side implementation is here: [PR #9233](https://github.com/temporalio/temporal/pull/9233).
I have a functional test that simulates SDK nexus poller and verifies it receives the cancel task.

temporal/api/worker/v1/worker_nexus_service_commands.proto
```diff
@@ -0,0 +1,49 @@
...
+
+// (--
+/////////////////////////////////////////////////////////////////////
+// This file contains:
+// - Conventions between server and worker.
+// - Definitions for commands and payloads for server-worker communication via Nexus
+//
+// COMMUNICATION PROTOCOL:
+// - Transport: Nexus tasks on task queue
+// - Server identifier: "sys-worker-service"
```
cretz: This is the Nexus service name?

temporal/api/worker/v1/worker_nexus_service_commands.proto
```diff
@@ -0,0 +1,49 @@
...
+//
+// - In/out payloads namings follow the same convention as the regular API:
+// - CommandNameRequest (input payload)
+// - CommandNameResponse (output payload).
+// - Empty payload if response is not needed/not expected
+//
+// --)
+
+// Request payload for the "cancel-activities" Nexus operation.
+message CancelActivitiesRequestPayload {
```
cretz: Pedantic, but I think you can remove the `Payload` suffixes (payloads have a pretty different meaning in our API)

temporal/api/nexusservices/workerservice/v1/request_response.proto
cretz, line 1: Arguably this should be defined in terms of https://github.com/nexus-rpc/nexus-rpc-gen, but this is technically fine too if others are ok with this. Ideally we should dogfood our own Nexus service definition file format (it can also teach us things about its usability).

cretz (commented):
LGTM, though would want to see end-to-end working including in an SDK if possible before approving

temporal/api/worker/v1/worker_nexus_service_commands.proto
rkannan82: Yes. I took this from Yuri's PR. Andrew, can you clarify if this is OK.

bergundy: You should use `temporal.api.workerservice.v1.WorkerService` IMHO to be consistent with other services defined in this repo.

temporal/api/worker/v1/worker_nexus_service_commands.proto
bergundy: Don't think it's pedantic. Please remove and be consistent with other request response definitions in this repo.

rkannan82: Done.

temporal/api/worker/v1/worker_nexus_service_commands.proto
rkannan82: Changed to temporal.api.worker.v1.WorkerService

temporal/api/nexusservices/workerservice/v1/request_response.proto
rkannan82: Added a yaml file for nexus service definition.

temporal/api/worker/v1/message.proto
rkannan82: We have the server side implementation pending review in [PR #9233](https://github.com/temporalio/temporal/pull/9233). I have an e2e test that simulates the SDK behavior by polling on the control queue. Let me know if this suffices to unblock this PR.

bergundy (approved):
LGTM, don't merge this until you've generated the code and verified that this API works across SDK and server.

There will probably be a separate code generation project that will construct the YAML from protos to prevent the need to define services in separate places but this is a good start IMHO.

temporal/api/worker/v1/worker_service.nexusrpc.yaml
```diff
@@ -0,0 +1,25 @@
...
+#
+# Nexus service definition for server-to-worker communication.
+# See worker_nexus_service_commands.proto for protocol details and conventions.
+#
+# Task queue format: /temporal-sys/worker-commands/{namespace}/{worker_grouping_key}
+
+nexusrpc: 1.0.0
+
+services:
+  temporal.api.worker.v1.WorkerService:
```
cretz: Not sure we have to follow our gRPC naming here. We don't have a best practice since this is the first one, but I think just `WorkerControlService` is fine or `TemporalWorkerControlService` (task queue is enough to keep things separated) and similar file name

temporal/api/worker/v1/worker_service.nexusrpc.yaml
```diff
@@ -0,0 +1,25 @@
...
+    description: >
+      Internal Nexus service for server-to-worker communication.
+      Used by the Temporal server to send commands to workers.
+    operations:
+      cancel-activities:
```
cretz: We don't have a best practice since this is the first one, but I think camelCase is reasonable, e.g. `cancelActivities`

temporal/api/worker/v1/worker_service.nexusrpc.yaml
```diff
@@ -0,0 +1,25 @@
...
+        description: >
+          Cancels running activities on a worker. The server sends this when
+          activities need to be cancelled (e.g., workflow cancellation or
+          explicit activity cancellation).
+        input:
+          $goRef: "go.temporal.io/api/worker/v1.CancelActivitiesRequest"
```
cretz: I think we should consider using JSON schema here and no protos (but if we must use protos, we should reference all languages)

temporal/api/worker/v1/worker_service.nexusrpc.yaml
cretz, line 1: I wonder if Nexus services should be in a separate place. We don't have a best practice since this is the first one, but maybe a `nexus-services/` top-level dir?

cretz (commented):
This is our first Nexus service, so lots of things undecided. I added comments, but would like to hear others' opinions.

temporal/api/worker/v1/worker_service.nexusrpc.yaml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nexus-rpc/nexus-rpc-gen/main/schemas/nexus-rpc-gen.json
+#
+# Nexus service definition for server-to-worker communication.
+# See worker_nexus_service_commands.proto for protocol details and conventions.
+#
+# Task queue format: /temporal-sys/worker-commands/{namespace}/{worker_grouping_key}
+
+nexusrpc: 1.0.0
```
yuandrew, line 8: Should this start as 1.0.0 or should we start with 0.1.0 until things are more solidified? Not sure how much this matters, or even what this is really used for

temporal/api/worker/v1/worker_service.nexusrpc.yaml
rkannan82: Roey recommended this: `You should use temporal.api.workerservice.v1.WorkerService IMHO to be consistent with other services defined in this repo.`

cretz: What other services defined in this repo? gRPC ones? To confirm, are we saying Nexus services have to somehow be consistent with gRPC services? Surely it's understood they are already not consistent because they are _completely different_ protocols/contracts.

Nexus services should be consistent with Nexus services and if there are gaps in expected consistency as Nexus use grows, we remedy it, but it's not correct to pretend completely unrelated proto/gRPC's rules/approaches have any bearing on Nexus.

temporal/api/worker/v1/worker_service.nexusrpc.yaml
rkannan82: Done.

temporal/api/worker/v1/worker_service.nexusrpc.yaml
rkannan82: Done.

temporal/api/worker/v1/worker_service.nexusrpc.yaml
rkannan82: Updated to have ref for all languages.

temporal/api/nexusservices/workerservice/v1/service.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+nexusrpc: 1.0.0
+
+services:
+  temporal.api.nexusservices.workerservice.v1.WorkerService:
+    description: >
+      Internal Nexus service for server-to-worker communication.
+      Used by the Temporal server to send commands to workers.
+    operations:
+      executeCommands:
```
bergundy:
```suggestion
      ExecuteCommands:
```

temporal/api/worker/v1/worker_service.nexusrpc.yaml
rkannan82: I believe this refers to the version of Nexus rpc spec. So it should be 1.0.0. The samples code has it.

https://github.com/nexus-rpc/nexus-rpc-gen/blob/b3e29919d21caba9d2b4cfadc84f743c34382b39/samples/user-service/user-service.nexusrpc.yaml#L2

temporal/api/nexusservices/workerservice/v1/request_response.proto
```diff
@@ -0,0 +1,45 @@
...
+// (--
+// Internal Nexus service for server-to-worker communication.
+// See service.yaml for the service definition.
+// --)
+
+// Request payload for the "executeCommands" Nexus operation.
+message WorkerCommandsRequest {
+    repeated WorkerCommand commands = 1;
+
+    message WorkerCommand {
```
bergundy: I would pull this out to a top level message and put it in `temporal.api.worker.v1` to make it easier to use and available in other contexts, traditionally we have only put requests and responses in `request_response.proto` files.

temporal/api/nexusservices/workerservice/v1/request_response.proto
```diff
@@ -0,0 +1,45 @@
...
+        oneof type {
+            CancelActivity cancel_activity = 1;
+        }
+    }
+
+    message CancelActivity {
```
bergundy: Same, make this a top level `temporal.api.worker.v1.CancelActivityCommand`.

temporal/api/nexusservices/workerservice/v1/request_response.proto
```diff
@@ -0,0 +1,45 @@
...
+        bytes task_token = 1;
+    }
+}
+
+// Response payload for the "executeCommands" Nexus operation.
+message WorkerCommandsResponse {
+    repeated WorkerCommandResult results = 1;
+
+    message WorkerCommandResult {
```
bergundy: Same here.

temporal/api/nexusservices/workerservice/v1/request_response.proto
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    message CancelActivity {
+        bytes task_token = 1;
+    }
+}
+
+// Response payload for the "executeCommands" Nexus operation.
+message WorkerCommandsResponse {
+    repeated WorkerCommandResult results = 1;
```
Sushisource: We should document the expectation (at least, what I assume is the expectation... hence the need for a docstrings) that this results list must be 1:1 with the request list, of the same size and order.

temporal/api/nexusservices/workerservice/v1/request_response.proto
```diff
@@ -0,0 +1,45 @@
...
+
+    message WorkerCommandResult {
+        oneof type {
+            CancelActivity cancel_activity = 1;
```
bergundy: `temporal.api.worker.v1.CancelActivityResult`

temporal/api/nexusservices/workerservice/v1/request_response.proto
```diff
@@ -0,0 +1,45 @@
...
+option ruby_package = "Temporalio::Api::Nexusservices::Workerservice::V1";
+option csharp_namespace = "Temporalio.Api.Nexusservices.Workerservice.V1";
+
+// (--
+// Internal Nexus service for server-to-worker communication.
+// See service.yaml for the service definition.
+// --)
+
+// Request payload for the "executeCommands" Nexus operation.
+message WorkerCommandsRequest {
```
bergundy: If the API is called `ExecuteCommands`, the request and response should be named accordingly.

```suggestion
message ExecuteCommandsRequest {
```

temporal/api/nexusservices/workerservice/v1/request_response.proto
yuandrew: But since this is a new nexus service, shouldn't this be decoupled from the existing gRPC `temporal.api.worker.v1` protos and exist in the separate `temporal/api/nexusservices`? Maybe `temporal.api.nexusservices.worker.v1`?

Do we know what other contexts we might want to use this message, or is that hypothetical?

rkannan82: I have made the changes that Roey suggested. Here is my interpretation.
- Top level request/response is about the RPC. In this case, it is nexus protocol. So we keep it under nexusservices dir.
- WorkerCommand is not strictly tied to this protocol. So we move it out of nexusservices dir.
-As to whether this is useful in other places: I am not sure. We could repurpose these command protos in the server side when we create the internal tasks. But it is not good to couple them.

temporal/api/worker/v1/worker_service.nexusrpc.yaml
rkannan82: Having a package name style similar to grpc seems reasonable to me. Unless there is a strong reason to not use, I suggest following this style. Please reopen if you feel otherwise.

Sushisource (approved):
Looking good to me

nexus-rpc/temporal-proto-models-nexusrpc.yaml
bergundy, line 1: Rename this to `nexus-rpc/temporal-proto-models-nexusrpc.yaml`. @spkane31 is working on a protoc plugin that will generate all Nexus service definitions from proto. We will want to replace this file with a generated version later.

bergundy (approved):
Just one comment on the location of the schema file.

PR #710, opened 2026-02-09
Updated proto with API changes

### LINKED ISSUE #778: temporal-web displays incorrect timeout for workflow tasks pulled off of Sticky queue
image

See image above, which indicates "10" seconds, however the server is hardcoded to 5 seconds and the task actually does timeout in 5 seconds, which can be confusing to the user:

image

saad1998:

**What changed?**
Added optional schedule_to_start_timeout (google.protobuf.Duration, field number 4) to WorkflowTaskScheduledEventAttributes in temporal/api/history/v1/message.proto.
Regenerated Go and OpenAPI in this repo (via make proto / project tooling).

**Why?**

Fixes temporalio/temporal#778: the server applies a schedule-to-start timeout (e.g. 5s for speculative workflow tasks, or the sticky timeout) but the event did not record it, so UIs could show a different value (e.g. 10s) and users were confused when the task timed out earlier. Storing the timeout on the event makes the server’s behavior visible and consistent for clients and UIs.

**Breaking changes**

None. New field is optional; existing events and clients remain valid.

**Server PR**

Server PR (to temporalio/temporal) that populates and uses this field: https://github.com/temporalio/temporal/pull/9257

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -261,6 +261,13 @@ message WorkflowTaskScheduledEventAttributes {
     google.protobuf.Duration start_to_close_timeout = 2;
     // Starting at 1, how many attempts there have been to complete this task
     int32 attempt = 3;
+    // Limits time a workflow task can stay in a task queue before a worker picks it up.
+    // Only set when the server enforces a schedule-to-start timeout (e.g. sticky or speculative WT on normal queue).
+    //
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_start_timeout = 4;
+
 }
 
 message WorkflowTaskStartedEventAttributes {
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -17957,6 +17957,10 @@
           "type": "integer",
           "format": "int32",
           "title": "Starting at 1, how many attempts there have been to complete this task"
+        },
+        "scheduleToStartTimeout": {
+          "type": "string",
+          "description": "Limits time a workflow task can stay in a task queue before a worker picks it up.\nOnly set when the server enforces a schedule-to-start timeout (e.g. sticky or speculative WT on normal queue).\n"
         }
       }
     },
--- a/openapi/openapiv3.yaml
+++ b/openapi/openapiv3.yaml
@@ -16718,6 +16718,15 @@ components:
           type: integer
           description: Starting at 1, how many attempts there have been to complete this task
           format: int32
+        scheduleToStartTimeout:
+          pattern: ^-?(?:0|[1-9][0-9]{0,11})(?:\.[0-9]{1,9})?s$
+          type: string
+          description: |-
+            Limits time a workflow task can stay in a task queue before a worker picks it up.
+             Only set when the server enforces a schedule-to-start timeout (e.g. sticky or speculative WT on normal queue).
+
+             (-- api-linter: core::0140::prepositions=disabled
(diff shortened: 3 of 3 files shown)


temporal/api/history/v1/message.proto
```diff
@@ -261,6 +261,13 @@ message WorkflowTaskScheduledEventAttributes {
     google.protobuf.Duration start_to_close_timeout = 2;
     // Starting at 1, how many attempts there have been to complete this task
     int32 attempt = 3;
+    // Limits time a workflow task can stay in a task queue before a worker picks it up.
+    // Only set when the server enforces a schedule-to-start timeout (e.g. sticky or speculative WT on normal queue).
+    //
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_start_timeout = 4;
```
cretz, line 269: Can you clarify whether this has a specific recent need, or was this just something found on the backlog at https://github.com/temporalio/temporal/issues/778? History fields are pretty scrutinized, so would want to confirm there is a pressing reason for this.

Also, the PR description is missing the accompanying server PR showing how this will be implemented.

cretz, line 269: Would like to understand details of how this value is present and/or updated or not on successive task failures or timeouts. I wouldn't expect speculative task failure is ever recorded, or is it? And as for sticky, I suppose it can only occur once because rescheduled task does not have a schedule to start?

Any reason not to add this field to `WorkflowTaskTimedOutEventAttributes` instead?

saad1998: This change came from the backlog. 
There isn’t a new production incident driving it. 
The original issue was observability: the UI could show a user-configured schedule-to-start timeout (e.g. 10s) while the server used a different (e.g. hardcoded) value, so users couldn’t see what was actually in effect from history. The goal is to make history reflect the schedule-to-start timeout when the server enforces one, so there’s no pressing reason beyond that observability/consistency improvement. If you’d rather not add history fields for that alone, I’m happy to close or repurpose the PR.

Also thanks for pointing out the missing server PR I thought I added it in the description.

saad1998: This is what I think is happening so far the server only sets schedule_to_start_timeout when it actually enforces a schedule-to-start timeout i.e. for sticky workflow tasks and for speculative workflow tasks on the normal queue. For a normal task on the normal queue the field stays unset, so it isn’t “updated” across successive timeouts in the usual retry sense; it’s only present on the scheduled events where that timeout applies.

For speculative tasks, the schedule-to-start timeout is recorded: when the timer fires we create the WORKFLOW_TASK_SCHEDULED event at that moment (in AddWorkflowTaskScheduleToStartTimeoutEvent) and then add the WORKFLOW_TASK_TIMED_OUT event, so the scheduled event and its timeout are written together. 

For sticky, schedule-to-start only applies to the sticky task; after it times out we reschedule on the normal queue, and that new scheduled event doesn’t have a schedule-to-start timeout, so the field wouldn’t be set there.

I’m happy to put the duration on WorkflowTaskTimedOutEventAttributes instead it would make the timeout event self-contained (you’d see both the timeout type and the duration that fired), avoid ambiguity about when the field is set on the scheduled event, and better match “this is what timed out.” I can switch the change to add the field there (and only populate it when timeout_type is SCHEDULE_TO_START) if that’s the direction you prefer; otherwise I can keep it on the scheduled event and add a short comment documenting this lifecycle.

saad1998:
@cretz just following up on this issue. Please let me know if the PR for this issue is acceptable or not, and if there is anything I missed out then do let me know.

PR #729, opened 2026-03-11
Add CountWorkers API

rkannan82:
## What
Add `CountWorkers` RPC to `WorkflowService` for counting workers matching a query filter.

- Request: `namespace` + `query` + `include_system_workers` (same filter fields as `ListWorkers`)
- Response: `int64 count`
- HTTP route: `GET /namespaces/{namespace}/worker-count`

[Server PR](https://github.com/temporalio/temporal/pull/9476)

## Why
The worker UI needs to display a total worker count (e.g. in tabs) without fetching the full list. Follows the established pattern used by `CountWorkflowExecutions`, `CountSchedules`, and `CountActivityExecutions`.

## How did you test it?
- [x] built (`make ci-build`)
- [x] passes buf linting and breaking change checks
- [x] regenerated OpenAPI specs

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -2190,6 +2190,44 @@
         ]
       }
     },
+    "/api/v1/namespaces/{namespace}/worker-count": {
+      "get": {
+        "summary": "CountWorkers is a visibility API to count workers in a specific namespace.",
+        "operationId": "CountWorkers2",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1CountWorkersResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpcStatus"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "namespace",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "query",
+            "description": "Query to filter workers before counting.\nSupported filter fields are the same as in ListWorkersRequest.",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          }
+        ],
+        "tags": [
+          "WorkflowService"
+        ]
+      }
+    },
     "/api/v1/namespaces/{namespace}/worker-deployment-versions/{deploymentVersion.deploymentName}/{deploymentVersion.buildId}": {
       "get": {
         "summary": "Describes a worker deployment version.\nExperimental. This API might significantly change or be removed in a future release.",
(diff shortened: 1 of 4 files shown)


yuandrew (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2684,6 +2684,18 @@ message DescribeWorkerResponse {
...
 
+message CountWorkersRequest {
+    string namespace = 1;
+    // Query to filter workers before counting.
+    // Supported filter fields are the same as in ListWorkersRequest.
+    string query = 2;
+}
+
+message CountWorkersResponse {
+    // Approximate number of workers matching the query.
```
Sushisource: More detail on why this is approximate is always useful in this sort of comment

openapi/openapiv2.json
```diff
@@ -2190,6 +2190,44 @@
         ]
       }
     },
+    "/api/v1/namespaces/{namespace}/worker-count": {
```
laurakwhit: @rkannan82 just wanted to check-in on the status of this endpoint as some UI changes depend on it. Are there currently any blockers here?

temporal/api/workflowservice/v1/request_response.proto
rkannan82: My bad, its not approximate.

openapi/openapiv2.json
rkannan82: On it now. Very sorry, I neglected this PR for so long.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3062,6 +3062,21 @@ message DescribeWorkerResponse {
...
+    // Supported filter fields are the same as in ListWorkersRequest.
+    string query = 2;
+    // When true, the count will include system workers that are created implicitly
+    // by the server and not by the user. By default, system workers are excluded.
+    bool include_system_workers = 3;
+}
+
+message CountWorkersResponse {
+    // Number of workers matching the query.
+    int64 count = 1;
```
laurakwhit, line 3077: Does this count include both running and shutting down workers?

ShahabT (approved):

temporal/api/workflowservice/v1/request_response.proto
rkannan82: Yes. Same as what ListWorkers returns. The request proto for CountWorkers contains same query options as ListWorkers.

laurakwhit: Not a blocker, just wondering if it would be at all valuable to have something similar to [CountWorkflowExecutionsRequest](https://github.com/temporalio/api/blob/main/temporal/api/workflowservice/v1/request_response.proto#L1045) so we can visually show the count grouped by worker status similar to the Workflows list view in the UI. Maybe if we do want it grouped in the UI we can just do an additional API call with a query for now, since there are currently only two statuses.

laurakwhit (approved):

PR #730, opened 2026-03-17
Add completed_by_worker_shutdown field to poll responses

rkannan82:
## What changed?
Add `completed_by_worker_shutdown` bool field to `PollWorkflowTaskQueueResponse`, `PollActivityTaskQueueResponse`, and `PollNexusTaskQueueResponse`.

[Server PR](https://github.com/temporalio/temporal/pull/9551)

## Why?
When the server completes a poll because the worker was shut down via `ShutdownWorker`, SDKs currently receive an empty response indistinguishable from a normal poll timeout. This gives SDKs an explicit signal to stop re-polling, and is useful for debugging.

## How did you test it?
- [x] built
- [ ] server-side implementation in separate PR

## Potential risks
None — additive proto field, fully backward compatible.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -14580,6 +14580,10 @@
         "pollerScalingDecision": {
           "$ref": "#/definitions/v1PollerScalingDecision",
           "description": "Server-advised information the SDK may use to adjust its poller count."
+        },
+        "shutdownWorkerRejected": {
+          "type": "boolean",
+          "description": "When true, this empty response was caused by the server rejecting the poll\nbecause the worker has been shut down via the ShutdownWorker API. The SDK\nshould stop polling on this task queue."
         }
       }
     },
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -15356,9 +15356,9 @@
           "$ref": "#/definitions/v1PollerScalingDecision",
           "description": "Server-advised information the SDK may use to adjust its poller count."
         },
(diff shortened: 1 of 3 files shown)


dnr (commented):
Does it make more sense to do this with a specific grpc error code instead? Maybe NotFound or FailedPrecondition? Something that would make old SDKs abort or at least back off?

rkannan82:
> Does it make more sense to do this with a specific grpc error code instead? Maybe NotFound or FailedPrecondition? Something that would make old SDKs abort or at least back off?

We agreed to not introduce a new error as this interferes with older sdks.

Sushisource (commented):
I don't think we should do this. In theory we will only be sending these responses to new SDKs, but as mentioned in the context of using the scaling down decision on empty responses before - I don't like violation the current assumption that empty responses are always purely empty.

We could, instead, attach a header? If the impetus for this is purely debugging, that seems reasonable?

rkannan82:
> I don't think we should do this. In theory we will only be sending these responses to new SDKs, but as mentioned in the context of using the scaling down decision on empty responses before - I don't like violation the current assumption that empty responses are always purely empty.
> 
> We could, instead, attach a header? If the impetus for this is purely debugging, that seems reasonable?

Closing this PR after for the following reason:
1. SDK already knows to not repoll. So this bit is only needed for debugging; don't need at this time.
2. Server-initiated shutdown should be delivered via a worker nexus command.

PR #736, opened 2026-03-18
Add "exposed" operation for system nexus endpoint

spkane31:
**What changed?**
Add `nexusannotations/v1` package to define service operations as exposed for Nexus handlers.

A sample of a code generator that takes use of this command is [here](https://github.com/temporalio/api-go/commit/da6df036bed0ca1b4932ee9583214ac43c489c9c)

**Why?**
Enables the system nexus endpoint work in server repo

**Breaking changes**
NA

**Server PR**
NA


temporal/api/workflowservice/v1/service.proto
```diff
@@ -487,6 +488,8 @@ service WorkflowService {
     // (-- api-linter: core::0136::prepositions=disabled
     //     aip.dev/not-precedent: "With" is used to indicate combined operation. --)
     rpc SignalWithStartWorkflowExecution (SignalWithStartWorkflowExecutionRequest) returns (SignalWithStartWorkflowExecutionResponse) {
+        option (nexus.v1.operation).tags = "exposed";
```
spkane31: For reviewer: this tag can be used by our code generation tool to generate Nexus handler interfaces.

lina-temporal (approved):
I like the annotation approach here, I think that even with the downside of a few of the fields being irrelevant in Nexus operation context, it's still the best option for maintainability and adoption across our APIs. Think we're still in discussion on the nature of the codegen, but this part LGTM.

nexusannotations/v1/options.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
+
+package nexusannotations.v1;
+
+import "google/protobuf/descriptor.proto";
+
+option go_package = "github.com/bergundy/nexus-proto-annotations/go/nexus/v1";
```
bergundy: The package name here would be different depending on the verdict of where we would want to put these annotations. I don't think we should reference anything in my personal repo.

bergundy: Also note that the annotations already exist in the referenced GH repo, I had to inline them here because the package name `nexus` was conflicting and confusing our protoc generation pipeline.

bergundy (commented):
We have some questions to answer before we can merge this.

Makefile
```diff
@@ -95,6 +95,11 @@ buf-install:
 	printf $(COLOR) "Install/update buf..."
 	go install github.com/bufbuild/buf/cmd/buf@v1.27.0
 
+##### Sync external proto dependencies #####
+sync-nexus-annotations:
```
bergundy, line 99: Why don't we have the same make target for google? Seems like we can sync both. But probably don't sync google as part of this PR if there are any changes there, better to do that separately.

bergundy (approved):
Approved, but let's make sure this is at minimum prototyped in the UI and SDK before merging.

cmd/protoc-gen-nexus-rpc-yaml/generator.go
bergundy, line 1: This should go in nexus-rpc-gen eventually but it's okay if it starts out in this repo.

cmd/protoc-gen-nexus-rpc-yaml/generator.go
```diff
@@ -0,0 +1,203 @@
...
+			for _, m := range svc.Methods {
+				opts, ok := m.Desc.Options().(*descriptorpb.MethodOptions)
+				if !ok || opts == nil {
+					continue
+				}
+				if !proto.HasExtension(opts, nexusannotationsv1.E_Operation) {
+					continue
+				}
+				opOpts := proto.GetExtension(opts, nexusannotationsv1.E_Operation).(*nexusannotationsv1.OperationOptions)
+				if !slices.Contains(opOpts.GetTags(), "exposed") {
```
bergundy: See what https://github.com/bergundy/protoc-gen-go-nexus did here, don't hardcode `exposed` in the tool, it should be passed as include and exclude tag arguments.

cmd/protoc-gen-nexus-rpc-yaml/generator.go
```diff
@@ -0,0 +1,203 @@
...
+					langRefs(f.Desc, m.Input.Desc),
+					langRefs(f.Desc, m.Output.Desc),
+				)
+			}
+		}
+	}
+	if err := writeFile(gen, "nexus/temporal-json-schema-models-nexusrpc.yaml", nexusDoc); err != nil {
+		return err
+	}
+	return writeFile(gen, "nexus/temporal-proto-models-nexusrpc.yaml", langsDoc)
```
bergundy: These paths needs to be parameters, not hardcoded.

cmd/protoc-gen-nexus-rpc-yaml/generator.go
bergundy: I'm fine with what you have here because it gets the job done but the task was to add a proper tool that anybody can use, not something that is specific to our API repo.

cmd/protoc-gen-nexus-rpc-yaml/generator.go
```diff
@@ -0,0 +1,203 @@
...
+// openAPIRef returns the nexus-rpc-gen multi-file $ref string for a message type,
+// referencing the openapiv3.yaml components/schemas entry. The path is relative to nexus/.
+//
+// Schema key convention used by protoc-gen-openapi (v3):
+//
+//	{MessageName}  (no package prefix)
+//
+// e.g. message "SignalWithStartWorkflowExecutionRequest"
+// → "../openapi/openapiv3.yaml#/components/schemas/SignalWithStartWorkflowExecutionRequest"
+func openAPIRef(msg protoreflect.MessageDescriptor) string {
```
bergundy: When you write a proper protoc plugin that is reusable, you wouldn't be able to rely on openapi schemas being available in these paths.

spkane31: I've duplicated this code into a draft PR in nexus-rpc/nexus-rpc-gen while we sort out that story: https://github.com/nexus-rpc/nexus-rpc-gen/pull/36

cmd/protoc-gen-nexus-rpc-yaml/generator.go
spkane31: I am tracking this work: https://github.com/nexus-rpc/nexus-rpc-gen/issues/2, this is a prototype to make sure we have the generated code we want. It will be easily to copy this to the nexus-rpc/nexus-rpc-gen once we have that set up

cmd/protoc-gen-nexus-rpc-yaml/generator.go
spkane31: Made into params

tconley1428 (approved):

Makefile
spkane31: There are a lot of changes in the google/api definitions that we don't have in this repo, do we want to add all of those all at once?
https://buf.build/googleapis/googleapis/docs/main%3Agoogle.api

bergundy: No, separate PR please.

cmd/protoc-gen-nexus-rpc-yaml/generator.go
bergundy, line 1: Feels like this file deserves the most scrutiny but if we end up not using the JSON schema based models it's not worth putting more effort into it.

Where did you come up with this logic? Did you copy it from grpc-gateway?

nexusannotations/v1/options.proto
```diff
@@ -0,0 +1,29 @@
...
+}
+
+message OperationOptions {
+  // Nexus operation name (defaults to proto method name).
+  string name = 1;
+  // Tags to attach to the operation. Used by code generators to include and exclude operations.
+  repeated string tags = 2;
+}
+
+message ServiceOptions {
```
bergundy, line 24: We have tags on services too but you didn't support excluding or including based on tags.

That's fine for now, but the "official" plugin would need to have all of this functionality, similar to what `bergundy/protoc-gen-go-nexus` does.

bergundy (approved):
Looks good.

cmd/protoc-gen-nexus-rpc-yaml/generator.go
spkane31: This is mostly generated with claude

PR #742, opened 2026-03-24
Add Callbacks and Links to Workflow Update

Quinn-With-Two-Ns:

**What changed?**
Add Callbacks and Links to Workflow Update.

**Why?**
Will allow starting async updates from a Nexus Operation Handler

**Breaking changes**

**Server PR**

https://github.com/temporalio/temporal/pull/9614


temporal/api/history/v1/message.proto
```diff
@@ -873,6 +880,8 @@ message WorkflowExecutionOptionsUpdatedEventAttributes {
     // Priority override upserted in this event. Represents the full priority; not just partial fields.
     // Ignored if nil.
     temporal.api.common.v1.Priority priority = 6;
+
+    map<string, WorkflowUpdateOptionsUpdate> workflow_update_options = 7;
```
bergundy: Is the key here the update ID? I slightly prefer putting the update ID in the `WorkflowUpdateOptionsUpdate` event so it's clear what the field just by the name. That would turn this field's type to `repeated WorkflowUpdateOptionsUpdate`.

temporal/api/history/v1/message.proto
```diff
@@ -909,6 +918,9 @@ message WorkflowExecutionUpdateAcceptedEventAttributes {
     // The message payload of the original request message that initiated this
     // update.
     temporal.api.update.v1.Request accepted_request = 4;
+    // Completion callbacks attached when this update was started.
+    // TODO(quinn) maybe not needed
```
bergundy: Looks like it's needed? My only concern is that the event time here is for when the update was accepted but there's no indication of when the callback was received for attachment.

temporal/api/history/v1/message.proto
Quinn-With-Two-Ns: Yeah I thought we would need to index into this but we never ended up needing to

temporal/api/common/v1/message.proto
```diff
@@ -240,9 +241,20 @@ message Link {
...
     }
 
+    // A link to a workflow execution. This is a more general version of WorkflowEvent that doesn't specify a 
+    // particular event within the workflow, useful when you want to link to a workflow but there is no particular event to link to,
+    // such as a Query or a Rejected Update.
+    message Workflow {
+        string namespace = 1;
+        string workflow_id = 2;
+        string run_id = 3;
+        temporal.api.enums.v1.LinkReason reason = 4;
```
bergundy: I kinda prefer a string here, it's a bit more overhead but doesn't require translating enums to strings and can flow newer values through older SDK and server versions that don't have up-to-date protos.

temporal/api/history/v1/message.proto
Quinn-With-Two-Ns: We currently don't keep any time for when the update was received, so if we don't care about when the update was received I don't think a callback changes that.

Quinn-With-Two-Ns: We already attach the callback to the request so we can remove the callback here.

temporal/api/common/v1/message.proto
Quinn-With-Two-Ns: I don't have a strong opinion here. I thought enums were just easier to keep consistent across the SDK, Server and UI.

temporal/api/history/v1/message.proto
bergundy: Agree knowing the request time could be useful and is a separate concern.

bergundy (approved):
Nothing blocking AFAIC

yuandrew (approved):

PR #746, opened 2026-03-27
feat: add worker runtime info to heartbeat

chris-olszewski:
Add worker runtime information to worker heartbeat. Will be used to track both runtime version (Node 20 vs 24) and different runtimes (Node vs Bun).

**Why?**
Record which runtimes are being used in the SDK. Needed to complete https://github.com/temporalio/sdk-typescript/issues/1964

**Breaking changes**
No

**Server PR**

### DIFF
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -130,6 +130,9 @@ message WorkerHeartbeat {
 
   // Storage drivers in use by this SDK.
   repeated StorageDriverInfo drivers = 24;
+
+  // Information about the language runtime executing the worker.
+  RuntimeInfo runtime_info = 25;
 }
 
 // Detailed worker information.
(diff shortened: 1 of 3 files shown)


temporal/api/worker/v1/message.proto
```diff
@@ -193,3 +199,13 @@ message StorageDriverInfo {
...
   string type = 1;
 }
+
+// Information about the language runtime executing the worker.
+message RuntimeInfo {
+  // Runtime name
+  string name = 1;
+  // Runtime version
+  string version = 2;
+  // Name of a known provider
```
tconley1428: What does provider mean here?

chris-olszewski: Hosting provider e.g. Vercel

yuandrew (approved):
LGTM, but do we typically want to wait for server implementation before merging API PRs?

PR #755, opened 2026-04-04
Fix first_execution_run_id comments on Cancel and Terminate RPCs

dandavison:
The proto comments on `first_execution_run_id` in `RequestCancelWorkflowExecutionRequest` and `TerminateWorkflowExecutionRequest` claimed that when a run ID is set on `workflow_execution`, validation would demand that that run descends from `first_execution_run_id`. In reality, the server unconditionally ignores the `workflow_execution` run ID when `first_execution_run_id` is set, resolves to the most recent execution, and then checks that *that* descends from `first_execution_run_id`:

[`requestcancelworkflow/api.go:32-34`](https://github.com/temporalio/temporal/blob/f8bbf7c7803874fe454c6db270dfbe2f69bd9e79/service/history/api/requestcancelworkflow/api.go#L32-L34)
```go
if len(firstExecutionRunID) != 0 {
    runID = ""
}
```

[`terminateworkflow/api.go:33-35`](https://github.com/temporalio/temporal/blob/f8bbf7c7803874fe454c6db270dfbe2f69bd9e79/service/history/api/terminateworkflow/api.go#L33-L35)
```go
if len(request.FirstExecutionRunId) != 0 {
    runID = ""
}
```

The `UpdateWorkflowExecutionRequest` comment is left unchanged because its [server implementation](https://github.com/temporalio/temporal/blob/f8bbf7c7803874fe454c6db270dfbe2f69bd9e79/service/history/api/updateworkflow/api.go#L77-L81) does pass the `workflow_execution` run ID through to execution resolution, and validates `first_execution_run_id` separately [after loading the execution](https://github.com/temporalio/temporal/blob/f8bbf7c7803874fe454c6db270dfbe2f69bd9e79/service/history/api/updateworkflow/api.go#L109-L111).

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -9156,6 +9156,19 @@
       },
       "description": "Target a worker polling on a Nexus task queue in a specific namespace."
     },
+    "ExecuteMultiOperationRequestOperation": {
+      "type": "object",
+      "properties": {
+        "startWorkflow": {
+          "$ref": "#/definitions/v1StartWorkflowExecutionRequest",
+          "title": "Additional restrictions:\n- setting `cron_schedule` is invalid\n- setting `request_eager_execution` is invalid\n- setting `workflow_start_delay` is invalid"
+        },
+        "updateWorkflow": {
+          "$ref": "#/definitions/v1UpdateWorkflowExecutionRequest",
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
Huh. Well that's fun. Do you think we need to update doctrings in SDKs?

PR #759, opened 2026-04-07
Add links and callbacks support for standalone activity

fretz12:
**What changed?**
  - Added completion_callbacks and links fields to StartActivityExecutionRequest.                                                                                                                                                                                                                                                                                                                                           
  - Added link field to StartActivityExecutionResponse.                                                                                                                                                                                                                                                                                                                                                                                     
  - Added callbacks field to DescribeActivityExecutionResponse.
  - Added links field to ActivityExecutionInfo.                                                                                                                                                                                                                                                                                                                                                                                             
  - Added new Link.Activity variant to the Link message type.           
  - Added new CallbackInfo base message                                                                                                                                                                                                                                                                                                                                                                
  - Updated OpenAPI v2/v3 specs to reflect the above changes.

**Why?**
To support links and callbacks for standalone activity executions (SAA), bringing feature parity with workflow-based activity scheduling.

### DIFF
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -162,7 +162,7 @@ message ActivityExecutionInfo {
     // Set if activity cancelation was requested.
     string canceled_reason = 32;
 
-    // Links to be associated with the activity.
+    // Links associated with the activity.
     repeated temporal.api.common.v1.Link links = 33;
 }
 
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -15,7 +15,7 @@ import "google/protobuf/timestamp.proto";
 import "temporal/api/common/v1/message.proto";
 import "temporal/api/deployment/v1/message.proto";
 import "temporal/api/enums/v1/activity.proto";
-import "temporal/api/enums/v1/common.proto";
+import "temporal/api/callback/v1/message.proto";
 import "temporal/api/enums/v1/workflow.proto";
 import "temporal/api/failure/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
@@ -211,26 +211,8 @@ message CallbackInfo {
         }
     }
 
-    // Information on how this callback should be invoked (e.g. its URL and type).
-    temporal.api.common.v1.Callback callback = 1;
     // Trigger for this callback.
-    Trigger trigger = 2;
-    // The time when the callback was registered.
-    google.protobuf.Timestamp registration_time = 3;
-
-    // The current state of the callback.
-    temporal.api.enums.v1.CallbackState state = 4;
-    // The number of attempts made to deliver the callback.
-    // This number represents a minimum bound since the attempt is incremented after the callback request completes.
-    int32 attempt = 5;
-
-    // The time when the last attempt completed.
-    google.protobuf.Timestamp last_attempt_complete_time = 6;
-    // The last attempt's failure, if any.
-    temporal.api.failure.v1.Failure last_attempt_failure = 7;
-    // The time when the next attempt is scheduled.
-    google.protobuf.Timestamp next_attempt_schedule_time = 8;
-
-    // If the state is BLOCKED, blocked reason provides additional information.
-    string blocked_reason = 9;
+    Trigger trigger = 1;
+    // Common callback info.
+    temporal.api.callback.v1.CallbackInfo callback_info = 2;
 }
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -161,6 +161,9 @@ message ActivityExecutionInfo {
 
     // Set if activity cancelation was requested.
     string canceled_reason = 32;
+
+    // Links to be associated with the activity.
+    repeated temporal.api.common.v1.Link links = 33;
 }
 
 // Limited activity information returned in the list response.
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -214,5 +214,5 @@ message CallbackInfo {
     // Trigger for this callback.
     Trigger trigger = 1;
     // Common callback info.
-    temporal.api.callback.v1.CallbackInfo callback_info = 2;
+    temporal.api.callback.v1.CallbackInfo info = 2;
 }
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -15,6 +15,7 @@ import "google/protobuf/timestamp.proto";
 import "temporal/api/common/v1/message.proto";
 import "temporal/api/deployment/v1/message.proto";
 import "temporal/api/enums/v1/activity.proto";
+import "temporal/api/enums/v1/common.proto";
 import "temporal/api/enums/v1/workflow.proto";
 import "temporal/api/failure/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
@@ -162,7 +163,7 @@ message ActivityExecutionInfo {
     // Set if activity cancelation was requested.
     string canceled_reason = 32;
 
-    // Links associated with the activity.
+    // Links to related entities, such as the entity that started this activity.
     repeated temporal.api.common.v1.Link links = 33;
 }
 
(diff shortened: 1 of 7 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2931,13 +2931,22 @@ message StartActivityExecutionRequest {
...
+    // Links to be associated with the activity.
+    repeated temporal.api.common.v1.Link links = 21;
 }
 
 message StartActivityExecutionResponse {
     // The run ID of the activity that was started - or used (via ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING).
     string run_id = 1;
     // If true, a new activity was started.
     bool started = 2;
+    // Link to the activity event.
```
Quinn-With-Two-Ns:
```suggestion
    // Link to the started activity.
```

I don't think stand alone activities have an "event"?

temporal/api/activity/v1/message.proto
```diff
@@ -161,6 +161,9 @@ message ActivityExecutionInfo {
 
     // Set if activity cancelation was requested.
     string canceled_reason = 32;
+
+    // Links to be associated with the activity.
+    repeated temporal.api.common.v1.Link links = 33;
```
Sushisource:
```suggestion
    // Links associated with the activity.
    repeated temporal.api.common.v1.Link links = 33;
```

Since this is current activity info, they already _are_ associated with the activity

fretz12: Makes sense, thanks

temporal/api/workflowservice/v1/request_response.proto
fretz12: Makes sense, thanks

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2931,13 +2931,22 @@ message StartActivityExecutionRequest {
     temporal.api.sdk.v1.UserMetadata user_metadata = 17;
     // Priority metadata.
     temporal.api.common.v1.Priority priority = 18;
+    // Time to wait before dispatching the first activity.
+    google.protobuf.Duration start_delay = 19;
```
bergundy: I would separate this out to a new PR. It's still unclear whether we will support this or not.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2931,13 +2931,22 @@ message StartActivityExecutionRequest {
...
+    // Callbacks to be called by the server when this activity reaches a terminal state.
+    // Callback addresses must be whitelisted in the server's dynamic configuration.
+    repeated temporal.api.common.v1.Callback completion_callbacks = 20;
+    // Links to be associated with the activity.
```
bergundy: I would make note that callbacks also have links that can be associated with them. These links should only be attached if not included with a callback.

fretz12: Added comment

temporal/api/workflowservice/v1/request_response.proto
fretz12: Removed from PR

bergundy (approved):

Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2974,6 +2982,9 @@ message DescribeActivityExecutionResponse {
 
     // Token for follow-on long-poll requests. Absent only if the activity is complete.
     bytes long_poll_token = 5;
+
+    // Callbacks attached to this activity execution and their current state.
+    repeated temporal.api.workflow.v1.CallbackInfo callbacks = 6;
```
dandavison: This `CallbackInfo` is currently workflow-specific:

```proto
// CallbackInfo contains the state of an attached workflow callback.
message CallbackInfo {
    // Trigger for when the workflow is closed.
    message WorkflowClosed {}

    message Trigger {
        oneof variant {
            WorkflowClosed workflow_closed = 1;
        }
    }
```

We either need to make it more general, or add an activity-specific `CallbackInfo`. I'm thinking we'd want the former -- add an `ActivityClosed` or `ExecutionClosed` trigger. We're going to want to straighten this out, and it would be a breaking API change to do it later.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2931,13 +2931,21 @@ message StartActivityExecutionRequest {
     temporal.api.sdk.v1.UserMetadata user_metadata = 17;
     // Priority metadata.
     temporal.api.common.v1.Priority priority = 18;
+    // Callbacks to be called by the server when this activity reaches a terminal state.
+    // Callback addresses must be whitelisted in the server's dynamic configuration.
+    repeated temporal.api.common.v1.Callback completion_callbacks = 19;
+    // Links to be associated with the activity. Callbacks may also have associated links;
+    // links already included with a callback should not be duplicated here.
+    repeated temporal.api.common.v1.Link links = 20;
```
dandavison, line 2939: The SAA PR only uses links on the callback. Is it necessary/appropriate to add these top-level links in addition to the callback links? If so, could the Nexus team give a refresher on what the respective roles are of the two links (I had it in my head that top-level were deprecated).

temporal/api/activity/v1/message.proto
```diff
@@ -161,6 +161,9 @@ message ActivityExecutionInfo {
 
     // Set if activity cancelation was requested.
     string canceled_reason = 32;
+
+    // Links associated with the activity.
```
dandavison: A reader is going to think initially these are links that in some way refer to this activity. However, they're not -- they're links pointing _to_ other entities associated with the activity; typically the entity that started the activity. Let's clarify this for the reader, here and in the workflow counterpart `HistoryEvent.links`. As an author of a Nexus SDK I found this confusing at first, so there's not much doubt that other users will benefit from some help here.


```suggestion
    // Links to entities associated with this activity, for example
    // to the entity that started this activity.
```

dandavison (commented):
Looks great but I think that we need to make `CallbackInfo` handle Activity before merging.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Ah I totally missed this, this callback info struct is only for workflows. You'll need to create a callback info struct for activities instead.

temporal/api/workflowservice/v1/request_response.proto
bergundy: Not deprecated, I commented on this before. It's possible to link without attaching callbacks.

fretz12: For reference: https://github.com/temporalio/api/pull/759#discussion_r3046811175

temporal/api/activity/v1/message.proto
fretz12: Gotcha, changed here and at HistoryEvent

temporal/api/workflowservice/v1/request_response.proto
fretz12: Good call, added an activity specific cb

temporal/api/activity/v1/message.proto
```diff
@@ -195,3 +199,38 @@ message ActivityExecutionListInfo {
...
+    // The time when the last attempt completed.
+    google.protobuf.Timestamp last_attempt_complete_time = 6;
+    // The last attempt's failure, if any.
+    temporal.api.failure.v1.Failure last_attempt_failure = 7;
+    // The time when the next attempt is scheduled.
+    google.protobuf.Timestamp next_attempt_schedule_time = 8;
+
+    // If the state is BLOCKED, blocked reason provides additional information.
+    string blocked_reason = 9;
+}
```
dandavison, line 218: I think this is the right design; just describing it to make it explicit. So each CHASM component will have its own `CallbackInfo`, the role of which is to describe a callback's state and triggering condition. Since different CHASM components will have their own component-specific sets of possible triggering conditions, we define component-specific `CallbackInfo` types. In practice the only field that differs for now is that `trigger` `oneof`; they'll share identical callback state fields.

dandavison (approved):

temporal/api/activity/v1/message.proto
bergundy: Agree, we should ideally have a single type for callback info since it will be used across multiple archetypes. The trigger is the only archetype specific information.

I'm going to walk back what I wrote yesterday and ask you to create a `temporal.api.callback.v1.CallbackInfo` which has all of the common fields and put only `Trigger` and `temporal.api.callback.v1.CallbackInfo` in this message.

fretz12: I refactored the common fields to `temporal.api.callback.v1.CallbackInfo`. Left workflow callback info as is to avoid breaking changes

bergundy (approved):

temporal/api/activity/v1/message.proto
```diff
@@ -195,3 +199,20 @@ message ActivityExecutionListInfo {
...
+    message Trigger {
+        oneof variant {
+            ActivityClosed activity_closed = 1;
+        }
+    }
+
+    // Trigger for this callback.
+    Trigger trigger = 1;
+    // Common callback info.
+    temporal.api.callback.v1.CallbackInfo callback_info = 2;
```
dandavison: We're thinking let's call this `info`.

fretz12: Renamed to info

dandavison (approved):
LGTM

PR #760, opened 2026-04-08
Add `CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_USE_RAMPING_VERSION`

carlydf:
Add `CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_USE_RAMPING_VERSION`

To enable the use case of Continue-as-new-upgrading a user-specified cohort of Pinned workflows before any other workflows are upgraded. To do this, user would first set Ramp percentage to zero, then tell (likely via signal) their cohort of workflows to Continue-as-new with this initial behavior, watch that bake, repeat for whatever other cohorts are desired, and finally do a percentage-based ramp which applies to new workflows as well as workflows doing `CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_AUTO_UPGRADE`

No

Does not break the server


temporal/api/enums/v1/workflow.proto
```diff
@@ -178,16 +178,31 @@ enum VersioningBehavior {
...
-    // command. If a Pinned override is inherited by the new run, and the new run starts with AutoUpgrade
-    // behavior, the base version of the new run will be the Target Version as described above, but the
-    // effective version will be whatever is specified by the Versioning Override until the override is removed.
+    // command. Versioning Override always takes precedence until it's removed manually via UpdateWorkflowExecutionOptions.
     CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_AUTO_UPGRADE = 1;
+
+    // Experimental.
+    // Use the Ramping Version of the workflow's task queue at start time, regardless of the workflow's
+    // Target Version (according to f(workflow_id, ramp_percentage)). After the first workflow task completes,
+    // use whatever Versioning Behavior the workflow is annotated with in the workflow code. If there is no Ramping
```
Sushisource:
```suggestion
    // the workflow will use whatever Versioning Behavior it is annotated with. If there is no Ramping
```

temporal/api/enums/v1/workflow.proto
```diff
@@ -178,16 +178,31 @@ enum VersioningBehavior {
...
+    // Version by the time that the first workflow task is dispatched, it will be sent to the Current Version.
+    //
+    // It is highly discouraged to use this if the workflow is annotated with AutoUpgrade behavior, because
+    // this setting ONLY applies to the first task of the workflow. If, after the first task, the workflow
+    // is AutoUpgrade, it will behave like a normal AutoUpgrade workflow and go to the Target Version, which
+    // may be the Current Version instead of the Ramping Version.
+    //
+    // Note that if the previous workflow had a Pinned override, that override will be inherited by the
```
Sushisource:
```suggestion
    // Note that if the workflow being continued has a Pinned override, that override will be inherited by the
```

Sushisource (approved):

temporal/api/deployment/v1/message.proto
```diff
@@ -316,9 +317,14 @@ message RoutingConfig {
...
+// Also used for Upgrade-on-CaN behaviors AutoUpgrade and UseRampingVersion.
 message InheritedAutoUpgradeInfo {
     // The source deployment version of the parent/previous workflow.
     temporal.api.deployment.v1.WorkerDeploymentVersion source_deployment_version = 1;
     // The revision number of the source deployment version of the parent/previous workflow.
     int64 source_deployment_revision_number = 2;
+    // The initial continue-as-new versioning behavior.
+    // Not passed to children or to future continue-as-new.
+    // Only used for the initial task of this run and the initial task of any retries of this run.
+    temporal.api.enums.v1.ContinueAsNewVersioningBehavior initial_versioning_behavior = 3;
```
ShahabT: should we say UNSPECIFIED means AUTO_UPGRADE for backwards compatibility with current SDK?

ShahabT (approved):

temporal/api/deployment/v1/message.proto
Shivs11: oooooo good point i think

Shivs11 (approved):

temporal/api/deployment/v1/message.proto
carlydf: Do you mean compatibility with the current server? The SDK does not populate `InheritedAutoUpgradeInfo`, the server does. 

But yeah, essentially the default "continue-as-new-initial-versioning-behavior" is AutoUpgrade if `InheritedAutoUpgradeInfo` is non-nil and this field is unspecified. Calling that out would be for compatibility with older server generated history events, not with older SDK

carlydf: I added a note on the two fields, hopefully it makes sense!

PR #761, opened 2026-04-09
Add link field to propagate backlink in signal and signal-with-start responses

long-nt-tran:
**What changed?**

Added `link` fields on `SignalWorkflowExecutionResponse` and `SignalWithStartWorkflowExecutionRequest`. I think a singular link should suffice but if reviewers feel otherwise (especially w.r.t. `SignalWithStartWorkflowExecutionRequest`) let me know.

**Why?**

To propagate backlinks on signal and signal-with-start executions back to the caller.

**Breaking changes**

Not a breaking change (unused field)

**Server PR**

https://github.com/temporalio/temporal/pull/9897

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -612,6 +612,9 @@ message WorkflowExecutionSignaledEventAttributes {
     bool skip_generate_workflow_task = 5 [deprecated = true];
     // When signal origin is a workflow execution, this field is set.
     temporal.api.common.v1.WorkflowExecution external_workflow_execution = 6;
+    // The request ID of the Signal request, used by the server to attach this to
+    // the correct Event ID when generating link.
+    string request_id = 7;
 }
 
 message WorkflowExecutionTerminatedEventAttributes {
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -793,6 +793,10 @@ message SignalWorkflowExecutionRequest {
 }
 
 message SignalWorkflowExecutionResponse {
+    // Link to be associated with the WorkflowExecutionSignaled event.
+    // Added on the response to propagate the backlink.
+    // Available from Temporal server 1.31 and up.
+    temporal.api.common.v1.Link link = 1;
 }
 
 message SignalWithStartWorkflowExecutionRequest {
@@ -865,6 +869,10 @@ message SignalWithStartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 2;
+    // Link to be associated with the WorkflowExecutionSignaled event.
+    // Added on the response to propagate the backlink.
+    // Available from Temporal server 1.31 and up.
+    temporal.api.common.v1.Link signal_link = 3;
 }
 
 message ResetWorkflowExecutionRequest {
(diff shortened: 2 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -865,6 +868,9 @@ message SignalWithStartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 2;
+    // Link to be associated with the WorkflowExecutionSignaled event.
+    // Added on the response to propagate the backlink.
+    temporal.api.common.v1.Link link = 3;
```
Quinn-With-Two-Ns:
```suggestion
    temporal.api.common.v1.Link signal_link = 3;
```
We may also want to link to the workflow started event at some point as well

long-nt-tran: that's a good point, let me just add both

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -865,6 +868,11 @@ message SignalWithStartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 2;
+    // Link to be associated with the WorkflowExecutionSignaled event.
+    // Added on the response to propagate the backlink.
+    temporal.api.common.v1.Link signal_link = 3;
+    // Workflow start link to be associated with the WorkflowExecutionStarted event.
```
bergundy: Would this be propagated back to the caller if the workflow was already started? Seems like that would be confusing. I would call out the behavior in the docstring.

Quinn-With-Two-Ns: IMO it should behave like the workflow event link on a normal start workflow, so link to the start if the workflow is started, link to the workflow options update if a callback was attached after the workflow was started.

long-nt-tran: commenting for posterity: chatted w/ both offline and we decided to add this in a later PR

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -865,6 +868,9 @@ message SignalWithStartWorkflowExecutionResponse {
     string run_id = 1;
     // If true, a new workflow was started.
     bool started = 2;
+    // Link to be associated with the WorkflowExecutionSignaled event.
+    // Added on the response to propagate the backlink.
```
bergundy, line 873:
```suggestion
    // Added on the response to propagate the backlink.
    // Available from Temporal server 1.31 and up.
```

bergundy (approved):
Approved with a small comment.

Quinn-With-Two-Ns (approved):

THardy98 (approved):

temporal/api/history/v1/message.proto
```diff
@@ -612,6 +612,9 @@ message WorkflowExecutionSignaledEventAttributes {
     bool skip_generate_workflow_task = 5 [deprecated = true];
     // When signal origin is a workflow execution, this field is set.
     temporal.api.common.v1.WorkflowExecution external_workflow_execution = 6;
+    // The request ID of the Signal request, used by the server to attach this to
+    // the correct Event ID when generating link.
+    string request_id = 7;
```
long-nt-tran, line 617: commenting for posterity: discussed offline w/ server, this change (post-server-approval) is ok, [see this comment for more details](https://github.com/temporalio/temporal/pull/9897#issuecomment-4240391884)

PR #763, opened 2026-04-10
Add new task queue kind to send worker commands

rkannan82:
## What

Add `TASK_QUEUE_KIND_WORKER_COMMANDS` (value 3) to `TaskQueueKind` enum.

## Why

Worker-commands task queues are used for server-to-worker communication (e.g. activity cancellations). They need to be distinguished from user-defined queues so the server can handle them differently in metrics and other areas. SDKs will set this kind when polling worker-commands queues.

## How did you test it?

Proto-only change, no runtime behavior yet.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dnr (commented):
We need a bigger server side design discussion before we add new kinds. That has lots of implications. We need to look at all the different types of ephemeral queues we want to support and figure out what properties are needed in what combinations.

ShahabT:
@dnr what you're saying makes sense, but I don't know why adding this WORKER_COMMANDS kind should be blocked by it. we know already that we want to do some custom things about this particular kind of task queues in server so better to pass the info in a clean and formal way rather than string prefix checks.

dnr:
Sure, but we should think about the scope of this new kind. E.g. could we/should we use the same kind for other ephemeral queue types

rkannan82:
> Sure, but we should think about the scope of this new kind. E.g. could we/should we use the same kind for other ephemeral queue types

Proposed server implementation:
- Define ephemeral property on the partition interface: https://github.com/temporalio/temporal/pull/9900 
- Implement the new kind: https://github.com/temporalio/temporal/pull/9899

Sushisource (approved):
I'm on board with this from an SDK perspective but I make no comments about server. I don't want to bypass David's concerns.

temporal/api/enums/v1/task_queue.proto
```diff
@@ -31,6 +31,10 @@ enum TaskQueueKind {
     //
     // Sticky queues are only for workflow tasks. There are no sticky task queues for activities.
     TASK_QUEUE_KIND_STICKY = 2;
+    // A worker-commands task queue is used for server-to-worker communication (e.g. activity
+    // cancellations). These queues are per-worker and only exist for the lifetime of the worker
+    // process.
+    TASK_QUEUE_KIND_WORKER_COMMANDS = 3;
```
dnr, line 38: add a comment about allowable types? can these be any type?

dnr (approved):
We talked and this is good from the api pov

temporal/api/enums/v1/task_queue.proto
rkannan82: Are you referring to task queue type? From what I see, this is set by the server based on the RPC being called. So I have just updated the comment to say this queue kind is used with PollNexusTaskQueue.

dnr: Sometimes it's implicit in the RPC, sometimes it's not (e.g. DescribeTaskQueue)

I mean similar to how TASK_QUEUE_KIND_STICKY is only valid with TASK_QUEUE_TYPE_WORKFLOW, we could say TASK_QUEUE_KIND_WORKER_COMMANDS is only used with TASK_QUEUE_TYPE_NEXUS

rkannan82: Done.

PR #766, opened 2026-04-17
change the config propagation design of ts-config

feiyang3cat:
**What changed?**
(1) Only allowed the flexibility of disable_propagation for child workflows
(2) Updated comments to clarify the behavior of bound in TimeSkippingConfig

**Why?**
mainly for simplicity: 
- Ensure the default policy reflects the most common use case → (2)
- Eliminate unnecessary flexibility → (1)

For reviewers: check the inline comment for detailed explanation

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -608,14 +608,18 @@ message TimeSkippingConfig {
     //
     // The bound is propagated to transitively related workflows differently
     // depending on the type and semantics of the related workflow:
-    // - For child workflows, the bound is propagated with the same value as the parent workflow.
-    //   If duration-based bound is used, the child workflow will have the full duration as the bound.
-    //   If target-time based bound is used, the child workflow will have the same target time as the parent workflow.
-    // - For continue-as-new workflows, if a duration-based bound is used, the remaining
-    //   duration is propagated, so the continue-as-new workflow shares the initial bound
-    //   with the parent workflow.
-    // - For reset workflows, the entire TimeSkippingConfig is restored to the value it
-    //   had at the event this workflow is reset to.
+    // - Child workflows: the bound is propagated with the same value as the
+    //   parent workflow, and the child begins execution at the parent's virtual
+    //   time. If the bound is duration-based, each child calculates its skipped
+    //   duration independently from its own start.
+    //
+    // - Continue-as-new workflows: the bound is propagated with the same value
+    //   as the previous workflow. If the bound is duration-based, the
+    //   continue-as-new workflow shares the accumulated duration with the
+    //   previous workflow.
+    //
+    // - Reset workflows: the entire TimeSkippingConfig is restored to the value
+    //   it had at the event this workflow was reset to.
     oneof bound {
 
         // Maximum total virtual time that can be skipped.
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -594,32 +594,27 @@ message TimeSkippingConfig {
     // Enables or disables time skipping for this workflow execution.
     bool enabled = 1;
 
-    // By default, the time skipping configuration is propagated to transitively related workflows.
-    // If set, transitively related workflows will be started with time skipping disabled.
-    bool disable_propagation = 2;
+    reserved 2;
+    reserved "disable_propagation";
 
-    // Optional bound that limits how long time skipping remains active.
-    // Once the bound is reached, time skipping is automatically disabled.
-    // It can later be re-enabled via UpdateWorkflowExecutionOptions.
+    // By default, the time skipping configuration (enabled and bound) is propagated to child workflows,
+    // using the same values as the parent workflow.
+    // If set, child workflows will be started with time skipping disabled.
+    // Regardless of this field, the start time of the child workflow uses the parent's virtual time.
+    bool disable_child_workflow_propagation = 3;
+
+    // Optional bound that limits how far virtual time can advance while time skipping is active.
+    // Once the bound is reached, time skipping is automatically disabled,
+    // but can be re-enabled via UpdateWorkflowExecutionOptions.
     //
-    // This is particularly useful in testing scenarios where workflows
-    // are expected to receive signals, updates, or other events while
-    // timers are in progress.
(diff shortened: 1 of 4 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -608,10 +606,20 @@ message TimeSkippingConfig {
     // are expected to receive signals, updates, or other events while
     // timers are in progress.
     //
-    // This bound is not propagated to transitively related workflows. 
-    // When bound is also needed for transitively related workflows,
-    // it is recommended to set disable_propagation to true 
-    // and configure TimeSkippingConfig explicitly for transitively related workflows.
```
feiyang3cat, line 614: Reset workflows & continue-as-new don't produce another workflow execution the way child workflows do. Instead, they either fix the current workflow (reset) or extend the current wf for technical reasons (continue-as-new).

Therefore:

(1) The only flexibility to disable propagation applies to child workflows. **disable_propagation → rename to disable_child_propagation**

(2) Reset workflow shouldn't be highlighted here — reset rewinds execution to a specific event, and what happens next depends entirely on the event history. (**removed to make the comments concise**)

(3) For continue-as-new, propagation cannot be flexibly configured. The new workflow execution is treated as an extension of the original — **they share the same bound. This should be called out explicitly in the comments.**

temporal/api/workflow/v1/message.proto
```diff
@@ -592,26 +592,29 @@ message WorkflowExecutionOptions {
...
+    reserved "disable_propagation";
 
-    // Optional bound that limits how long time skipping remains active.
-    // Once the bound is reached, time skipping is automatically disabled.
-    // It can later be re-enabled via UpdateWorkflowExecutionOptions.
+    // By default, the time skipping configuration (enabled and bound) is propagated to child workflows,
+    // using the same values as the parent workflow.
+    // If set, child workflows will be started with time skipping disabled.
+    // Regardless of this field, the start time of the child workflow uses the parent's virtual time.
+    bool disable_child_workflow_propagation = 3;
```
yycptt, line 604: lol interesting that 3 is not used before.

temporal/api/workflow/v1/message.proto
```diff
@@ -592,26 +592,29 @@ message WorkflowExecutionOptions {
...
-    // are expected to receive signals, updates, or other events while
-    // timers are in progress.
+    // This is useful in testing scenarios where a workflow is expected to receive
+    // signals, updates, or other external events while timers are in progress.
     //
-    // This bound is not propagated to transitively related workflows. 
-    // When bound is also needed for transitively related workflows,
-    // it is recommended to set disable_propagation to true 
-    // and configure TimeSkippingConfig explicitly for transitively related workflows.
+    // Bound scope:
```
yycptt, line 613: Probably also clarify the behavior of bound on reset. 
Will, for example, the skipped/elapsed duration get reset as well?

yycptt (approved):

temporal/api/workflow/v1/message.proto
feiyang3cat: yes, the 3 was somehow saved by fate

temporal/api/workflow/v1/message.proto
feiyang3cat: For this part, I haven't found a clean way to document the full "propagation" behavior without being verbose or exposing too many internal details. As we agreed a sound design ties time skipping propagation closely to the type of workflow execution being triggered, and there are roughly three categories:

(1) Feature-wise propagation — config and virtual time are carried over to a new, distinct workflow execution:
- **Child workflows**
- _Nexus_ (only if the Nexus operation maps to a workflow execution — so we drop this )

(2) Only implementation-wise propagation:
- **Reset**: returns execution to a point in the current execution's event history. The resulting config and virtual time depend on the target event and the state of the execution at that time.
- **Continue-as-new**: will be treated as one extended execution run, and all CANs share the config and inherit the virtual time from previous one;
- **Retry**: won't inherit the virtual time, and only inherit the initial configuration at the time previous workflow started 

(3) Grey area
- Cron / Scheduler: best strategy TBD

So IMH, we postpone add an explanation of this whole strategy until all of them are soundly completed.

feiyang3cat: alternatively, we may hold this pr until I finish the propagation part of the project

temporal/api/workflow/v1/message.proto
```diff
@@ -592,26 +592,29 @@ message WorkflowExecutionOptions {
...
     // Enables or disables time skipping for this workflow execution.
-    // By default, this field is propagated to transitively related workflows (child workflows/start-as-new/reset) 
-    // at the time they are started.
-    // Changes made after a transitively related workflow has started are not propagated.
     bool enabled = 1;
 
-    // If set, the enabled field is not propagated to transitively related workflows.
-    bool disable_propagation = 2;
+    reserved 2;
+    reserved "disable_propagation";
```
Sushisource, line 598: Do you actually need to bother with reserving 2? We never shipped any of this so it should be fine to break

temporal/api/workflow/v1/message.proto
Sushisource: IMO there's kind of no reason in general to have any of these API changes in master until we actually have the MVP working.

That said, we already do, and it's just updating a comment, so not a big deal either way.

But to my point in my other comment about not needing `reserved` - this is kinda why it's easier to just keep API stuff unmerged until we're happy with the implementation.

Sushisource (approved):

temporal/api/workflow/v1/message.proto
feiyang3cat: Get it.

I will hold this PR until the propagation feature set is completed and ready to merge into server main. 
And future updates to API will hold for a while until related features are ready and stable.

We took an approach of merging feature-set by feature-set of this project into main to avoid maintaining a long lasting pr and merging a super huge pr into the server.

feiyang3cat:
close this pr, and merge until the propagation feature set is completed

temporal/api/workflow/v1/message.proto
feiyang3cat: But it seems we cannot pass the lint check without reserving it.

PR #767, opened 2026-04-21
Add start delay field to StartActivityExecutionRequest

fretz12:
**What changed?**
Added a start_delay field to StartActivityExecutionRequest in the proto definition and updated the OpenAPI specs accordingly.

**Why?**
To allow callers to specify a delay before the first standalone activity task is dispatched, matching the existing start delay capability available for workflow-started activities.


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3000,6 +3000,8 @@ message StartActivityExecutionRequest {
     repeated temporal.api.common.v1.Link links = 20;
     // Options for handling conflicts when using ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING.
     temporal.api.common.v1.OnConflictOptions on_conflict_options = 21;
+    // Time to wait before dispatching the first activity.
+    google.protobuf.Duration start_delay = 22;
```
```suggestion
    // Time to wait before dispatching the activity.
    google.protobuf.Duration start_delay = 22;
```

What does "first" mean there? Attempt? I would just elide it

fretz12: That's better

Sushisource (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3000,6 +3000,8 @@ message StartActivityExecutionRequest {
     repeated temporal.api.common.v1.Link links = 20;
     // Options for handling conflicts when using ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING.
     temporal.api.common.v1.OnConflictOptions on_conflict_options = 21;
+    // Time to wait before dispatching the activity.
```
dandavison: nit: this is a bit clearer to me and matches the language in the workflow `start_delay` comments.

```suggestion
    // Time to wait before dispatching the first activity task.
```

dandavison (approved):

temporal/api/workflowservice/v1/request_response.proto
fretz12: I'm going to make a bit clearer. @Sushisource had previously mentioned the first sounded confusing.
```
// Time to wait before dispatching the activity. This delay is not applied to retry attempts.
```
LMK thoughts

fretz12: Per discussion Dan still prefers the task word. "So the more explicit/technical version made sense to me; for "activities" themselves, users are familiar with "starting" them, but even there there's confusion with "scheduling" , so I thought best not to add a 3rd verb"

PR #770, opened 2026-04-22
api changes for time-skipping propagation

feiyang3cat:
**What changed and why?**

Propagation refers to how time-skipping config is carried over to all workflows started by the current workflow: retry, continue-as-new, child workflows, reset, etc.

1. Remove **disable_propagation** and **max_target_time**, because neither of them can be soundly defined when time-skipping propagates transitively across related workflows. One of the many examples:

- max_target_time: as an absolute time point, it may already have passed by the time a retry or downstream workflow starts, making it ambiguous whether to disable or preserve time-skipping — neither is a clearly correct default
- disable_propagation: it is unclear which type of transitively triggered workflow this should apply to, and treating all cases uniformly is unreasonable; we defer exposing this flexibility until user demand and use cases are better understood

2. Add **comments of the propagation behavior** for each feature (retry, continue-as-new, child workflow, reset, cron)
3. Add **comments to the bound field**
4. Add **propagatedSkippedDuration** to time skipping config so that new workflows can inherit the virtual time

### DIFF
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -204,6 +204,10 @@ message WorkflowExecutionStartedEventAttributes {
     // The configuration may be updated after start via UpdateWorkflowExecutionOptions, which
     // will be reflected in the WorkflowExecutionOptionsUpdatedEvent.
     temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 41;
+
+    // The time skipped by the previous execution that started this workflow.
+    // It can happen in cases of child workflows and continue-as-new workflows.
+    google.protobuf.Duration initial_skipped_duration = 42;
 }
 
 // Wrapper for a target deployment version that the SDK declined to upgrade to.
@@ -773,6 +777,9 @@ message StartChildWorkflowExecutionInitiatedEventAttributes {
 
     // The propagated time-skipping configuration for the child workflow.
     temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 21;
+
+    // Propagate the duration skipped to the child workflow.
+    google.protobuf.Duration initial_skipped_duration = 30;
 }
 
 message StartChildWorkflowExecutionFailedEventAttributes {
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -770,6 +770,9 @@ message StartChildWorkflowExecutionInitiatedEventAttributes {
     bool inherit_build_id = 19 [deprecated = true];
     // Priority metadata
     temporal.api.common.v1.Priority priority = 20;
+
+    // The propagated time-skipping config for the child workflow.
+    temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 21;
 }
 
 message StartChildWorkflowExecutionFailedEventAttributes {
--- a/temporal/api/history/v1/message.proto
+++ b/temporal/api/history/v1/message.proto
@@ -771,7 +771,7 @@ message StartChildWorkflowExecutionInitiatedEventAttributes {
     // Priority metadata
     temporal.api.common.v1.Priority priority = 20;
 
-    // The propagated time-skipping config for the child workflow.
+    // The propagated time-skipping configuration for the child workflow.
     temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 21;
 }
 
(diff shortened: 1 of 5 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -590,28 +590,28 @@ message WorkflowExecutionOptions {
 // User timers are not classified as in-flight work and will be skipped over.
 // When time advances, it skips to the earlier of the next user timer or the configured bound, if either exists.
 message TimeSkippingConfig {
```
feiyang3cat, line 605: these reserved ones can be removed
but `buf-breaking step in Makefile:115-117` stops this kind of changes

temporal/api/workflow/v1/message.proto
```diff
@@ -591,43 +591,46 @@ message WorkflowExecutionOptions {
 // and possibly other features added in the future.
 // User timers are not classified as in-flight work and will be skipped over.
 // When time advances, it skips to the earlier of the next user timer or the configured bound, if either exists.
+//
```
feiyang3cat, line 594: cc @yycptt added here but yet still this cannot replace the power of an educational document with examples in our web

temporal/api/workflow/v1/message.proto
```diff
@@ -591,43 +591,45 @@ message WorkflowExecutionOptions {
...
         google.protobuf.Duration max_elapsed_duration = 5;
-
-        // Absolute virtual timestamp at which time skipping is disabled.
-        // Time skipping will not advance beyond this point.
-        google.protobuf.Timestamp max_target_time = 6;
     }
+
+    // If this execution was started by a previous execution that had already skipped some time,
+    // it inherits the accumulated skipped duration from that execution through this field.
+    // This field is set internally by the server and cannot be configured by the user.
```
simvlad: Should this even be here then? Maybe it should be a separate field on StartWorkflowExecutionRequest and WorkflowExecutionStartedEventAttributes. Then you can also avoid all those packing/unpacking/cloning changes.

feiyang3cat: oh, yes. I think this makes sense. Let me change it.

temporal/api/history/v1/message.proto
```diff
@@ -773,6 +777,9 @@ message StartChildWorkflowExecutionInitiatedEventAttributes {
 
     // The propagated time-skipping configuration for the child workflow.
     temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 21;
+
+    // Propagate the duration skipped to the child workflow.
```
feiyang3cat, line 781: need snapshot for child workflow as the parent may still be running
but not needed for CaN as the CaN Event is the closing event of the previous wf

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -199,8 +199,15 @@ message StartWorkflowExecutionRequest {
...
     // Deployment Options of the worker who will process the eager task. Passed when `request_eager_execution=true`.
     temporal.api.deployment.v1.WorkerDeploymentOptions eager_worker_deployment_options = 28;
+
     // Time-skipping configuration. If not set, time skipping is disabled.
     temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 29;
+
+    // If a workflow execution is started by a previous execution (parent-child workflow or continue-as-new) 
+    // that has already skipped some time, the accumulated skipped duration from that execution 
+    // can be passed to the new workflow execution through this field.
+    google.protobuf.Duration initial_skipped_duration = 30;
```
prathyushpv: Do we need this in this request proto? We have internal proto definition for the history service's StartWorkflowExecutionRequest. If SDK's doesn't have to provide this value, we can move this to that proto. For the use cases of StartChildWorkflow and Continue-As-New, we only call this API on history service without going through frontend.

feiyang3cat: I see. this makes sense

feiyang3cat: udpated

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -199,8 +199,10 @@ message StartWorkflowExecutionRequest {
     temporal.api.common.v1.Priority priority = 27;
     // Deployment Options of the worker who will process the eager task. Passed when `request_eager_execution=true`.
     temporal.api.deployment.v1.WorkerDeploymentOptions eager_worker_deployment_options = 28;
+
```
simvlad, line 202: revert

simvlad (approved):

prathyushpv (approved):

temporal/api/workflow/v1/message.proto
Sushisource: Typically the way we do this is you can disable the break check and then re-enable it in an immediate follow-up PR.

Alternatively, not the end of the world to just leave them. 

-->noted

Sushisource (approved):

PR #753, opened 2026-04-03
Add Stand Alone Callbacks

Quinn-With-Two-Ns:
Add Stand Alone Callbacks.

**Why?**
Allow users to implement custom async Nexus operations

**Server PR**
https://github.com/temporalio/temporal/pull/9805

PR #774, opened 2026-04-27
Add Standalone Callbacks (rebase)

chrsmith:
> ℹ️ This PR is the same as https://github.com/temporalio/api/pull/753, only squashed and rebased. It's the same set of changes, modulo trivial edits to doc comments.

**What changed?**

Add the API for standalone callbacks. (AKA Manual Completions for Nexus.)

**Why?**

Standalone callbacks support connecting non-Temporal services to Nexus. e.g. supporting a way for, say, a Slackbot directly complete a Nexus operation rather than needing to add an intermediary workflow.

**Breaking changes**

None, this is purely additive.

**Server PR**

These changes also require updates in the `api-go` package as well, since the use of `oneof` requires some minor codegen tweaks.
https://github.com/temporalio/api-go/pull/255

The server-side changes are here:
https://github.com/temporalio/temporal/pull/10192

### DIFF
--- a/temporal/api/callback/v1/message.proto
+++ b/temporal/api/callback/v1/message.proto
@@ -9,11 +9,114 @@ option java_outer_classname = "MessageProto";
 option ruby_package = "Temporalio::Api::Callback::V1";
 option csharp_namespace = "Temporalio.Api.Callback.V1";
 
+import "google/protobuf/duration.proto";
+import "google/protobuf/empty.proto";
 import "google/protobuf/timestamp.proto";
 
 import "temporal/api/common/v1/message.proto";
 import "temporal/api/enums/v1/common.proto";
 import "temporal/api/failure/v1/message.proto";
+
+// The outcome of a completed callback execution: either success or a failure.
+message CallbackExecutionOutcome {
+    oneof value {
+        // The callback completed successfully.
+        google.protobuf.Empty success = 1;
+        // The failure if the callback completed unsuccessfully.
+        temporal.api.failure.v1.Failure failure = 2;
+    }
+}
+
+// The Nexus completion data that a standalone callback execution will deliver to its target URL.
+// Exactly one of success or failure should be set.
+message CallbackExecutionCompletion {
+    oneof result {
+        // Deliver a successful Nexus operation completion with this result payload.
+        // If set, the callback delivers a successful completion to the target URL.
+        temporal.api.common.v1.Payload success = 1;
+        // Deliver a failed Nexus operation completion with this failure.
+        // If set, the callback delivers a failed completion to the target URL.
+        // If CanceledFailureInfo is set, the target operation is resolved is canceled instead of failed.
+        temporal.api.failure.v1.Failure failure = 2;
+    }
+}
+
+// Information about a standalone callback execution.
+message CallbackExecutionInfo {
+    // Unique identifier of this callback within its namespace.
+    string callback_id = 1;
+
+    // Run ID of the callback execution.
+    string run_id = 2;
+
+    // Information on how this callback should be invoked (e.g. its URL and type).
+    temporal.api.common.v1.Callback callback = 3;
+
+    // A general status for this callback, indicates whether it is currently running or in one of the terminal statuses.
+    temporal.api.enums.v1.CallbackExecutionStatus status = 4;
+
+    // The detailed state of this callback, provides more granular information than the general status.
+    temporal.api.enums.v1.CallbackState state = 5;
+
+    // The number of attempts made to deliver the callback.
+    // This number represents a minimum bound since the attempt is incremented after the callback request completes.
+    int32 attempt = 6;
+
+    // The time when the callback was created/scheduled.
+    google.protobuf.Timestamp create_time = 7;
+
+    // The time when the last attempt completed.
+    google.protobuf.Timestamp last_attempt_complete_time = 8;
+
+    // The last attempt's failure, if any.
+    temporal.api.failure.v1.Failure last_attempt_failure = 9;
+
+    // The time when the next attempt is scheduled.
+    google.protobuf.Timestamp next_attempt_schedule_time = 10;
+
+    // If the state is BLOCKED, provides additional information.
+    string blocked_reason = 11;
+
+    // Time when the callback transitioned to a terminal state.
+    google.protobuf.Timestamp close_time = 12;
+
+    // Search attributes for indexing.
+    temporal.api.common.v1.SearchAttributes search_attributes = 13;
+
+    // Schedule-to-close timeout for this callback.
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration schedule_to_close_timeout = 14;
+
+    // Incremented each time the callback's state is mutated in persistence.
+    int64 state_transition_count = 15;
+}
+
+// Limited callback information returned in the list response.
+message CallbackExecutionListInfo {
+    // Unique identifier of this callback within its namespace.
+    string callback_id = 1;
+
+    // Run ID of the callback execution.
+    string run_id = 2;
+
+    // Only running and terminal statuses appear here. More detailed information in CallbackExecutionInfo but not
+    // available in the list response.
+    temporal.api.enums.v1.CallbackExecutionStatus status = 3;
+
+    // The time when the callback was created/scheduled.
+    google.protobuf.Timestamp create_time = 4;
+
+    // Time when the callback transitioned to a terminal state.
+    google.protobuf.Timestamp close_time = 5;
+
+    // Search attributes from the start request.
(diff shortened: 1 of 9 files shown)


temporal/api/callback/v1/message.proto
```diff
@@ -9,12 +9,116 @@ option java_outer_classname = "MessageProto";
...
+    temporal.api.common.v1.Callback callback = 3;
+
+    // A general status for this callback, indicates whether it is currently running or in one of the terminal statuses.
+    temporal.api.enums.v1.CallbackExecutionStatus status = 4;
+
+    // The detailed state of this callback, provides more granular information than the general status.
+    temporal.api.enums.v1.CallbackState state = 5;
+
+    // The number of attempts made to deliver the callback.
+    // This number represents a minimum bound since the attempt is incremented after the callback request completes.
```
bergundy: We should take a pass at this docstring both here and for Nexus ops. I changed the behavior for Nexus to start from attempt 1 and have a better explanation here:

https://github.com/temporalio/temporal/blob/a05038cf9993ebe043466d3597952e0e29622467/chasm/lib/nexusoperation/proto/v1/operation.proto#L44-L47

temporal/api/callback/v1/message.proto
```diff
@@ -9,12 +9,116 @@ option java_outer_classname = "MessageProto";
...
+    temporal.api.failure.v1.Failure last_attempt_failure = 9;
+
+    // The time when the next attempt is scheduled.
+    google.protobuf.Timestamp next_attempt_schedule_time = 10;
+
+    // If the state is BLOCKED, provides additional information.
+    string blocked_reason = 11;
+
+    // Time when the callback transitioned to a terminal state.
+    google.protobuf.Timestamp close_time = 12;
```
bergundy: nit: could you put this next to `create_time`? Logically those should be grouped.

temporal/api/common/v1/message.proto
```diff
@@ -181,6 +181,8 @@ message Callback {
         string url = 1;
         // Header to attach to callback request.
         map<string, string> header = 2;
+        // Standard token to use for identifying the callback, used for completion.
```
bergundy, line 184: Maybe put a comment here that implementations should _also_ populate the token header field for compatibility with older servers?

temporal/api/enums/v1/common.proto
```diff
@@ -49,6 +49,26 @@ enum CallbackState {
...
+    // Default value, unspecified status.
+    CALLBACK_EXECUTION_STATUS_UNSPECIFIED = 0;
+    // Callback execution is running.
+    CALLBACK_EXECUTION_STATUS_RUNNING = 1;
+    // Callback has succeeded.
+    CALLBACK_EXECUTION_STATUS_SUCCEEDED = 2;
+    // Callback has failed.
+    CALLBACK_EXECUTION_STATUS_FAILED = 3;
+    // Callback was terminated via TerminateCallbackExecution.
+    CALLBACK_EXECUTION_STATUS_TERMINATED = 4;
```
bergundy, line 73: What about timed out?

temporal/api/enums/v1/common.proto
```diff
@@ -49,6 +49,26 @@ enum CallbackState {
     CALLBACK_STATE_SUCCEEDED = 5;
     // Callback is blocked (eg: by circuit breaker).
     CALLBACK_STATE_BLOCKED = 6;
+    // Callback was terminated via TerminateCallbackExecution. Only possible for standalone callbacks.
+    CALLBACK_STATE_TERMINATED = 7;
```
bergundy, line 53: What about timed out?

temporal/api/errordetails/v1/message.proto
```diff
@@ -130,6 +130,14 @@ message ActivityExecutionAlreadyStartedFailure {
     string run_id = 2;
 }
 
+// An error indicating that a callback execution failed to start because a callback with the given
+// callback ID already exists in this namespace.
+message CallbackExecutionAlreadyStartedFailure {
+    string start_request_id = 1;
+    string run_id = 2;
+    string callback_id = 3;
```
bergundy: The other already started failures don't include the business ID, let's not add that here.

temporal/api/nexus/v1/message.proto
```diff
@@ -64,6 +64,8 @@ message StartOperationRequest {
     map<string, string> callback_header = 6;
     // Links contain caller information and can be attached to the operations started by the handler.
     repeated Link links = 7;
+    // Callback token, to uniquely identify the callback as applicable.
```
bergundy, line 67: nit: document the semantics of when to use this field and when to populate the header value.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3544,3 +3545,144 @@ message DeleteNexusOperationExecutionRequest {
...
+    // The run ID of the callback that was started.
+    string run_id = 1;
+}
+
+message DescribeCallbackExecutionRequest {
+    string namespace = 1;
+    // Identifier for the callback
+    string callback_id = 2;
+    // Run ID of the callback execution to describe. If empty, the latest run will be described.
+    string run_id = 3;
```
bergundy, line 3584: Can you check what the semantics of a request with an empty run ID with a long poll token? Is that valid in the SAA code?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3544,3 +3545,144 @@ message DeleteNexusOperationExecutionRequest {
...
+}
+
+message StartCallbackExecutionResponse {
+    // The run ID of the callback that was started.
+    string run_id = 1;
+}
+
+message DescribeCallbackExecutionRequest {
+    string namespace = 1;
+    // Identifier for the callback
```
bergundy: nit: All other fields have a punctuation. I see this is missing in other messages for the callback_id field too.

```suggestion
    // Identifier for the callback.
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3544,3 +3545,144 @@ message DeleteNexusOperationExecutionRequest {
...
+    temporal.api.callback.v1.CallbackExecutionOutcome outcome = 2;
+}
+
+message ListCallbackExecutionsRequest {
+    string namespace = 1;
+    // Max number of executions to return per page.
+    int32 page_size = 2;
+    // Token returned in ListCallbackExecutionsResponse.
+    bytes next_page_token = 3;
+    // Visibility query, see https://docs.temporal.io/list-filter for the syntax.
```
bergundy, line 3629: Please document which search attributes are valid here (and for the count API).

bergundy (commented):
Mostly LGTM

temporal/api/enums/v1/common.proto
chrsmith: If the callback times out, we just use the `STATUS_FAILED` state, and set the `TimeoutFailureInfo` field in the terminal failure proto.

But I'm assuming you'd like to add it, and I cannot think of a compelling argument not to. 😄

temporal/api/workflowservice/v1/request_response.proto
chrsmith: That comment is incorrect. Looking at the implementation and doc comments for SAA, it checks that run_id is present if long_poll_token is set. And it's also called out in the doc comment. Will fix.

temporal/api/workflowservice/v1/request_response.proto
chrsmith: Nice catch, fixed.

chrsmith:
@bergundy PTAL. I rebased with `master`, so the diff from the earlier review is commit
https://github.com/temporalio/api/pull/774/commits/7c4b67edf545110b09a64632eef0aabcaf02823e.

temporal/api/callback/v1/message.proto
```diff
@@ -9,12 +9,116 @@ option java_outer_classname = "MessageProto";
...
+
+    // A general status for this callback, indicates whether it is currently running or in one of the terminal statuses.
+    temporal.api.enums.v1.CallbackExecutionStatus status = 4;
+
+    // The detailed state of this callback, provides more granular information than the general status.
+    temporal.api.enums.v1.CallbackState state = 5;
+
+    // The number of attempts made to deliver the start callback execution request.
+    // This number is approximate. There could be more attempts if callback invocation fails, or fewer
+    // if it was terminated.
```
bergundy:
```suggestion
    // This number is approximate. There could be more attempts if the server crashes before recording the attempt's completion, or fewer
    // if the callback was terminated or timed out after the counter has been incremented and before an attempt could be made.
```

bergundy: Feel free to reword this but the text I added is more accurate.

bergundy (approved):
Left just one small comment.

chrsmith:
This PR is pretty much complete, however we opted to go with a different approach for implementing standalone completions. (And will look into using SAA for this functionality.) This code might be useful down the road, but this particular PR can be closed out.

PR #775, opened 2026-04-28
Add weak consistency option to DescribeNamespaceRequest

simvlad:
**What changed?**
Added weak_consistency boolean field to `DescribeNamespaceRequest`. When set to `true`, the server may serve the response from an eventually-consistent source (e.g. the in-memory namespace registry) instead of reading through to persistence.

**Why?**
`DescribeNamespace` currently always reads through to the persistence store. The new field lets clients opt into eventually-consistent reads where read-after-write semantics aren't required, while preserving the existing strong-consistency contract for operator/admin callers (default behavior unchanged).

**Breaking changes**
None. Additive field with default `false`, preserving current behavior for clients that don't set it.

**Server PR**
https://github.com/temporalio/temporal/pull/10103

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -150,6 +150,13 @@
             "in": "query",
             "required": false,
             "type": "string"
+          },
+          {
+            "name": "weakConsistency",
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
The comment should probably reference that the SDK should use this when fetching namespace capabilities on startup since I believe that's the main reason we're adding this flag

yycptt (approved):

PR #778, opened 2026-04-30
Add API for GetWorkflowExecutionResult

long-nt-tran:
Add frontend API and request/response protos for getting workflow execution result.

Since this is intended to be used via Nexus, the API allows caller to attach callbacks for an ongoing workflow execution.

**Why?**
New API to enable Nexus ops to map more ergonomically in the handler.

**Breaking changes**
Not breaking, new (unused) API

**Server PR**
https://github.com/temporalio/temporal/pull/10173

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -10796,10 +10796,18 @@
       },
       "description": "Target a worker polling on a Nexus task queue in a specific namespace."
     },
-    "GetWorkflowExecutionResultResponseNotCompleted": {
+    "GetWorkflowExecutionResultResponseFailed": {
+      "type": "object",
+      "properties": {
+        "failure": {
+          "$ref": "#/definitions/v1Failure"
+        }
+      }
+    },
+    "GetWorkflowExecutionResultResponseRunning": {
       "type": "object"
     },
-    "GetWorkflowExecutionResultResponseSuccess": {
+    "GetWorkflowExecutionResultResponseSucceed": {
       "type": "object",
       "properties": {
         "result": {
@@ -13108,7 +13116,7 @@
           "description": "The time when the last attempt completed."
         },
         "lastAttemptFailure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "description": "The last attempt's failure, if any."
         },
         "nextAttemptScheduleTime": {
@@ -13171,7 +13179,7 @@
           "description": "The time when the last attempt completed."
         },
         "lastAttemptFailure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "description": "The last attempt's failure, if any."
         },
         "nextAttemptScheduleTime": {
@@ -13211,58 +13219,6 @@
         }
       },
       "description": "When StartWorkflowExecution uses the conflict policy WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING and\nthere is already an existing running workflow, OnConflictOptions defines actions to be taken on\nthe existing running workflow. In this case, it will create a WorkflowExecutionOptionsUpdatedEvent\nhistory event in the running workflow with the changes requested in this object."
-    },
-    "failureV1Failure": {
-      "type": "object",
-      "properties": {
-        "message": {
-          "type": "string"
-        },
-        "source": {
-          "type": "string",
-          "description": "The source this Failure originated in, e.g. TypeScriptSDK / JavaSDK\nIn some SDKs this is used to rehydrate the stack trace into an exception object."
-        },
-        "stackTrace": {
-          "type": "string"
-        },
-        "encodedAttributes": {
-          "$ref": "#/definitions/v1Payload",
-          "description": "Alternative way to supply `message` and `stack_trace` and possibly other attributes, used for encryption of\nerrors originating in user code which might contain sensitive information.\nThe `encoded_attributes` Payload could represent any serializable object, e.g. JSON object or a `Failure` proto\nmessage.\n\nSDK authors:\n- The SDK should provide a default `encodeFailureAttributes` and `decodeFailureAttributes` implementation that:\n  - Uses a JSON object to represent `{ message, stack_trace }`.\n  - Overwrites the original message with \"Encoded failure\" to indicate that more information could be extracted.\n  - Overwrites the original stack_trace with an empty string.\n  - The resulting JSON object is converted to Payload using the default PayloadConverter and should be processed\n    by the user-provided PayloadCodec\n\n- If there's demand, we could allow overriding the default SDK implementation to encode other opaque Failure attributes."
-        },
-        "cause": {
-          "$ref": "#/definitions/failureV1Failure"
-        },
-        "applicationFailureInfo": {
-          "$ref": "#/definitions/v1ApplicationFailureInfo"
-        },
-        "timeoutFailureInfo": {
-          "$ref": "#/definitions/v1TimeoutFailureInfo"
-        },
-        "canceledFailureInfo": {
-          "$ref": "#/definitions/v1CanceledFailureInfo"
-        },
-        "terminatedFailureInfo": {
-          "$ref": "#/definitions/v1TerminatedFailureInfo"
-        },
-        "serverFailureInfo": {
-          "$ref": "#/definitions/v1ServerFailureInfo"
-        },
-        "resetWorkflowFailureInfo": {
-          "$ref": "#/definitions/v1ResetWorkflowFailureInfo"
-        },
-        "activityFailureInfo": {
-          "$ref": "#/definitions/v1ActivityFailureInfo"
-        },
-        "childWorkflowExecutionFailureInfo": {
-          "$ref": "#/definitions/v1ChildWorkflowExecutionFailureInfo"
-        },
-        "nexusOperationExecutionFailureInfo": {
-          "$ref": "#/definitions/v1NexusOperationFailureInfo"
-        },
-        "nexusHandlerFailureInfo": {
-          "$ref": "#/definitions/v1NexusHandlerFailureInfo"
-        }
-      }
     },
     "protobufAny": {
       "type": "object",
@@ -13378,7 +13334,7 @@
           "description": "Time when the activity transitioned to a closed state."
         },
         "lastFailure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "description": "Failure details from the last failed attempt."
         },
         "lastWorkerIdentity": {
@@ -13508,7 +13464,7 @@
           "description": "The result if the activity completed successfully."
         },
         "failure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "description": "The failure if the activity completed unsuccessfully."
         }
       },
@@ -13698,7 +13654,7 @@
       "type": "object",
       "properties": {
         "failure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "title": "Failure details"
         },
         "scheduledEventId": {
@@ -13800,7 +13756,7 @@
           "title": "Starting at 1, the number of times this task has been attempted"
         },
         "lastFailure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "description": "Will be set to the most recent failure details, if this task has previously failed and then\nbeen retried."
         },
         "workerVersion": {
@@ -13818,7 +13774,7 @@
       "type": "object",
       "properties": {
         "failure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "description": "If this activity had failed, was retried, and then timed out, that failure is stored as the\n`cause` in here."
         },
         "scheduledEventId": {
@@ -14386,7 +14342,7 @@
       "type": "object",
       "properties": {
         "failure": {
-          "$ref": "#/definitions/failureV1Failure"
+          "$ref": "#/definitions/v1Failure"
         },
         "namespace": {
           "type": "string",
@@ -15171,7 +15127,7 @@
           "description": "The result if the operation completed successfully."
         },
         "failure": {
-          "$ref": "#/definitions/failureV1Failure",
+          "$ref": "#/definitions/v1Failure",
           "description": "The failure if the operation completed unsuccessfully."
         },
         "longPollToken": {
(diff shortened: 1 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,44 @@ message DescribeWorkflowExecutionResponse {
...
+    // Link to the relevant history event:
+    // - completion event, if the workflow has finished
+    // - options-updated event, if callbacks were registered on a running workflow
+    repeated temporal.api.common.v1.Link links = 1;
+
+    // Completion status of the workflow, which carries relevant information for the caller.
+    oneof completion {
+        NotCompleted not_completed = 2;
+        Completed completed = 3;
+    }
```
long-nt-tran, line 1199: also possible to have

```
temporal.api.common.v1.WorkflowExecution execution = 1;
temporal.api.enums.v1.WorkflowExecutionStatus status    = 2;
temporal.api.common.v1.Payloads result = 3;
temporal.api.failure.v1.Failure failure = 4;
```
all at base level (not wrapped inside `NotCompleted` / `Completed`), but I like the cleaner `oneof`, rather than having these fields at the base level and leaving `result` and `failure` as `nil` if not applicable.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,44 @@ message DescribeWorkflowExecutionResponse {
...
 }
 
+message GetWorkflowExecutionResultRequest {
+    string namespace = 1;
+    temporal.api.common.v1.WorkflowExecution execution = 2;
+    // Used to de-dup requests (i.e., so we don't register the same callbacks multiple times).
+    string request_id = 3;
+    string identity = 4;
+    // Callbacks and links to register if the workflow is not yet completed.
+    repeated temporal.api.common.v1.Callback callbacks = 5;
```
stephanos: I see `StartWorkflowExecutionRequest` and `StartActivityExecutionRequest` are both calling this `completion_callbacks`; I'd suggest doing the same here.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,44 @@ message DescribeWorkflowExecutionResponse {
...
+    }
+
+    message NotCompleted {
+        temporal.api.common.v1.WorkflowExecution execution = 1;
+        temporal.api.enums.v1.WorkflowExecutionStatus status = 2;
+    }
+
+    message Completed {
+        temporal.api.common.v1.WorkflowExecution execution = 1;
+        temporal.api.enums.v1.WorkflowExecutionStatus status    = 2;
```
stephanos: 🤖 `nit` **[complexity]**: Inconsistent whitespace alignment.

**Issue:** `status    = 2` has extra spaces here, while `NotCompleted` above uses `status = 2`.

**Suggestion:** Remove the extra whitespace to match `NotCompleted`.

temporal/api/workflowservice/v1/request_response.proto
Quinn-With-Two-Ns: Hm I agree with using a `oneof` here, but why put `execution` and `status` in each case instead of in `GetWorkflowExecutionResultResponse`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,44 @@ message DescribeWorkflowExecutionResponse {
...
+        // Set only when status = COMPLETED.
+        temporal.api.common.v1.Payloads result = 3;
+        // Set for FAILED / CANCELED / TERMINATED / TIMED_OUT.
+        temporal.api.failure.v1.Failure failure = 4;
```
Quinn-With-Two-Ns: If we are already using a `oneof`, why not separate `Completed` into `Success` and `Failure`?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,47 @@ message DescribeWorkflowExecutionResponse {
...
+    repeated temporal.api.common.v1.Link links = 3;
+
+    // Completion status of the workflow, which carries relevant information for the caller.
+    oneof completion {
+        NotCompleted not_completed = 4;
+        Success success = 5;
+        Failure failure = 6;
+    }
+
+    message NotCompleted {}
```
long-nt-tran: I think empty message is ok (basically using this as an enum)? I see it used elsewhere, i.e.,

https://github.com/temporalio/api/blob/8e0453c3a17693d7abb78853c3ccccf0c632e782/temporal/api/workflow/v1/message.proto#L460

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,47 @@ message DescribeWorkflowExecutionResponse {
...
+
+    message Success {
+        temporal.api.common.v1.Payloads result = 1;
+    }
+
+    message Failure {
+        temporal.api.failure.v1.Failure failure = 2;
```
stephanos: This should be `1` AFAICT

temporal/api/workflowservice/v1/service.proto
```diff
@@ -2023,4 +2023,20 @@ service WorkflowService {
     // (-- api-linter: core::0127::http-annotation=disabled
     //     aip.dev/not-precedent: Nexus operation deletion not exposed to HTTP, users should use cancel or terminate. --)
     rpc DeleteNexusOperationExecution (DeleteNexusOperationExecutionRequest) returns (DeleteNexusOperationExecutionResponse) {}
+
+    // GetWorkflowExecutionResult returns the result of a workflow execution. If the workflow
+    // is still running, returns a NotCompleted response and optionally registers callbacks
+    // to be invoked when the workflow completes.
+    rpc GetWorkflowExecutionResult (GetWorkflowExecutionResultRequest) returns (GetWorkflowExecutionResultResponse) {
+          option (google.api.http) = {
```
stephanos: The indentation here is slightly off

(I have added a formatter to `temporal` but didn't get around to adding it to this repo yet)

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,47 @@ message DescribeWorkflowExecutionResponse {
...
+message GetWorkflowExecutionResultRequest {
+    string namespace = 1;
+    // For now, server will always provide the result of the latest run of the workflow_id
+    // provided even if the run_id is set here. We use WorkflowExecution message for forward
+    // compatibility if we want to enable getting the result of a specific run_id in the future.
+    temporal.api.common.v1.WorkflowExecution execution = 2;
+    // Used to de-dup requests (i.e., so we don't register the same callbacks multiple times).
+    string request_id = 3;
+    string identity = 4;
+    // Callbacks and links to register if the workflow is not yet completed.
```
stephanos: I'd split the comments and not combine one for two fields. If someone looks at the docs for `links` they won't see this.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,47 @@ message DescribeWorkflowExecutionResponse {
...
+    repeated temporal.api.common.v1.Callback callbacks = 5;
+    repeated temporal.api.common.v1.Link links = 6;
```
stephanos, line 1182: I see `StartActivityExecutionRequest` has a detailed note about the link behavior which seems worth copying here (assuming it's true here, too): https://github.com/temporalio/api/blob/8e0453c3a17693d7abb78853c3ccccf0c632e782/temporal/api/workflowservice/v1/request_response.proto#L3157C5-L3158C77

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,47 @@ message DescribeWorkflowExecutionResponse {
...
+    temporal.api.enums.v1.WorkflowExecutionStatus status = 2;
+
+    // Link to the relevant history event:
+    // - completion event, if the workflow has finished
+    // - options-updated event, if callbacks were registered on a running workflow
+    repeated temporal.api.common.v1.Link links = 3;
+
+    // Completion status of the workflow, which carries relevant information for the caller.
+    oneof completion {
+        NotCompleted not_completed = 4;
```
stephanos: What's the purpose of this field if there's already `status`?

I understand that there might not be a success or failure yet, I just wonder if that should be expressed as an empty/nil `completion` field. Not sure, maybe more an SDK question as it might impact its user-facing API impl.

long-nt-tran: I'm using this as an "enum that carries extra info" (i.e., `Success` carries the result, `Failure` carries the failure). I was thinking that `status` is a more granular set of enums in case the user really wants to see the status (I.e., what kind of failure it is), although maybe this is also inspectable via the failure itself.

Leaning towards removing `status` field since the "enum with extra info" is a bit easier to carry extra info in success/failure cases

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,51 @@ message DescribeWorkflowExecutionResponse {
...
+    string namespace = 1;
+    // For now, server will always provide the result of the latest run of the workflow_id
+    // provided even if the run_id is set here. We use WorkflowExecution message for forward
+    // compatibility if we want to enable getting the result of a specific run_id in the future.
+    temporal.api.common.v1.WorkflowExecution execution = 2;
+    // A unique identifier for this create request for idempotence. Typically UUIDv4.
+    string request_id = 3;
+    // The identity of the client who initiated this request.
+    string identity = 4;
+    // Callbacks to be called by the server when this activity reaches a terminal state.
```
stephanos: nit: this says "activity"?

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,51 @@ message DescribeWorkflowExecutionResponse {
     temporal.api.workflow.v1.WorkflowExecutionExtendedInfo workflow_extended_info = 8;
 }
 
+message GetWorkflowExecutionResultRequest {
+    string namespace = 1;
+    // For now, server will always provide the result of the latest run of the workflow_id
+    // provided even if the run_id is set here. We use WorkflowExecution message for forward
+    // compatibility if we want to enable getting the result of a specific run_id in the future.
+    temporal.api.common.v1.WorkflowExecution execution = 2;
+    // A unique identifier for this create request for idempotence. Typically UUIDv4.
```
stephanos: nit: this says "create request"?

temporal/api/workflowservice/v1/service.proto
```diff
@@ -2023,4 +2023,20 @@ service WorkflowService {
     // (-- api-linter: core::0127::http-annotation=disabled
     //     aip.dev/not-precedent: Nexus operation deletion not exposed to HTTP, users should use cancel or terminate. --)
     rpc DeleteNexusOperationExecution (DeleteNexusOperationExecutionRequest) returns (DeleteNexusOperationExecutionResponse) {}
+
+    // GetWorkflowExecutionResult returns the result of a workflow execution. If the workflow
+    // is still running, returns a NotCompleted response and optionally registers callbacks
+    // to be invoked when the workflow completes.
+    rpc GetWorkflowExecutionResult (GetWorkflowExecutionResultRequest) returns (GetWorkflowExecutionResultResponse) {
+        option (google.api.http) = {
+            get: "/namespaces/{namespace}/workflows/{execution.workflow_id}/result"
```
stephanos: I was surprised to see this was GET, but I see `GetWorkflowExecutionHistoryRequest` is also a GET. Worth checking if this should be a POST, though.

temporal/api/workflowservice/v1/request_response.proto
chrsmith: +1 to using an empty message instead of `protobuf.Empty{}` or similar. This allows us to add more data later if needed. e.g. maybe we'll want to wire in a "time_until_timeout" or "started_at" field later.

temporal/api/workflowservice/v1/service.proto
```diff
@@ -2023,4 +2023,20 @@ service WorkflowService {
     // (-- api-linter: core::0127::http-annotation=disabled
     //     aip.dev/not-precedent: Nexus operation deletion not exposed to HTTP, users should use cancel or terminate. --)
     rpc DeleteNexusOperationExecution (DeleteNexusOperationExecutionRequest) returns (DeleteNexusOperationExecutionResponse) {}
+
+    // GetWorkflowExecutionResult returns the result of a workflow execution. If the workflow
+    // is still running, returns a NotCompleted response and optionally registers callbacks
```
chrsmith: Nit: It doesn't return a "NotCompleted" response. It will always return the same `GetWorkflowExecutionResultResponse`, which will describe it's status which may be NotComplete.

I'd slightly argue that we can just remove the sentence "If the workflow is still running..." since it doesn't clarify anything. It will always return the result of a workflow execution.

What do you think of the following. Is this strictly better?

>     // GetWorkflowExecutionResult returns the state of a workflow execution, which may be still running, completed, or failed. It can optionally register new
>     // callbacks to be invoked when Workflow reaches a terminal state.
>     // (Adding a callback to a completed workflow will return an XXX error.)

chrsmith (commented):
I'll defer to Stephan for approval. But modulo some nits this LGTM!

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1163,6 +1163,51 @@ message DescribeWorkflowExecutionResponse {
...
+    string request_id = 3;
+    // The identity of the client who initiated this request.
+    string identity = 4;
+    // Callbacks to be called by the server when this activity reaches a terminal state.
+    // Callback addresses must be allowlisted in the server's dynamic configuration.
+    repeated temporal.api.common.v1.Callback completion_callbacks = 5;
```
chrsmith, line 1179: I don't know if this is quite right, but reserve the right to be wrong 😅 

You would expect an operation like `Get*` to not mutate the underlying resource. (At least that's the expectation in the HTTP world.) So it seems strange that calling `Get*` could actually modify the workflow's state, and register new callbacks.

- Is there a particular reason we want to allow registering new callbacks? And if not, maybe we can remove this capability?
- If we do want to enable registering new callbacks to a Workflow (which seems like it'd be useful), should we instead add a separate `UpdateWorkflowExecution` operation?

long-nt-tran: I think from the user's POV, this request is semantically a "get" operation even if the implementation detail requires some mutation to the target's state to deliver that result. I think Nexus-mapped operation would have to attach some sort of callback to deliver result async. An example is:

1. Alice uses a Nexus op in their namespace that maps to `GetWorkflowExecutionResult` for a workflow in Bob's namespace
2. Workflow hasn't completed. As an implementation detail, this request now attaches a "completion callback" which is essentially a URL
3. Once the workflow in Bob's namespace finishes, the result will be delivered to this callback URL (for example this is [its CHASM implementation which is a retryable RPC/HTTP call wrapped in a state machine](https://github.com/temporalio/temporal/blob/main/chasm/lib/callback/))

------------------

I do agree that it can sound strange to have the "Get" modify the workflow, although it's not really modifying the workflow's state, but rather attaches a listener onto the workflow.

From the user's POV, how the result gets delivered is kinda opaque so I think `Get*` sounds ok here?

openapi/openapiv2.json
```diff
@@ -12969,7 +13094,7 @@
           "description": "The time when the last attempt completed."
         },
         "lastAttemptFailure": {
-          "$ref": "#/definitions/v1Failure",
+          "$ref": "#/definitions/failureV1Failure",
```
stephanos: hm, this is odd?

temporal/api/workflowservice/v1/service.proto
long-nt-tran: yeah... i guess attaching callback is a mutation to the backend (even if an implementation detail)

temporal/api/workflowservice/v1/service.proto
long-nt-tran: thank you! I edited to be a variant copied from your comment

temporal/api/workflowservice/v1/request_response.proto
long-nt-tran: bad copypasta from me

temporal/api/workflowservice/v1/request_response.proto
long-nt-tran: bad copypasta from me

openapi/openapiv2.json
long-nt-tran: from claude, seems legit:

```
The root cause is that your nested message Failure adds a second message ending in Failure to the generator's scope, which bumps the minimum-unique-depth for temporal.api.failure.v1.Failure and cascades.

  The fix is to rename the nested message so it doesn't end in Failure:

  // Before
  message Failure {
      temporal.api.failure.v1.Failure failure = 1;
  }

  // After
  message Failed {
      temporal.api.failure.v1.Failure failure = 1;
  }

  And update the oneof:

  oneof completion_status {
      NotCompleted not_completed = 3;
      Success success = 4;
      Failed failure = 5;   // field name stays "failure", type is now "Failed"
  }

  This way the generator's scope still has only one message whose FQN ends in .Failure (temporal.api.failure.v1.Failure), so its minimum-unique-depth stays at 1 and it keeps the name v1Failure.

  Alternatively, if the wrapper is unnecessary, you could use temporal.api.failure.v1.Failure directly in the oneof and drop the nested message entirely — but that depends on whether you need the wrapper for forward compatibility.
```

It seems harmless(?), but I renamed the enums for now

temporal/api/workflowservice/v1/service.proto
```diff
@@ -2026,4 +2026,24 @@ service WorkflowService {
...
     //     aip.dev/not-precedent: Nexus operation deletion not exposed to HTTP, users should use cancel or terminate. --)
     rpc DeleteNexusOperationExecution (DeleteNexusOperationExecutionRequest) returns (DeleteNexusOperationExecutionResponse) {}
+
+    // GetWorkflowExecutionResult returns the state of a workflow execution, which may be still running,
+    // completed, or failed. It can optionally register new callbacks to be invoked when Workflow reaches
+    // a terminal state.
+    // If the workflow is completed, it returns the result of the completed workflow.
+    // NOTE: This has to be a POST since we're optionally attaching callbacks to the running workflow, which mutates it.
+    rpc GetWorkflowExecutionResult (GetWorkflowExecutionResultRequest) returns (GetWorkflowExecutionResultResponse) {
+        option (google.api.http) = {
```
spkane31, line 2036: Can you add 
```
option (nexusannotations.v1.operation).tags = "exposed";
```
at the top of this rpc so we can expose it as a system nexus operation?

spkane31: same as the `SignalWithStartWorkflowExecution` rpc

bergundy (commented):
Let's wait with this because we are going to define a completely separate API for the system Nexus endpoint for the same purpose and there's still open discussions around #720 that prevents us from having a single API surface.

PR #781, opened 2026-05-15
add GetWorkflowExecutionResult as a system nexus endpoint

spkane31:
**What changed?**
Add the `GetWorkflowExecutionResult` RPC and expose it as a system nexus operation.

**Why?**
New API for workflows

**Breaking changes**
No breaking changes

**Server PR**
TBA


temporal/nexus/api/workflowservice/v1/request_response.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+
+package temporal.nexus.api.workflowservice.v1;
```
spkane31: I chose to do this as `temporal.nexus.api....` instead of `nexus.api.workflowservice` because the latter would create confusion for our language packages. For example if we chose the latter would the go_package be `go.nexus.io/api/workflowservice`, which makes me think the service is nexus workflow specific. This also lacks any indication of temporal in the naming.

temporal/api/applicationservice/v1/service.proto
```diff
@@ -0,0 +1,51 @@
...
+    option (nexusannotations.v1.operation).tags = "exposed";
+
+    option (google.api.http) = {
+      post: "/namespaces/{namespace}/workflows/{execution.workflow_id}/wait-for-external-workflow"
+      body: "*"
+      additional_bindings {
+        post: "/api/v1/namespaces/{namespace}/workflows/{execution.workflow_id}/wait-for-external-workflow"
+        body: "*"
+      }
+    };
```
bergundy: I don't think we need HTTP for any of the system nexus APIs.

temporal/api/applicationservice/v1/service.proto
```diff
@@ -0,0 +1,51 @@
...
+    option (temporal.api.protometa.v1.request_header) = {
+      header: "temporal-resource-id"
+      value : "workflow:{execution.workflow_id}"
+    };
+  }
```
bergundy: Need to understand how the SDK will use this later when we support direct calls for the system nexus endpoint. For now it doesn't give us any value to add this.

api-linter.yaml
```diff
@@ -42,6 +42,7 @@
 
 - included_paths:
     - "**/workflowservice/v1/service.proto"
+    - "**/workflownexusservice/v1/service.proto"
```
bergundy:
```suggestion
    - "**/applicationservice/v1/service.proto"
```

temporal/api/applicationservice/v1/request_response.proto
```diff
@@ -0,0 +1,35 @@
...
+message GetWorkflowExecutionResultRequest {
+    string namespace = 1;
+    // The workflow execution to wait for.
+    temporal.api.common.v1.WorkflowExecution execution = 2;
+}
+
+
+message GetWorkflowExecutionResultResponse {
+    // The status of the workflow execution being waited on. This is needed to determine whether the workflow completed successfully or failed.
+    temporal.api.enums.v1.WorkflowExecutionStatus status = 1;
```
bergundy: Didn't we say we're not adding this?

temporal/api/applicationservice/v1/request_response.proto
```diff
@@ -0,0 +1,35 @@
...
+option csharp_namespace = "Temporalio.Api.Applicationservice.V1";
+
+import "temporal/api/enums/v1/workflow.proto";
+import "temporal/api/common/v1/message.proto";
+import "temporal/api/failure/v1/message.proto";
+
+// (-- api-linter: core::0131::request-name-required=disabled --)
+// (-- api-linter: core::0131::request-unknown-fields=disabled --)
+message GetWorkflowExecutionResultRequest {
+    string namespace = 1;
```
bergundy: We said not to add this. It's always same namespace.

temporal/api/applicationservice/v1/request_response.proto
```diff
@@ -0,0 +1,35 @@
...
+    // The workflow execution to wait for.
+    temporal.api.common.v1.WorkflowExecution execution = 2;
```
bergundy: We should decide and then document what to do if run ID is provided. We may not be able to follow the runs in some cases and we would want a consistent experience.

spkane31: I think we should start by not accepting RunID and defaulting to the latest execution. If we get signal that there's a need to expose specific RunIDs in the future we can add support later.

PR #783, opened 2026-05-20
Add system Nexus WIT operation metadata

tconley1428:
## What changed
- Adds generation step for producing/updating WIT files for exposed API definitions.
- Documents that Rust is required when running make because system Nexus WIT generation installs and runs nexus-api-gen.


nexus/workflow-service.wit
```diff
@@ -0,0 +1,122 @@
...
+  ///   typescript="workflow.getExternalWorkflowHandle(request.id, result.runId ?? undefined)"
+  ///   typescript-package="@temporalio/workflow"
+  /// @nexus.operation name="SignalWithStartWorkflowExecution"
+  /// @nexus.experimental
+  signal-with-start-workflow: func(
+    request: signal-with-start-workflow-request,
+  ) -> signal-with-start-workflow-response;
+  signal-with-start-workflow-execution: func(
+    request: signal-with-start-workflow-request,
+  ) -> signal-with-start-workflow-response;
```
tconley1428: Need to fix this bug with add-rpc.

Sushisource:
Is this PR gonna have the check that new things were added or is that a follow up PR?

tconley1428:
It already does! The API repo already checks for _any_ dirty diff after the makefile build.

Sushisource:
> It already does! The API repo already checks for _any_ dirty diff after the makefile build.

Aaaa, how about that then. Nice. What about manual changes to the output though? Isn't it going to complain about a diff there and we need some way to indicate it's OK?

tconley1428:
The make command actually does a merge operation into the output rather than fully regenerating. It will leave customizations in place. There might be some teething with that but it has been working so far.

tconley1428:
You can tell by the fact that this heavily customized output is _not_ failing the diff check.

Sushisource (approved):

spkane31 (approved):

PR #784, opened 2026-05-22
ShutdownWorker: Consolidate task_queue fields into single repeated field

yuandrew:
**What changed?**
Consolidated task_queue and sticky_task_queue fields into single `task_queues` field

**Why?**
Need to add support for Go Sessions, which polls on 2 separate internal task queues, easiest way is to just have a single repeated `task_queues` field that we can encapsulate, also making things easier in case we ever add even more task queues

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1109,6 +1109,11 @@ message ResetStickyTaskQueueResponse {
 }
 
 message ShutdownWorkerRequest {
+    message WorkerPollerTaskQueue {
+        string name = 1;
+        temporal.api.enums.v1.TaskQueueType type = 2;
+    }
+
     string namespace = 1;
     // sticky_task_queue may not always be populated. We want to ensure all workers
     // send a shutdown request to update worker state for heartbeating, as well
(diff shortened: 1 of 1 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1124,10 +1124,16 @@ message ShutdownWorkerRequest {
     // Task queue name the worker is polling on. This allows server to cancel
     // all outstanding poll RPC calls from SDK. This avoids a race condition that
     // can lead to tasks being lost.
-    string task_queue = 7;
+    string task_queue = 7 [deprecated = true];
```
dnr, line 1132: First, you can just make this repeated, no need to add a new field.

But, it's useful to know which is the sticky one since then we don't have to fan out the cancels to an activity queue with the sticky name (which doesn't exist). And the session queue is only an activity queue, right? So it seems like we should just have a list of name+type pairs.

yuandrew: Thanks for the context on server's behavior, I added a new `WorkerPollerTaskQueue` type to be name+type

PR #785, opened 2026-05-22
Add optional activity outcome in start activity response

ks-temporal:
**What changed?**
Added a new outcome field of type ActivityExecutionOutcome in
StartActivityExecutionResponse and a new boolean flag of 
id_reuse_policy_return_existing_outcome_on_reject in
StartActivityExecutionRequest.

**Why?**
So that the outcome is included if the execution is already started,
and the activity is already completed, to avoid another query for
the outcome. When a caller's `StartActivityExecution` short-circuits
to an existing run with the same activity ID because of an ID reuse,
the caller often wants the outcome of the existing run, not just the
identifier or an error pointing at it. Retrieving the outcome in the
response avoids the caller to use a second RPC of 
`DescribeActivityExecution` to retrieve a result that the server
already has cached.

**Breaking changes**
No

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -3165,7 +3165,7 @@ message StartActivityExecutionRequest {
     // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
     google.protobuf.Duration start_delay = 22;
 
-    // Include the outcome (result/failure) in the response if the activity has completed.
+    // Include the outcome in the response if the activity has completed.
     bool include_outcome = 23;
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -3165,7 +3165,11 @@ message StartActivityExecutionRequest {
     // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
     google.protobuf.Duration start_delay = 22;
 
-    // Include the outcome in the response if the activity has completed.
+    // Include the outcome in a success response if the activity has completed, and
+    // id_reuse_policy or id_conflict_policy would have caused a failure response otherwise.
+    // This is applicable only when the id_reuse_policy is ACTIVITY_ID_REUSE_POLICY_REJECT_DUPLICATE,
+    // or the activity succeeded and id_reuse_policy is ACTIVITY_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY,
+    // or id_conflict_policy is ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING.
     bool include_outcome = 23;
 }
 
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -3164,6 +3164,9 @@ message StartActivityExecutionRequest {
     temporal.api.common.v1.OnConflictOptions on_conflict_options = 21;
     // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
     google.protobuf.Duration start_delay = 22;
+
+    // Include the outcome (result/failure) in the response if the activity has completed.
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3164,6 +3164,9 @@ message StartActivityExecutionRequest {
     temporal.api.common.v1.OnConflictOptions on_conflict_options = 21;
     // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
     google.protobuf.Duration start_delay = 22;
+
+    // Include the outcome (result/failure) in the response if the activity has completed.
```
fretz12:
```suggestion
    // Include the outcome in the response if the activity has completed.
```
result/failure feels it's implied already

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3164,6 +3164,9 @@ message StartActivityExecutionRequest {
     temporal.api.common.v1.OnConflictOptions on_conflict_options = 21;
     // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
     google.protobuf.Duration start_delay = 22;
+
+    // Include the outcome in the response if the activity has completed.
```
bergundy: I would mention how this interacts with reuse policy and conflict policy here.

ks-temporal: updated the comment.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3169,6 +3169,12 @@ message StartActivityExecutionRequest {
     temporal.api.common.v1.OnConflictOptions on_conflict_options = 21;
     // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
     google.protobuf.Duration start_delay = 22;
+
+    // Include the outcome in a success response if the activity has completed, and
+    // id_reuse_policy would have caused a failure response otherwise.
+    // This is applicable only when the id_reuse_policy is ACTIVITY_ID_REUSE_POLICY_REJECT_DUPLICATE,
```
dandavison, line 3175: Suppose the activity failed, and the reuse policy is `ACTIVITY_ID_REUSE_POLICY_REJECT_DUPLICATE`. This scenario meets the condition

> the activity has completed, and id_reuse_policy would have caused a failure response

However, we will return a failure response, not a success response, which contradicts the docstring.

dandavison: My comment above was confused. I agree we should return a success response containing the failure outcome.

dandavison (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3178,6 +3184,8 @@ message StartActivityExecutionResponse {
     bool started = 2;
     // Link to the started activity.
     temporal.api.common.v1.Link link = 3;
+    // Only set if the activity is already completed and include_outcome was true in the request.
+    temporal.api.activity.v1.ActivityExecutionOutcome outcome = 4;
```
Quinn-With-Two-Ns, line 3192: If this is populated will the `run_id` and `link` be populated? I would think so, but the comments on both those fields imply differently

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3178,6 +3184,8 @@ message StartActivityExecutionResponse {
     bool started = 2;
     // Link to the started activity.
     temporal.api.common.v1.Link link = 3;
+    // Only set if the activity is already completed and include_outcome was true in the request.
```
stephanos: is `include_outcome` a stale reference?

ks-temporal: Yes, will fix. Thanks.

temporal/api/workflowservice/v1/request_response.proto
ks-temporal: Updated comment in all related fields.

PR #786, opened 2026-05-22
[breaking-change] time skipping fast forward

feiyang3cat:
**What changed?**

1. Removed obsolete fields — safe to delete as this feature has not been released (disable compatibility checking temporarily)
2. Renamed `bound` to `fast forward` to define the concept from the perspective of users' behavior
3. Add `disable_propagation_for_child` to give users a way to stop propagation
4. Add `update_time` to `UpdateWorkflowExecutionOptionsRequest`
5. Moving TimeSkippingConfig from workflow path to common path

**Why?**
1. for 1) make the API cleaner
2. for 2) a neat name to reduce misunderstanding
3. for 3) a child workflow is a different execution and has its own virtual clock, and users may need an option to change the default propagation behavior
4.  for 4) this can be used when SDK verify the time in `TimeSkippingTransitionedEvent` match the sleep duration
5. for 5) other executions like schedulers will also use the config

**Breaking changes**
1. removed the feature `maxSkippedDuration`-> the server has already implemented this feature need to delete the code first 
2. restructured the `maxElapsedDuration`to `fast_forward`-> the server need to do renaming

**Server PR**
https://github.com/temporalio/temporal/pull/10576


temporal/api/workflow/v1/message.proto
```diff
@@ -591,46 +591,32 @@ message WorkflowExecutionOptions {
...
-        // (-- api-linter: core::0142::time-field-names=disabled --)
-        google.protobuf.Duration max_elapsed_duration = 5;
-    }
+    // Optionally register a sleep for the current workflow execution.
+    // When the sleep completes, time skipping is disabled and this action is recorded in
+    // the WorkflowExecutionTimeSkippingTransitionedEvent, but it can be re-enabled by
+    // setting `enabled` to true via UpdateWorkflowExecutionOptions.
+    // The current workflow execution is a chain of runs (retries, cron, continue-as-new);
+    // child workflows are separate executions, so this sleep won't affect them.
+    google.protobuf.Duration sleep = 2;
```
feiyang3cat: other names like sleep_timer, timer, timeskipping_timer

openapi/openapiv2.json
```diff
@@ -19487,16 +19487,16 @@
...
+      "description": "The configuration for time skipping of a workflow execution (a chain of runs including retries, cron, continue-as-new).\nWhen time skipping is enabled, virtual time advances automatically whenever there is no in-flight work.\nIn-flight work includes activities, child workflows, Nexus operations, signal/cancel external workflow operations,\nand possibly other features added in the future.\nUser timers are not classified as in-flight work and will be skipped over; the virtual clock may also skip to the\ntime point of the registered sleep when there is no in-flight work.\n\nFor child workflows, by default, if the parent execution is skipping time, the child execution will also skip time,\nbut the parent's sleep won't affect the child execution. A flag is provided to disable propagation of the\n\"enabled\" flag to child workflows; regardless of that flag, a child workflow inherits the virtual time from the\nparent execution as its start time."
```
GregoryTravis: "to the time point of the registered sleep" -- does this mean to end point of the sleep period?

openapi/openapiv2.json
```diff
@@ -19487,16 +19487,16 @@
           "type": "boolean",
           "description": "Enables or disables time skipping for this workflow execution."
         },
-        "maxSkippedDuration": {
+        "sleep": {
           "type": "string",
-          "description": "Maximum total virtual time that can be skipped."
+          "description": "Optionally register a sleep for the current workflow execution.\nWhen the sleep completes, time skipping is disabled and this action is recorded in\nthe WorkflowExecutionTimeSkippingTransitionedEvent, but it can be re-enabled by\nsetting `enabled` to true via UpdateWorkflowExecutionOptions.\nThe current workflow execution is a chain of runs (retries, cron, continue-as-new);\nchild workflows are separate executions, so this sleep won't affect them."
```
GregoryTravis: Is the implication here that the sleep will actually take place, and not be skipped?

openapi/openapiv2.json
feiyang3cat: 1. oh I see, this is better expression. yes. will change to  "to the end point of the sleep period"

2. and I think I may need to add another thing is if the workflow execution (the whole chain of runs) completes before having a chance to skip time, the sleep may never take place -> check this example

openapi/openapiv2.json
feiyang3cat: yeah, totally, to be specific `sleep` counts both natural time passing and time skipping. 

for example, the user calls `sleep(1h)` at 10am, and then the sleep is expected to take place in 11am if the workflow execution hasn't completed.  And at 11am (virtual workflow time) the sleep takes place, but for the past 1hour the workflow may takes 30 minutes busy with activities (and this is real 30 minutes), and then waiting for some user timer/condition so we skip 30min(this is skipped) to fire the 1hour.

feiyang3cat: let me write some sleep examples

openapi/openapiv2.json
```diff
@@ -19487,16 +19487,16 @@
           "type": "boolean",
           "description": "Enables or disables time skipping for this workflow execution."
         },
-        "maxSkippedDuration": {
+        "sleep": {
```
JLDLaughlin: do you think users might confuse this sleep with sleeps in their actual workflow code?

I was thinking it might be useful to tie this more closely to the skipping? but I see that you're moving away from that!

feiyang3cat: yeah, totally. do you have better language recommendations? I have considered several other options but not pass as better:
- "bound": I feel at first we think "bound" is better, but when we call it "bound" engineers have different understanding how this bound should be propagated through a chain of runs and a tree of parent-child workflows and many interpretations make sense.
- "variants of sleep" I am also thinking about time_skipping_timer, times_kipping_sleep_timer, time_skipping_sleep if not very verbose? yet right now the testing server already exposes a sleep method for the users but it indeed seemed to have caused some confusion users
- "skip": not accurate as the 1hour may contain some duration that is not skipped and we believe sleep is easier

Sushisource: `warp`? `jump`?

I agree we should avoid overloading sleep. I think `skip` is actually totally fine too, and we just need to explain that the whole duration may not be skipped as you said.

feiyang3cat: Changed to fast_forward, reserving skip in case we need to control the exact skip duration, and leaving sleep as-is.

temporal/api/history/v1/message.proto
```diff
@@ -1002,17 +1002,16 @@ message WorkflowExecutionUnpausedEventAttributes {
...
+// either time was advanced or time skipping was disabled automatically due to the fast_forward completing.
 // The worker_may_ignore field in HistoryEvent should always be set true for this event.
 message WorkflowExecutionTimeSkippingTransitionedEventAttributes {
-
-    // The virtual time after time skipping was applied.
+    // The virtual time point that time skipping advanced to.
     google.protobuf.Timestamp target_time = 1;
 
-    // when true, time skipping was disabled automatically due to a bound being reached.
+    // When true, time skipping has been disabled automatically due to the fast_forward completing.
```
Sushisource:
```suggestion
    // When true, time skipping has been disabled automatically due to a call to fast_forward completing.
```

temporal/api/workflow/v1/message.proto
```diff
@@ -591,46 +591,32 @@ message WorkflowExecutionOptions {
...
-// (2) Retry and cron: the configuration and accumulated skipped duration are inherited as recorded when the
-//     current workflow started; the accumulated skipped duration of the current run is not propagated.
-// (3) Reset: the new run retains the time-skipping configuration of the current execution. Because reset replays
-//     all events up to the reset point and re-applies any UpdateWorkflowExecutionOptions changes made after that
-//     point, the resulting run ends up with the same final time-skipping configuration as the previous run.
+// User timers are not classified as in-flight work and will be skipped over; the virtual clock may also skip to the
+// time point of the registered fast_forward when there is no in-flight work.
+//
+// For child workflows, by default, if the parent execution is skipping time, the child execution will also skip time,
+// but the parent's fast_forward won't affect the child execution. A flag is provided to disable propagation of the
```
Sushisource:
```suggestion
// but a parent's fast_forward won't affect its child's execution. A flag is provided to disable propagation of the
```

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2387,6 +2387,10 @@ message UpdateWorkflowExecutionOptionsRequest {
 message UpdateWorkflowExecutionOptionsResponse {
     // Workflow Execution options after update.
     temporal.api.workflow.v1.WorkflowExecutionOptions workflow_execution_options = 1;
+
+    // The Workflow Execution time when the options were updated. When time skipping is
+    // enabled, this is the workflow's virtual time rather than wall-clock time.
+    google.protobuf.Timestamp update_time = 2;
```
Sushisource, line 2393: How does this get used? Does SDK need to use this?

feiyang3cat: I think so but it is just a refinement/enhancement that when SDK checks the time skipping event it can re-verifies the time is correct. this is only needed in edgy case:
-> the user set fast-forward 2hours using SDK and waiting on it (1) 
-> and the user issues another request to change it to 4hours (2)
-> and when client (1) finds the time skipping transitioned event it has a way to know it is a correct time to return

temporal/api/history/v1/message.proto
feiyang3cat: why do we need to add "a call" in this comment as there won't be any extra calls from the client side to disable it ?

Sushisource: "a call" here refers to the call that initiated the fast forward

temporal/api/workflow/v1/message.proto
```diff
@@ -591,46 +591,32 @@ message WorkflowExecutionOptions {
...
-    oneof bound {
-        // Maximum total virtual time that can be skipped.
-        google.protobuf.Duration max_skipped_duration = 4;
-
-        // Maximum elapsed time since time skipping was enabled.
-        // This includes both skipped time and real time elapsing.
-        // (-- api-linter: core::0142::time-field-names=disabled --)
-        google.protobuf.Duration max_elapsed_duration = 5;
-    }
+    // Optionally fast-forward the current workflow execution to this point in time.
```
simvlad: I don't think this should say "point in time", since this is a duration.

temporal/api/workflowservice/v1/request_response.proto
simvlad: Can you go into more details what would happen in those edge cases (if fast-forward is updated to 2h before and after another user issues a request to update to 4h). From reading the server PR, I thought the fast-forward time is replace (either you will skip by 2h or 4h depending on which one comes first). How does the update_time helps there

feiyang3cat: (1) fast-forward is updated to 2h (update_time = A) before and after another user issues a request to update to 1h (update_time = B)  **assuming the 2h fast-forward hasn't taken place** 
(2) fast-forward takes place in B+1h **as we only allows one fast forward in time skipping**
(3) the user calling A+2h is still using SDK function to wait for the **TimeSkippingTransitionedEvent**  and  not the event comes, if SDK doesn't check the event it doesn't know the time it will think A+2h has arrived. if the user has update_time A it can check if the **event time = update_time + duration**

temporal/api/workflow/v1/message.proto
feiyang3cat: changed

temporal/api/history/v1/message.proto
feiyang3cat: get it. updated.

temporal/api/workflowservice/v1/request_response.proto
simvlad: ok, so it will just sleep for the remaining of the 1h?

feiyang3cat: have resolved in huddle

Makefile
```diff
@@ -41,7 +41,7 @@ $(PROTO_OUT):
 	mkdir $(PROTO_OUT)
 
 ##### Compile proto files for go #####
-grpc: buf-lint api-linter buf-breaking clean go-grpc fix-path
```
simvlad, line 44: intentional?

temporal/api/workflow/v1/message.proto
```diff
@@ -591,48 +591,48 @@ message WorkflowExecutionOptions {
...
-        // (-- api-linter: core::0142::time-field-names=disabled --)
-        google.protobuf.Duration max_elapsed_duration = 5;
-    }
+    // Optionally fast-forward the current workflow execution by this duration ahead of current workflow execution time.
+    // When the fast-forward completes, time skipping is disabled by the call that initiated the fast-forward, and this
+    // action is recorded in the WorkflowExecutionTimeSkippingTransitionedEvent. It can be re-enabled by
+    // setting `enabled` to true via UpdateWorkflowExecutionOptions.
+    // The current workflow execution is a chain of runs (retries, cron, continue-as-new);
+    // child workflows are separate executions, so this fast_forward won't affect them.
+    google.protobuf.Duration fast_forward = 2;
```
simvlad: should we document the behavior when you override the existing fast_forward?

temporal/api/history/v1/message.proto
```diff
@@ -206,11 +206,16 @@ message WorkflowExecutionStartedEventAttributes {
     // will be reflected in the WorkflowExecutionOptionsUpdatedEvent.
     temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 41;
 
-    // The time skipped by the previous execution that started this workflow.
-    // It can happen in cases of child workflows and continue-as-new workflows.
-    google.protobuf.Duration initial_skipped_duration = 42;
+    reserved 42;
+    reserved "initial_skipped_duration";
+
+  // The time-skipping state propagated from a previous run of this workflow. This can be nil
```
simvlad: nit: something is wrong with indentation

simvlad (approved):
Looks good from server side

Makefile
feiyang3cat: oh yeah, I learnt that the current process to delete deprecated fields is to allow breaking and add this back in the next PR.

temporal/api/history/v1/message.proto
```diff
@@ -906,8 +915,14 @@ message WorkflowExecutionOptionsUpdatedEventAttributes {
     // Priority override upserted in this event. Represents the full priority; not just partial fields.
     // Ignored if nil.
     temporal.api.common.v1.Priority priority = 6;
-    // If set, the time-skipping configuration was changed. Contains the full updated configuration.
-    temporal.api.workflow.v1.TimeSkippingConfig time_skipping_config = 7;
+
+    // TimeSkippingConfig override upserted in this event. Represents the full config.
+    temporal.api.common.v1.TimeSkippingConfig time_skipping_config = 7;
+
+    // Indicates the time skipping config was updated by the recent call to update
```
simvlad, line 922: Seems to me it is more clear to say that this boolean set to true indicates that this update set the new `time_skipping_config` value and is needed to differentiate between clearing time_skipping_config and not touching it.

temporal/api/common/v1/message.proto
```diff
@@ -393,3 +394,52 @@ message OnConflictOptions {
...
+// For child workflows, by default, if the parent execution is skipping time, the child execution will also skip time,
+// but a parent's fast_forward won't affect its child's execution. A flag is provided to disable propagation of the
+// "enabled" flag to child workflows; regardless of that flag, a child workflow inherits the virtual time from the
+// parent execution as its start time.
+message TimeSkippingConfig {
+    // Enables or disables time skipping for this workflow execution.
+    bool enabled = 1;
+
+    // Optionally fast-forward the current workflow execution by this duration ahead of current workflow execution time.
+    // When the fast-forward completes, time skipping is disabled by the call that initiated the fast-forward, and this
```
Sushisource: "by the call" implies that the client is responsible for disabling the fast-forward. Maybe just 

```suggestion
    // After the fast-forward completes, time skipping is disabled, and this
```

temporal/api/common/v1/message.proto
```diff
@@ -393,3 +394,52 @@ message OnConflictOptions {
...
+    // action is recorded in the WorkflowExecutionTimeSkippingTransitionedEvent. It can be re-enabled by
+    // setting `enabled` to true via UpdateWorkflowExecutionOptions.
```
Sushisource:
```suggestion
    // setting `enabled` to true or setting `fast_forward` again via UpdateWorkflowExecutionOptions.
```

Sushisource (approved):
I still think we should just remove the field reservations here and just have it be breaking, but, not a huge deal either way.

temporal/api/history/v1/message.proto
feiyang3cat: yet it is actually needed to differentiate between "keeping the current time_skipping_config and not touching it" ?

PR #787, opened 2026-05-26
delete max elasped duration

feiyang3cat:
**What changed?**
delete max_elapsed_duration still in a compatible way in this pr

**Why?**
1. we deem this this feature is not necessary given we already have max_skip_duration and its variation in duration calculation adds comprehension efforts than given customers real benefits
2. this feature also complicates server side implementation to a higher level

**Breaking changes**
this feature is already implemented and merged in server as this project is huge we merge it feature by feature, so this api should be merged only after the corresponding server pr to delete it merged

**Server PR**
https://github.com/temporalio/temporal/pull/10387

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -609,8 +609,8 @@ message WorkflowExecutionOptions {
 //     all events up to the reset point and re-applies any UpdateWorkflowExecutionOptions changes made after that
 //     point, the resulting run ends up with the same final time-skipping configuration as the previous run.
 message TimeSkippingConfig {
-    reserved 2, 6;
-    reserved "disable_propagation", "max_target_time";
+    reserved 2, 5, 6;
+    reserved "disable_propagation", "max_elapsed_duration", "max_target_time";
 
     // Enables or disables time skipping for this workflow execution.
     bool enabled = 1;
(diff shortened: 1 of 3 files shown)


temporal/api/workflow/v1/message.proto
```diff
@@ -609,8 +609,8 @@ message WorkflowExecutionOptions {
 //     all events up to the reset point and re-applies any UpdateWorkflowExecutionOptions changes made after that
 //     point, the resulting run ends up with the same final time-skipping configuration as the previous run.
 message TimeSkippingConfig {
-    reserved 2, 6;
-    reserved "disable_propagation", "max_target_time";
+    reserved 2, 5, 6;
```
prathyushpv, line 612: Nobody uses this feature so far. So I wonder do we need to reserve these fields, and just reuse these numbers?

prathyushpv (approved):

temporal/api/workflow/v1/message.proto
```diff
@@ -625,11 +625,6 @@ message TimeSkippingConfig {
     oneof bound {
```
prathyushpv, line 625: Do we need this oneof field if we only have `max_skipped_duration`? Are we planning to keep add more option in the future?

PR #790, opened 2026-05-27
Add external payload transfer stats for workflow tasks

simvlad:
**What changed?**
Added `ExternalPayloadTransferStats` and a corresponding `external_payload_transfer_stats` field on `RespondWorkflowTaskCompletedRequest`.  The message carries a list of per-storage-driver entries, each with upload/download counts and byte totals. `ExternalPayloadTransferStats` is in common/v1 so we can extend it to other requests in the future if needed, e.g. `StartWorkflowExecutionRequest`. 

**Why?**
Lets SDKs report external payload transfer activity per workflow task so the server can emit workflow task latency metrics tagged by driver, and correlate latency with transferred bytes/count.

**Breaking changes**
None.

**Server PR**
WIP

### DIFF
--- a/temporal/api/common/v1/message.proto
+++ b/temporal/api/common/v1/message.proto
@@ -393,3 +393,24 @@ message OnConflictOptions {
     // Attaches the links to the running execution.
     bool attach_links = 3;
 }
+
+// External payload transfer stats, broken down by storage driver.
+message ExternalPayloadTransferStats {
+    repeated StorageDriverStats drivers = 1;
+
+    message StorageDriverStats {
+        // Storage driver these stats are for. Matches `StorageDriverInfo.type`.
+        string driver_type = 1;
+        // Accumulated upload stats.
+        TransferStats uploaded = 2;
+        // Accumulated download stats.
+        TransferStats downloaded = 3;
+    }
+
+    message TransferStats {
+        // Number of payloads transferred
+        int64 total_count = 1;
+        // Total size of payloads transferred in bytes
+        int64 total_size_bytes = 2;
+    }
+}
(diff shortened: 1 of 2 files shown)


temporal/api/common/v1/message.proto
```diff
@@ -393,3 +393,24 @@ message OnConflictOptions {
...
+    message StorageDriverStats {
+        // Storage driver these stats are for. Matches `StorageDriverInfo.type`.
+        string driver_type = 1;
+        // Accumulated upload stats.
+        TransferStats uploaded = 2;
+        // Accumulated download stats.
+        TransferStats downloaded = 3;
+    }
+
+    message TransferStats {
```
jmaeagle99, line 410: Do we want a latency/duration field for each driver type upload/download? Might not be for metrics, but maybe for Web UI?

temporal/api/common/v1/message.proto
```diff
@@ -393,3 +393,24 @@ message OnConflictOptions {
     // Attaches the links to the running execution.
     bool attach_links = 3;
 }
+
+// External payload transfer stats, broken down by storage driver.
+message ExternalPayloadTransferStats {
```
jmaeagle99, line 398: Do we want latency/duration information for wall-clock upload/download? Might not be for metrics, but maybe for Web UI?

temporal/api/common/v1/message.proto
simvlad: For others: we discussed, and decided to add later when necessary

temporal/api/common/v1/message.proto
simvlad: Same ^^

PR #793, opened 2026-06-01
Add Event Groups

mjameswh:
**What changed?**

- Introduce the `temporal.api.sdk.v1.EventGroupMarker` message.
- Add `event_group_markers` on `Command` and `HistoryEvent`.

**Why?**

Event Groups is a new form of metadata that allows regrouping of logically related Workflow Events, according to user-defined or system-inferred criteria, providing users with improved visibility, analysis, and debugging capabilities.

**Breaking changes**

No.

**Server PR**

https://github.com/temporalio/temporal/pull/10472

### DIFF
--- a/temporal/api/command/v1/message.proto
+++ b/temporal/api/command/v1/message.proto
@@ -17,6 +17,7 @@ import "temporal/api/common/v1/message.proto";
 import "temporal/api/failure/v1/message.proto";
 import "temporal/api/taskqueue/v1/message.proto";
 import "temporal/api/sdk/v1/user_metadata.proto";
+import "temporal/api/sdk/v1/event_group_marker.proto";
 
 message ScheduleActivityTaskCommandAttributes {
     string activity_id = 1;
@@ -294,6 +295,10 @@ message Command {
     //  * start_timer_command_attributes - populates temporal.api.history.v1.HistoryEvent for timer
     //    started where the summary is used to identify the timer.
     temporal.api.sdk.v1.UserMetadata user_metadata = 301;
+
+    // Event Group Markers attached to the command by the workflow author.
+    repeated temporal.api.sdk.v1.EventGroupMarker event_group_markers = 302;
+
     // The command details. The type must match that in `command_type`.
     oneof attributes {
         ScheduleActivityTaskCommandAttributes schedule_activity_task_command_attributes = 2;
--- a/temporal/api/sdk/v1/event_group_marker.proto
+++ b/temporal/api/sdk/v1/event_group_marker.proto
@@ -1,0 +1,64 @@
+syntax = "proto3";
+
+package temporal.api.sdk.v1;
+
+option go_package = "go.temporal.io/api/sdk/v1;sdk";
+option java_package = "io.temporal.api.sdk.v1";
+option java_multiple_files = true;
+option java_outer_classname = "EventGroupMarkerProto";
+option ruby_package = "Temporalio::Api::Sdk::V1";
+option csharp_namespace = "Temporalio.Api.Sdk.V1";
+
+
+import "temporal/api/common/v1/message.proto";
+
+message EventGroupMarker {
+  // What this Marker represents. The variant determines whether the Marker was
+  // created explicitly by user code (label) or implicitly by the SDK on inbound
+  // signals/events (inbound_event) or update handlers (inbound_update).
+  oneof variant {
+    Label label = 1;
+    InboundEvent inbound_event = 2;
+    InboundUpdate inbound_update = 3;
+  }
+
+  // A user-defined short-form string value to be used as the group's label.
+  message Label {
+    // Opaque identifier assigned by the SDK.
+    string id = 1;
+
+    // This payload should be a "json/plain"-encoded payload that is a single
+    // JSON string for use in user interfaces. User interface formatting may not
+    // apply to this text when used in "label" situations. The payload data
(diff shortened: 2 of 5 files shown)


temporal/api/command/v1/message.proto
```diff
@@ -294,6 +295,10 @@ message Command {
     //  * start_timer_command_attributes - populates temporal.api.history.v1.HistoryEvent for timer
     //    started where the summary is used to identify the timer.
     temporal.api.sdk.v1.UserMetadata user_metadata = 301;
+
+    // Event Group Markers attached to the command by the workflow author.
+    repeated temporal.api.sdk.v1.EventGroupMarker event_group_markers = 302;
```
Sushisource, line 300: Why here rather than in `UserMetadata`? Not necessarily opposed, just w/ UserMetadata I think we probably can avoid touching server at all.

temporal/api/sdk/v1/event_group_marker.proto
```diff
@@ -0,0 +1,65 @@
...
+  // marker ID; further references to an existing marker ID reuse existing
+  // attributes of the referenced marker.
+  //
+  // In the case where a later group marker does carry attributes, then those
+  // attributes override the previous attributes of the same marker ID for all
+  // events carrying that marker ID, including preceding events.
+  oneof attributes {
+    LabelAttributes label_attributes = 2;
+    InboundEventAttributes inbound_event_attributes = 3;
+    InboundUpdateAttributes inbound_update_attributes = 4;
```
Sushisource: I get event/update being mutually exclusive, but, I think it'd be nice to be able to have explicit labels at the same time as it being auto-triggered? We could have some kind of `getCurrentGroup` API that you could then attach labels to?

Or, does the comment imply you could have two of these markers, one label, and one inbound, for the same ID and that enables this?

temporal/api/sdk/v1/event_group_marker.proto
```diff
@@ -0,0 +1,65 @@
...
+  // A user-defined short-form string value to be used as the group's label.
+  message LabelAttributes {
+    // This payload should be a "json/plain"-encoded payload that is a single
+    // JSON string for use in user interfaces. User interface formatting may not
+    // apply to this text when used in "label" situations. The payload data
+    // section is limited to 400 bytes by default.
+    //
+    // Note that it is valid to have distinct markers (i.e. distinct marker IDs)
+    // in a given workflow execution that carry the same label.
+    temporal.api.common.v1.Payload label = 1;
```
Sushisource: This is `Payload` just to allow for encryption I assume?

mjameswh: Indeed. Basically, this follows the convention we used for `summary` and `details`.

temporal/api/sdk/v1/event_group_marker.proto
mjameswh: > Does the comment imply you could have two of these markers, one label, and one inbound, for the same ID and that enables this?

Well, that would be technically feasible, though definitely not what I had in mind.

> I think it'd be nice to be able to have explicit labels at the same time as it being auto-triggered? We could have some kind of getCurrentGroup API that you could then attach labels to?

An event can be attached to multiple groups, so `getCurrentGroup()` is not exactly possible. We could have `getCurrentGroups()`, but then that would be painful for users to do what you describe. Or we could have some form of `getCurrentSignalUpdateOrWorkflowEventGroup()` (just need to figure out a name), for which there is always one and only one at any point...

But at that point, I'd simply suggest the user to explicitly create a labelled group wrapping the complete signal/update handler's code.

temporal/api/command/v1/message.proto
mjameswh: Short story is that really depends on where we want to go with Event Groups. What we have in the MVP is really just an extension of the existing User Metadata concept, only providing minor enrichment to the workflow history. If we stop there, adding this to `UserMetadata` would indeed be a possibility (that was actually my initial direction).

However, envisioned goals for the proposal include some features that could benefit from Event Groups being their own thing rather than piggybacked on User Metadata. For example, one of the near-term goal is to allow requesting cancellation of an Event Group. ~Another one is to be able to follow Event Group across workflows.~ In both cases, there are some implicit requirements regarding the validity of the data stored in event groups, stronger than what we currently have on user metadata.

Even given the immediate MVP scope, the fact is that we accept `UserMetadata` on many APIs that can't accept Event Groups. So we'd need to spread server-side checks everywhere to confirm that supplied user metadata doesn't contain event groups. That's way more painful than adding this field here and add explicit copy assignments to the server in the very few places (three to be exact) where that's pertinent.

Sushisource: >the fact is that we accept UserMetadata on many APIs that can't accept Event Groups.

Yeah, that makes sense. The previous points I feel like we still have the data well formed even if it's in UserMetadata, but this makes a lot of sense to me.

temporal/api/sdk/v1/event_group_marker.proto
Sushisource: > I'd simply suggest the user to explicitly create a labelled group wrapping the complete signal/update handler's code.

Yeah, reasonable, and you're saying those commands would be part of both the auto and manual group. Makes sense.

Sushisource (approved):
Super

Quinn-With-Two-Ns (approved):

PR #798, opened 2026-06-10
add VerisioningOverride to Child Workflow's on start

Shivs11:
**What changed?**
- WISOTT

**Why?**
- Versioning

**Breaking changes**
- None

**Server PR**
https://github.com/temporalio/temporal/pull/10646


temporal/api/enums/v1/failed_cause.proto
```diff
@@ -89,6 +89,7 @@ enum StartChildWorkflowExecutionFailedCause {
     START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_UNSPECIFIED = 0;
     START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_EXISTS = 1;
     START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_NAMESPACE_NOT_FOUND = 2;
+    START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_INVALID_VERSIONING_OVERRIDE = 3;
```
Shivs11, line 97: the main reason I chose to have an additional cause here is because there could be a world where the version the user is trying to pin their child workflow on does not exist, and so I thought it's best to have a new cause out here since we should not be starting the child workflow in that case

note: if a user tries to start a non child workflow with a versioning override where the version does not exist, we right now do not start that workflow. The intention was to keep the behaviour consistent.

Sushisource (approved):

rkannan82 (approved):

PR #799, opened 2026-06-11
Add server_scaled_provider_cloud_run capability

rossnelson:
## What changed
Adds `server_scaled_provider_cloud_run = 13` to `GetSystemInfoResponse.Capabilities`, following the `server_scaled_deployments` pattern (#752).

## Why
The Temporal UI shows Google Cloud Run as a compute provider option for server-scaled worker deployments (temporalio/ui#3518), gated behind this capability — disabled with a "Coming Soon" badge until the server advertises support. The server should only set this once the Cloud Run scaling algorithm (worker-set support in temporal-auto-scaled-workers) ships and is enabled via server configuration.

### DIFF
--- a/temporal/api/workflowservice/v1/request_response.proto
+++ b/temporal/api/workflowservice/v1/request_response.proto
@@ -1343,6 +1343,11 @@ message GetSystemInfoResponse {
         // to be enabled via server configuration.
         bool server_scaled_deployments = 12;
 
+        // True if the server supports the Cloud Run compute provider for
+        // server-scaled deployments. Dependent on server version and the
+        // provider being enabled via server configuration.
+        bool server_scaled_provider_cloud_run = 13;
+
     }
 }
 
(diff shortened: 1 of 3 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -1343,6 +1343,11 @@ message GetSystemInfoResponse {
         // to be enabled via server configuration.
         bool server_scaled_deployments = 12;
 
+        // True if the server supports the Cloud Run compute provider for
+        // server-scaled deployments. Dependent on server version and the
+        // provider being enabled via server configuration.
+        bool server_scaled_provider_cloud_run = 13;
```
02strich, line 1349: is adding a single new value for this for each provider a scaleable approach? any thoughts on having the server return a list of supported providers?

laurakwhit: Definitely think a list of providers could work and might be more extensible! I'm not aware of any non-boolean values for system_Info > capabilities, but seems like we could potentially add an `enum supported_server_scaled_providers` or similar. 

Not sure if this is something we want to keep configurable though. E.g. wondering if the eventual state should be that every provider is enabled by default and these are just temporary flags that help in the release process.

02strich (approved):

bergundy (approved):

eamsden (approved):

PR #800, opened 2026-06-11
Workflow Command Pagination API

simvlad:
**What changed?**
Added a paginated workflow task completion API:
- `page_number` and `intermediate_page` fields on `RespondWorkflowTaskCompletedRequest`, letting a workflow task completion be split across multiple "page" requests sharing one task token.
- `workflow_task_completion_pagination` namespace capability flag that SDKs must check before sending paginated requests.
- `WorkflowTaskCompletionBufferLostFailure` signaling the server dropped buffered pages and the client should resend all pages from page 0.

**Why?**
Large workflow task completions can exceed request size limits. Pagination lets workers split a single completion across multiple requests.

**Breaking changes**
None. New fields/messages are additive; pagination is gated behind the namespace capability.

**Server PR**
https://github.com/temporalio/temporal/pull/10880

### DIFF
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -58,6 +58,8 @@ message NamespaceInfo {
         bool workflow_update_callbacks = 12;
         // When true, workers should use poller autoscaling by default unless explicitly configured otherwise.
         bool poller_autoscaling_auto_enroll = 13;
+        // True if the namespace supports pagination of `RespondWorkflowTaskCompleted` request.
+        bool workflow_task_completion_pagination = 14;
     }
 
     // Namespace configured limits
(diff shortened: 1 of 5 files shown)


jmaeagle99 (approved):

temporal/api/namespace/v1/message.proto
```diff
@@ -58,6 +58,10 @@ message NamespaceInfo {
         bool workflow_update_callbacks = 12;
         // When true, workers should use poller autoscaling by default unless explicitly configured otherwise.
         bool poller_autoscaling_auto_enroll = 13;
+        // True if the namespace supports paginated workflow task completion, i.e.
+        // `RespondWorkflowTaskCompleted` requests split across multiple pages. SDKs must check this
+        // capability before sending a request with `intermediate_page` set to true.
```
Sushisource:
```suggestion
        // capability before sending a request with `intermediate_page` set to true or with a nonzero value in `page_number`.
```

To be complete

Alternatively you could just remove this bit since I think the documentation on the actual fields is more important.

Sushisource (approved):

yycptt (approved):

PR #804, opened 2026-06-15
Add start_delay to ActivityOptions for update support

fretz12:
**What changed?**
Added a start_delay field to the ActivityOptions message in temporal/api/activity/v1/message.proto. OpenAPI specs (v2/v3) were regenerated accordingly.

**Why?**
Expose start_delay through UpdateActivityExecutionOptions so users can modify or zero out the delay on a still-pending standalone activity. The new value is anchored to the activity's original schedule_time, so the field has the same semantics whether set at creation or via update. Server-side implementation lands in a follow-up PR.

### DIFF
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -63,6 +63,11 @@ message ActivityOptions {
     // Priority metadata. If this message is not present, or any fields are not
     // present, they inherit the values from the workflow.
     temporal.api.common.v1.Priority priority = 7;
+
+    // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
+    // When updated, anchored to the original `schedule_time`. If the resulting time is in the past, dispatches
+    // immediately.
+    google.protobuf.Duration start_delay = 8;
 }
 
 // Information about a standalone activity.
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -64,9 +64,9 @@ message ActivityOptions {
     // present, they inherit the values from the workflow.
     temporal.api.common.v1.Priority priority = 7;
 
-    // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
-    // When updated, anchored to the original `schedule_time`. If the resulting time is in the past, dispatches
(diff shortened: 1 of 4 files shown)


temporal/api/activity/v1/message.proto
```diff
@@ -63,6 +63,11 @@ message ActivityOptions {
     // Priority metadata. If this message is not present, or any fields are not
     // present, they inherit the values from the workflow.
     temporal.api.common.v1.Priority priority = 7;
+
+    // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
+    // When updated, anchored to the original `schedule_time`. If the resulting time is in the past, dispatches
+    // immediately.
+    google.protobuf.Duration start_delay = 8;
```
dandavison, line 70: How about we word this as
> Time to wait before making the first activity task available for dispatch

Internally, there are two dispatch times: (1) the time that History dispatches a task to Matching and (2) the time that Matching dispatches a task to a user worker. Only (2) is directly relevant to users, and (1) != (2) because of the possibility of non-sync match (as well as workers being offline).

If you agree with this suggestion, it should also be applied to the StartWorkflow, SignalWithStartWorkflow, StartActivity doc comments.

temporal/api/activity/v1/message.proto
```diff
@@ -63,6 +63,11 @@ message ActivityOptions {
     // Priority metadata. If this message is not present, or any fields are not
     // present, they inherit the values from the workflow.
     temporal.api.common.v1.Priority priority = 7;
+
+    // Time to wait before dispatching the first activity task. This delay is not applied to retry attempts.
+    // When updated, anchored to the original `schedule_time`. If the resulting time is in the past, dispatches
```
dandavison:
```suggestion
    // When updated, the time is added to the original `schedule_time`, not to the current time. If the resulting time is in the past, dispatches
```

temporal/api/activity/v1/message.proto
fretz12: Makes sense. Applied comment where applicable.

maciejdudko (approved):

dandavison (approved):

PR #808, opened 2026-06-17
Add one-time move to the versioning override API!

Shivs11:
**What changed?**
- WISOTT

**Why?**
- So that operators can have a versioning override that is not sticky. The idea is that the next task of the workflow execution moves to the target version, after which the behaviour/version of the workflow are dictated by the worker response.

**Breaking changes**
- No, since this is a new option that I am adding whose default value is false (it's opt-in)

**Server PR**
- [here](https://github.com/temporalio/temporal/pull/10763)

### DIFF
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -619,6 +619,8 @@ message VersioningOverride {
         // until one Workflow Task completes there. After completion, the workflow
         // execution's Versioning Behavior and Deployment Version come from the
         // worker's completion response.
+        // (-- api-linter: core::0142::time-field-type=disabled
+        //     aip.dev/not-precedent: one_time describes one-time routing semantics, not a timestamp or duration. --)
         OneTimeOverride one_time = 5;
     }
 
--- a/temporal/api/workflow/v1/message.proto
+++ b/temporal/api/workflow/v1/message.proto
@@ -615,8 +615,9 @@ message VersioningOverride {
         // Override the workflow to have AutoUpgrade behavior.
         bool auto_upgrade = 4;
 
-        // Move the workflow to a specific Worker Deployment Version for one successful Workflow Task.
-        OneTimeMove one_time = 5;
+        // Temporarily override workflow task routing to a specific Worker Deployment Version
+        // until one Workflow Task completes there.
+        OneTimeOverride one_time = 5;
     }
 
     // Required.
@@ -665,7 +666,7 @@ message VersioningOverride {
     //
     // If no Workflow Task completes on `target_deployment_version`, this override
     // remains pending.
-    message OneTimeMove {
+    message OneTimeOverride {
         // Required. Worker Deployment Version to receive the one-time Workflow Task.
         temporal.api.deployment.v1.WorkerDeploymentVersion target_deployment_version = 1;
     }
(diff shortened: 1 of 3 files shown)


temporal/api/workflow/v1/message.proto
yiminc, line 615: so this would continue to be sticky pin behavior? I think we need to update the doc so it is more clear. It would ignore even when SDK do CAN and wants change to AU to current version?

Shivs11: >  It would ignore even when SDK do CAN and wants change to AU to current version

yes, the PinnedOverride is meant to be sticky. As long as it exists on a workflow execution, it shall be pinned to a specific workflow execution

temporal/api/workflow/v1/message.proto
```diff
@@ -614,6 +614,12 @@ message VersioningOverride {
 
         // Override the workflow to have AutoUpgrade behavior.
         bool auto_upgrade = 4;
+
+        // Temporarily override workflow task routing to a specific Worker Deployment Version
+        // until one Workflow Task completes there. After completion, the workflow
+        // execution's Versioning Behavior and Deployment Version come from the
+        // worker's completion response.
+        OneTimeOverride one_time = 5;
```
rkannan82, line 626: I have given this naming some more thought. What do you think about?
- MoveAndKeepDefaultBehavior: this clearly tells the behavior. move to version X, then respect whatever is the default versioning configuration defined in code. It tells this is not sticky unlike PinnedOverride where it moves to version X and stays put.


- Also add the following to the comment:
When you moves the wf from version x to y, and a new version z arrives.

  - if a workflow is pinned, it will stay on the y
  - if wf is auto upgrade -- it will move to z
  - if wf uses upgrade on can -- it will move to z when can happens

- Personally, I would avoid using temporary.

Shivs11: I quite like the model that you took when naming this, however, the only thing that bothers me is the word "Default" -- the thing is, in worker-versioning, you have two things that could be the "Default" for the true versioning behaviour of a workflow; they are:

- The workflow annotation on the workflow type
- The default workflow behaviour on the worker that is assigned to execute this workflow

When you have a versioning behaviour that is defined on the workflow type, that's the behaviour that is always followed. 

Having said that, here are some of the things that came to my head the second I read your suggestion:

 - What do we mean by "Default Behaviour" here? I have a feeling that an operator might could get confused thinking about which specific behaviour (workflow type annotation, worker registration) here. I worry that the word is quite overloaded here.
 - I could also see a world where someone could interpret this as: "keep the old/default behavior this workflow had before the move."
 -  I am still inclined to using OneTimeOverride only because from the lens of an operator, it portrays the message of moving their workflows one time to a new version. While the after math of this move is subtle, I think we can rely on the operator reading the comments to have a stronger clarity

curious to hear your thoughts on this @rkannan82

rkannan82: Sounds good Shivam. We can keep OneTimeOverride, and document the semantics of what happens after this move. We should clearly call out how this differs from Pinned -- see if what I suggested earlier makes sense to add to the comment.

rkannan82: One last thing... From earlier conversation .. we didn't wan't MoveOverride. Is it because pin also moves? Can we frame it as Pin is sticky, whereas move is not? Reason is eventually we are going to expose this using a move API or CLI. I don't think we will call it as one time there.

rkannan82 (approved):

Sushisource (approved):
I think the comments do a great job explaining it all now.

temporal/api/workflow/v1/message.proto
Shivs11: discussed offline, have decided to stick with OneTimeOverride for now!

PR #809, opened 2026-06-18
Add runtime type information to worker heartbeats

Sushisource:
**What changed?**
Title

**Why?**
Help us understand customer environments. SDKs will include an opt-out flag.

**Breaking changes**

**Server PR**

### DIFF
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -130,6 +130,10 @@ message WorkerHeartbeat {
 
   // Storage drivers in use by this SDK.
   repeated StorageDriverInfo drivers = 24;
+
+  // Runtime environments in use by this SDK. Often will only be one of these, but it's allowed to
+  // be repeated for scenarios like "Python inside a Lambda".
+  repeated RuntimeInfo runtimes = 25;
 }
 
 // Detailed worker information.
@@ -192,6 +196,25 @@ message PluginInfo {
 message StorageDriverInfo {
   // The type of the driver, required.
   string type = 1;
+}
+
+message RuntimeInfo {
+  enum RuntimeType {
+    RUNTIME_TYPE_UNSPECIFIED = 0;
+    RUNTIME_TYPE_JVM = 1;
+    RUNTIME_TYPE_PYTHON = 2;
+    RUNTIME_TYPE_NODE = 3;
+    RUNTIME_TYPE_BUN = 4;
+    RUNTIME_TYPE_RUBY = 5;
+    RUNTIME_TYPE_GO = 6;
+    RUNTIME_TYPE_DOCKER = 7;
+    RUNTIME_TYPE_LAMBDA = 8;
+    RUNTIME_TYPE_GCR = 9;
+  }
+  // The type of the runtime.
+  RuntimeType type = 1;
+  // The version of the runtime.
+  string version = 2;
 }
 
 // A command sent from the server to a worker.
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -131,9 +131,8 @@ message WorkerHeartbeat {
   // Storage drivers in use by this SDK.
   repeated StorageDriverInfo drivers = 24;
 
-  // Runtime environments in use by this SDK. Often will only be one of these, but it's allowed to
-  // be repeated for scenarios like "Python inside a Lambda".
-  repeated RuntimeInfo runtimes = 25;
+  // Information about the environment this SDK is running in.
+  EnvironmentInfo environment = 25;
 }
 
 // Detailed worker information.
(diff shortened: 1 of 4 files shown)


temporal/api/worker/v1/message.proto
```diff
@@ -194,6 +198,25 @@ message StorageDriverInfo {
   string type = 1;
 }
 
+message RuntimeInfo {
+  enum RuntimeType {
```
jmaeagle99: Missing DOTNET

jmaeagle99: I would also like to differentiate between .NET types e.g. Framework, Core, NativeAOT, etc and Unknown. So I think there should be a secondary field for this. It would be good to keep that it is .NET separate from the type of .NET in case we can't accurately identify it so we can still filter based on .NET and know we have a known gap.

jmaeagle99: Okay, I see we have an sdk_name field above. So maybe add DOTNET_FRAMEWORK and DOTNET_CORE.

temporal/api/worker/v1/message.proto
```diff
@@ -194,6 +198,25 @@ message StorageDriverInfo {
...
+    RUNTIME_TYPE_UNSPECIFIED = 0;
+    RUNTIME_TYPE_JVM = 1;
+    RUNTIME_TYPE_PYTHON = 2;
+    RUNTIME_TYPE_NODE = 3;
+    RUNTIME_TYPE_BUN = 4;
+    RUNTIME_TYPE_RUBY = 5;
+    RUNTIME_TYPE_GO = 6;
+    RUNTIME_TYPE_DOCKER = 7;
```
02strich: Is there more than one type possible? Like python in docker?

yuandrew: the field in the heartbeat is `repeated`, so we can include multiple
>   // be repeated for scenarios like "Python inside a Lambda".

yuandrew: Actually, does it make sense to have `RuntimeType type = 1;` to be repeated here? Would we expect the version of multiple runtimes in the heartbeat to diverge, or will they always be the same?

yuandrew (approved):

temporal/api/worker/v1/message.proto
```diff
@@ -194,6 +197,97 @@ message StorageDriverInfo {
...
+    string version = 2;
+  }
+
+  enum Architecture {
+    ARCHITECTURE_UNSPECIFIED = 0;
+    ARCHITECTURE_AMD64 = 1;
+    ARCHITECTURE_ARM64 = 2;
+  }
+
+  message Platform {
```
jmaeagle99, line 261: Should there be an UnspecifiedPlatform in case someone is running on something that isn't Windows, Linux, or MacOS (e.g. BSD). Idk if we'd want an optional name field (so then is it really unspecified?) and version. Maybe named OtherPlatform?

temporal/api/worker/v1/message.proto
```diff
@@ -194,6 +197,97 @@ message StorageDriverInfo {
...
+      RUNTIME_TYPE_DOTNET_CORE = 8;
+    }
+    // The type of the runtime.
+    RuntimeType type = 1;
+    // The version of the runtime, if obtainable.
+    string version = 2;
+  }
+
+  message HostingEnvironment {
+    enum HostingEnvironmentType {
```
jmaeagle99, line 226: I think these are overly specific and orthogonal sometimes:
- Are we going to list every single branded cloud *aaS environment we want to know about? For example, Lambda, ECS, EKS, Fargate, EC2, etc. That's just AWS and probably doesn't enumerate all of their compute environments. I know Azure its own variety (e.g. Functions, App Service for Windows, App Service for Linux (yes, they are separate), Container Apps, Kubernetes, Virtual Machines, etc). I'm sure GCP has their own too. It really depends on the granularity that is needed.
- Even within one *aaS environment, there are multiple deployment technologies that are allowed. For example, Lambda supports both container images (which might not just be Docker) and zip archives: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-lambda-function.html

I think that the cloud service should at least be separate from the deployment/delivery technology.

temporal/api/worker/v1/message.proto
```diff
@@ -194,6 +197,97 @@ message StorageDriverInfo {
   string type = 1;
 }
 
+message EnvironmentInfo {
+  message Runtime {
+    enum RuntimeType {
```
jmaeagle99, line 202: Some of these are "runtimes" and some of these are languages. Should we separate them as such? It'll make it easier to describe things like:
- TS on Node vs TS on Bun
- Java on JVM vs Kotlin on JVM

So I think RuntimeType should be:
- RUNTIME_TYPE_UNSPECIFIED
- RUNTIME_TYPE_NATIVE (for Go and Rust)
- RUNTIME_TYPE_JVM
- RUNTIME_TYPE_CPYTHON
- RUNTIME_TYPE_NODE (for TS)
- RUNTIME_TYPE_BUN (for TS)
- RUNTIME_TYPE_DOTNET_CORE (for .NET)
- RUNTIME_TYPE_DOTNET_FRAMEWORK (for .NET)
- RUNTIME_TYPE_CRUBY
- RUNTIME_TYPE_ZEND (for PHP)

The above are actual runtimes, independent of which languages run on them. For languages, we either just infer that from the sdk_name or create a separate enum and message for those.

Sushisource: Ya, makes sense

temporal/api/worker/v1/message.proto
Sushisource: OtherPlatform makes sense

temporal/api/worker/v1/message.proto
Sushisource: I was just going with what we're supporting for serverless right so far, but, yes totally valid point. 

The fact that it's repeated allows for things like docker-in-lambda.

However, maybe we don't really care about this since we have serverless usage info via the fact that they're registered as configurations on the server already. At the same time, I dunno how well we can correlate that with _other_ SDK info and having it all in once place is nice.

So, kinda my inclination is this can just be things we know we care about, and not everything else, and we can be fine leaving it at that, but I'm definitely open to other suggestions.

02strich: +1 to this being independent of the serverless config as that will be harder to reason about in a few cases. I would also probably be more generous in the API definition than we'll be distinguishing in the code for now - so having Lambda, ECS, EKS, EC2, AgentCore Runtime should cover AWS. 

The one question this raises for me, whether it would be simpler to have this a string field, but I am not sold on that personally.

Sushisource: I want to avoid strings because they will bloat wire size significantly (as well as, obviously, be hard to make consistent).

I'll add more variants.

temporal/api/worker/v1/message.proto
Sushisource: Updated (though used RoadRunner instead of Zend since the PHP SDK only works w/ RoadRunner)

temporal/api/worker/v1/message.proto
```diff
@@ -194,6 +197,120 @@ message StorageDriverInfo {
...
+      RUNTIME_TYPE_UNSPECIFIED = 0;
+      RUNTIME_TYPE_JVM = 1;
+      RUNTIME_TYPE_CPYTHON = 2;
+      RUNTIME_TYPE_NODE = 3;
+      RUNTIME_TYPE_BUN = 4;
+      RUNTIME_TYPE_CRUBY = 5;
+      RUNTIME_TYPE_GO = 6;
+      RUNTIME_TYPE_DOTNET_FRAMEWORK = 7;
+      RUNTIME_TYPE_DOTNET_CORE = 8;
+      RUNTIME_TYPE_NATIVE = 9;
```
02strich, line 214: what does NATIVE mean here?

02strich (approved):

temporal/api/worker/v1/message.proto
Sushisource: Compiled. So, Rust mostly. Possibly GraalVM?

I kept Go even though it's compiled since the version does have some bearing on the runtime behavior.

Sushisource:
This might sit for a bit since @rkannan82 wants to possibly line it up with the register worker API

Sushisource:
Relates to https://github.com/temporalio/sdk-typescript/issues/1964

Sushisource:
We're no longer gonna wait on register to be created first

jmaeagle99 (approved):

rkannan82 (approved):

PR #810, opened 2026-06-19
Add compute provider validation to worker deployment version summary

smuneebahmad:
**What changed?**
Added `ComputeStatus` to the Worker Deployment API with a `ProviderValidationStatus` that tracks the result of the most recent connectivity check between Temporal and a customer's compute resource. An empty error message means validation passed; a non-empty message describes what failed.

**Why?**
Enables surfacing the connectivity health between Temporal and a customer's compute resource (e.g. Lambda) through the Worker Deployment API, so the UI can show ongoing validations status of compute configs without additional queries.

**Breaking changes**
None. All additions are new optional fields; existing clients are unaffected.

**Server PR**
https://github.com/temporalio/temporal/pull/10778


temporal/api/compute/v1/validation.proto
```diff
@@ -0,0 +1,36 @@
...
+option go_package = "go.temporal.io/api/compute/v1;compute";
+option java_package = "io.temporal.api.compute.v1";
+option java_multiple_files = true;
+option java_outer_classname = "ValidationProto";
+option ruby_package = "Temporalio::Api::Compute::V1";
+option csharp_namespace = "Temporalio.Api.Compute.V1";
+
+import "google/protobuf/timestamp.proto";
+
+// WorkerDeploymentValidationSummary aggregates validation status counts across all
```
jaypipes: This and the other definitions in this file should be in the `deployment/v1/message.proto` file and be prefixed with `Compute` instead of `WorkerDeployment`.

temporal/api/compute/v1/validation.proto
```diff
@@ -0,0 +1,36 @@
...
+        HEALTH_OK = 1;
+        HEALTH_ERROR = 2;
+    }
+    // The health of the compute resource connection.
+    Health health = 1;
+    // Human-readable error message if health is HEALTH_ERROR.
+    string error_message = 2;
+    // Timestamp of the last validation check.
+    google.protobuf.Timestamp last_check_time = 3;
+}
```
jaypipes: See my related comment in the `deployment/v1/message.proto` file. Instead of having a struct for just the "validation status", it's more future-proof to name the struct `ComputeStatus` and have a field called `provider_validation` in that struct. That way we don't have a situation where we're adding multiple status fields to the WorkerDeploymentInfo struct in the future for each distinct health-check for compute.

```proto
// ComputeStatus represents compute-related configuration and healthchecks.
message ComputeStatus {
    // ProviderValidationStatus represents the result of the most
    // recent connectivity check between Temporal and a customer's compute provider.
    message ProviderValidationStatus {
        // Human-readable error message connectivity validation failed.
        string error_message = 1;
        // Timestamp of the last validation check.
        google.protobuf.Timestamp last_check_time = 2;
    }
    // ProviderValidation encapsulates the health signal for validating the compute provider
    ProviderValidationStatus provider_validation = 1;
```

temporal/api/deployment/v1/message.proto
```diff
@@ -255,6 +256,9 @@ message WorkerDeploymentInfo {
         // Cleared if the version becomes current or ramping again.
         google.protobuf.Timestamp last_deactivation_time = 10;
         temporal.api.compute.v1.ComputeConfigSummary compute_config = 13;
+        // The result of the most recent connectivity check for this version's compute resource.
+        // Only populated for versions that have compute config.
+        temporal.api.compute.v1.WorkerDeploymentVersionValidationStatus validation_status = 14;
```
jaypipes: This should be `compute_status` and should refer to a struct that encapsulates all compute-related status/health signals:

```proto
    // ComputeStatus represents compute-related configuration and healthchecks.
    ComputeStatus compute_status = 14;
```

temporal/api/compute/v1/validation.proto
```diff
@@ -0,0 +1,36 @@
...
+message WorkerDeploymentVersionValidationStatus {
+    enum Health {
+        HEALTH_UNSPECIFIED = 0;
+        HEALTH_OK = 1;
+        HEALTH_ERROR = 2;
+    }
+    // The health of the compute resource connection.
+    Health health = 1;
+    // Human-readable error message if health is HEALTH_ERROR.
+    string error_message = 2;
```
jaypipes: I would recommend reducing these to just an `error_message` field. The provider is either validated or it isn't. Having an empty `error_message` field means the provider is validated. See below...

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2641,6 +2642,8 @@ message ListWorkerDeploymentsResponse {
         temporal.api.deployment.v1.WorkerDeploymentInfo.WorkerDeploymentVersionSummary current_version_summary = 5;
         // Summary of the ramping version of the Worker Deployment.
         temporal.api.deployment.v1.WorkerDeploymentInfo.WorkerDeploymentVersionSummary ramping_version_summary = 6;
+        // Aggregated WCI validation status counts across all versions with compute config.
+        temporal.api.compute.v1.WorkerDeploymentValidationSummary validation_summary = 7;
```
jaypipes: This is a mismatch of data modeling. The existing summary fields are aggregated on the version dimension. But the WorkerDeploymentValidationSummary aggregates across all versions. This makes it harder to reason about how to use this information in the UI and CLI...

jaypipes (changes requested):

temporal/api/compute/v1/validation.proto
smuneebahmad: Good point! I'll go with your recommended approach here.

jaypipes (approved):
👍

02strich (approved):

ShahabT (approved):

temporal/api/deployment/v1/message.proto
```diff
@@ -191,6 +191,21 @@ message VersionDrainageInfo {
...
+        // An empty string means validation passed.
+        string error_message = 1;
+        // Timestamp of the last validation check.
+        google.protobuf.Timestamp last_check_time = 2;
+    }
+    // provider_validation encapsulates the health signal for validating the compute provider.
+    ProviderValidationStatus provider_validation = 1;
+}
+
 message WorkerDeploymentInfo {
```
yuandrew, line 209: Was it intentional for `WorkerDeploymentInfo`'s docstring to be fully moved to `ComputeStatus`? Wouldn't we still want somesthing describing this?

temporal/api/deployment/v1/message.proto
tconley1428, line 187: Seems like these comments are no longer located correctly.

yuandrew (approved):
One question for docs, but not blocking

PR #812, opened 2026-06-29
Add first_execution_run_id to StartWorkflowExecutionResponse

### LINKED ISSUE #8537: Add first execution run ID for already-started start workflow results
**Describe the solution you'd like**

SDK needs `StartWorkflowExecutionResponse.first_execution_run_id` for on-conflict-use-existing situations, and `WorkflowExecutionAlreadyStartedFailure.first_execution_run_id` for other already-started situations. A lot of the time this will be populated with the same value as `run_id`.

Today, when SDK gets `StartWorkflowExecutionResponse` with `started` as `false` or `WorkflowExecutionAlreadyStartedFailure`, it doesn't include the first execution run ID that we can bind successive calls to. It only provides the run ID. If you start a workflow today successfully, SDK uses the run ID as the first execution run ID on successive calls like cancel and signal and such. But if you start a workflow today w/ conflict policy of use existing and it doesn't start, that run ID is not acceptable for first execution run ID.

Quinn-With-Two-Ns:
**What changed?**
Add FirstExecutionRunID to start response

**Why?**

closes https://github.com/temporalio/temporal/issues/8537

>Today, when SDK gets StartWorkflowExecutionResponse with started as false or WorkflowExecutionAlreadyStartedFailure, it doesn't include the first execution run ID that we can bind successive calls to. It only provides the run ID. If you start a workflow today successfully, SDK uses the run ID as the first execution run ID on successive calls like cancel and signal and such. But if you start a workflow today w/ conflict policy of use existing and it doesn't start, that run ID is not acceptable for first execution run ID.

**Server PR**
https://github.com/temporalio/temporal/pull/10873

### DIFF
--- a/nexus/workflow-service.wit
+++ b/nexus/workflow-service.wit
@@ -104,6 +104,7 @@ interface workflow-service {
     started: option<bool>,
     /// @nexus.omit
     signal-link: placeholder,
+    first-execution-run-id: string,
   }
 
   /// @nexus.doc
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -19206,6 +19206,10 @@
           "type": "string",
           "description": "The run id of the workflow that was started - or just signaled, if it was already running."
         },
+        "firstExecutionRunId": {
+          "type": "string",
+          "description": "If the workflow was started as a result of a de-dupe, this field will contain the run id of the first execution in the chain."
+        },
         "started": {
           "type": "boolean",
           "description": "If true, a new workflow was started."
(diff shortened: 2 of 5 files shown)


tconley1428 (approved):

tconley1428:
Interesting first example of the WIT failing CI due to a change. You should regenerate it and we can decide if we want to omit it or not.

bergundy (approved):
I would add this to cancel, query and signal RPCs too but this is a good step in the right direction.

bergundy:
And we need a way to pass in the first execution run ID for more interactions. Not blocking this PR.

PR #816, opened 2026-07-01
Add activity task failure cause

jmaeagle99:
**What changed?**

Add new activity task failure cause to to activity task failed requests and events.

**Why?**

We have a few well-known failure causes that we can attach for better debugability.


temporal/api/enums/v1/failed_cause.proto
```diff
@@ -85,6 +85,19 @@ enum WorkflowTaskFailedCause {
     WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE = 37;
 }
 
+// Activity tasks can fail for various reasons. Note that some of these reasons can only originate
+// from the server, and some of them can only originate from the SDK/worker.
+enum ActivityTaskFailedCause {
+    ACTIVITY_TASK_FAILED_CAUSE_UNSPECIFIED = 0;
+    // An activity task failed because a grpc message was too large.
```
dandavison: > a grpc message
Is this specifically "the RespondActivityTaskCompleted gRPC message"?

temporal/api/enums/v1/failed_cause.proto
```diff
@@ -85,6 +85,19 @@ enum WorkflowTaskFailedCause {
...
+    ACTIVITY_TASK_FAILED_CAUSE_GRPC_MESSAGE_TOO_LARGE = 1;
+    // An activity task failed because payloads were too large.
```
dandavison:
```suggestion
    // An activity task failed because one or more payloads were too large.
```

maybe?

temporal/api/enums/v1/failed_cause.proto
jmaeagle99: Yes, the intent is to handle the gRPC too large response for `RespondActivityTaskCompletedRequest` and submit an `RespondActivityTaskFailedRequest` failure; similar to how we do for WFT completions. Today, I believe we don't do anything in the worker when it receives the gRPC too large response for activity completions, and the server eventually times out the activity however it is configured.

jmaeagle99: I decided to drop the gRPC too large value for now since workers do not proactively fail the activity task at this time. We can add back later if we decide to change that behavior.

temporal/api/enums/v1/failed_cause.proto
jmaeagle99: Updated the comments to be more precise about what is too large without being overly prescriptive.

Quinn-With-Two-Ns (approved):

dandavison (approved):

PR #817, opened 2026-07-06
vts: rename propagation field

feiyang3cat:
**What changed?**
in `TimeSkippingConfig` message, `disable_child_propagation` was changed to `disable_propagation`

**Why?**
this field should be generic than being specific to one execution type

**Breaking changes**
yes just for static code dependence

**Server PR**
https://github.com/temporalio/temporal/pull/10940


simvlad:
But we still propagate for chain of runs, correct?

simvlad (approved):

GregoryTravis (approved):

PR #818, opened 2026-07-07
Add rate_limiting_active field to TaskQueueStats

rkannan82:
## What

Adds `bool rate_limiting_active` to `TaskQueueStats` proto message.

## Why

Auto-scaling systems need to know when backlog growth is caused by rate limiting rather than insufficient worker capacity. When the whole-queue rate limit is blocking dispatches, scaling up workers doesn't help — the bottleneck is the rate limit, not worker count.

[Server PR](https://github.com/temporalio/temporal/pull/10944)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


temporal/api/taskqueue/v1/message.proto
```diff
@@ -143,6 +143,15 @@ message TaskQueueStats {
     //   workflow goes to a normal queue, and the rest workflow tasks go to the Sticky queue associated with a specific
     //   worker instance.
     float tasks_dispatch_rate = 4;
+
+    // Whether the task queue's whole-queue rate limit is currently blocking dispatches. When true, adding more
```
Sushisource: What are the semantics of "currently" here? Will this flicker on and off if worker throughput is nearly equal to the rate limit? I think adding a bit more detail to the doc would be useful here.

And, internally, if that is possible, it might make sense to apply some smoothing/debounce or something so scalers don't oscillate

rkannan82: (updated comment) This is instantaneous value, and can flicker. We expect the consumer to handle any smoothing. I didn't want to add complexity right now to server.

temporal/api/taskqueue/v1/message.proto
```diff
@@ -143,6 +143,19 @@ message TaskQueueStats {
...
     //   worker instance.
     float tasks_dispatch_rate = 4;
+
+    // Whether the task queue's whole-queue rate limit is blocking dispatches. When true, adding more
+    // workers will not increase throughput — the bottleneck is the rate limit, not worker count. This field is
+    // useful for auto-scaling systems to avoid unnecessary scale-up.
+    //
+    // This reflects instantaneous state at the time of the query. If worker throughput is close to the rate
+    // limit, this value may alternate between true and false across successive queries. Consumers should
+    // apply their own smoothing or hysteresis to avoid oscillating scaling decisions.
```
ShahabT: wondering if this is the right semantics or more of a sliding window (say 30s to be consistent with task add/dispatch rate fields)?

Also, what instantaneous mean? does it mean that at this moment a task is blocked behind the limiter.allow?

ShahabT: if the concern is complexity in server, we already have sliding window `taskTracker` used for add/dispatch rates. so it shouldn't be much work.

rkannan82: Thanks. Updated the semantics to reflect -- was rate limited in the last 30 second window.

temporal/api/taskqueue/v1/message.proto
rkannan82: After discussing with Shahab, changed the semantics from instantaneous to a window (last 30 seconds).

Sushisource (approved):

ShahabT (approved):

PR #820, opened 2026-07-07
remove reset_heartbeats field in ResetActivityExecutionRequest

spkane31:
**What changed?**
Remove the `reset_heartbeat` field from `temporal.api.workflowservice.v1.ResetActivityExecutionRequest`.

**Why?**
The new default behavior will be to reset heartbeats with a reset request so this field is no longer needed.

**Breaking changes**
This is _technically_ a breaking change, but this endpoint has not been published in the server yet so no users can actually use this field.

**Server PR**
N/A

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -13,6 +13,18 @@ breaking:
     - WIRE_JSON
   ignore:
     - google
+  ignore_only:
+    # TODO(seankane): Remove these temporary breaking-change ignores after this PR merges.
+    FIELD_SAME_JSON_NAME:
+      - temporal/api/workflowservice/v1/request_response.proto
+    FIELD_SAME_NAME:
+      - temporal/api/workflowservice/v1/request_response.proto
+    FIELD_WIRE_JSON_COMPATIBLE_TYPE:
+      - temporal/api/workflowservice/v1/request_response.proto
+    FIELD_NO_DELETE_UNLESS_NAME_RESERVED:
+      - temporal/api/workflowservice/v1/request_response.proto
+    FIELD_NO_DELETE_UNLESS_NUMBER_RESERVED:
+      - temporal/api/workflowservice/v1/request_response.proto
 lint:
   use:
     - DEFAULT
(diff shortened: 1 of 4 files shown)


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -2350,24 +2328,20 @@ message ResetActivityExecutionRequest {
     // The identity of the client who initiated this request.
     string identity = 5;
 
-    // Indicates that activity should reset heartbeat details.
-    // This flag will be applied only to the new instance of the activity.
-    bool reset_heartbeat = 6;
-
     // If activity is paused, it will remain paused after reset
-    bool keep_paused = 7;
+    bool keep_paused = 6;
```
fretz12, line 2390: I'm assuming swapping is ok if no one uses this API? 
Safe thing to do is: 
```
    reserved 6;
    reserved "reset_heartbeat";

    // If activity is paused, it will remain paused after reset
    bool keep_paused = 7;
...
```

buf.yaml
```diff
@@ -3,22 +3,30 @@ name: buf.build/temporalio/api
...
     # to run api-linter, so just tell buf it ignore it
     - google
-    # Same for nexusannotations - local copy for api-linter, BSR dep for buf
-    - nexusannotations
 breaking:
   use:
     - WIRE_JSON
   ignore:
     - google
+  ignore_only:
```
fretz12, line 19: if we did my above comment, I don't think we need this

spkane31: Agree we can swap, but since this has never been used server side I think it is cleaner to remove

temporal/api/workflowservice/v1/request_response.proto
dandavison: Agree it's OK to renumber: `ResetActivityExecutionRequest` has always returned an error.

dandavison (approved):

maciejdudko (approved):

PR #823, opened 2026-07-09
Add scale_decisions to WorkerPollerInfo

veeral-patel:
## What changed

Adds a `map<string, int64> scale_decisions` field (tag 4) to `WorkerPollerInfo`.

It carries a per-poller-type breakdown of poller autoscaler scale decisions **since the last heartbeat**, keyed by reason:
`server_up`, `server_up_suppressed`, `server_down`, `empty_poll_down`, `re_halve`, `error_down`. Only set when the poller count is auto-scaled.

## Why

Poller autoscaling decisions (including when a server-requested scale-up is *suppressed* because the worker's throughput gate is closed) are currently only visible via SDK metrics. Surfacing them in the worker heartbeat makes scaling behavior observable server-side without a metrics pipeline.

## Consumer

Go SDK: temporalio/sdk-go (companion PR wires up the SDK to populate this field per interval).

### DIFF
--- a/temporal/api/worker/v1/message.proto
+++ b/temporal/api/worker/v1/message.proto
@@ -22,6 +22,11 @@ message WorkerPollerInfo {
 
   // Set true if the number of concurrent pollers is auto-scaled
   bool is_autoscaling = 3;
+
+  // Counts of poller autoscaler scale decisions since the last heartbeat, keyed by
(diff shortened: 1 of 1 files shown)


veeral-patel:
Closed in favor of a opt-in server side metric: https://github.com/temporalio/temporal/pull/11084

PR #833, opened 2026-07-17
Add standalone activity start delay to namespace capabilities

fretz12:
**What changed?**
Added the standalone_activity_start_delay field to NamespaceInfo.Capabilities.

**Why?**
Clients need to determine whether standalone activity start delay is effectively available for a namespace without relying on server-version checks. This lets UI clients hide start-delay controls when the feature is disabled by server configuration.

### DIFF
--- a/temporal/api/namespace/v1/message.proto
+++ b/temporal/api/namespace/v1/message.proto
@@ -60,6 +60,8 @@ message NamespaceInfo {
         bool poller_autoscaling_auto_enroll = 13;
         // True if the namespace supports pagination of `RespondWorkflowTaskCompleted` request.
         bool workflow_task_completion_pagination = 14;
+        // True if the namespace supports start delay for standalone activities.
+        bool standalone_activity_start_delay = 15;
     }
 
     // Namespace configured limits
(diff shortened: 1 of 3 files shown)


temporal/api/namespace/v1/message.proto
```diff
@@ -60,6 +60,8 @@ message NamespaceInfo {
         bool poller_autoscaling_auto_enroll = 13;
         // True if the namespace supports pagination of `RespondWorkflowTaskCompleted` request.
         bool workflow_task_completion_pagination = 14;
+        // True if the namespace supports start delay for standalone activities.
```
spkane31, line 63: Thoughts on adding a comment that says this is also used for operator commands?

spkane31 (approved):

temporal/api/namespace/v1/message.proto
```diff
@@ -60,6 +60,10 @@ message NamespaceInfo {
...
+        bool standalone_activity_start_delay = 15;
+        // True if the namespace supports batch operations for standalone activities.
+        bool standalone_activity_start_batch_operations = 16;
```
fretz12: nit:
```suggestion
        bool standalone_activity_batch_operations = 16;
```

THardy98 (approved):

PR #834, opened 2026-07-17
Add ACTIVITY_EXECUTION_STATUS_PAUSED

spkane31:
**What changed?**
WISOTT

**Why?**
This will be returned in `ListActivityExecution` responses

**Breaking changes**
NA

**Server PR**
NA


dandavison (approved):

mjameswh (approved):

temporal/api/enums/v1/activity.proto
cconstable: I think this comment is no longer true with PAUSED.

cconstable:
```suggestion
// The status is updated when the activity is originally scheduled, on pause, and when the activity reaches a terminal status.
```

cconstable (approved):

temporal/api/enums/v1/activity.proto
spkane31: Good call, removing

PR #835, opened 2026-07-19
time skipping supports max skip, polling, describe

feiyang3cat:
**What changed?**

1. add a max skip field  to `TimeSkippingConfig`
2. add `TimeSkippingInfo` to `DescribeWorkflowExecution` (contains virtual current time and time-skipping runtime status) 
3. add `PollWorkflowExecutionTimeSkipping` for fast-forward completion 
4. comment refinements: remove the workflow keyword from time-skipping message comments under the common path 
5. change `fast_forward` from a duration field to a message with both duration and id 

**Breaking changes**
Change #5

**Server PR**
https://github.com/temporalio/temporal/pull/11220


temporal/api/common/v1/message.proto
```diff
@@ -420,35 +422,68 @@ message TimeSkippingConfig {
...
+    // current execution and through a chain of runs (CaN, cron, retry).
     google.protobuf.Duration initial_skipped_duration = 1;
 
-    // If there is a fast-forward action set for the previous run in a chain of runs,
-    // the target time should be propagated to the next run as well.
+    // The fast-forward target time. It only propagates across a chain of runs within the same execution.
     google.protobuf.Timestamp fast_forward_target_time = 2;
+
+    // The initial skip count. It only propagates across a chain of runs within the same execution.
+    int32 initial_skip_count = 3;
```
feiyang3cat, line 478: for reviewers:  this shall be propagated so that we can stop workflow retry as workflow retry generates a new run instead retrying within the same run

temporal/api/common/v1/message.proto
```diff
@@ -420,35 +422,71 @@ message TimeSkippingConfig {
...
+    // unlimited retries when backoff is skipped.
+    // Every time the execution skips time, the skip count is incremented by one, and when it reaches
+    // max_skip_per_session, time skipping is disabled.
+    // For an execution with a chain of runs (retry, cron, continue-as-new), the count is accumulated
+    // across all runs within the same session. The count resets to 0 at the start of each new session,
+    // i.e. each time this config is updated to re-enable time skipping.
+    //
+    // If this field is not set, the server applies a large default value (e.g. 100). The default can
+    // be changed through dynamic config, and is overridden by this field when set.
+    int32 max_skip_per_session = 4;
```
feiyang3cat: we use a per_session max_skip instead of a total/delta max_skip to make this feature easy

the unneeded complexity introduced by other options:
1. total: when the value is updated, both the client and server need to read current skip number to decide if the new value is valid and we may need to expose the current skip number
2. delta: as we set a default value when the field is not set, if the user set TSC multiple times, delta may silently accumulate the total number to a large number

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3634,3 +3634,31 @@ message DeleteNexusOperationExecutionRequest {
...
+    enum Result {
+        // The poll timed out server-side before any time-skipping state change occurred to fast forward.
+        // The caller may poll again.
+        RESULT_UNSPECIFIED = 0;
+        // The execution's time-skipping config changed while the poll was waiting.
+        RESULT_TIMESKIPPING_CONFIG_UPDATED = 1;
+        // The fast-forward that was active when the poll started reached its target time and completed.
+        RESULT_FAST_FORWARD_COMPLETED = 2;
+        // No fast-forward was in progress when the poll started, so there was nothing to wait for.
+        RESULT_NO_PENDING_FAST_FORWARD = 3;
```
feiyang3cat: also need to add an result of workflow end (status = completed/failed/canceled/TERMINATED/TIMED_OUT)
but we need the whole execution (a chain of runs) to end instead of a single run's status

GregoryTravis (approved):

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3634,3 +3634,31 @@ message DeleteNexusOperationExecutionRequest {
...
+// A long-poll request that blocks until the time-skipping state of an execution changes, so callers
+// can observe fast-forward progress and time-skipping config updates without busy-polling.
+message PollWorkflowExecutionTimeSkippingRequest {
+    string namespace = 1;
+    temporal.api.common.v1.WorkflowExecution workflow_execution = 2;
+}
+
+message PollWorkflowExecutionTimeSkippingResponse {
+    enum Result {
+        // The poll timed out server-side before any time-skipping state change occurred to fast forward.
```
simvlad: Sound like this should be different than RESULT_UNSPECIFIED, maybe RESULT_TIMED_OUT?

temporal/api/common/v1/message.proto
```diff
@@ -403,15 +403,17 @@ message OnConflictOptions {
...
-// added to the workflow history to capture the state changes.
+// Options like fast_forward, disable_propagation, and max_skip_per_session are provided for granular
+// control of the execution's time skipping behavior. See each field's comment for a detailed explanation.
 //
+// An example of workflows with time skipping:
+// For workflows, an execution is a chain of runs including retries, cron, and continue-as-new.
+// In-flight work includes activities, child workflows, Nexus operations, signal/cancel external workflow operations, etc.
+// User timers are not classified as in-flight work and will be skipped over; the virtual clock may also skip to the
+// time point of the registered fast-forward when there is no in-flight work.
+// Every time time is skipped, the skip count is incremented by one; max_skip_per_session bounds the number of skips allowed within a single time-skipping session.
```
simvlad: time time :)

temporal/api/common/v1/message.proto
```diff
@@ -433,6 +433,11 @@ message TimeSkippingConfig {
     // pass beyond the end of the execution, and fast_forward won't have a chance to complete.
     google.protobuf.Duration fast_forward = 2;
 
+    // A client-supplied ID that must be set together with `fast_forward`. It is used to poll for
+    // fast-forward completion via PollWorkflowExecutionTimeSkipping.
```
feiyang3cat: for reviewers:  explicitly making this id required if fast_forward is used for simplicity

GregoryTravis: Is there a reason not to make it optional? The SDK only needs this to wait on the result of a FF, and it doesn't always need to do that.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3641,27 +3641,32 @@ message DeleteNexusOperationExecutionResponse {
...
-        // The poll timed out server-side before any time-skipping state change occurred to fast forward.
-        // The caller may poll again.
+        // Never returned; guards against an unset result.
         RESULT_UNSPECIFIED = 0;
-        // The execution's time-skipping config changed while the poll was waiting.
-        RESULT_TIMESKIPPING_CONFIG_UPDATED = 1;
-        // The fast-forward that was active when the poll started reached its target time and completed.
+        // The poll timed out server-side before the fast-forward completed. The caller may poll again.
+        RESULT_POLL_TIMEOUT = 1;
+        // The fast-forward identified by the request's `fast_forward_id` reached its target time and completed.
```
GregoryTravis: Maybe repeat here that the server only keeps the most recently started FF ID.

temporal/api/common/v1/message.proto
```diff
@@ -420,35 +422,81 @@ message TimeSkippingConfig {
...
+// Describes the current time-skipping state of a workflow execution.
+message TimeSkippingInfo {
+    // Current virtual time of the execution. If the execution hasn't skipped
+    // any time yet, it will be the same as wall clock time.
+    google.protobuf.Timestamp current_time = 1;
+
+    // If the execution is actively trying to skip time automatically when there is a chance,
+    // this field will be set to true. If time has stopped skipping either by fast-forward completion,
+    // max skip allowed checking, or user configuration, it will be false.
+    bool is_running = 2;
```
feiyang3cat: @tegan-temporal

temporal/api/workflowservice/v1/request_response.proto
feiyang3cat: udpated

temporal/api/common/v1/message.proto
```diff
@@ -403,52 +403,111 @@ message OnConflictOptions {
...
-    // If a new fast-forward is set via UpdateWorkflowExecutionOptions before the previous
-    // one completes, the new one will override the previous one.
-    // If the fast-forward duration exceeds the remaining execution timeout, time will only
-    // be fast-forwarded up to the end of the execution.
-    google.protobuf.Duration fast_forward = 2;
+    // An optional opt-in to control time-skipping behavior through fast-forward; see its definition for details.
+    FastForwardConfig fast_forward_config = 2;
 
     // By default, executions started by another execution (e.g. a child workflow of a parent workflow or
-    // a schedule with the timeskipping policy enabled), inherit the "enabled" flag and skip time when possible.
```
feiyang3cat, line 438: refine the comment: this is about **config** propagation for **different** executions (not different runs within the same execution

temporal/api/common/v1/message.proto
```diff
@@ -403,52 +403,111 @@ message OnConflictOptions {
...
+    // a schedule with the time-skipping policy enabled) inherit the "enabled" flag and skip time when possible.
     // This flag disables that inheritance.
     bool disable_propagation = 3;
+
+    // The maximum number of skips allowed every time this field is updated. It protects the execution from
```
feiyang3cat, line 434: @GregoryTravis

temporal/api/common/v1/message.proto
```diff
@@ -403,52 +403,111 @@ message OnConflictOptions {
...
+    // situations like unlimited retries when backoff is skipped.
+    //
+    // Every time the execution skips time, the skip count is incremented by one, and when it reaches
+    // max_session_skip_count, time skipping stops. Whenever this config field is updated, the accumulated
+    // skip count is cleared, marking the start of a new session.
+    // For an execution with a chain of runs (retry, cron, continue-as-new), the count is accumulated
+    // across all runs within the same session.
```
GregoryTravis, line 441: I wondering about an edge case where you execute a single fast forward, but across e.g. 3 retries: does this count as 3 skips against the max, or just one?

If it's just one, perhaps add something like "multiple retries during a single fast-forward only count as one skip".

feiyang3cat: it will count as 3, and the default value right now is 250. I think it will be fine users can change the default value or override the value using api. But indeed it is hard for user to estimate the number accurately if they don't want to study the internal server timer mechanism so it will be rough number.

GregoryTravis: How about a workflow that does 3 multiple sequential sleeps? Does each one count against the max?

I realize that this field is probably not one that users will set very often, but I want to figure out a way to write a test for it, if the SDK is going to expose it.

GregoryTravis: Or perhaps it is enough to test this by setting the field, and reading the configuration back and confirming it? Rather than verifying server behavior?

feiyang3cat: 1. skipping 3 sleeps will count as 3. server won't skip multiple timers at each skip.

2. I agree that users most of the time don't need to feel the existence of this option. And even if this max skip takes effect, they may not want to set this field but to figure out if their tests have unlimited retires.

temporal/api/common/v1/message.proto
feiyang3cat: I think both ways work:

1. server doesn't allow empty ff-id, and if users don't want to wait on it they can set a default id or random id
2. server allow setting empty ff-id,  and we need to define a sound behavior for polling with ff-id empty. I think the simplest way is we don't allow polling an empty ff-id 

     - 2.1. so polling with ff-id in the request will have an argument error 
     - 2.2. polling with a valid ff-id but the current ff-id is empty will return ff-id not matching
     - 2.3. in comments we indicate if users want to poll for the ff they need to set a ff-id for it

GregoryTravis: On the SDK side I'm currently always generated an ID and that's working, so I think it's fine to leave it as required.

simvlad (approved):

PR #836, opened 2026-07-20
API surface area for worker callbacks

chrsmith:
⚠️ This isn't meant to be merged any time soon. This is just part of the worker callbacks implementation. After the code review, I'll replace the `feature/worker-callbacks` branch with this.

---

**What changed?**

Adds the API surface area for the WIP worker callbacks feature.

Specifically, the changes are:

1. Add a `completion_callbacks` field to `StartNexusOperationExecutionRequest`. (How worker callbacks will be sent to the server.)
2. Introduce a new `Worker` variant of `commonpb.Callback`. (How worker callbacks will be encoded.)
3. Add a new `notificationservice/v1/request_response.proto` to define the input/output shape for worker callbacks. (The payload that will be delivered to workers.)

**Why?**

Today it is cumbersome to durably invoke Nexus operations and execute logic based on their results. "Worker callbacks" is a capability being developed that would allow developers to attach a Nexus operation to a SANO. Later, when the SANO reaches a terminal state, the attached Nexus operation would be invoked.

**Breaking changes**

None.

**Server PR**

TBD. The current prototype (based on the `feature/worker-callbacks` branch in this repo) can be found at:
https://github.com/temporalio/temporal/compare/feature/worker-callbacks?expand=1


temporal/api/common/v1/message.proto
```diff
@@ -201,10 +201,41 @@ message Callback {
...
+
+        // Target Nexus service, e.g. "temporal.api.notificationservice.v1.NotificationService".
+        string service = 2;
+        // Target operation, e.g. "OnComplete".
+        string operation = 3;
+
+        // Arbitrary user-supplied data from the source operation's callsite. (As applicable, not all operations
+        // support attaching context data.)
+        //
+        // There is a relatively small maximum size the source context can be, e.g. 32KiB.
```
bergundy, line 230: nit: You can also mention which dynamic config controls this limit for self-hosters.

temporal/api/common/v1/message.proto
```diff
@@ -201,10 +201,41 @@ message Callback {
...
+    // are purely "caller-side".)
+    //
+    // Worker callbacks are only supported for certain types of operations, e.g. standalone Nexus operations.
+    // Attempting to attach a Worker callback for an unsupported operation will result in an INVALID_ARGUMENT
+    // error from the server.
+    message Worker {
+        // Nexus task queue the Temporal worker is listening on.
+        //
+        // NOTE: This is not a temporal.api.taskqueue.v1.TaskQueue to avoid a circular dependency.
+        string taskqueue_name = 1;
```
bergundy: nit: we call this field `task_queue` everywhere, it will generate better across languages with the underscore. Rename to `task_queue_name`.

temporal/api/common/v1/message.proto
```diff
@@ -201,10 +201,41 @@ message Callback {
...
+
+        // Target Nexus service, e.g. "temporal.api.notificationservice.v1.NotificationService".
+        string service = 2;
+        // Target operation, e.g. "OnComplete".
+        string operation = 3;
```
bergundy: Let's just have a TODO to see if we want to allow customizing this from the get-go or not. The default can be on the server and it's easier to add fields later than remove them.

temporal/api/notificationservice/v1/request_response.proto
```diff
@@ -0,0 +1,57 @@
...
+    message Outcome {
+        oneof result {
+            // The operation was successful, and resulted in the given payload(s).
+            temporal.api.common.v1.Payloads success = 1;
+            // The operation failed. Includes timeout, cancellation, and application errors.
+            temporal.api.failure.v1.Failure failure = 2;
+        }
+    }
+
+    SourceOperation source_operation = 1;
```
bergundy: As discussed in person, I would leave this out for now. When we add support for attaching callbacks to other primitives, I am not convinced we will want to provide information per primitive here.

The link + source context get you quite a bit of information already.

If we do decide to keep this for the first iteration, I would name this differently because it's not necessarily an operation in the nexus sense, it's a primitive or component within a primitive (workflow run, workflow update, activity, schedule, nexus operation).

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3393,6 +3393,9 @@ message StartNexusOperationExecutionRequest {
     map<string, string> nexus_header = 15;
     // Metadata for use by user interfaces to display the fixed as-of-start summary and details of the operation.
     temporal.api.sdk.v1.UserMetadata user_metadata = 16;
+
+    // Completion callbacks to be invoked once the Nexus operation reaches a terminal state.
+    repeated temporal.api.common.v1.Callback completion_callbacks = 17;
```
bergundy: You'll need to add the full attachment support that we have for request IDs and on conflict options for workflows and activities.

bergundy (commented):
You're missing a new link definition to a callback on an execution.

temporal/api/common/v1/message.proto
chrsmith: Good call. I haven't added that on the server-side yet, but will file a ticket and make a point to update this comment when that dynamic config value actually exists.

I was thinking we'd go with a new dynamic config value like `limit.callbackContext.warn` and `limit.callbackContext.error`. (Patterned after `limit.blobSize.*`, `limit.memoSize.*`, etc.)

temporal/api/common/v1/message.proto
chrsmith: Asking because I'm not sure: is it a problem that we aren't using the proper `TaskQueue` proto type here?

This is the full proto definition. We'd be missing out on the `kind` and `normal_name` of task queue. And just assuming it's `TASK_QUEUE_KIND_NORMAL`. (And never `TASK_QUEUE_KIND_WORKER_COMMANDS` or `TASK_QUEUE_KIND_STICKY`.)

Could that be a problem down the road?

temporal/api/common/v1/message.proto
chrsmith: I'm inclined to keep this as-is, and allow for callbacks to specify an any Nexus service and operation. By fixing those values it makes using worker callbacks much more painful. (Since you need to spin up multiple Temporal workers, with different versions of the `NotificationService` listening to different task queues.)

I could be convinced otherwise, but I'd rather start with this and see if we can convince ourselves that it really isn't needed than leave it as a TODO. (e.g. let's bring this up during the the Cross-SDK review.)

temporal/api/notificationservice/v1/request_response.proto
chrsmith: I could be convinced otherwise, but I don't think the "link + source context" is sufficient. (But it depends on what the link actually _is_, and the information that's available.)

The next rev of this PR is specifically about exposing the `CallbackExecutionInfo` and explaining why I don't think we should be adding links. (So let's resolve those bigger topics, before we get into the details of this particular proto and the link it would be sent with.)

temporal/api/workflowservice/v1/request_response.proto
chrsmith: Would you mind elaborating on what you mean by this? What do you mean by "full attachment" here?

I'm patterning this after the way we added completion callbacks to `StartActivityExecutionRequest` and `StartWorkflowExecutionRequest`. I see in both of those other request types there are `id_conflict_policy` / `id_reuse_policy` fields. But `StartNexusOperationExecutionRequest` has similar fields as well.

Are you saying that each individual `commonpb.Callback`s should have its own (potentially user-supplied) `callback_id` ? And ID Conflict/Reuse policies supplied too? That doesn't make sense to me. Since if the SANO is accepted by Temporal, the callbacks should be accepted as well. (And I don't see any sort of `UpdateNexusOperationExecution`-type command, that would allow us to merge/update callbacks?)

chrsmith:
@bergundy Thanks for the review, please take another look.

- Expose a new `CallbackExecutionInfo` proto from `DescribeNexusOperationExecutionResponse`. (It's a subset of what was going to be added for _Manual Completion_ in https://github.com/temporalio/api/pull/774.) 
- Add a new `Callback` variant of `Link`.

temporal/api/common/v1/message.proto
bergundy: Discussed on slack. Agreed that we should take a service and operation name. But with that point, let's replace the `temporal.api.notificationservice.v1.NotificationService` example with something that is more likely to be written by our users.

temporal/api/common/v1/message.proto
```diff
@@ -273,12 +304,28 @@ message Link {
...
+    // A link to a callback or completion handler for an operation. e.g. not a link to a standalone Nexus operation,
+    // but the callback that was executed once that operation finished.
+    message Callback {
+        message NexusOperationCompletion {
+            string operation_id = 1;
+            string run_id = 2;
+            string request_id  = 3;
+        }
+
+        string callback_id = 1;
```
bergundy: There's no `callback_id` to refer to in all of our models.

temporal/api/common/v1/message.proto
```diff
@@ -273,12 +304,28 @@ message Link {
...
+        oneof variant {
+            NexusOperationCompletion nexus_operation_completion = 2;
```
bergundy: This should be field number 1.

temporal/api/common/v1/message.proto
```diff
@@ -273,12 +304,28 @@ message Link {
...
+        }
+    }
+
     oneof variant {
         WorkflowEvent workflow_event = 1;
         BatchJob batch_job = 2;
         Activity activity = 3;
         NexusOperation nexus_operation = 4;
         Workflow workflow = 5;
+        Callback callback = 6;
```
bergundy: Maybe just flatten the `Callback into this variant`? `NexusOperationCompletionCallback`? Or just `NexusOperationCallback`?

temporal/api/notificationservice/v1/request_response.proto
bergundy: Given the recent discussion on slack, are you still thinking we won't expose links?

temporal/api/workflowservice/v1/request_response.proto
bergundy: Let's sync up on this

temporal/api/workflowservice/v1/message.proto
```diff
@@ -0,0 +1,74 @@
...
+        // The callback completed successfully. (Which may include delivering a "failed" result successfully.)
+        google.protobuf.Empty success = 1;
+        // The failure if the callback was not able to complete successfully. e.g. timed out, received an
+        // unretriable error, etc.
+        temporal.api.failure.v1.Failure failure = 2;
+    }
+}
+
+// Information about a callback's execution.
+message CallbackExecutionInfo {
```
bergundy: These message should not live in workflowservice, this package only contains service definitions and requests and responses. There's also no need to redefine these. Reuse https://github.com/temporalio/api/blob/1a137206f19d45f77bf029dce97f68fa2abf630b/temporal/api/callback/v1/message.proto#L19 instead.

temporal/api/workflowservice/v1/message.proto
```diff
@@ -0,0 +1,74 @@
...
+    // Identifier of the callback's execution. Unique within the containing namespace.
+    string callback_id = 1;
+    // Run ID of the callback execution.
+    string run_id = 2;
```
bergundy: These callbacks are not separate execution records, they are embedded within an execution. We may add standalone callbacks later which will have their own business ID and run ID, but, if you remember, we punted on that feature.

temporal/api/notificationservice/v1/request_response.proto
chrsmith: I have been convinced that we should be exposing links here. And so when the developer were to go and implement their Nexus handler's `Start(...)` method they would have a link to the source callback. (But not a link to the source SANO which completed, that triggered the callback.)

And from that `Link` object the user should be able to fetch any additional data they require. So I'll remove this `source_operation` field and add a comment about links in the next rev.

temporal/api/common/v1/message.proto
chrsmith: A SANO operation could have multiple worker callbacks associated with it. So in order for the link to not be ambiguous, we need some sort of unique identifier for _which_ callback on the SANO.

The user would be able to see those IDs on the `workflowservice.CallbackExecutionInfo` messages, returned from `DescribeNexusOperationExecution`.

Does that sound right? Or is it that we don't actually need to have `callback_id` here?

temporal/api/common/v1/message.proto
chrsmith: I can be convinced otherwise, but I'm inclined to keep it as-is. Let me know if you disagree.

When we later support attaching worker callbacks to Workflows and Activities, it would be kind of lame to have unique kinds of links for those. e.g. `WorkflowCompletionCallback` and `ActivityCompletionCallback`. Do we want to treat those as three different kinds of Links or just a Link to a callback, which can be pointing to different resources?

temporal/api/workflowservice/v1/message.proto
chrsmith: 👍  Good catch. I had missed there was already a `temporal/api/callback/v1/message.proto`. I only moved this to `workflowservice` because I couldn't add the required dependency on `failurepb.Failure` in the `common/v1` namespace.

temporal/api/workflowservice/v1/message.proto
chrsmith: I'll admit the my understanding of "separate execution records" or the benefit of "standalone callbacks" is hazy. But are you saying that all SANO callbacks will have the same `run_id`? (Which would be the same as the source SANO?)

That makes sense. And the implication is just that we should remove the `run_id` field here? (Since it would be implied/required by the call to `DescribeNexusOperationExecution` that is returning these `CallbackExecutionInfo` objects?)

temporal/api/common/v1/message.proto
chrsmith: Done. #namingishard

```
        // Target Nexus service, e.g. "HTTPAdaptor".
        string service = 2;
        // Target operation, e.g. "DeliverAsWebhook".
        string operation = 3;
```

But this does beg the question, do we still want to define the request/response protos in a `notificationservice/v1/` namespace? Or would "worker callback" types be more applicable to being in the `workflowservice` or `nexus` namespaces?

temporal/api/common/v1/message.proto
chrsmith: Resolving this to cut down on the noise. As the numbering of this field is dependent on whether or not we keep `callback_id = 1`.

temporal/api/workflowservice/v1/message.proto
chrsmith: Removed `run_id` from `CallbackExecutionInfo` that was added to `callback/v1/message.proto`.

chrsmith:
Rebased and force-pushed. @bergundy PTAL. Though that last one re: "full attachment" will probably need a sync conversation, as I'm not sure what you're asking for there.

Here are the open questions/threads remaining:

- Exposing a `callback_id` field on `Link_Callback` objects. We need that, right?
    https://github.com/temporalio/api/pull/836#discussion_r3639214652

- Happy to have `Link_NexusOperationCompletionCallback`, but are you sure we want that _and_ `Link_WorkflowCompletionCallback` later?
    https://github.com/temporalio/api/pull/836#discussion_r3639232950

- Is there any problem with us only having the task queue _name_ and not its kind?
    https://github.com/temporalio/api/pull/836#discussion_r3623659637

- Adding "full attachment support that we have for request IDs and on conflict options". I'm not sure what that means.
    https://github.com/temporalio/api/pull/836#discussion_r3623302367

temporal/api/callback/v1/message.proto
```diff
@@ -34,4 +35,23 @@ message CallbackInfo {
...
+}
+
+message CallbackExecutionInfo {
+    // Identifier of the callback's execution. Unique within the containing namespace.
+    string callback_id = 1;
+
+    CallbackInfo info = 2;
+    CallbackExecutionOutcome outcome = 3;
+}
+
```
bergundy:
```suggestion
    // TODO document me
    CallbackExecutionOutcome outcome = 9;
}

message CallbackExecutionOutcome {
    oneof value {
        // The callback completed successfully. (Which may include delivering a "failed" result successfully.)
        google.protobuf.Empty success = 1;
        // The failure if the callback was not able to complete successfully. e.g. timed out, received an
        // unretriable error, etc.
        temporal.api.failure.v1.Failure failure = 2;
    }
}

File: temporal/api/common/v1/message.proto [hunk 12]
```diff
@@ -201,10 +201,41 @@ message Callback {
...
+    // Worker callbacks are only supported for certain types of operations, e.g. standalone Nexus operations.
+    // Attempting to attach a Worker callback for an unsupported operation will result in an INVALID_ARGUMENT
+    // error from the server.
+    message Worker {
+        // Nexus task queue the Temporal worker is listening on.
+        //
+        // NOTE: This is not a temporal.api.taskqueue.v1.TaskQueue to avoid a circular dependency.
+        string task_queue_name = 1;
+
+        // Target Nexus service, e.g. "HTTPAdaptor".
```
bergundy: nit:

```suggestion
        // Target Nexus service, e.g. "HTTPAdapter".
```

temporal/api/common/v1/message.proto
bergundy: That's what the request ID is for.

temporal/api/common/v1/message.proto
bergundy: This works too.

temporal/api/notificationservice/v1/request_response.proto
```diff
@@ -0,0 +1,40 @@
...
+// allowing for defining completion handlers for arbitrary asynchronous operations.
+//
+// Information about the source operation will be available in the form of a commonpb.Link,
+// which will be available separately from this OnCompleteRequest. e.g. a link to the source
+// standalone Nexus operation would be found in the nexuspb.StartOperationRequest parameter
+// sent to the worker callback. (In addition to this OnCompleteRequest.)
+message OnCompleteRequest {
+
+    // The outcome of the source operation.
+    message Outcome {
```
bergundy, line 25: Might be worth defining Outcome in `common` at this point.

temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -3441,6 +3445,10 @@ message DescribeNexusOperationExecutionResponse {
 
     // Token for follow-on long-poll requests. Absent only if the operation is complete.
     bytes long_poll_token = 6;
+
+    // Completion callbacks to be invoked once the Nexus operation reaches a terminal state.
+    // They will remain in the CALLBACK_STATE_STANDBY state until the Nexus operation is finished.
+    repeated temporal.api.callback.v1.CallbackExecutionInfo completion_callbacks = 7;
```
bergundy:
```suggestion
    repeated temporal.api.nexusoperation.v1.CallbackInfo callbacks = 7;
```

And copy from https://github.com/temporalio/api/blob/f53963d4489c8a73aa30bd7091fe758f9896c08c/temporal/api/activity/v1/message.proto#L233

temporal/api/callback/v1/message.proto
chrsmith: I wasn't sure how concerned we'd be with including the failure objects in `CallbackInfo`, so I created the separate `-ExecutionInfo`. But I'm happy to make the change to just add the `-ExecutionOutcome` as a field to `CallbackInfo`. (And existing callsites would just leave it blank, and maybe we'd wire through a `bool include_outcome` field in the request object or something.)

temporal/api/common/v1/message.proto
chrsmith: I don't follow. Would you mind elaborating on that? ... in fact, what does `NexusOperationCompletion::request_id` mean in this context?

Would that be the `request_id` that was sent when we go and invoke the worker callback? (e.g. the `nexuspb.StartOperationRequest::request_id` that is sent to the worker?)

temporal/api/notificationservice/v1/request_response.proto
chrsmith: I tried, but it would require some more refactoring since that would lead to a circular dependency between `commonpb` and `failurepb`. Maybe we could expose some _other_ or simpler error type in the `Outcome::failure` case. But the `failurepb.Failure` seems kinda useful in how it strongly types various categories of errors.

temporal/api/workflowservice/v1/request_response.proto
chrsmith: Ack, will get to in the next rev. And having different `CallbackInfo` objects defined in different namespaces (e.g. `/activity`, `/nexusoperation` will sidestep some other problems we might have.)

temporal/api/common/v1/message.proto
chrsmith: Circling back, since we'll remove the `callback_id` I agree it makes more sense to just flatten these. I'll introduce a top-level `NexusOperationCompletionCallback` link kind. (Since we wouldn't be adding anything by wrapping them as variants of `Link_Callback`.)

chrsmith:
@bergundy PTAL. I moved some of the types around, e.g. introducing a new `nexusoperations.v1` namespace for a specialized `CallbackInfo` type. But I suspect we might want a sync meeting to get the question around if/how/why we would want to expose ID reuse/conflict policies when attaching worker callbacks. And double check that the `Link_NexusOperationCallback` proto is sufficient for tracking which worker callback execution is being linked to.

temporal/api/workflowservice/v1/request_response.proto
chrsmith: Done. The latest commit adds a new `temporal.api.nexusoperation.v1.OnConflictOptions on_conflict_options = 18;` field.

temporal/api/callback/v1/message.proto
```diff
@@ -34,4 +35,17 @@ message CallbackInfo {
...
     // If the state is BLOCKED, blocked reason provides additional information.
     string blocked_reason = 8;
-}
\ No newline at end of file
+
+    // Result of the callback's execution, only set when the callback reaches a terminal state.
+    CallbackExecutionOutcome outcome = 9;
+}
+
+message CallbackExecutionOutcome {
```
bergundy: nit: "execution" refers to a top level entity in our system that has a business ID and run ID.

```suggestion
message CallbackOutcome {
```

temporal/api/nexusoperation/v1/message.proto
```diff
@@ -0,0 +1,42 @@
...
+// When StartNexusOperationExecutionRequest uses the conflict policy NEXUS_OPERATION_ID_CONFLICT_POLICY_USE_EXISTING
+// and there is already an existing, running standalone Nexus operation, OnConflictOptions defines actions to be
+// taken.
+message OnConflictOptions {
+    // Attaches the request ID to the running operation.
+    bool attach_request_id = 1;
+    // Attaches the completion callbacks to the running operation.
+    bool attach_completion_callbacks = 2;
+    // Attaches any new links to the running operation.
+    bool attach_links = 3;
```
bergundy, line 41: We don't support (start request) link attachment yet. Callbacks can come with their own link but other start APIS support adding links without callbacks as well. I would add support for only-link attachment as part of this change for completeness. See `StartActivityExecution` and `StartWorkflowExecution` for reference.

bergundy (approved):
Overall LGTM. Added a couple of minor comments.

temporal/api/callback/v1/message.proto
chrsmith: Good to know, thanks.

chrsmith:
With @bergundy giving this PR an LGTM, I'll consider this the "official" v1 of the worker callbacks API surface area.

I'll merge this into the long-lived feature branch `feature/worker-callbacks` and close this PR out.

PR #840, opened 2026-07-23
💥 Update documention of current_retry_interval and next_attempt_schedule_time

dandavison:
**What changed?**
- Update definition of `current_retry_interval` for WFA to reflect breaking behavior change made in https://github.com/temporalio/temporal/pull/11203
- Use equivalent wording for SAA

**Why?**
The current behavior of WFA was deemed to be a bug and changed in https://github.com/temporalio/temporal/pull/11203

### DIFF
--- a/temporal/api/activity/v1/message.proto
+++ b/temporal/api/activity/v1/message.proto
@@ -139,19 +139,20 @@ message ActivityExecutionInfo {
     temporal.api.failure.v1.Failure last_failure = 20;
     string last_worker_identity = 21;
 
-    // Time from the last attempt failure to the next activity retry.
-    // If the activity is currently running, this represents the next retry interval in case the attempt fails.
-    // If activity is currently backing off between attempt, this represents the current retry interval.
-    // If there is no next retry allowed, this field will be null.
-    // This interval is typically calculated from the specified retry policy, but may be modified if an activity fails
-    // with a retryable application failure specifying a retry delay.
+    // The retry interval for a currently pending retry: the amount of time between the last attempt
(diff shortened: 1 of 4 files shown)


Quinn-With-Two-Ns (approved):

temporal/api/activity/v1/message.proto
```diff
@@ -130,23 +124,23 @@ message ActivityExecutionInfo {
...
-    // If the activity is currently running, this represents the next retry interval in case the attempt fails.
-    // If activity is currently backing off between attempt, this represents the current retry interval.
-    // If there is no next retry allowed, this field will be null.
-    // This interval is typically calculated from the specified retry policy, but may be modified if an activity fails
-    // with a retryable application failure specifying a retry delay.
+    // The retry interval for a currently pending retry: the amount of time between the last attempt
+    // failure and when the next attempt will become available to worker poll requests. Set only
+    // during retry backoff, i.e. when an activity is currently waiting for a retry, and the next
+    // activity task has not yet become available to worker poll requests. Null at all other times,
+    // including when paused and during start delay. This is calculated from the active retry policy,
```
fretz12, line 146: I believe when paused, the server still returns the stale future next_attempt_schedule_time and current_retry_interval. So I think we should clear those during pause for server

fretz12 (approved):

temporal/api/activity/v1/message.proto
dandavison: Thanks updated the server PR to do that.

spkane31 (approved):

PR #846, opened 2026-08-01
Drop `reset_attempts` and `reset_heartbeat` from `UnpauseActivityExecution`

dandavison:
**What changed?**
- Drop `reset_attempts` and `reset_heartbeat` from `UnpauseActivityExecution`

**Why?**
- They are confusing: they mix `Unpause` and `Reset` functionality in a confusing way
- We have so far been unable to assign desirable and consistent semantics to them during implementation
- No known user demand
- They can be added later

A commit for extending this to WFA is reverted in history; we can do this later.

**Breaking changes**
- This API has always been rejected by the server. When server starts to accept it, an old client could submit these options and they would be ignored. Operator API is not GA.

### DIFF
--- a/openapi/openapiv2.json
+++ b/openapi/openapiv2.json
@@ -303,7 +303,7 @@
     "/api/v1/namespaces/{namespace}/activities-deprecated/unpause": {
       "post": {
         "summary": "UnpauseActivity unpauses the execution of an activity specified by its ID or type.\nIf there are multiple pending activities of the provided type - all of them will be unpaused.",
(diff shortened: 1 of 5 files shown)


spkane31 (approved):
Do we need to `reserve` these keywords if this API has never been used in the server? IMO we can keep the api surface a bit cleaner by not using the `reserve` keyword and acknowledging this API has never been exposed and isn't a real breaking change.

dandavison:
> Do we need to reserve these keywords if this API has never been used in the server?

We should, because the API protos have been released, and our SDKs have released autogenerated clients that will send those fields in positions 6 and 7.

fretz12 (approved):

maciejdudko (approved):

GregoryTravis (approved):

PR #853, opened 2026-08-18
WIP: support draft APIs 

mavemuri:
**What changed?**

**Why?**

**Breaking changes**

**Server PR**


cmd/generate-stable-protos/tags.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+
+	"github.com/bufbuild/protocompile/experimental/ir"
+	"github.com/bufbuild/protocompile/experimental/seq"
+)
+
+// draftTags is the set of fully-qualified extension names that mark a
+// declaration as experimental/draft.
```
stephanos, line 11: nit: we should align on "experimental" or "draft" and not use them interchangeably. I think "draft" makes most sense as we already use "experimental" in stable API comments.

temporal/api/batch/v1/message.proto
```diff
@@ -1,14 +1,10 @@
+// Code generated by generate-stable-protos. DO NOT EDIT.
+// @generated
+
 syntax = "proto3";
 
 package temporal.api.batch.v1;
 
-option go_package = "go.temporal.io/api/batch/v1;batch";
```
stephanos, line 5: Given there's some churn in these files (e.g. moved `option`s, changed indentation), why don't we land a PR before any of this to only format these files the same way already? That makes the change here much smaller. If I see this right, a simple `buf format -w` should do the trick. That also ensures that both set of files (api and api_next) look the same structurally; it'd be odd if they didn't.

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+	for path, text := range protos {
+		protos[path] = generatedHeader + text
+	}
+}
+
+func stripDrafts(ctx context.Context, root string, paths []string) ([]string, map[string]string, error) {
+	session := &ir.Session{}
+	opener := baseOpener(root)
+
+	// first, lower the api_next set into ir
```
stephanos, line 91: nit: IMO it's a good habit to write all comments as full sentences ie capitalized and with `.`. For consistency.

temporal/api_next/workflow/v1/message.proto
```diff
@@ -0,0 +1,762 @@
+syntax = "proto3";
```
stephanos, line 1: nit: calling out the obvious but still calling it out; these files should me "moved" as far as git must be concerned.

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+		return nil, nil, fmt.Errorf("%s not found in the corpus; cannot discover the draft-tag extension set", experimentalTagsPath)
+	}
+	tags, err := getDraftTags(experimentalFile)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	stableOrder := make([]string, 0, len(paths))
+	protoTexts := make(map[string]string, len(paths))
+	for p, irFile := range byPath {
```
stephanos, line 126: do we need `stableOrder`? I'd simplify it and only return `protoTexts` and use `paths := slices.Sorted(maps.Keys(texts))` as needed. Less efficient; but that doesn't matter here and it's simpler.

cmd/generate-stable-protos/opener.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// diskOpener resolves repo-relative proto paths ("temporal/api_next/x.proto")
+// against a directory on disk.
+func diskOpener(root string) source.Opener {
+	return &source.FS{FS: os.DirFS(root)}
+}
+
+// baseOpener is the disk corpus plus well-known types, shared by every pass
+// that reads real files unmodified (currently just stripExperimentalChanges,
```
stephanos, line 16: nit: `stripExperimentalChanges` doesn't exist; same for some of the other references in these comments

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
+package main
```
stephanos, line 1: WDYT about having a folder `testdata`/`fixtures` that includes a folder for each variation of a draft API and has a `input.proto` and `output.proto` file? I think that would be a great way to clearly test it and make it easy to inspect visually.

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+		return nil, fmt.Errorf("walk %q: %w", dir, err)
+	}
+	sort.Strings(sources)
+	return sources, nil
+}
+
+// rewriteImportPaths rewrites the surviving imports from api_next to api
+func rewriteImportPaths(text string, nextPaths []string) string {
+	for _, path := range nextPaths {
+		text = strings.ReplaceAll(text, `"`+path+`"`, `"`+stablePath(path)+`"`)
```
stephanos, line 318: I think
```
import public "new.proto";
import 'other.proto'; // single quotes
```
are also valid, though? Could/should we support or fail those?

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/bufbuild/protocompile/experimental/ast/printer"
+	"github.com/bufbuild/protocompile/experimental/incremental"
+	"github.com/bufbuild/protocompile/experimental/incremental/queries"
+	"github.com/bufbuild/protocompile/experimental/ir"
+	"github.com/bufbuild/protocompile/experimental/report"
+)
+
+const (
+	apiNextPrefix   = "temporal/api_next/"
+	stableAPIPrefix = "temporal/api/"
```
stephanos, line 24: What about making the tool not know about "temporal"? If we make the only assumptions that there's an `api_next` and `api` folder, and that the `api_next` one contains `protometa/v1/experimental.proto` it would work on anything.

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+	addGeneratedHeader(stableProtos)
+
+	if err := verifyGeneratedProtos(ctx, root, stableProtos); err != nil {
+		return err
+	}
+
+	return writeFiles(outDir, stableProtos)
+}
+
+func addGeneratedHeader(protos map[string]string) {
```
stephanos, line 81: nit: personally; I'd inline a function like this (ie both fairly short and only used once) as I find it easier to read from top to bottom than jump back and forth between function calls.

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("print %s: %w", p, err)
+		}
+		stableTexts[stablePath(p)] = rewriteImportPaths(protoText, survivingNextPaths)
+	}
+	return stableTexts, nil
+}
+
+// re-lower one more time to ensure there are no compilation issues
+func verifyGeneratedProtos(ctx context.Context, root string, stableTexts map[string]string) error {
```
stephanos, line 220: Good idea!

cmd/generate-stable-protos/opener.go
```diff
@@ -0,0 +1,43 @@
...
+// note on memOpener below for why.
+func baseOpener(root string) source.Opener {
+	return &source.Openers{source.WKTs(), diskOpener(root)}
+}
+
+// memOpener serves an entire in-memory set of final texts (keyed by their
+// paths in whatever path space the caller is currently working in) ahead
+// of the disk corpus and WKTs. Used by stripUnusedImports (re-lowering
+// draft-stripped text) and verifyStrippedProtos (combined cross-file
+// verification of the final stable tree).
```
stephanos, line 28: nit: I strip the `used by ...` part as (a) it's easy to lookup where a function is used and (b) it's easy to become stale

cmd/generate-stable-protos/tags.go
stephanos: This also applies to the proto file name.

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+	opener := memOpener(root, texts)
+
+	relowerQueries := make([]incremental.Query[*ir.File], len(order))
+	for i, p := range order {
+		relowerQueries[i] = queries.IR{Opener: opener, Session: session, Path: p}
+	}
+	results, err := lowerFiles(ctx, relowerQueries...)
+	if err != nil {
+		return nil, fmt.Errorf("re-lower draft-stripped corpus: %w", err)
+	}
```
stephanos, line 170: There might be a helper opportunity here; I see three places create a session, construct queries, lower.

cmd/generate-stable-protos/main.go
```diff
@@ -0,0 +1,321 @@
...
+
+	lowerQueries := make([]incremental.Query[*ir.File], len(paths))
+	for i, p := range paths {
+		lowerQueries[i] = queries.IR{Opener: opener, Session: session, Path: p}
+	}
+	results, err := lowerFiles(ctx, lowerQueries...)
+	if err != nil {
+		return fmt.Errorf("verify stable tree failed: %w", err)
+	}
+	var jErr error
```
stephanos, line 238: what's the `j` for?

PR #854, opened 2026-08-18
Adding field for gradual ramp requests

stuart-wells:
**What changed?**
Adding field for requesting a gradual ramp duration on a namespace.

**Why?**
Gradual ramp will allow a namespace that was just connected for replication to come up slowly and not overwhelm the connected passive cell. It is best to do this on a per-namespace basis instead of having a cell-wide policy.

**Breaking changes**
n/a

**Server PR**
n/a


temporal/api/workflowservice/v1/request_response.proto
```diff
@@ -124,6 +124,9 @@ message UpdateNamespaceRequest {
     string delete_bad_binary = 6;
     // promote local namespace to global namespace. Ignored if namespace is already global namespace.
     bool promote_namespace = 7;
+    // Duration over which replication traffic should ramp to 100% for passive clusters added by
+    // this request. An unset or non-positive duration disables the ramp for this request.
+    google.protobuf.Duration replication_ramp_duration = 8;
```
xwduan: Will it be better to put it inside NamespaceReplicationConfig?

stuart-wells: Sure that works too. I left it out because it's not included in the response (not a general property of the namespace, only added to clusters added by this request), but I'm fine with it being in NamespaceReplicationConfig

xwduan (approved):

tconley1428 (approved):

PR #859, opened 2026-08-25
buf v2 config

dandavison:
**What changed?**

- Migrated `buf.yaml` and `buf.lock` from config v1 to v2 using `buf config migrate --module .`.

- Bumped buf pinned in the Makefile from v1.27.0 to v1.49.0, and renamed `buf mod prune` to `buf dep prune`.

- Dropped the unused `buf.build/grpc-ecosystem/grpc-gateway` dependency.

- Renamed the lint category `DEFAULT` to `STANDARD`, and replaced the two comments on `excludes` with one line.

- Deleted `buf.gen.yaml`

**Why?**

- [LSP support](https://buf.build/blog/protobuf-lsp): The buf language server does not discover module roots from a v1 `buf.yaml`.

- pin has to move because v1.27.0 predates v2 and rejects it. v1.49.0 is the version `.github/workflows/push-to-buf.yml` already runs against this repo.

- `DEFAULT` was renamed to `STANDARD` in buf v1.40.0.

- AFAIK no-one calls `buf generate`; we use `protoc`

**Breaking changes**

Rerun `make buf-install` to pick up the new buf. Otherwise none.


buf.yaml
```diff
@@ -1,24 +1,28 @@
...
+version: v2
+modules:
+  - path: .
+    name: buf.build/temporalio/api
+    excludes:
+      # Vendored for api-linter (can't read the BSR); excluded so buf sees them once.
+      - google
+      - nexusannotations
 deps:
-  - buf.build/grpc-ecosystem/grpc-gateway
```
stephanos, line 4: I guess we never used this?

buf.yaml
```diff
@@ -1,24 +1,28 @@
...
-    # Same for nexusannotations - local copy for api-linter, BSR dep for buf
-    - nexusannotations
-breaking:
+lint:
   use:
-    - WIRE_JSON
+    - STANDARD
+  except:
+    - FIELD_NOT_REQUIRED
+    - PACKAGE_NO_IMPORT_CYCLE
```
stephanos: Are these precautionary exclusions or would these actually break the lint today?

dandavison: Thanks. They were added automatically by `buf config migrate --module .` in d81350b, but I've removed them now.

buf.yaml
dandavison: Correct.

buf.yaml
```diff
@@ -1,24 +1,25 @@
...
+    - cmd
     - google
-lint:
+  disallow_comment_ignores: true
+breaking:
   use:
-    - DEFAULT
+    - WIRE_JSON
+  except:
+    - FIELD_SAME_DEFAULT
```
stephanos: Could also check if this is needed.

stephanos (approved):
LGTM with limited `buf` knowledge. Fact that behavior seems the same is good.

mjameswh (approved):

PR #868, opened 2026-09-09
Allow releasing specific commits, and publishing immediately

stpierre:
**What changed?**

Add `api_ref` and `api_go_ref` so a caller can name the commits to release, and `auto_publish` so it can skip the manual publish step. All three default to the existing behavior, so a manual dispatch is unaffected.

**Why?**

These changes will allow us to automate API releases as part of server releases.

`create-release.yml` could only release the head of a branch, and only as a draft. Neither works for an automated caller: a cloud release branch pins a specific api-go commit, and a draft release creates no tag, which is the thing the release invariant actually requires.

**Breaking changes**

None.

**Server PR**

N/A

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -191,4 +191,4 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.generate_token.outputs.token }}
           TAG: ${{ github.event.inputs.tag }}
-        run: gh release edit "$TAG" --draft=false -R "$GITHUB_REPOSITORY"
+        run: gh release edit "$TAG" --draft=false -R "$GITHUB_REPOSITORY" --latest
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -20,6 +20,22 @@ on:
         description: An ID used by external tools to identify workflow runs(can be left empty when running manually)
         default: "none"
         type: string
+      api_ref:
+        description: "api commit or ref to release; defaults to `branch`"
+        default: ""
+        type: string
+      api_go_ref:
+        description: "api-go commit or ref to release; defaults to `branch`"
+        default: ""
+        type: string
+      auto_publish:
+        description: "Publish the release instead of leaving a draft. A draft creates no tag, so automated callers need this."
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+
 jobs:
   dispatch:
     runs-on: ubuntu-latest
@@ -33,10 +49,15 @@ jobs:
       api_commit_sha: ${{ steps.pin_commits.outputs.api_commit_sha }}
       api_go_commit_sha: ${{ steps.pin_commits.outputs.api_go_commit_sha }}
     steps:
+      # api_ref/api_go_ref let a caller release specific commits rather than
+      # the head of a branch. They must name a corresponding pair: the "Pin
+      # commits sha" step below asserts that the api-go commit's proto/api
+      # submodule points at the api commit. Left empty, both fall back to
+      # `branch`.
       - name: Checkout api
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
-          ref: ${{ github.event.inputs.branch }}
+          ref: ${{ inputs.api_ref || inputs.branch }}
           fetch-depth: 0
           fetch-tags: true
           path: api
@@ -45,7 +66,7 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           repository: temporalio/api-go
-          ref: ${{ github.event.inputs.branch }}
+          ref: ${{ inputs.api_go_ref || inputs.branch }}
           submodules: true
           path: api-go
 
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yml
```diff
@@ -141,3 +162,33 @@ jobs:
...
+        with:
+          app-id: ${{ secrets.TEMPORAL_CICD_APP_ID }}
+          private-key: ${{ secrets.TEMPORAL_CICD_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+
+      - name: Publish release
+        env:
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+          TAG: ${{ github.event.inputs.tag }}
+        run: gh release edit "$TAG" --draft=false -R "$GITHUB_REPOSITORY"
```
stephanos: Does this always set the new release as "latest"? If so, maybe we need an extra input to decide if this is "latest" or not.

.github/workflows/create-release.yml
```diff
@@ -141,3 +162,33 @@ jobs:
       api_commit_sha: ${{ needs.prepare-inputs.outputs.api_commit_sha }}
       base_tag: ${{ github.event.inputs.base_tag }}
     secrets: inherit
+
+  # Publishing this repo's release fires `release: published`, which
+  # trigger-api-go-publish-release.yml turns into api-go's publish-release.yml
+  # -- the step that actually creates the api-go tag.
```
stephanos: 🤖

<details>
<summary><strong>med</strong> — A green run no longer implies the api-go tag exists; the publish chain is fire-and-forget.</summary>

The comment is right about the chain, but the causality is asynchronous. `gh release edit --draft=false` returns as soon as the API call succeeds; `trigger-api-go-publish-release.yml` → api-go `publish-release.yml` then runs in a *separate* workflow run that this run neither waits on nor observes.

That downstream job can fail independently — it looks the draft up by name:

```sh
RELEASE_TMP_NAME="${TAG}-${API_COMMIT_SHA}"
RELEASE_ID=$(gh api /repos/temporalio/api-go/releases --jq '.[] | select(.name == "'"${RELEASE_TMP_NAME}"'") | .id')
if [[ -z "${RELEASE_ID}" ]]; then echo "Release not found"; exit 1; fi
```

with `api_commit_sha` sourced from `github.event.release.target_commitish`. If that lookup misses, or the PATCH fails, this workflow still reports success. An automated caller that waits on this run and then assumes `go.temporal.io/api@$TAG` is resolvable will proceed against a tag that does not exist — which is precisely the invariant the PR description says automated callers need.

</details>

**Suggestion:** Have the `publish-release` job confirm the end state before exiting — after `gh release edit`, poll for the api-go tag (or for the api-go release with `draft == false`) with a bounded timeout and fail the run if it never appears. Alternatively call `temporalio/api-go/.github/workflows/publish-release.yml` directly as a `needs`-ordered job so its result is part of this run, and let the `release: published` trigger remain the path for manual publishes only.

.github/workflows/create-release.yml
```diff
@@ -33,10 +49,15 @@ jobs:
...
+      # api_ref/api_go_ref let a caller release specific commits rather than
+      # the head of a branch. They must name a corresponding pair: the "Pin
+      # commits sha" step below asserts that the api-go commit's proto/api
+      # submodule points at the api commit. Left empty, both fall back to
+      # `branch`.
       - name: Checkout api
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
-          ref: ${{ github.event.inputs.branch }}
+          ref: ${{ inputs.api_ref || inputs.branch }}
```
stephanos: 🤖

<details>
<summary><strong>small</strong> — <code>Pin commits sha</code> still declares a <code>BRANCH</code> env var that is now actively misleading.</summary>

Now that the checkouts can resolve to `api_ref`/`api_go_ref`, the `BRANCH: ${{ github.event.inputs.branch }}` env on the `Pin commits sha` step (line 99) no longer describes what is checked out. It is also unused by that step's script, as is `GH_TOKEN` there — the script only shells out to `git rev-parse`.

</details>

**Suggestion:** Drop both unused `env` entries from `Pin commits sha` so nothing implies the pinned shas came from `branch`.

.github/workflows/create-release.yml
```diff
@@ -141,3 +162,33 @@ jobs:
...
+  # trigger-api-go-publish-release.yml turns into api-go's publish-release.yml
+  # -- the step that actually creates the api-go tag.
+  publish-release:
+    name: "Publish release"
+    needs:
+      - create-release
+      - release-api-go
+    if: |
+      !cancelled() &&
+      (inputs.auto_publish == true || inputs.auto_publish == 'true') &&
```
stephanos, line 176: When would we need the `== 'true'` part? Isn't it always boolean since it's an input?

stpierre: > Isn't it always boolean since it's an input?

If Github Actions was even just a _moderately_ terrible system, yes. But instead, no. `workflow_call` events pass booleans as booleans, but `workflow_dispatch` events pass them as strings. Even though this WF currently only has a `workflow_dispatch` trigger, adding both is smart forward-compatibility, because someone adding a `workflow_call` trigger would not necessarily suspect that the exact same input variable might be a boolean in one case and a string in another, even though it's declared as a boolean in both.

.github/workflows/create-release.yml
stpierre: As written, this does not set it as latest. How should we approach that? I can see four options; two are defensible:

* Always set it as latest. I think this is what we've done historically, but that doesn't mean it's right (but it does mean that it at least isn't so wrong that it's caused issues so far).
* Add an input to allow setting latest, default to true. Preserves historical behavior but gives us an escape hatch; downside is extra complexity that I'm not sure we'd ever actually use.

Two are less defensible:

* Add an input to allow setting latest, default to false. Diverges from historical behavior, adds complexity, and I'm not sure we gain anything.
* Never set it as latest (the current approach). Diverges from historical behavior for no good reason.

Happy to defer to your judgment on this -- RelEng just wants to cut releases however releases should be cut.

.github/workflows/create-release.yml
stephanos: Gotcha; ok that makes sense 👍

.github/workflows/create-release.yml
stpierre: I've updated this to always mark as latest, which is at least more correct than never marking it as latest. Happy to adjust that if needed.

stephanos (approved):

.github/workflows/create-release.yml
stephanos: LGTM!

maciejdudko (approved):

PR #871, opened 2026-09-17
Clarify namespace poller group snapshots

yuandrew:
**What changed?**
Make it clear that poller groups are namespace wide, not task queue unique.

**Why?**
Codex got confused (and I also forgot), i think we can make this more explicitly clear in commends

**Breaking changes**
No, comment only change

**Server PR**


michaely520 (approved):

tconley1428 (approved):

ShahabT (changes requested):
I don't think we should mark the info Poll APIs as NS level. The one returned by the DescribeNamespace sure is NS level, as the seed value for all TQs in the NS.

True that as of today the weights are all the same for all task queues. But SDK should not assume so. We are adding backlog awareness inside each cell's partition selections. Next natural step is to extend that to make MCN weights backlog-aware. Once that is done, the weights for each TQ would change.

Also note that things like cell-pinning would change the set of cells for a given TQ.

